PR #382, opened 2013-01-09
Add fact for EC2 region

rdark:
Stock set of facts don't include the current region.

This patch just parses the AZ and strips [a-z] from the end of the string to give you the region.


mpdude:
@rdark This would indeed be nice to have :-)

PR #563, opened 2013-11-28
  Fixes #23314 - Add Docker to virtual facts

jamtur01:
Adds support to the virtual facts to detect if Facter is running inside a Docker container.

It uses the /proc/1/environ file and checks for the string:

```
docker=true
```

See http://www.freedesktop.org/wiki/Software/systemd/ContainerInterface/ for more details of this.

See related PR in Docker - https://github.com/dotcloud/docker/pull/2939

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 require 'rubygems'
 require 'mocha'
 require 'rspec'
+require 'rspec/expectations'
 require 'facter'
 require 'fileutils'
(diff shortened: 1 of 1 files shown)


michakrause:
@jamtur1 why not check /proc/1/cgroup for a docker string?

jamtur01:
@michakrause I am no longer involved any way in this PR. :)

kbarber:
@jamtur01 gone, but not forgotten :-).

PR #731, opened 2014-07-16
Display facts for aliases in *BSD

vamme:
- interfaces.rb - aliases on *BSD are iterated over up to their actual count - slightly modified Michelle Sullivan's version
- util/ip.rb - added get_alias_no method to return alias count; method get_interface_value modified by Michelle Sullivan to allow obtaining values for specific aliases

Originated here:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=190796

### DIFF
--- a/lib/facter/interfaces.rb
+++ b/lib/facter/interfaces.rb
@@ -47,3 +47,16 @@ Facter::Util::IP.get_interfaces.each do |interface|
     end
   end
 end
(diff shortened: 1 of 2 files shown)


misullivan:
Did this ever end up in the main code base?

mmoll:
I would say no

PR #832, opened 2014-12-29
(FACT-546) Enclose interface name in single quotes

vinzent:
oVirt/RHEV creates a ";vdsmdummy;" network interface. 

facter outputs many messages: 
sh: vdsmdummy: command not found

ifconfig and ip command lines are called like "/sbin/ifconfig ;vdsmdummy;" which is interpreted as "ifconfig" and "vdsmdummy" (same for the ip command). 

The interface names should be enclosed by single quotes.

### DIFF
--- a/lib/facter/util/ip.rb
+++ b/lib/facter/util/ip.rb
@@ -141,7 +141,7 @@ module Facter::Util::IP
     if File.exists?("/sys/class/net/#{interface}/address") then
       ib_mac_address = `cat /sys/class/net/#{interface}/address`.chomp
     elsif File.exists?("/sbin/ip") then
-      ip_output = %x{/sbin/ip link show #{interface}}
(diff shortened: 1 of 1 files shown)


whopper:
@vinzent thanks for the contribution!

We'll be having our weekly pull request triage this Wednesday, so we'll discuss getting this merged in then. Using single quotes with `ifconfig` and `ip` worked on every platform that I tested, so :+1: from me.

kylog:
@whopper do you know if cfacter has any prob with interface names with semi-colons?

whopper:
@kylog I'm not sure - I'll go poke around and see if I can replicate this.

whopper:
@kylog it looks like cfacter is unaffected since it uses the `getifaddrs()` function to grab interface info directly rather than having to shell out to `ifconfig` or `ip`. So, I believe we should be good on that front!

PR #834, opened 2014-12-29
(FACT-753) Update default external fact directories for unified agent layout

whopper:
This commit updates the default search directories for external facts.
We now look in `/opt/puppetlabs/agent/facts.d` for both open source and
PE versions of Facter. When Facter 3 ships, we will default to only
searching in this directory, but to facilitate a smoother transition
Facter 2.4 will look in both old and new external fact search paths.

Note that this change only affects Unix agents, as Windows paths are
staying the same.

This change is part of the unified agent effort.

### DIFF
--- a/acceptance/tests/runs_external_facts_once.rb
+++ b/acceptance/tests/runs_external_facts_once.rb
@@ -24,7 +24,7 @@ agents.each do |agent|
     ext = '.bat'
     content = win_content
   else
-    factsd = '/etc/facter/facts.d'
+    factsd = '/opt/puppetlabs/agent/facts.d'
     ext = '.sh'
     content = unix_content
   end
--- a/lib/facter/util/config.rb
+++ b/lib/facter/util/config.rb
@@ -41,7 +41,11 @@ module Facter::Util::Config
     if Facter::Util::Root.root?
       windows_dir = windows_data_dir
       if windows_dir.nil? then
-        @external_facts_dirs = ["/etc/facter/facts.d", "/etc/puppetlabs/facter/facts.d"]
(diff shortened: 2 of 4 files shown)


joshcooper:
@whopper @haus do we actually need to change the windows location? If we do change it, we will need to update puppet_for_the_win so that the installer can apply the correct permissions to the directory, i.e. we don't want unprivileged users being able to drop external executable facts into the directory, and have LocalSystem execute it. /cc @Iristyle

Iristyle:
Hopefully we can leave it alone for Windows - I don't think that we're manipulating the directory structure for Windows agents based on conversations I've had with @melissaanne, but please let me know if that understanding is incorrect @haus

haus:
nope, windows shouldn't be changing right now. all clear.

joshcooper:
Thinking out loud here... is this a breaking change that requires a facter 3.0 release? Or will this go out in 2.4?

kylog:
@joshcooper great question. I was thinking that this PR continues to honor the current directories (just adds another one) so it would *not* be a breaking change. (And then the complementary change in cfacter/facter3 is breaking but that will be released in a major.)

But I'd like that reasoning double checked so: thoughts?

joshcooper:
@kylog :+1:

whopper:
@joshcooper @kylog @haus so just to be totally sure before the merge - I opted to make the path `/opt/puppetlabs/agent/facts.d`, but we could also do `/opt/puppetlabs/agent/facter/facts.d`. Do you think leaving out the `facter` directory is a good idea, or should I put it back in?

kylog:
I like `/opt/puppetlabs/agent/facts.d` - it's a shorter path, and I don't think specifying what sub-package uses the directory is all that useful. Anyone else with opinions?

PR #835, opened 2014-12-31
(FACT-775) do not replace dot by underscore in interface name

pmisiak:
because facter should return eth0.100 vlan interface instead of eth0_100

### DIFF
--- a/lib/facter/util/ip.rb
+++ b/lib/facter/util/ip.rb
@@ -43,7 +43,7 @@ module Facter::Util::IP
 
   # Convert an interface name into purely alphanumeric characters.
   def self.alphafy(interface)
-    interface.gsub(/[^a-z0-9_]/i, '_')
+    interface.gsub(/[^a-z0-9_\.]/i, '_')
   end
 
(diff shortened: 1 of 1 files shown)


pmisiak:
issue created: FACT-775
commit message updated

whopper:
Thanks @pmisiak! We'll be having our weekly pull request triage meeting this Wednesday, so we'll discuss getting this merged in then.

MikaelSmith:
PR Triage Notes:
* This is a breaking change for Facter (changing interface names), so we will target it to Facter 3.
* Some discussion on what other fact name assumptions may be changed in Facter 3, need to consider what constraints are required by Puppet. We'll start a discussion on puppet-dev.

MikaelSmith:
Thanks for the contribution, but as this is a breaking change that would have to be incorporated in Facter 3 - and Facter 3 will be implemented in C++ (see puppetlabs/cfacter) - I'm going to close this pull request. I've continued the discussion on this request in FACT-775.

PR #836, opened 2015-01-09
(FACT-776) Retrieve filesystem label if it exists for partitions fact, in Linux

vide:
Hi
this is to implement this feature request https://tickets.puppetlabs.com/browse/FACT-776 (which I opened too), I hope the implementation is OK for you. If I need to add something more, just ask (and explain how for example if it's unit testing, I'm not very good at it)

### DIFF
--- a/lib/facter/util/partitions/linux.rb
+++ b/lib/facter/util/partitions/linux.rb
@@ -55,6 +55,12 @@ module Facter::Util::Partitions
       end
     end
     
+    def self.label(partition)
+      if Facter::Core::Execution.which('blkid')
+        Facter::Core::Execution.exec("blkid #{File.join('/dev', partition)}").scan(/LABEL="([^"]*)"/).flatten.first
+      end
+    end
+    
     private
     def self.read_size(partition)
       if device = partition.match(/(\D+)/)[1] and File.readable?(File.join(SYSFS_BLOCK_DIRECTORY, device, partition, 'size'))
--- a/lib/facter/util/partitions/openbsd.rb
+++ b/lib/facter/util/partitions/openbsd.rb
@@ -22,6 +22,11 @@ module Facter::Util::Partitions
     def self.filesystem(partition)
       scan_mount(/\/dev\/#{partition}\son\s\S+\stype\s(\S+)/)
     end
+   
+    # On OpenBSD there are no labels for partitions
+    def self.label(partition)
+      nil
+    end
 
     private
     def self.scan_mount(scan_regex)
--- a/lib/facter/partitions.rb
+++ b/lib/facter/partitions.rb
@@ -29,6 +29,7 @@ Facter.add(:partitions) do
       details['uuid']       = Facter::Util::Partitions.uuid(part)
       details['size']       = Facter::Util::Partitions.size(part)
       details['mount']      = Facter::Util::Partitions.mount(part)
+      details['label']      = Facter::Util::Partitions.label(part)
       details['filesystem'] = Facter::Util::Partitions.filesystem(part)
       details.reject! {|k,v| v.nil? || v.to_s.empty? }
       partitions[part] = details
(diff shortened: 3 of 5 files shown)


vide:
CLA signed

MikaelSmith:
This seems like a reasonable extension. Please update your commit message to follow the contributor guidelines at https://github.com/puppetlabs/facter/blob/master/CONTRIBUTING.md.

lib/facter/util/partitions/linux.rb
```diff
@@ -55,6 +55,13 @@ def self.filesystem(partition)
       end
     end
     
+    def self.label(partition)
+      label = nil
```
MikaelSmith: This looks like an orphaned variable. Can it be removed?

lib/facter/util/partitions/openbsd.rb
```diff
@@ -23,6 +23,10 @@ def self.filesystem(partition)
       scan_mount(/\/dev\/#{partition}\son\s\S+\stype\s(\S+)/)
     end
 
+    def self.label(partition)
```
MikaelSmith: Like uuid, please add a comment explaining why this is nil on OpenBSD.

vide:
@MikaelSmith I thought about doing it but I don't have any OpenBSD installation at hand so I wasn't sure about it. Googling didn't clarify that. Anyway I will add some comment

lib/facter/util/partitions/linux.rb
```diff
@@ -55,6 +55,13 @@ def self.filesystem(partition)
...
+      if Facter::Core::Execution.which('blkid')
+        Facter::Core::Execution.exec("blkid #{File.join('/dev', partition)}").scan(/LABEL="([^"]*)"/).flatten.first
```
MikaelSmith: Have you tried this with an unlabeled partition?

MikaelSmith:
If you'd still like to try OpenBSD, http://www.vagrantbox.es/ has some Vagrant boxes that may work.

lib/facter/util/partitions/linux.rb
vide: Yes, it doesn't return any key, so it's completely backwards compatible

lib/facter/util/partitions/linux.rb
vide: Actually that should ensure that if there's no label everything works as
always. I can try to see if even without this assignment that scenario is
covered as wel

MikaelSmith: An empty function returns nil. Both #mount and #filesystem rely on that.

vide:
I changed the path as you suggested @MikaelSmith , hope it's cool now!

MikaelSmith:
It was just pointed out to me that we have specs that cover the partitions facts at https://github.com/puppetlabs/facter/blob/master/spec/unit/partitions_spec.rb#L21-L49. They should be extended to cover the `label`, following the same pattern used for the other facts in `partitions`.

vide:
Ok, added the test. I put the label just in one to check if it works OK with an unlabeled filesystem too

MikaelSmith:
Thanks for the contribution!

vide:
Thanks! I'm quite new to gitHub so, may I ask why I don't appear as author of this commit nor contributor to facter code now?

kylog:
@vide git can be confusing :) Every merge results in a "merge commit" whose author may (in fact, almost always is) different from the author of the individual commit(s) that make up the pull request.

So here @MikaelSmith is the author of the *merge commit* but you are the author of the code commit. To visualize this better, see https://github.com/puppetlabs/facter/commits/master where you'll see both those two commits under "Commits on Jan 15, 2015".

Hope that makes sense, and thanks again for the contribution!

vide:
Thanks @kylog  :)

PR #837, opened 2015-01-13
(FACT-779) Fix issue with Facter failing silently with --puppet option on Puppet 4

whopper:
In commit a5363dd0 in Puppet, the deprecated parse_config class method
was removed from Puppet. Until now, Facter still called this method
when the --puppet flag was invoked, and was failing silently. This
commit updates that call to the newer Puppet.initialize_settings method.


haus:
:+1: would use again

PR #838, opened 2015-01-13
(FACT-184) Facter can't detect long interface names

kstreitova:
The current way how to detect interface names is using the "ifconfig"
but this command shortened the long interface names. As Facter reads
"ifconfig" output with a shortened interface names puppet is not
able to work with them anymore. There is a patch that uses "ip link"
instead of "ifconfig" to retrieve interface names because "ip link"
provides the full interface names.

### DIFF
--- a/lib/facter/util/ip.rb
+++ b/lib/facter/util/ip.rb
@@ -80,7 +80,10 @@ module Facter::Util::IP
 
   def self.get_all_interface_output
     case Facter.value(:kernel)
-    when 'Linux', 'OpenBSD', 'NetBSD', 'FreeBSD', 'Darwin', 'GNU/kFreeBSD', 'DragonFly', 'AIX'
+    when 'Linux'
+      output = Facter::Util::IP.exec_ip(["link"])
+      output.gsub!(/^\d+:\s*/, "")				# delete leading number
+    when 'OpenBSD', 'NetBSD', 'FreeBSD', 'Darwin', 'GNU/kFreeBSD', 'DragonFly', 'AIX'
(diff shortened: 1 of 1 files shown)


kstreitova:
We have signed CLA as a company (SUSE)

lib/facter/util/ip.rb
```diff
@@ -80,7 +80,10 @@ def self.get_interfaces
 
   def self.get_all_interface_output
     case Facter.value(:kernel)
-    when 'Linux', 'OpenBSD', 'NetBSD', 'FreeBSD', 'Darwin', 'GNU/kFreeBSD', 'DragonFly', 'AIX'
+    when 'Linux'
+      output = Facter::Util::IP.exec_ip(["link"])
```
MikaelSmith: Based on the discussion in FACT-184, it's likely safer to first try `ip link`, then if it's not available call `ifconfig`.

MikaelSmith:
Travis is failing because the spec test for this code stubs ifconfig, but not ip. Please look at fixing the test.

kylog:
@kstreitova have you had a chance to look at the spec failures? Thanks!

kylog:
@peterhuene i'm guessing that facter 3 (cfacter) won't have this problem. Thoughts?

peterhuene:
That's correct, as it does not parse ifconfig/ip output.

MikaelSmith:
I made a few notes on the ticket to that effect.

kylog:
This will be fixed in facter 3.0 using the native facter implementation. Thanks for raising the issue!

PR #841, opened 2015-01-22
(FACT-797) Fix KVM detection on Linux

mouchar:
Non-root users must rely on output of ```lspci``` command to detect virtualization. However, PCI devices have ```Virtio``` in its description, so we need case-insensitive match (as I'm not aware how the output on other platforms look like).

### DIFF
--- a/spec/unit/virtual_spec.rb
+++ b/spec/unit/virtual_spec.rb
@@ -210,6 +210,11 @@ describe "Virtual fact" do
       Facter.fact(:virtual).value.should == "hyperv"
     end
 
+    it "should be kvm with virtio device lspci 2>/dev/null" do
+      Facter::Util::Virtual.expects(:lspci).returns("00:03.0 Ethernet controller: Red Hat, Inc Virtio network device")
+      Facter.fact(:virtual).value.should == "kvm"
+    end
+
     context "In a Linux Container (LXC)" do
       before :each do
         Facter.fact(:kernel).stubs(:value).returns("Linux")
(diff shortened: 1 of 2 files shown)


ScottGarman:
:+1:

kylog:
@mouchar this makes sense. Thanks for the contribution! 

One small thing, can you add a spec test for the mixed-case that you're fixing here? This helps ensure that we don't regress. Thanks!

mouchar:
@kylog, just added new rpec test and reformatted commit message. If you have any further objections, let me know.
R.

spec/unit/virtual_spec.rb
```diff
@@ -210,6 +210,11 @@
       Facter.fact(:virtual).value.should == "hyperv"
     end
 
+    it "should be kvm with virtio device lspci 2>/dev/null" do
+      Facter::Core::Execution.stubs(:exec).with('lspci 2>/dev/null').returns("00:03.0 Ethernet controller: Red Hat, Inc Virtio network device")
```
ScottGarman: Can you change this .stubs to .expects instead? It will still stub out the method. This will help ensure the test is more independent of the environment it's being run on.

mouchar:
@ScottGarman Patch updated. Sorry, I don't have much experience with writing fancy rspecs, so I used .stubs() because it is being used in the rest of spec/unit/virtual_spec.rb file. :-)

joshcooper:
@ScottGarman we might want to merge this to facter/stable instead? I don't know that facter/master will ever be released. /cc @kylog @peterhuene

peterhuene:
This and commit 06196412a655b5103011249297368aa5d5543bf2 need to be ported to cfacter.

PR #842, opened 2015-01-27
(maint) add yarjuf to Gemfile

justinstoller:
There's a ci:spec task that requires the yarjuf gem, but there's no yarjuf gem in the Gemfile. This fixes that.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ platforms :ruby do
 end
 
 group :development, :test do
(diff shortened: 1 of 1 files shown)


justinstoller:
This should fix the issues seen [here](https://jenkins.puppetlabs.com/job/platform_aio-facter_unit-ruby_master-linux/RUBY_VERSION=1.9.3-p484,SLAVE_LABEL=unit/2/)

PR #843, opened 2015-01-30
(PUP-3618) Add supporting facts for pkgng provider

xaque208:
To support the Puppet package provider for PkgNG, these facts help
determine if pkg is supported and enabled on a given system.

This was a community effort:

    Alex Hornung <alex@alexhornung.com>
    Brice Lopez <brice.lopez@banh-bao.fr>
    Dennis Weewer <d.weewer@amsio.com>
    Jiří Kubíček <jiri.kubicek@kraxnet.cz>
    Reinier Schoof <reinier@skoef.nl>
    Ryan Langseth <ryan.langseth@gmail.com>
    Ryan Melcer <rmelcer@iteris.com>
    Tim Bishop <tim@bishnet.net>
    clinta <github@clintarmstrong.net>

Thanks all!

### DIFF
--- a/lib/facter/pkgng.rb
+++ b/lib/facter/pkgng.rb
@@ -1,0 +1,49 @@
+
+Facter.add("pkgng_supported") do
+  confine :kernel => ["FreeBSD", "DragonFly"]
+
+  setcode do
+    os = Facter.value('kernel')
+    kernel = Facter.value('kernelversion')
+    if os == "FreeBSD"
+      if kernel =~ /^(8|9|10|11)(\.[0-9])?/
+        "true"
+      end
+    elsif os == "DragonFly"
+      maj, min = kernel.split('.').map { |s| s.to_i }
+      if maj > 3 or (maj == 3 and min >= 4)
+        "true"
+      end
+    end
+  end
+
+end
+
+Facter.add("pkgng_enabled") do
+  confine :kernel => ["FreeBSD", "DragonFly"]
+
+  setcode do
+    if system("TMPDIR=/dev/null ASSUME_ALWAYS_YES=1 PACKAGESITE=file:///nonexistent pkg info pkg >/dev/null 2>&1")
+      "true"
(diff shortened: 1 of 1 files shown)


peterhuene:
@xaque208 If this is primarily to enable default provider behavior, I really think we should be putting this logic in the provider and not having it as a fact, especially a default-shipping fact.  Is there something I'm missing here (short of perhaps having to implement additional functionality in provider confines to support your needs) that requires this to be a fact?

xaque208:
Well, I suppose when the provider gets merged into core, some of this will change, but I've been able to do:

```Puppet
if $pkg_enabled and $kernel == 'FreeBSD' {
  Package {
    provider => 'pkgng',
  }
}
```

The base system, while supporting pkgng, does not enable it by default.  You must still bootstrap the system to get the package manager fully functional on a new system.

I'm fine with moving the logic of detecting enablement in the provider if thats the preferred way to do this.  I didn't really get feedback on how to handle the facts, and I am just now catching up on your messages in the provider PR.

ScottGarman:
Triage notes: this PR depends on Puppet [PR-3276](https://github.com/puppetlabs/puppet/pull/3276)

kylog:
Yes, this depends on the resolution of Puppet PR 3276 (where I think the recommendation was to not rely on facter extensions).

However, if there is consensus around a facter component to the solution, two bits of logistics:
1) this would need a FACT ticket for this PR (rather than a PUP ticket)
2) it would need to be re-targeted at the facter/2.x branch?

Backstory on the latter: https://groups.google.com/forum/#!msg/puppet-dev/tBqHZB9M308/QqdaldMsJlkJ

Thanks for the contribution!

xaque208:
With a slight change I've made this morning to Puppet PR 3276, this should no longer be needed.

kylog:
Okay thanks for the update @xaque208. Closing.

PR #846, opened 2015-02-06
(FACT-797) Fix KVM detection on Linux

ScottGarman:
Non-root users must rely on output of lspci command
to detect virtualization. However, PCI devices have
"Virtio" in its description, so we need case-insensitive
match (as I'm not aware how the output on other platforms
look like). Added rspec to improve test coverage.


peterhuene:
Didn't we already merge this in https://github.com/puppetlabs/facter/commit/151fb43c9092b491ebb9c6113f416fddde8b3359?

ScottGarman:
Sorry @peterhuene, I should have made it more obvious that this PR is for stable. The commit went into master but we're not going to be doing any more facter releases based on master (presumably).

peterhuene:
Erm whoops, sorry.  I often don't check the target branch.

PR #847, opened 2015-02-09
Catching a timeout exception

AndrewBallinger:
A very trivial patch.

This exception can bubble up and cause confusion in places like: https://github.com/puppetlabs/facter/blob/master/lib/facter/core/resolvable.rb#L68
Where it throws "Timed out after #{limit} seconds while resolving #{qualified_name}" even if it timed out for a different reason after a different amount of time.

It looks best to catch it here with the other connection errors.

### DIFF
--- a/lib/facter/ec2/rest.rb
+++ b/lib/facter/ec2/rest.rb
@@ -90,6 +90,9 @@ module Facter
       rescue *CONNECTION_ERRORS => e
         Facter.log_exception(e, "Failed to fetch ec2 uri #{uri}: #{e.message}")
         return nil
+      rescue Timeout::Error => e
+        Facter.log_exception(e, "Failed to fetch ec2 uri #{uri}: #{e.message}")
+        return nil
       end
 
       private


AndrewBallinger:
I feel that this meets the trivial qualification. As it is "Adding logging messages or debugging output."

I will look into the CLA in any case.

ahpook:
+1 to this meeting trivial patch standard. 

@AndrewBallinger FYI, the CLA-bot auto replies to every PR in our namespace with the CLA status of each contributor, its response was not directed at you personally :)

kylog:
This points out that we have logic to retry on timeouts in the `reachable?` method, but *not* in the subsequent `fetch` methods. That inconsistency seems undesirable.

I wonder if the right thing to do is to commonize the timeout/retry/exception logic across those methods so that we consistently apply the same criteria when trying to hit these endpoints.

AndrewBallinger:
I agree! I thought about going for the bigger change, but I wanted to go in for the inch before the yard. This little error message is helping us debug a broader stack issue already.

kylog:
@peterhuene do you know if the native facter implementation of ec2 facts has timeout handling?

kylog:
Holding off on further review of this pending our sorting out timeline for facter 3. And we hope to sort that out in the next week or so.

peterhuene:
@kylog They currently timeout at 200ms (probably too short :/).

kylog:
We're planning to have facter 3 out soon-ish (on the order of weeks rather than months) so this may have been overtaken by events.

However, @AndrewBallinger, if you'd still like to see this fix in a facter 2.4.z release, can you re-target at the facter/2.x branch?

Thanks for the contribution!

AndrewBallinger:
I suppose I'll close this one then.  Thanks for being open to my tiny contribution!

PR #844, opened 2015-02-04
(FACT-596) Update selinux_config_policy fact

frconil:
The newer version of selinux use the term "Loaded policy name"
in lieu of "Policy from config file".

This updated version matches on both for backard compatibility
reasons.

This should fix FACT-756 as well.

PR #848, opened 2015-02-10
(FACT-596) Update selinux_config_policy fact

frconil:
The newer version of selinux use the term "Loaded policy name" in lieu
of "Policy from config file".

This updated version matches on both for backard compatibility reasons

Also added a fixture containing an example run from the newer output
for selinux on centos 7/debian/rhel 7.

### DIFF
--- a/lib/facter/selinux.rb
+++ b/lib/facter/selinux.rb
@@ -148,7 +148,7 @@ Facter.add("selinux_config_policy") do
   setcode do
     result = 'unknown'
     mode = Facter::Core::Execution.exec(sestatus_cmd)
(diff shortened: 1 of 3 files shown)


frconil:
As requested in https://github.com/puppetlabs/facter/pull/844

kylog:
And ... one last one: @peterhuene do we handle both formats in native facter?

peterhuene:
@kylog I believe so, but native facter searches through `/etc/selinux/config` itself rather than relying on the existence of the `sestatus` program.  It looks at `SELINUX_TYPE`.

PR #849, opened 2015-02-15
Properly detect KVM when CPU type is not qemu32/qemu64

levic:
kvm detection code in util/virtual.rb examines cpuid data and looks for 'QEMU' string. However, if KVM CPU emulation is not set to qemu32 or qemu64 then this does not work. For example, if CPU type is set to 'host' then /proc/cpuinfo simply passes through the underlying host CPU details.

If a QEMU cpu is not found, kvm can be detected by looking at dmidecode data.


ScottGarman:
:+1:

MikaelSmith:
This should also be targeted against CFACT (facter 3). I can do that once a JIRA ticket exists.

kylog:
One more logistic: can you re-target at the facter/2.x branch?

Thanks for the contribution!

melissa:
Closing in favor of https://github.com/puppetlabs/facter/pull/896 retargeted at facter/2.x

PR #852, opened 2015-02-19
WIP (FACT-822) Default to AIO virt-what location

MikaelSmith:
Without this change, facter in the AIO agent may use a different
virt-what implementation than intended.

Check for virt-what in the AIO agent install location. If not found,
then search for it on PATH.


kylog:
Actually can you target at the 2.x branch? That's what puppet-agent is now using.

MikaelSmith:
Sure. I'm reconsidering this to mirror the CFacter behavior.

PR #853, opened 2015-02-19
(FACT-822) Prioritize AIO binaries over PATH

MikaelSmith:
Without this change, facter in the AIO agent may use system binaries
instead of AIO package binaries.

Prefix the search path with the AIO binary location so we will use AIO
binaries when they're available.

### DIFF
--- a/lib/facter/util/config.rb
+++ b/lib/facter/util/config.rb
@@ -64,4 +64,22 @@ module Facter::Util::Config
   end
 
   setup_default_ext_facts_dirs
+
+  def self.override_binary_dir=(dir)
+    @override_binary_dir = dir
+  end
+
+  def self.override_binary_dir
+    @override_binary_dir
+  end
+
+  def self.setup_default_override_binary_dir
+    if Facter::Util::Config.is_windows?
+      @override_binary_dir = nil
+    else
+      @override_binary_dir = "/opt/puppetlabs/puppet/bin"
+    end
+  end
+
+  setup_default_override_binary_dir
 end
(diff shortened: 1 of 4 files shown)


stschulte:
Just out of curiousity: Which binary is meant here? In case it is shipped with the AIO package wouldn't it make sense to specificially use an absolute path to the binary in the fact then?

kylog:
This does seem like it could open the door to some odd corner cases where we add `foo` to /opt/puppetlabs/puppet/bin, but there's also a system `foo`, and that's the one that, say, a custom fact is written to assume.

MikaelSmith:
That's a fair point. I didn't have a specific case in mind, so hard-coding this specific fact would be fine. Adding a custom fact that gets the wrong executable seems pretty unlikely, but so is adding more executables that facter will call. And it's good not to confuse users just to make maintenance easier.

kylog:
Okey doke, let's go that route. Same for the cfacter impl.

peterhuene:
Besides `virt-what`, are there other executables we plan on having Facter rely on from AIO?  Just curious.

kylog:
I don't know of are any other executables, now or planned.

haus:
I don't know if/when we will, but pciutils and dmidecode are likely candidates for AIO inclusion.

MikaelSmith:
Oh, another twist: prefixing PATH with install dirs is already done in the Windows MSI at https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L25.

MikaelSmith:
Alright, answer about PATH on Windows is that the AIO agent has public and private bin dirs. We'll need to keep that approach in mind if we start adding our own versions of binaries on Windows.

MikaelSmith:
Updated to only affect virt-what.

kylog:
:+1:

PR #854, opened 2015-02-24
(FACT-826) Use /opt/puppetlabs/facter/facts.d

joshcooper:
Previously, facter would load external facts from
`/opt/puppetlabs/agent/facts.d`, but this was wrong for two reasons. The
`agent` directory was renamed to `puppet` in AIO. Also, it was
conflating facter's `facts.d` directory, with puppet's `facts.d`
directory. Puppet pluginsyncs external facts in modules to the latter
directory, which defaults to `$vardir/facts.d`

This updates facter to use `/opt/puppetlabs/facter/facts.d`. This
directory is private to facter.

### DIFF
--- a/acceptance/tests/runs_external_facts_once.rb
+++ b/acceptance/tests/runs_external_facts_once.rb
@@ -24,7 +24,7 @@ agents.each do |agent|
     ext = '.bat'
     content = win_content
   else
-    factsd = '/opt/puppetlabs/agent/facts.d'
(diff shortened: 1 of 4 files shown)


joshcooper:
ah rebase issues, fixing

kylog:
Can you retarget at 2.x branch and we'll merge up?

kylog:
Btw, somehow the spec file picked up execute permissions.

PR #862, opened 2015-03-03
(maint) Update beaker pin and allow env var override

kylog:
This adds support for BEAKER_VERSION as was added in puppet at commit:

puppetlabs/puppet@072891f

However, the beaker pin is to master, since it has updated AIO paths.

This change also updates the multi-json gem version - this was pinned to 1.7.7
over a year ago to avoid a bug since fixed, so we can return to normal
pessimistic versioning. This is necessary because newer beakers pick up a
requirement for a newer version of multi-json via a transitive dependency.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -1,12 +1,22 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-gem "beaker", "~> 2.2.0"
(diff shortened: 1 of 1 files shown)


kylog:
@joshcooper this is the updated PR for facter with beaker at master. Note this is intentionally pointed at facter/2.x because that's what puppet-agent is using.

PR #863, opened 2015-03-04
(FACT-825) Increase prtdiag timeout for Solaris virtual fact

whopper:
This commit bumps up the default timeout for the virtual fact in
Solaris to 20 seconds. `prtdiag` on large Solaris systems can often
take upwards of 10 seconds, so 20 seconds is a reasonably safe timeout.

This fix was previously merged (d82f695a0) and then lost in the
Facter-2 / master branch merge.

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -73,7 +73,7 @@ end
 Facter.add("virtual") do
   confine :kernel => 'SunOS'
   has_weight 10
-  self.timeout = 6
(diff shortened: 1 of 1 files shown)


kylog:
Sorry, can you actually target this at facter/2.x? 

This is confusing atm because we're coming up on the transition to facter 3. Email to puppet-dev@ coming soon!

whopper:
Aha, sure! Retargeting

PR #864, opened 2015-03-04
(FACT-825) Increase prtdiag timeout for Solaris virtual fact

whopper:
This commit bumps up the default timeout for the virtual fact in
Solaris to 20 seconds. `prtdiag` on large Solaris systems can often
take upwards of 10 seconds, so 20 seconds is a reasonably safe timeout.

This fix was previously merged (d82f695a0) and then lost in the
Facter-2 / master branch merge.

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -73,7 +73,7 @@ end
 Facter.add("virtual") do
   confine :kernel => 'SunOS'
   has_weight 10
-  self.timeout = 6
+  self.timeout = 20
 
   setcode do
     next "zone" if Facter::Util::Virtual.zone?


whopper:
@kylog retargeted!

branan:
Can `prtdiag` hang in ways that it doesn't return to us? Why do we have a timeout at all?

branan:
I'm fine with how we came to the new timeout number (largest seen *2), I'm just curious why it exists in the first place

whopper:
@branan according to bfc16f6b, it hangs in specific cases which would cause facter to hang indefinitely.

branan:
@adrienthebo You will pay for your lame commit message :stuck_out_tongue: 

@whopper cool, we should probably ticket to add that to cfacter too, then? Unless we've bypassed prtdiag and are just hitting system APIs, in which case it might not matter at all.

branan:
I guess since we apparently do need a timeout, I'm :+1:

PR #865, opened 2015-03-05
(FACT-832) Update ubuntu acceptance test for AIO

kylog:
This passed for precise at my desk.

### DIFF
--- a/acceptance/tests/operatingsystem_detection_after_clear_on_ubuntu.rb
+++ b/acceptance/tests/operatingsystem_detection_after_clear_on_ubuntu.rb
@@ -14,5 +14,5 @@ agents.each do |agent|
 
   create_remote_file(agent, script_name, script_contents)
 
-  on(agent, "ruby #{script_name}")
+  on(agent, "#{agent['privatebindir']}/ruby #{script_name}")
 end


kylog:
@haus this should fix the acceptance test failure on ubuntus that we noted.

/cc @justinstoller

haus:
@kylog I don't know the standard terms for these things in beaker-speak. @joshcooper was mentioning something about this in the AIO standup today. I thought this was going to be called privatebindir or something?

kylog:
Yes it will be `privatebindir` once QENG-1946 is resolved. I don't know the timeframe for that, so filed this, and then FACT-833 to update this once there's a beaker release with `privatebindir`.

This temporary solution is also in place in puppet, e.g. https://github.com/puppetlabs/puppet/blob/8eb9557420bd10cfb268c55a62698136338c9b79/acceptance/config/nodes/rhel7.yaml#L14

Yaks all the way down.

haus:
:+1:

kylog:
@joshcooper thoughts?

joshcooper:
It should be possible to specify `privatebindir` in the host configs. Beaker doesn't care if it's a setting it currently knows about or not. That way we don't have to fix the acceptance test as part of FACT-833, instead we just pull the pin and remove the `privatebindir` overrides from the host configs in one commit.

kylog:
Retested with the new and improved beaker#master, reducing this to a one-line change.

joshcooper:
```
$ be rake ci:test:aio SHA=25a146bb621c20f1878d0e14b66c235df4a0e969 CONFIG=config/nodes/precise.yaml
...
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 8.89 seconds
      Average Test Time: 1.78 seconds
              Attempted: 5
                 Passed: 5
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 5

PR #845, opened 2015-02-04
(FACT-805) use /etc/os-release to identify CoreOS

jumanjiman:
It is possible to run an unprivileged docker container on CoreOS
with selective mounts (incl. `-v /etc/os-release:/etc/os-release:ro`)
and have the puppet agent *inside* the container affect the host
*outside* the container.

It is also nice to be able to run such a container on CoreOS
merely to report inventory to the puppet master and puppetdb
to track inventory.

Before this commit, Facter falls back to generic Linux facts:

    # facter -p | egrep -i '^(os|operating)'
    operatingsystemmajrelease => 3
    operatingsystemrelease => 3.18.2
    os => {"family"=>"Linux", "release"=>{"major"=>"3", "minor"=>"18", "full"=>"3.18.2"}}
    osfamily => Linux

After this commit, Facter uses /etc/os-release when /etc/os-release
is mounted inside the container and recognizes the underlying
coreos inventory data. This commit is based on the existing
logic for Cumulus Linux and adds corresponding rspec tests.

    # facter -p | egrep -i '^(os|operating)'
    operatingsystem => CoreOS
    operatingsystemmajrelease => 575
    operatingsystemrelease => 575.0.0
    os => {"name"=>"CoreOS", "family"=>"CoreOS", "release"=>{"major"=>"575", "minor"=>"0", "full"=>"575.0.0"}}
    osfamily => CoreOS

New unit tests:

    Facter::Operatingsystem::Linux
      Operatingsystemmajrelease key
        on CoreOS operatingsystems
          should be derived from operatingsystemrelease

    Facter::Util::Operatingsystem
      reading the os-release file
        correctly parses the file on CoreOS Linux

It would be nice to DRY up the code so that Cumulus and CoreOS
use a common implementation, but this commit attempts to keep
the logic separate to avoid touching any existing rspec for Cumulus.

PR #866, opened 2015-03-05
(FACT-805) use /etc/os-release to identify CoreOS

jumanjiman:
It is possible to run an unprivileged docker container on CoreOS
with selective mounts (incl. `-v /etc/os-release:/etc/os-release:ro`)
and have the puppet agent *inside* the container affect the host
*outside* the container.

It is also nice to be able to run such a container on CoreOS
merely to report inventory to the puppet master and puppetdb
to track inventory.

Before this commit, Facter falls back to generic Linux facts:

    # facter -p | egrep -i '^(os|operating)'
    operatingsystemmajrelease => 3
    operatingsystemrelease => 3.18.2
    os => {"family"=>"Linux", "release"=>{"major"=>"3", "minor"=>"18", "full"=>"3.18.2"}}
    osfamily => Linux

After this commit, Facter uses /etc/os-release when /etc/os-release
is mounted inside the container and recognizes the underlying
coreos inventory data. This commit is based on the existing
logic for Cumulus Linux and adds corresponding rspec tests.

    # facter -p | egrep -i '^(os|operating)'
    operatingsystem => CoreOS
    operatingsystemmajrelease => 575
    operatingsystemrelease => 575.0.0
    os => {"name"=>"CoreOS", "family"=>"CoreOS", "release"=>{"major"=>"575", "minor"=>"0", "full"=>"575.0.0"}}
    osfamily => CoreOS

New unit tests:

    Facter::Operatingsystem::Linux
      Operatingsystemmajrelease key
        on CoreOS operatingsystems
          should be derived from operatingsystemrelease

    Facter::Util::Operatingsystem
      reading the os-release file
        correctly parses the file on CoreOS Linux

It would be nice to DRY up the code so that Cumulus and CoreOS
use a common implementation, but this commit attempts to keep
the logic separate to avoid touching any existing rspec for Cumulus.

Related to: CFACT-215 puppetlabs/cfacter@c8c9b7a7af7

Related to: puppetlabs/facter#845


branan:
:+1: This is already in cfacter, and includes tests. Can't ask for much better than that :)

PR #869, opened 2015-03-06
(FACT-824) Reduce calls to 'ip link show' in Facter::Util::IP

holstvoogd:
Only call get_bonding_interface when it is result is actually used.


branan:
I'd like to double-check that we don't have the same bug in cfacter before we merge this, so we don't forget to fix it if it's there. Once we know we're safe on that side we can get this in.

@peterhuene  or @MikaelSmith, does one of you have time to check the cfacter behavior here?

MikaelSmith:
cfacter uses a system call `getifaddrs` that returns all this info; no room for unnecessary calls there, so this bug doesn't exist in cfacter.

branan:
I'm :+1:, then

PR #870, opened 2015-03-06
(FACT-834) Avoid double loading ec2 warnings

activars:
When implementing custom facter module, I usually require external dependencies. But when requiring 'facter/ec2',  I get the following warning:

```
Already evaluated rest at true, reevaluating anyways
Already evaluated rest at true, reevaluating anyways
```

This PR ensure the `facter/ec2/rest.rb` is only loaded once when I require 'facter/ec2' module. 

Singed CLA.

### DIFF
--- a/lib/facter/ec2.rb
+++ b/lib/facter/ec2.rb
@@ -15,7 +15,7 @@
 #   Directly queries the EC2 metadata endpoint.
 #
 
-require 'facter/ec2/rest'
+require 'facter/ec2/rest' unless defined?(Facter::EC2)
 
 Facter.define_fact(:ec2_metadata) do
   define_resolution(:rest) do


branan:
I'd like to spend some time looking at the `rest.rb` file to figure out what's actually causing this warning. It should always be safe to require something more than once, and I'd prefer to fix the underlying issue if we can sort it out.

Do agree we should silence this warning, though

activars:
I agree on your statement. I think it's still better to avoid loading twice because this would call amazon's http endpoint multiple times. I prefer http call is executed once per facter run.

branan:
I haven't been able to track down the root cause of the warning. At this point I guess I'm :+1: on merging this in.

activars:
I have this change running in production for a few weeks now, it's working for us.

melissa:
@activars it looks like you updated the title of this pull request, but could you update the commit message as well? It needs to be in the same format, referencing the ticket, i.e. `(FACT-834) Avoid double loading ec2 warnings`. Once that's done, we should be able to merge this pull request.

activars:
I corrected the commit message.

whopper:
@activars thanks! :+1: once travis goes green

PR #871, opened 2015-03-12
(FACT-852) Support for systemd and systemd_version

Spredzy:
This commit adds support for systemd and systemd_version facts.

### DIFF
--- a/lib/facter/systemd.rb
+++ b/lib/facter/systemd.rb
@@ -1,0 +1,39 @@
+# Fact: systemd
+#
+# Purpose: 
+#   Determine whether SystemD is the init system on the node
+#
+# Resolution:
+#   Check the name of the process 1 (ps -p 1)
+#
+# Caveats:
+#
+
+# Fact: systemd-version
+#
+# Purpose: 
+#   Determine the version of systemd installed
+#
+# Resolution:
+#  Check the output of systemctl --version
+#
+# Caveats:
+#
+
+Facter.add(:systemd) do
+  confine :kernel => :linux
+  setcode do
+    result = false
+    init_process_name = Facter::Core::Execution.exec('ps -p 1 -o comm=')
+    if init_process_name.eql? 'systemd'
+      result = true
+    end
+  end
+end
+
+Facter.add(:systemd_version) do
+  confine :systemd => true
+  setcode do
+    version = Facter::Core::Execution.exec("systemctl --version | grep 'systemd' | awk '{ print $2 }'")
+  end
+end
--- a/schema/facter.json
+++ b/schema/facter.json
@@ -181,6 +181,8 @@
         "swapsize" : { "type": "string" },
         "swapsize_mb" : { "type": "string" },
         "system32" : { "type": "string" },
+        "systemd" : { "type": "boolean" },
+        "systemd_version" : { "type": "string" },
         "system_uptime" : { "type": "object",
           "properties" : {
             "seconds" : { "type": "integer" },


lib/facter/systemd.rb
```diff
@@ -0,0 +1,39 @@
...
+    result = false
+    init_process_name = Facter::Core::Execution.exec('ps -p 1 -o comm=')
+    if init_process_name.eql? 'systemd'
+      result = true
+    end
+  end
+end
+
+Facter.add(:systemd_version) do
+  confine :kernel => :linux
```
daenney: Should this additionally be confined to `:systemd => true`?

Spredzy: @daenney you're right. Fixed.

lib/facter/systemd.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+
+Facter.add(:systemd) do
+  confine :kernel => :linux
+  setcode do
+    result = false
+    init_process_name = Facter::Core::Execution.exec('ps -p 1 -o comm=')
+    if init_process_name.eql? 'systemd'
+      result = true
+    end
```
joshcooper: I think you just want to do:

```
setcode do
   ...
   init_process_name == 'systemd'
end
```

In other words, the value of the fact should be the last value of the setcode block. If you're on a systemd system, then the last value is `true` due to evaluating `result = true`, e.g.

```
irb(main):001:0> result = true
=> true
```

If you're not on systemd, then the value is actually coming from the `if` check:

```
irb(main):002:0> if 1 == 2
irb(main):003:1> end
=> nil
```

lib/facter/systemd.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+end
+
+Facter.add(:systemd_version) do
+  confine :systemd => true
+  setcode do
+    version = Facter::Core::Execution.exec("systemctl --version | grep 'systemd' | awk '{ print $2 }'")
```
joshcooper: Remove the `version =` part, and just return the output of the exec call.

lib/facter/systemd.rb
daenney: Actually, wouldn't we want this instead:

```ruby
setcode do
  if init_process_name = Facter::Core::Execution.exec('ps -p 1 -o comm='.eql? 'systemd'
    true
  else
    false
  end
end
```

joshcooper: or just

```
setcode do
  Facter::Core::Execution.exec('ps -p 1 -o comm=') == 'systemd'
end

**Comment by melissa** [comment]
Thanks for the contribution!

However, we likely aren't going to do any more releases from the master branch of facter. This looks like work that should be targeted at Facter 3, our native implementation in C++. @MikaelSmith do you have any input on this?

File: lib/facter/systemd.rb (hunk 2 above)
Comment (daenney) [inline lib/facter/systemd.rb]: Oh good point!

**Comment by MikaelSmith** [comment]
Yes, this will need an additional ticket to cover Facter 3.

**Comment by MikaelSmith** [comment]
We plan to make that more obvious soon by moving the Facter 3 codebase to facter#master.

**Comment by kylog** [comment]
Agreed. I'd target this at Facter 3. And yes, the codebase move will make this all more obvious. Coming soon!

For backstory:  https://groups.google.com/forum/#!msg/puppet-dev/tBqHZB9M308/QqdaldMsJlkJ

**Comment by kylog** [comment]
One other thing: I'd add these two facts as a single new structured fact.

**Comment by daenney** [comment]
How would you name it then though? One is binary, one is data:

```
systemd { active => true, version => $version }
```

It feels awkward.

**Comment by kylog** [comment]
Hmm, well, naming is hard, but:

```
systemd { active => true, version => 212 }
```

actually makes sense to me. And then in code you could reference:

```
systemd[active]
systemd[version]
```

Although ... would it be worth generalizing it to `service_managers` or some such? *That* feels like an awkward name, but would allow sth like:

```
if service_manager[type] == 'systemd' ....
```

daenney:
```init_system````?

branan:
Per the existing discussion on this ticket and the concensus in the PR triage discussion meeting, I'm going to close this. New features should be targeted at cfacter as we get ready to release facter 3.

melissa:
Just an update to this, we moved the CFacter code into the facter/master branch. So this work should be targeted there.

PR #875, opened 2015-03-20
(CFACT-258) Add a timeout option to Facter::Core::Execution#execute.

peterhuene:
This adds support in the Ruby Facter API to pass a timeout option to
Facter::Core::Execution#execute.  If the timeout elapses, a
Facter::Core::Execution::ExecutionFailure is raised.

The timeout option is given in seconds.

Example:
```ruby
Facter::Core::Execution.execute('foo', timeout: 5)
```


MikaelSmith:
:+1:

PR #879, opened 2015-03-23
(FACT-855) Update acceptance tests for Facter 3

MikaelSmith:
Remove use of `--puppet`, as it's not needed for the test and is no
longer part of facter.

Remove the json-based schema test. This will be replaced later with
validation against the new schema.


acceptance/tests/ticket_7039_facter_multiple_facts_one_file.rb
```diff
@@ -21,14 +21,14 @@
...
+  on(agent, facter('test_fact1', :environment => env)) do
+    fail_test "Fact 1 not returned by facter test_fact1" unless
       stdout.include? 'test fact 1'
   end
 
   step "Agent: Verify test_fact2 from #{dir}/test_facts.rb"
-  on(agent, facter('--puppet', 'test_fact2', :environment => env)) do
-    fail_test "Fact 1 not returned by facter --puppet test_fact2" unless
+  on(agent, facter('test_fact2', :environment => env)) do
+    fail_test "Fact 1 not returned by facter test_fact2" unless
```
kylog: Should the message here be "Fact 2 ..."?

MikaelSmith: Yes, looks like it should.

PR #880, opened 2015-03-23
(FACT-853) Add epoch for Redhat packaging

MikaelSmith:
Facter used an epoch, so now we're stuck with it.


haus:
The changelog at the bottom also needs the epoch added.

PR #882, opened 2015-03-23
(FACT-865) Work around missing stderr output in tests

MikaelSmith:
FACT-865 documents a bug where stderr written by an external fact isn't
reproduced in facter output. Preserve the behavior of an existing
acceptance test that relied on that behavior by writing to a file
instead.

### DIFF
--- a/acceptance/tests/runs_external_facts_once.rb
+++ b/acceptance/tests/runs_external_facts_once.rb
@@ -1,18 +1,9 @@
 test_name "#22944: Facter executes external executable facts many times"
 
-unix_content = <<EOM
-#!/bin/sh
-echo "SCRIPT CALLED" >&2
-echo "test=value"
-EOM
-
-win_content = <<EOM
-echo "SCRIPT CALLED" >&2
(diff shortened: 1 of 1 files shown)


acceptance/tests/runs_external_facts_once.rb
```diff
@@ -47,13 +43,14 @@
   on(agent, "chmod +x '#{ext_fact}'")
 
   step "Agent #{agent}: ensure it only executes once"
-  on(agent, facter) do
-    lines = stderr.split('\n')
+  on(agent, facter('test'))
```
peterhuene: Should we leave this just `facter` to ensure that during all fact evaluation, the external fact is only executed once?

peterhuene:
Other than that, :+1:

acceptance/tests/runs_external_facts_once.rb
kylog: Makes sense.

peterhuene:
:+1:

PR #883, opened 2015-03-23
(CFACT-268) Don't resolve custom facts when --no-custom-facts is used

whopper:
This commit fixes a regression around the --no-custom-facts
command-line option. Currently, the option is not being honored,
and so custom facts will be resolved even if the option is used.
This commit ensures that custom facts will not be resolved if either
this option is passed or if Ruby is not present.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -237,7 +237,7 @@ int main(int argc, char **argv)
         // Add the environment facts
         facts.add_environment_facts();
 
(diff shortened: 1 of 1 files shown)


MikaelSmith:
:+1:

peterhuene:
:+1:

kylog:
Not too worried about coveralls for this one :)

PR #887, opened 2015-03-24
(maint) Fail hard if building native facter throws errors

melissa:
We are currently running into some errors where building native facter
will throw errors, but the script will continue to run. This isn't good,
especially because it means we ship an incomplete archive. This commit
is an attempt at having the native facter build script fail in a
consistent way when things go wrong.


MikaelSmith:
:+1:

ferventcoder:
Yay!!!

PR #888, opened 2015-03-24
(maint) Fix submodule init in Windows packaging

MikaelSmith:
Without this change, packaging on Windows initializes the submodules on
the master branch, no matter what ref is requested.

Do submodule init and update after changing to the requested branch, to
ensure the correct submodules are initialized and at the right ref.

### DIFF
--- a/contrib/facter.ps1
+++ b/contrib/facter.ps1
@@ -95,9 +95,10 @@ echo $env:PATH
 cd $sourceDir
 
 ## Download facter and setup build directories
-git clone --recursive $facterFork facter
+git clone $facterFork facter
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@melissa

ferventcoder:
This seems a bit weird to me, since we are already in facter that we would clone facter again

MikaelSmith:
Right now it's sourced directly from github... it should be retired as we come up with a better build process, but that hasn't happened yet.

ferventcoder:
Right on.

melissa:
:+1:

PR #889, opened 2015-03-24
(maint) Fix ordering in Windows script

MikaelSmith:
Without this commit, the script fails because `param` needs to come first.

Move all other commands after `param`.

### DIFF
--- a/contrib/facter.ps1
+++ b/contrib/facter.ps1
@@ -1,5 +1,3 @@
-$ErrorActionPreference = 'Stop'
-
 ### Set variables from command line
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@ferventcoder @melissa

MikaelSmith:
See http://jenkins-release.delivery.puppetlabs.net/job/BuildFacterWindows/ARCH=64,label=sane_builders/101/console for the error.

melissa:
:( it didn't do that when I tested it. 

Thanks for catching it. :+1:

PR #890, opened 2015-03-24
(FACT-855) Package schema for future acceptance test

MikaelSmith:
Without this commit, future acceptance tests will not be have access to
the fact schema. Package the schema as part of the acceptance artifacts
so they can be used for acceptance tests.


MikaelSmith:
It's useful to have this in first so I can test the acceptance test that uses it before checking merging that.

PR #892, opened 2015-03-25
(FACT-868) Fix YAML output for short IPv6 addresses

MikaelSmith:
The IPv6 address ::1 isn't quoted in Facter's YAML output, but parsing
it with Ruby YAML strips one of the ':' characters and is not read as a
valid IP address.

Update Facter to quote strings starting with ':'.


peterhuene:
:+1: waiting on CI.

PR #893, opened 2015-03-25
(CFACT-258) Fix Facter::Core::Execution#execute not raising by default.

peterhuene:
If an options hash was passed to Facter::Core::Execution#execute and the
options hash did not contain an :on_fail key, we were not properly
defaulting the value to :raise to raise an error.

This fix calls rb_hash_lookup2 to supply :raise if the lookup fails,
resulting in raising an exception if the :on_fail key is not present in
the options hash.

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -383,11 +383,7 @@ namespace facter { namespace ruby {
         volatile VALUE name = ruby.nil_value();
         VALUE options = argc == 2 ? argv[1] : ruby.nil_value();
         if (!ruby.is_nil(options)) {
-            name = ruby.rb_funcall(
-                    options,
-                    ruby.rb_intern("delete"),
-                    1,
-                    ruby.rb_funcall(ruby.utf8_value("name"), ruby.rb_intern("to_sym"), 0));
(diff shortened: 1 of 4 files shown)


lib/src/ruby/module.cc
```diff
@@ -696,15 +692,16 @@ namespace facter { namespace ruby {
...
         uint32_t timeout = 0;
-        volatile VALUE timeout_option = ruby.rb_hash_lookup(argv[1], ruby.rb_funcall(ruby.utf8_value("timeout"), ruby.rb_intern("to_sym"), 0));
+        volatile VALUE timeout_option = ruby.rb_hash_lookup(argv[1], ruby.to_symbol("timeout"));
         if (ruby.is_fixednum(timeout_option)) {
             timeout = static_cast<uint32_t>(ruby.rb_num2ulong(timeout_option));
         }
 
-        // Get the on_fail option
+        // Get the on_fail option (defaults to :raise)
         bool raise = false;
```
MikaelSmith: How does this default to :raise?

peterhuene: Third argument to `rb_hash_lookup2`, which returns the :raise symbol if not present in the options hash.

MikaelSmith: Ok, that works.

MikaelSmith:
:+1:

PR #894, opened 2015-03-26
(FACT-868) Add unit tests for output quoting

MikaelSmith:
Facter selectively quotes output for JSON, YAML, and text. Add tests to
confirm some of the quoting behavior.


er0ck:
(y)

PR #895, opened 2015-03-26
(FACT-889) Update locale handling

MikaelSmith:
An update to Leatherman removes setting global locale settings, and
introduces the get_locale helper method to get a UTF-8 system default
locale (important on Windows, where the default locale is not UTF-8).
Leatherman now provides an argument to set the locale when setting up
logging, that only affects the logging sink.

The change to setup_logging uses a UTF-8 default, so we don't change
calls to it. However boost::filesystem is no longer initialized in
Leatherman, so we initialize it locally as part of logging setup. We
specifically ensure boost::filesystem is imbued in the libfacter
library, in case boost is statically linked.

### DIFF
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -1,5 +1,7 @@
 #include <facter/logging/logging.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <leatherman/locale/locale.hpp>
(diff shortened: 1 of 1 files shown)


kylog:
Let's start moving tickets to FACT.

melissa:
Should this ticket be moved to the FACT project in Jira?

PR #896, opened 2015-03-27
(FACT-888) Properly detect KVM when CPU type is not qemu32/qemu64

melissa:
If CPU type is not set to a qemu type, /proc/cpuinfo will not contain
'QEMU'; in this case we have to look at dmidecode data to detect
kvm/qemu.


melissa:
On the previous pull request (#849), there was a comment about adding spec tests. It looks like a test for this functionality already exists https://github.com/puppetlabs/facter/blob/2.x/spec/unit/virtual_spec.rb#L343-L346, which is a little surprising to me. It currently does not fail if this addition is not available. I'm going to keep looking at this to see if I can figure out additional spec tests that would more accurately represent the situation this commit is addressing.

whopper:
@melissa it looks like the test we're missing is when `dmidecode` returns something with `Manufacturer: QEMU` in it for Linux, BSD, SunOS, HP-UX and GNU/kFreeBSD.

whopper:
:+1:

PR #900, opened 2015-03-30
(FACT-865) Split stdout and stderr handling in execution.

peterhuene:
This implements a new pipe for reading stderr in addition to stdout from
child processes.  Previously, stderr was optionally redirected to
stdout or to the null device if not merging the two streams.  This change
allows child process execution to treat stdout and stderr as two
separate streams, enabling the ability to process stdout while logging
stderr messages (e.g. external executable facts).

The execute functions now return a tuple of both the stdout and stderr
contents.  The each_line functions now take an additional function
object to call when a line of stderr is read.

### DIFF
--- a/lib/src/execution/execution.cc
+++ b/lib/src/execution/execution.cc
@@ -21,9 +21,10 @@ namespace facter { namespace execution {
     {
     }
 
-    execution_failure_exception::execution_failure_exception(string const& output, string const& message) :
+    execution_failure_exception::execution_failure_exception(string const& message, string output, string error) :
         execution_exception(message),
-        _output(output)
+        _output(move(output)),
+        _error(move(error))
     {
     }
 
@@ -32,8 +33,13 @@ namespace facter { namespace execution {
         return _output;
     }
 
-    child_exit_exception::child_exit_exception(int status_code, string const& output, string const& message) :
-        execution_failure_exception(output, message),
+    string const& execution_failure_exception::error() const
+    {
+        return _error;
+    }
+
+    child_exit_exception::child_exit_exception(string const& message, int status_code, string output, string error) :
+        execution_failure_exception(message, move(output), move(error)),
         _status_code(status_code)
     {
     }
@@ -43,8 +49,8 @@ namespace facter { namespace execution {
         return _status_code;
     }
 
-    child_signal_exception::child_signal_exception(int signal, string const& output, string const& message) :
-        execution_failure_exception(output, message),
+    child_signal_exception::child_signal_exception(string const& message, int signal, string output, string error) :
+        execution_failure_exception(message, move(output), move(error)),
         _signal(signal)
     {
     }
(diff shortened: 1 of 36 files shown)


peterhuene:
Looks like I missed a place we call `execute`.  Fixing.

lib/src/execution/execution.cc
```diff
@@ -121,157 +127,224 @@ namespace facter { namespace execution {
...
-                if (!callback(output)) {
+        // Buffers for all of the output or partial line output
+        string stdout_buffer;
+        string stderr_buffer;
+
+        // Read the streams
+        read_streams(
+            [&](string const& data) {
+                if (!process_data(trim, data, stdout_buffer, stdout_logger, stdout_callback)) {
                     LOG_DEBUG("completed processing output; closing child pipe.");
```
MikaelSmith: Should we specify which pipe we're closing?

lib/src/execution/execution.cc
```diff
@@ -121,157 +127,224 @@ namespace facter { namespace execution {
...
-            }
-
-            // Save the new trailing data
-            output.assign(buffer.begin()+lastNL, buffer.end());
-        }
+                return true;
+            },
+            [&](string const& data) {
+                if (!process_data(trim, data, stderr_buffer, stderr_logger, stderr_callback)) {
+                    LOG_DEBUG("completed processing output; closing child pipe.");
```
MikaelSmith: https://github.com/puppetlabs/puppet/pull/3752

lib/src/execution/posix/execution.cc
```diff
@@ -83,11 +84,97 @@ namespace facter { namespace execution {
...
+
+                // There is data to read
+                auto count = read(descriptor, &(*buffer)[0], buffer->size());
+                if (count < 0) {
+                    if (errno != EINTR) {
+                        LOG_ERROR("read failed: %1% (%2%).", strerror(errno), errno);
+                        throw execution_exception("failed to read child output.");
+                    }
+                    // Interrupted by signal
+                    LOG_DEBUG("child pipe read was interrupted and will be retried.");
```
MikaelSmith: Could be useful to know which child pipe was interrupted (stdout vs stderr).

lib/src/execution/windows/execution.cc
```diff
@@ -195,11 +195,151 @@ namespace facter { namespace execution {
         return commandline;
     }
 
-    pair<bool, string> execute(
+    // Typedef for pipe data
```
MikaelSmith: Bookmark.

lib/src/execution/execution.cc
peterhuene: It should really be "closing child pipes", as both pipes will close when we stop processing. Will fix.

lib/src/execution/execution.cc
peterhuene: Same comment as above.  Will fix.

lib/src/execution/posix/execution.cc
peterhuene: Will fix.

lib/src/execution/windows/execution.cc
```diff
@@ -195,11 +195,151 @@ namespace facter { namespace execution {
...
+    typedef tuple<HANDLE, OVERLAPPED*, bool, string*, function<bool(string const&)> const*> pipe_data;
```
MikaelSmith: Almost worth a struct :)

lib/src/execution/windows/execution.cc
```diff
@@ -195,11 +195,151 @@ namespace facter { namespace execution {
...
+                return;
+            }
+
+            if (timeout) {
+                wait_handles.push_back(timer);
+            }
+
+            // Wait for data (and, optionally, timeout)
+            auto result = WaitForMultipleObjects(wait_handles.size(), wait_handles.data(), FALSE, INFINITE);
+            if (result >= (WAIT_OBJECT_0 + wait_handles.size())) {
```
MikaelSmith: I was going to ask about assuming WAIT_OBJECT_0 is 0 and that result is unsigned, but the docs for WaitForMultipleObjects explicitly mention those things.

lib/src/facts/external/execution_resolver.cc
```diff
@@ -28,18 +28,39 @@ namespace facter { namespace facts { namespace external {
...
-            }, { execution_options::defaults, execution_options::throw_on_failure });
+                },
+                [&](string const& line) {
+                    if (!error.empty()) {
+                        error += "\n";
+                    }
+                    error += line;
+                    return true;
+                },
+                0,  // REVIEW: should we have a timeout for executable external facts?
```
MikaelSmith: I don't see a way to specify this, so I don't see a reason to add it. If someone's providing executable facts, they can add their own timeout.

As devil's advocate, it'd be easy to add, and we could provide a config/command-line option for it. No way to specify it on a per-script basis, which makes me lean away from that approach.

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -224,9 +224,11 @@ namespace facter { namespace facts { namespace linux {
 
         // For VMware ESX, execute the vmware tool
         if (value.empty() && name == os::vmware_esx) {
-            auto result = execute("vmware", { "-v" });
-            if (result.first) {
-                re_search(result.second, boost::regex("VMware ESX .*?(\\d.*)"), &value);
+            bool success;
+            string output, error;
+            tie(success, output, error) = execute("vmware", { "-v" });
```
MikaelSmith: Error redirects to null, so I'd rather have a variable name that reflects that choice here.

lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -170,12 +170,14 @@ namespace facter { namespace facts { namespace linux {
 
     string virtualization_resolver::get_vmware_vm()
     {
-        auto result = execute("vmware", { "-v" });
-        if (!result.first) {
+        bool success;
+        string output, error;
+        tie(success, output, error) = execute("vmware", { "-v" });
```
MikaelSmith: Same, change error to something that makes it obvious it's a dummy because stderr redirects to null.

lib/src/ruby/simple_resolution.cc
```diff
@@ -63,16 +63,24 @@ namespace facter { namespace ruby {
...
-                execution_options::redirect_stderr
+        bool success;
+        string output, error;
+        tie(success, output, error) = execute(
+            command_shell,
+            {
+                command_args,
+                expand_command(_command)
+            },
+            0,  // REVIEW: should setcode commands have a default timeout?
```
MikaelSmith: I don't have a good heuristic for a default, and would lean towards no. Ideally a timeout wouldn't be needed.

lib/tests/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ set(LIBFACTER_TESTS_COMMON_SOURCES
     "facts/schema.cc"
     "facts/string_value.cc"
     "logging/logging.cc"
+    "log_capture.cc"
```
MikaelSmith: bookmark

MikaelSmith:
I'd argue for always logging stderr if redirect_stderr_to_null is chosen and a severity level of DEBUG or more is chosen.

lib/tests/fixtures/facts/external/posix/execution/error_message
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/env sh
+echo error message! >&2
```
MikaelSmith: No equivalent test on Windows?

lib/tests/execution/windows/execution.cc
```diff
@@ -90,39 +90,61 @@ SCENARIO("executing commands with execution::execute") {
         });
         return variables;
     };
-     GIVEN("a command that succeeds") {
-         THEN("the output should be returned") {
-            auto result = execute("cmd.exe", { "/c", "type", normalize(LIBFACTER_TESTS_DIRECTORY "/fixtures/execution/ls/file3.txt") });
-            REQUIRE(result.first);
-            REQUIRE(result.second == "file3");
+    GIVEN("a command that succeeds") {
```
MikaelSmith: We're getting lots of duplicate code between the windows and posix tests... someday we might look for a way to clean that up.

MikaelSmith:
First review pass done, nice work.

lib/src/execution/windows/execution.cc
peterhuene: Yeah, this kind of organically grew as more and more data was needed in the helper function :/  I'll take a look at using a struct to clean it up a little.

lib/src/execution/windows/execution.cc
peterhuene: Yeah, originally I had a success check for `result >= WAIT_OBJECT_0 && result < (WAIT_OBJECT_0 + wait_handles.size())`, but the compiler was "too smart" and turned a warning that `result >= WAIT_OBJECT_0` is always true into an error (since result is unsigned and WAIT_OBJECT_0 is 0).  Had to then make the assumption in the code that WAIT_OBJECT_0 is 0 (I sincerely doubt that will ever change).

lib/src/facts/external/execution_resolver.cc
peterhuene: Yeah, it would need to be on the command line and affect all executable external facts (and perhaps just be a default timeout for all command executions, external fact or not).  May be something valuable to revisit in the future.

lib/src/facts/linux/operating_system_resolver.cc
peterhuene: Will change to "none" or "null" or some such.

lib/src/facts/linux/virtualization_resolver.cc
peterhuene: Will make same change as above.

lib/src/ruby/simple_resolution.cc
peterhuene: Same as my comment about external facts above; I think we could add a global execution timeout on the command line, if anything.  May revisit in the future, but for now will remove the REVIEW comments.

lib/tests/fixtures/facts/external/posix/execution/error_message
peterhuene: I'll add one for the PS resolver.

MikaelSmith: We also run batch and cmd files on Windows. That's more what I was thinking of, but batch and PS are good.

peterhuene: Yeah, was just going to add a batch + another test for the PS resolver specifically.

lib/src/execution/windows/execution.cc
peterhuene: I've pushed a commit with it changed to a struct for POSIX (I think it's cleaner).  Going to make the same changes to Windows now.

peterhuene:
I'll see about logging output when redirecting to null with a log level of >= debug.

peterhuene:
Okay, all feedback taken into account and this should now (hopefully) pass CI.  Ready for one last review.

lib/src/execution/windows/execution.cc
MikaelSmith: That does look cleaner :+1:

MikaelSmith:
:+1: Looks good to me.

PR #902, opened 2015-03-31
(FACT-893) prefer /etc/centos-release over /etc/redhat-release

duritong:
CentOS decided to change the format of /etc/redhat-release within
a minor release, which affected correct detection of CentOS systems
on a 7.1.1503 release.
This patch prefers /etc/centos-release over /etc/redhat-release if
present.
/etc/centos-release has already been present on EL6 and actually
/etc/redhat-release used to be a symlink on /etc/centos-release.
From CentOS 7.1.1503 on /etc/redhat-release id not anymore a
symlink, but comes with its own not (yet) parseable content.
Given that the CentOS project prefers /etc/centos-release over
/etc/redhat-release facter should also prefer that.

More background: http://lists.centos.org/pipermail/centos-devel/2015-February/012873.html

Parsing /etc/os-release might be the more forward looking change,
however this would require much more changes to facter, than such
a emergency hot-fix would justify.

### DIFF
--- a/lib/facter/operatingsystem/linux.rb
+++ b/lib/facter/operatingsystem/linux.rb
@@ -273,6 +273,8 @@ module Facter
             else
               operatingsystem = "OEL"
             end
+          elsif FileTest.exists?("/etc/centos-release")
+            operatingsystem = get_centos_operatingsystem_name
           elsif FileTest.exists?("/etc/redhat-release")
             operatingsystem = get_redhat_operatingsystem_name
           elsif FileTest.exists?("/etc/SuSE-release")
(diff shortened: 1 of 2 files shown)


whopper:
@duritong thanks a ton for jumping on this so fast! Could we get you to update your commit message to `(FACT-893) prefer /etc/centos-release over /etc/redhat-release`? This  helps with the bookkeeping in our release process.

duritong:
@whopper here you go

whopper:
also :+1: based on a bit of functional testing. In CentOS 7.1

    operatingsystem => CentOS
    operatingsystemmajrelease => 7
    operatingsystemrelease => 7.1.1503
    os => {"name"=>"CentOS", "family"=>"RedHat", "release"=>{"major"=>"7", "minor"=>"1",     "full"=>"7.1.1503"}}
    osfamily => RedHat

... and in CentOS 6:

    operatingsystem => CentOS
    operatingsystemmajrelease => 6
    operatingsystemrelease => 6.5
    os => {"name"=>"CentOS", "release"=>{"major"=>"6", "full"=>"6.5", "minor"=>"5"}, "family"=>"RedHat"}
    osfamily => RedHat
... and in RedHat 7, these facts remain correct.

PR #903, opened 2015-03-31
(FACT-894) Prepend new AIO path to search path for external facts

kylog:


MikaelSmith:
Is this pathing finalized? I'd like to see a PR against puppet-specifications.

PR #905, opened 2015-04-01
issue/version-fixup

kylog:


kylog:
I'm assuming we want to keep this in lockstep with the version in lib/CMakeLists.txt.

PR #906, opened 2015-04-01
(FACT-898) Fix schema test to accept integer as a valid double

MikaelSmith:
The schema acceptance test parses YAML output using Ruby's YAML parser,
and validates it against our YAML schema. The YAML output from facter
will drop the '.' if the value is a whole number, such as '0' - which
happened in the load_averages fact, causing an acceptance test failure.
This is valid YAML output for a floating point type, as defined at
http://www.yaml.org/spec/1.2/spec.html#id2804092.

Update the test to recognize all YAML floating point formats for a fact
expected to output a double.


peterhuene:
:+1: for merge after CI completes.

PR #907, opened 2015-04-02
(FACT-900) Fix external facts path for non-Administrator

MikaelSmith:
The external facts path for non-Administrator users was incorrect, it used
COMMON_APPDATA when the puppet specification states it should use
~/.puppetlabs/opt/facter/facts.d. One reason for this location is that
facts in COMMON_APPDATA may not be executable by non-Administrators.

Determine whether Facter is running with Administrator privileges; if not,
use the location in the HOME directory instead of COMMON_APPDATA.

### DIFF
--- a/lib/src/facts/windows/collection.cc
+++ b/lib/src/facts/windows/collection.cc
@@ -17,6 +17,7 @@
 #include <internal/facts/windows/uptime_resolver.hpp>
 #include <internal/facts/windows/virtualization_resolver.hpp>
 #include <internal/util/windows/system_error.hpp>
+#include <internal/util/windows/user.hpp>
 #include <internal/util/windows/windows.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/filesystem.hpp>
@@ -32,13 +33,26 @@ namespace facter { namespace facts {
 
     vector<string> collection::get_external_fact_directories()
     {
-        // Get the user data path
-        TCHAR szPath[MAX_PATH];
-        if (!SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath))) {
-            LOG_DEBUG("error finding COMMON_APPDATA: %1%", system_error());
+        if (user::is_admin()) {
+            // Get the common data path
+            TCHAR szPath[MAX_PATH];
+            if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath))) {
+                path p = path(szPath) / "PuppetLabs" / "facter" / "facts.d";
+                return {p.string()};
+            }
+
+            LOG_WARNING("error finding COMMON_APPDATA, external facts unavailable: %1%", system_error());
+        } else {
+            auto home = user::home_dir();
+            if (!home.empty()) {
+                path p = path(home) / ".puppetlabs" / "opt" / "facter" / "facts.d";
+                return {p.string()};
+            }
+
+            LOG_DEBUG("HOME environment variable not set, external facts unavailable");
         }
-        path p = path(szPath) / "PuppetLabs" / "facter" / "facts.d";
-        return vector<string>{p.string()};
+
+        return {};
     }
 
     vector<unique_ptr<external::resolver>> collection::get_external_resolvers()
--- a/lib/src/util/windows/user.cc
+++ b/lib/src/util/windows/user.cc
@@ -1,0 +1,58 @@
+#include <internal/util/windows/user.hpp>
+#include <internal/util/windows/process.hpp>
+#include <internal/util/windows/system_error.hpp>
+#include <internal/util/windows/windows.hpp>
+#include <facter/util/environment.hpp>
+#include <leatherman/logging/logging.hpp>
+
+using namespace std;
+
+namespace facter { namespace util { namespace windows { namespace user {
+
+    bool is_admin()
+    {
+        if (process::supports_elevated_security()) {
+            return process::has_elevated_security();
+        }
+
+        return check_token_membership();
+    }
+
+    bool check_token_membership()
+    {
+        DWORD sid_size = SECURITY_MAX_SID_SIZE;
+        unsigned char sid_buffer[SECURITY_MAX_SID_SIZE];
+        auto sid = static_cast<PSID>(&sid_buffer);
+        if (!CreateWellKnownSid(WinBuiltinAdministratorsSid, nullptr, sid, &sid_size)) {
+            LOG_DEBUG("Failed to create administrators SID: %1%", system_error());
+            return false;
+        }
+
+        if (!IsValidSid(sid)) {
+            LOG_DEBUG("Invalid SID");
+            return false;
+        }
+
+        BOOL is_member;
+        if (!CheckTokenMembership(nullptr, sid, &is_member)) {
+            LOG_DEBUG("Failed to check membership: %1%", system_error());
+            return false;
+        }
+
+        return is_member;
+    }
+
+    string home_dir()
+    {
+        string home, alt;
+        if (environment::get("HOME", home)) {
+            return home;
+        } else if (environment::get("HOMEDRIVE", home) && environment::get("HOMEPATH", alt)) {
+            return home + alt;
+        } else if (environment::get("USERPROFILE", home)) {
+            return home;
+        }
+        return {};
+    }
+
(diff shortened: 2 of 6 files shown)


MikaelSmith:
@peterhuene @joshcooper @Iristyle I could use some extra review on this.

I've pretty much copied the behavior out of Puppet's `admin?` check, see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/user.rb#L10, so double-checking that I've got that right.

Also, what should the errors coming out of the check be? Debug, Warning, Error?

MikaelSmith:
I've tested out on Windows 2012r2 Administrator and Guest accounts.

kylog:
@MikaelSmith nit: cpplint didn't like the newline situation.

MikaelSmith:
Thanks. I also need to test on Server 2003, since that has different logic.

MikaelSmith:
Behavior on Server 2003 looks good.

lib/src/facts/windows/collection.cc
```diff
@@ -32,13 +33,26 @@ namespace facter { namespace facts {
...
+            if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath))) {
+                path p = path(szPath) / "PuppetLabs" / "facter" / "facts.d";
+                return {p.string()};
+            }
+
+            LOG_WARNING("error finding COMMON_APPDATA, external facts unavailable: %1%", system_error());
+        } else {
+            auto home = user::home_dir();
+            if (!home.empty()) {
+                path p = path(home) / ".puppetlabs" / "opt" / "facter" / "facts.d";
```
peterhuene: This is really the directory we're using on Windows too?  I'd expect at least something that matches the admin path :/

peterhuene:
:+1: from me.

lib/src/facts/windows/collection.cc
branan: I'd have expected us to use local (or roaming, or a merge of both) application data for non-administrator external facts on windows. Using a unix path off of HOME seems weird/wrong

MikaelSmith: I defer to @kylog and @joshcooper, they had a hand in the spec. This will do whatever Facter 2.x does for compatibility.

lib/src/facts/windows/collection.cc
```diff
@@ -32,13 +33,26 @@ namespace facter { namespace facts {
...
+        if (user::is_admin()) {
+            // Get the common data path
+            TCHAR szPath[MAX_PATH];
+            if (SUCCEEDED(SHGetFolderPath(NULL, CSIDL_COMMON_APPDATA, NULL, 0, szPath))) {
+                path p = path(szPath) / "PuppetLabs" / "facter" / "facts.d";
+                return {p.string()};
+            }
+
+            LOG_WARNING("error finding COMMON_APPDATA, external facts unavailable: %1%", system_error());
+        } else {
```
branan: I feel like I'd expect it to try to load the global external facts regardless of your status as an admin, and also allow overriding with per-user things.

As it stands there's no way to add external facts that are resolved by all users on a machine.

branan:
So I'm tentatively :+1:, assuming both of my comments/complaints are "by design". I would like to get an understanding of our reasoning for this stuff, but I recognize right now we need to match facter 2 behavior so I'm not gonna get too hung up on things.

lib/src/facts/windows/collection.cc
joshcooper: I'm not opposed to a global + per-user idea, but posix external facts don't behave that way, so I'd prefer to keep windows consistent.

lib/src/facts/windows/collection.cc
joshcooper: @peterhuene I didn't follow the "matches the admin path" part?

That is the "correct" path, at least as we reached consensus on. Historically puppet, facter, etc, and ruby ecosystem supports unix style `dot` directories, e.g. `~/.puppet`, etc. The difference now is that there is more of hierarchy within. The reasoning behind this change is documented in PUP-4189.

peterhuene: @joshcooper It was a comment on how the sub-directories from the "root" (%AppData%\PuppetLabs vs. %HOME%\.puppetlabs) differ, namely the "opt" in one and not the other.  Just one more thing for users to remember / lookup when their external facts aren't loaded.

joshcooper: @peterhuene ah yeah, I agree that's a bit surprising... On the flip side, it is easy to document the per-user paths because it's always `~/.puppetlabs/opt/facter/facts.d` everywhere.

lib/src/facts/windows/collection.cc
branan: Word. I'd say it's also surprising on Posix systems that non-admin users won't load global external facts, so it's something to revisit.

lib/src/util/windows/user.cc
```diff
@@ -0,0 +1,58 @@
...
+            LOG_DEBUG("Failed to check membership: %1%", system_error());
+            return false;
+        }
+
+        return is_member;
+    }
+
+    string home_dir()
+    {
+        string home, alt;
```
Iristyle: Sorry - late to the party on this one.  I'm guessing you went off of the existing Puppet code that I pointed you to... but 

Why not use [GetUserProfileDirectory](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762280%28v=vs.85%29.aspx) here?

Alternatively, you could also use the shell function [SHGetFolderPath](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762181.aspx) with [CSIDL_PROFILE](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762494.aspx) - but looks like there are some caveats.

I'm guessing that we didn't use the API calls when the Windows code was originally written because it was an "easier" approach in Ruby (code wasn't FFI'd at the time for one thing).

I know we *probably* haven't run into issues within Puppet, but I think peeking at a couple of env vars is a fairly naive approach here.

Iristyle: Another note on the shell APIs (which were already in place at https://github.com/puppetlabs/facter/pull/907/files#diff-5264e60482bee614938ab1266bdf6ce4R39 as well)

* `SHGetFolderPath` has been deprecated by [SHGetKnownFolderPath](http://msdn.microsoft.com/en-us/library/bb762188%28v=vs.85%29.aspx)
* `SHGetFolderPath` / `SHGetKnownFolderPath` are actually not supported on `Server Core` as they're part of the shell APIs that have been removed.  I suspect the same will hold true for the newly announced `Nano Server`. See this [MSDN page](https://msdn.microsoft.com/en-us/library/hh846320\(v=vs.85\).aspx) for the list of APIs available in Server Core.  Note that `GetUserProfileDirectoryW` **is available** on that platform.

I know we don't explicitly make guarantees about running on `Server Core` (and certainly not `Nano Server`), but this is an area where I'd like to make sure we don't accidentally shoot ourselves in the foot.  In the future, I expect these Windows SKUs to become more important, so we don't want to leave Puppet out of the conversation on these platforms if we can help it.

/cc @joshcooper @ferventcoder  @kylog

peterhuene: We should probably open another ticket to support "core" SKUs and remove these deprecated APIs.

Iristyle: Agreed :+1:

kylog: Agreed on both ticketing the API change separately from this PR and on trying out facter 3 on nano server.

lib/src/facts/windows/collection.cc
kylog: Sorry, late to this conversation. Yeah, this approach was definitely a compromise between cross-platform (and docs friendly) path consistency vs platform-specific intelligibility (the embedded "opt" is a little weird on *nix, but very weird on windows). For now, I'm okay rolling with this compromise.

If we find it's a source of confusion/aggravation for users (I don't know how extensive the use case for a per-user default external facts path even is), we can always add a sane(r) path later.

lib/src/util/windows/user.cc
ferventcoder: I'm just gonna +1 this. :)

MikaelSmith: Ok, I'll add a new ticket for https://github.com/puppetlabs/facter/pull/911.

PR #909, opened 2015-04-08
(FACT-895) Include loadavg.h on Solaris

MikaelSmith:
getloadavg is not defined in the POSIX standard, though it's present on
all POSIX systems we support. On Solaris, it's not present in stdlib.h,
so compilation fails.

On Solaris, include sys/loadavg.h to resolve getloadavg.


peterhuene:
Manually verified on solaris.  We still have a link issue in the test executable (working on that now).

PR #911, opened 2015-04-08
(FACT-947) Use GetUserProfileDirectory

MikaelSmith:
Previously, we queried environment variables to find the User's profile
directory. This followed logic used in Puppet to do the same. That was
more error-prone and required querying multiple environment variables.

Use GetUserProfileDirectory to get the path instead, which should work on
all platforms. In particular this will be better supported in future
Windows versions.

### DIFF
--- a/lib/src/util/windows/process.cc
+++ b/lib/src/util/windows/process.cc
@@ -22,7 +22,7 @@ namespace facter { namespace util { namespace windows { namespace process {
 
     bool has_elevated_security()
     {
-        HANDLE temp_token;
+        HANDLE temp_token = INVALID_HANDLE_VALUE;
         if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
             // pre-Vista will return ERROR_NO_SUCH_PRIVILEGE
             if (GetLastError() != ERROR_NO_SUCH_PRIVILEGE) {
@@ -30,7 +30,7 @@ namespace facter { namespace util { namespace windows { namespace process {
             }
             return false;
         }
-        scoped_resource<HANDLE> token(move(temp_token), CloseHandle);
+        scoped_resource<HANDLE> token(temp_token, CloseHandle);
 
         TOKEN_ELEVATION token_elevation;
         DWORD token_elevation_length;
--- a/lib/src/util/windows/process.cc
+++ b/lib/src/util/windows/process.cc
@@ -24,10 +24,7 @@ namespace facter { namespace util { namespace windows { namespace process {
     {
         HANDLE temp_token = INVALID_HANDLE_VALUE;
         if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
-            // pre-Vista will return ERROR_NO_SUCH_PRIVILEGE
-            if (GetLastError() != ERROR_NO_SUCH_PRIVILEGE) {
-                LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
-            }
+            LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
             return false;
         }
         scoped_resource<HANDLE> token(temp_token, CloseHandle);
--- a/lib/src/util/windows/user.cc
+++ b/lib/src/util/windows/user.cc
@@ -2,8 +2,10 @@
 #include <internal/util/windows/process.hpp>
 #include <internal/util/windows/system_error.hpp>
 #include <internal/util/windows/windows.hpp>
-#include <facter/util/environment.hpp>
+#include <facter/util/scoped_resource.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <boost/nowide/convert.hpp>
+#include <userenv.h>
 
 using namespace std;
 
@@ -44,15 +46,31 @@ namespace facter { namespace util { namespace windows { namespace user {
 
     string home_dir()
     {
-        string home, alt;
-        if (environment::get("HOME", home)) {
-            return home;
-        } else if (environment::get("HOMEDRIVE", home) && environment::get("HOMEPATH", alt)) {
-            return home + alt;
-        } else if (environment::get("USERPROFILE", home)) {
-            return home;
+        HANDLE temp_token = INVALID_HANDLE_VALUE;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
+            LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
+            return {};
         }
-        return {};
+        scoped_resource<HANDLE> token(temp_token, CloseHandle);
+
+        DWORD pathLen = 0u;
+        if (GetUserProfileDirectoryW(token, nullptr, &pathLen)) {
+            LOG_DEBUG("GetUserProfileDirectoryW call returned unexpectedly");
+            return {};
+        } else if (GetLastError() != ERROR_INSUFFICIENT_BUFFER) {
+            LOG_DEBUG("GetUserProfileDirectoryW call failed: %1%", system_error());
+            return {};
+        }
+
+        wstring buffer(pathLen, '\0');
+        if (!GetUserProfileDirectoryW(token, &buffer[0], &pathLen)) {
+            LOG_DEBUG("GetUserProfileDirectoryW call failed: %1%", system_error());
+            return {};
+        }
+
+        // Strip the trailing null character.
+        buffer.resize(pathLen > 0u ? pathLen - 1u : 0u);
+        return boost::nowide::narrow(buffer);
     }
 
 }}}}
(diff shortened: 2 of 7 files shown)


MikaelSmith:
Verified on Windows Server 2012r2:
As Admin
```
2015-04-08 14:49:56.729314 DEBUG puppetlabs.facter - searching "C:/ProgramData\PuppetLabs\facter\facts.d" for external facts.
2015-04-08 14:49:56.729314 DEBUG puppetlabs.facter - checking execution on C:/ProgramData\PuppetLabs\facter\facts.d\foo.bat
2015-04-08 14:49:56.729314 DEBUG puppetlabs.facter - resolving facts from executable file "C:/ProgramData\PuppetLabs\facter\facts.d\foo.bat".
2015-04-08 14:49:56.729314 DEBUG puppetlabs.facter - executing command: C:/ProgramData\PuppetLabs\facter\facts.d\foo.bat
2015-04-08 14:49:56.744955 DEBUG | - zebra=C:\Program Files (x86)
2015-04-08 14:49:56.744955 DEBUG puppetlabs.facter - fact "zebra" has resolved to "C:\Program Files (x86)".
2015-04-08 14:49:56.744955 DEBUG puppetlabs.facter - process exited with exit code 0.
2015-04-08 14:49:56.744955 DEBUG puppetlabs.facter - completed resolving facts from executable file "C:/ProgramData\PuppetLabs\facter\facts.d\foo.bat".
```
As Standard User
```
2015-04-08 14:52:50.174710 DEBUG puppetlabs.facter - searching "C:/Users\michael\.puppetlabs\opt\facter\facts.d" for external facts.
2015-04-08 14:52:50.174710 DEBUG puppetlabs.facter - checking execution on C:/Users\michael\.puppetlabs\opt\facter\facts.d\foo.bat
2015-04-08 14:52:50.174710 DEBUG puppetlabs.facter - resolving facts from executable file "C:/Users\michael\.puppetlabs\opt\facter\facts.d\foo.bat".
2015-04-08 14:52:50.174710 DEBUG puppetlabs.facter - executing command: C:/Users\michael\.puppetlabs\opt\facter\facts.d\foo.bat
2015-04-08 14:52:50.190342 DEBUG | - zebra=C:\Users\michael
2015-04-08 14:52:50.190342 DEBUG puppetlabs.facter - fact "zebra" has resolved to "C:\Users\michael".
2015-04-08 14:52:50.190342 DEBUG puppetlabs.facter - process exited with exit code 0.
2015-04-08 14:52:50.190342 DEBUG puppetlabs.facter - completed resolving facts from executable file "C:/Users\michael\.puppetlabs\opt\facter\facts.d\foo.bat".
```

Verified on Windows Server 2003:
As Admin `2015-04-08 14:47:05.000000 DEBUG puppetlabs.facter - skipping external facts for "C:\Documents and Settings\All Users\Application Data\PuppetLabs\facter\facts.d": No such file or directory`
As Standard User `2015-04-08 14:51:35.828125 DEBUG puppetlabs.facter - skipping external facts for "C:\Documents and Settings\michael\.puppetlabs\opt\facter\facts.d": No such file or directory`

lib/src/util/windows/user.cc
```diff
@@ -44,15 +46,24 @@ namespace facter { namespace util { namespace windows { namespace user {
...
+        HANDLE temp_token;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
+            LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
+            return {};
         }
-        return {};
+        scoped_resource<HANDLE> token(move(temp_token), CloseHandle);
+
+        DWORD pathLen;
+        GetUserProfileDirectoryW(token, nullptr, &pathLen);
```
peterhuene: If this fails for some reason, `pathLen` is indeterminate.

peterhuene:
Other than above, :+1: provided the Windows folks are :+1: .

lib/src/util/windows/user.cc
MikaelSmith: Yes... I suppose I should be safer about it.

joshcooper: Although not specified in the msdn docs, the usual behavior is for the function to fail, and have GetLastError return ERROR_INSUFFICIENT_BUFFER and set the required buffer length in the in/out param. That way you can differentiate between "I need to call the API again with the new buffer length" vs "I should fail completely"

joshcooper:
:+1:

lib/src/util/windows/process.cc
```diff
@@ -24,10 +24,7 @@ namespace facter { namespace util { namespace windows { namespace process {
     {
         HANDLE temp_token;
         if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
-            // pre-Vista will return ERROR_NO_SUCH_PRIVILEGE
```
branan: This change seems unrelated to this pull. What's the reason for it?

lib/src/util/windows/user.cc
```diff
@@ -44,15 +46,31 @@ namespace facter { namespace util { namespace windows { namespace user {
...
-            return home + alt;
-        } else if (environment::get("USERPROFILE", home)) {
-            return home;
+        HANDLE temp_token;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
+            LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
+            return {};
         }
-        return {};
+        scoped_resource<HANDLE> token(move(temp_token), CloseHandle);
```
branan: It looks like scoped_resource takes its argument by value and `HANDLE` is just a typedef to a big integer. I don't think we need `move` here?

https://github.com/puppetlabs/facter/blob/master/lib/inc/facter/util/scoped_resource.hpp#L36

lib/src/util/windows/process.cc
MikaelSmith: It's related because I'm repeating the OpenProcessToken call.

I removed it because it came from https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/process.rb#L200, and reviewing the APIs further I don't see any place that error should be generated.

Actually, if anywhere it should be on the call to GetTokenInformation (where TokenElevation may not exist on 2003). This code is never called on 2003 because it's guarded by supports_elevated_security, so having the debug message shouldn't hurt.

lib/src/util/windows/user.cc
MikaelSmith: That seems true, I probably misused it elsewhere and copied it. I'll fix that.

lib/src/util/windows/process.cc
branan: Makes sense.

lib/src/util/windows/user.cc
```diff
@@ -44,15 +46,31 @@ namespace facter { namespace util { namespace windows { namespace user {
...
-            return home + alt;
-        } else if (environment::get("USERPROFILE", home)) {
-            return home;
+        HANDLE temp_token;
+        if (!OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &temp_token)) {
+            LOG_DEBUG("OpenProcessToken call failed: %1%", system_error());
+            return {};
         }
-        return {};
+        scoped_resource<HANDLE> token(temp_token, CloseHandle);
```
joshcooper: Thanks @branan for pointing out the `move` business. I noticed there are other places were we move, e.g. https://github.com/puppetlabs/facter/pull/911/files#diff-374aeb3d1d3ffe6398f65723565a063cR30. May be we fix those in a different commit?

Also, I'm wondering if we should be initializing `temp_token` to `INVALID_HANDLE_VALULE` so there's no chance of `CloseHandle` being called on an uninitialized value. I don't think there's a chance of that happening here, just a general best practice concern.

MikaelSmith:
I broke the misc changes into a bunch of separate commits.

Iristyle:
Looks great... :shipit:

lib/src/util/windows/wmi.cc
```diff
@@ -49,7 +49,7 @@ namespace facter { namespace util { namespace windows {
         if (FAILED(hres)) {
             throw wmi_exception(format_hresult("failed to create IWbemLocator object", hres));
         }
-        _pLoc = scoped_resource<IWbemLocator *>(move(pLoc),
+        _pLoc = scoped_resource<IWbemLocator *>(pLoc,
```
joshcooper: Do we want to remove `move`? In this case `pLoc` is an actual pointer, whereas `HANDLE` is an opaque index into a table of pointers.

MikaelSmith: `move` vs `copy` on a scalar primitive (such as a pointer or opaque index) is purely semantic. `move` is generally implemented as a `copy` because that's the simplest operation. There's no real negative to using `move` on a primitive if you don't plan to use it again (otherwise you violate the semantic contract, even though it'll work with all current compilers).

`move` only has a functional difference with objects or arrays (composite value types) that define `move` semantics via a move operator or constructor.

joshcooper: makes sense, good to know

PR #915, opened 2015-04-09
(FACT-914) Avoid locale on Solaris with GCC

MikaelSmith:
On Solaris when using GCC, std::locale is broken.
https://gcc.gnu.org/ml/gcc/2012-08/msg00284.html

Disable using std::locale on Solaris. The defaults support UTF-8.


MikaelSmith:
Waiting on leatherman update, which I'll amend into this PR.

peterhuene:
The leatherman commit has been merged.

PR #916, opened 2015-04-09
(FACT-946) Fix intermittent verify_facts failure

MikaelSmith:
YAML-CPP emits doubles without a trailing decimal when they round to an
integer. Our type checking expected more stringent type differentiation,
so it would fail occasionally when a double could be emitted as a whole
number.

Fix by accepting integers for types that are expected to be a double.


acceptance/tests/verify_facts.rb
```diff
@@ -26,7 +26,8 @@ def validate_fact(node, fact_value, hidden)
               # http://www.yaml.org/spec/1.2/spec.html#id2804092 states the form of a double in YAML
               # If a type isn't explicit, it's reasonable to output a single integer value for a double,
               # as in 0 and 1, instead of 0.0 or 1.0.
-              fact_value.is_a? Float or fact_value.to_s =~ /^([0-9]|\.inf|\.nan|-\.inf)$/
+              # YAML-CPP seems to drop the decimel whenever it feels like it, so just match Int too.
```
kylog: s/decimel/decimal/ :)

PR #918, opened 2015-04-10
(FACT-951) Build Boost.Chrono in Windows packaging

MikaelSmith:
Prior commits updated Facter to explicitly depend on Boost Chrono.
Update Windows packaging to be explicit about building it as well.

### DIFF
--- a/contrib/facter.ps1
+++ b/contrib/facter.ps1
@@ -127,6 +127,7 @@ if ($buildSource) {
     "--with-regex",
     "--with-log",
     "--with-locale",
+    "--with-chrono",
     "--prefix=`"$toolsDir\$boostPkg`"",
     "boost.locale.iconv=off"
     "-j$cores"


contrib/facter.ps1
```diff
@@ -127,6 +127,7 @@ if ($buildSource) {
     "--with-regex",
     "--with-log",
     "--with-locale",
+    "--with-chrono",
```
melissa: Is this something we're going to want to pull into the linux/solaris builds as well, or is it windows specific?

MikaelSmith: Yes. Some work for Solaris has already been done: https://github.com/puppetlabs/cfacter-build/blob/master/projects/boost/Makefile.SunOS-i386#L33

MikaelSmith: And everywhere it was already happening implicitly because Boost's build system new it was required by Boost.Log. That's why CI passed.

melissa:
@MikaelSmith I want to add this explicitly everywhere. Will you also add this to https://github.com/puppetlabs/pl-build-tools-vanagon/blob/master/configs/components/boost.rb ? Other than that, I'm :+1:

MikaelSmith:
Added https://github.com/puppetlabs/pl-build-tools-vanagon/pull/14

PR #921, opened 2015-04-13
(maint) Remove Fedora 19 from build targets

melissa:
Fedora 19 went EOL on 2015-01-06. We should not longer be providing
packages for this platform.


melissa:
I'm not positive what's happening with the stable branch of facter, but it looks like it's still ruby code. The commit to facter/2.x should be sufficient.

PR #923, opened 2015-04-14
(FACT-955) Fix hardwaremodel/architecture fact for WOW64.

peterhuene:
On 64-bit Windows, facter was reporting "i686" for the hardwaremodel and
architecture facts when running under WOW64 (i.e. facter itself was
32-bit).  There was an incorrect check in Facter for WOW64 that was
reporting the architecture of Facter and not the operating system.

This fix removes that check.  Additionally, removing architectures that
are no longer supported (Windows only supports x86, x64, ia64, and ARM).

### DIFF
--- a/lib/src/facts/windows/operating_system_resolver.cc
+++ b/lib/src/facts/windows/operating_system_resolver.cc
@@ -20,54 +20,27 @@ namespace facter { namespace facts { namespace windows {
 
     static string get_hardware()
     {
-        // IsWow64Process is not available on all supported versions of Windows.
-        // Use GetModuleHandle to get a handle to the DLL that contains the function
-        // and GetProcAddress to get a pointer to the function if available.
-        typedef BOOL (WINAPI *LPFN_ISWOW64PROCESS) (HANDLE, PBOOL);
-
-        BOOL isWow64 = FALSE;
-        LPFN_ISWOW64PROCESS fnIsWow64Process = (LPFN_ISWOW64PROCESS)
-                GetProcAddress(GetModuleHandleW(L"kernel32"), "IsWow64Process");
-        if (nullptr != fnIsWow64Process) {
-            if (!fnIsWow64Process(GetCurrentProcess(), &isWow64)) {
-                LOG_DEBUG("failure determining whether current process is WOW64, defaulting to false"
-                        ": %1%", system_error());
-            }
-        }
-
         SYSTEM_INFO sysInfo;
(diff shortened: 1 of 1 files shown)


lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -20,54 +20,27 @@ namespace facter { namespace facts { namespace windows {
...
-            case PROCESSOR_ARCHITECTURE_INTEL:          return "i" + to_string(archLevel) + "86";
-            default: return "unknown";  // PROCESSOR_ARCHITECTURE_UNKNOWN
+            case PROCESSOR_ARCHITECTURE_AMD64:
+                return "x64";
+            case PROCESSOR_ARCHITECTURE_ARM:
+                return "arm";
+            case PROCESSOR_ARCHITECTURE_IA64:
+                return "ia64";
+            case PROCESSOR_ARCHITECTURE_INTEL:
+                return "i" + to_string((sysInfo.wProcessorLevel > 5) ? 6 : sysInfo.wProcessorLevel) + "86";
```
MikaelSmith: Based on our earlier discussion about 32-bit OS on 64-bit hardware, how does this work?

peterhuene: GetNativeSystemInfo will always report the processor architecture of the OS, regardless of WOW, so it should be all we need.  On x86, this should report as `i?86`.  I'm setting up a 32-bit OS now.

peterhuene: Let's not merge until I've verified this.

peterhuene: * x86 Windows: `architecture => x86` and `hardwaremodel => i686`.
* x64 Windows with 32-bit Facter: `architecture => x64` and `hardwaremodel => x64`.
* x64 Windows with 64-bit FacteR: `architecture => x64` and `hardwaremodel => x64`.

Should be gtg.

MikaelSmith: You decided not to change this to hardwaremodel x86_64 on 64-bit? Or is that a different PR?

peterhuene: Different PR.

MikaelSmith:
We need an acceptance test covering some OS facts on different OS configs. I'll file that as a separate ticket however.

ferventcoder:
w00t!

PR #924, opened 2015-04-15
(FACT-952) Update Facter man page for Facter 3

whopper:
The Facter man page was last updated in February of 2014. This
commit updates it for new features and options in Facter 3.

### DIFF
--- a/man/man8/facter.8
+++ b/man/man8/facter.8
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "FACTER" "8" "February 2014" "" ""
+.TH "FACTER" "8" "April 2015" "" ""
 .
 .SH "NAME"
 \fBfacter\fR \- Gather system information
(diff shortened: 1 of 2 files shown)


man/man8/facter.8
```diff
@@ -45,14 +78,48 @@ hardwaremodel => x86_64
...
+.nf
+$ facter processors.isa
+x86_64
+.
+.fi
+.
+.IP "" 0
+.
+.P
+Display a single legacy fact:
```
whopper: Do we want to call out being able to call the old facts, despite the fact that they aren't included in the default output? Suggestions for terminology?

MikaelSmith: I think we should, and calling them legacy facts sounds good.

peterhuene: Yeah, we want to show that those facts are still available, but not displayed in the output because they'd show duplicated data from the newer structured facts.

man/man8/facter.8
```diff
@@ -45,14 +78,49 @@ hardwaremodel => x86_64
...
+$ facter processors.isa
+x86_64
+.
+.fi
+.
+.IP "" 0
+.
+.P
+Display a single legacy fact. Note that non-structured facts existing in previous versions of Facter are still available,
+but are not displayed by default due to redundancy with newer structured facts:
```
whopper: @MikaelSmith @peterhuene cool, added a bit of wording to the example explain legacy facts. How's this sound?

peterhuene: Makes sense to me.

MikaelSmith:
We also need to update the install script (CMakeLists.txt) to install the man file. A single `install(FILES ...)` should be sufficient.

MikaelSmith:
:+1: 
The man-page formatting isn't great, but I'm fine with leaving that for future work.

PR #925, opened 2015-04-15
(FACT-610) Make hardwaremodel fact on x64 Windows consistent.

peterhuene:
Most platforms use "x86_64" as the hardware model for 64-bit platforms.
Windows was using "x64" instead.  To make Windows consistent with other
platforms, the "hardwaremodel" will use the "x86_64" identifier, but
will keep "architecture" as x64, since Windows uses "x64" in its nomenclature in
favor of "amd64" and "x86_64".


MikaelSmith:
:+1:

PR #927, opened 2015-04-17
(maint) Fix boost::optional usage for Boost 1.58.

peterhuene:
Boost 1.58 stops treating nullptr_t as a substitute for boost::none, so
remove instances where nullptr was used.  Also changing a few places
where the default initialization syntax was used; let's stick to using
boost::none as it's more semantically clear.


peterhuene:
Reminder to myself to delete the branch on merge.  Should have been pushed to my fork but I accidentally cloned from PL and not from my fork.

PR #928, opened 2015-04-20
(FACT-942) Add acceptance test for external fact directory

whopper:
The Puppet 4 AIO pathing changes exposed a lack of tests
around the default external facts (facts.d) directory. This commit
adds an acceptance test to ensure that external facts placed into
the default facts.d directory are found and resolved by Facter.

### DIFF
--- a/acceptance/tests/external_facts/root_uses_default_external_fact_dir.rb
+++ b/acceptance/tests/external_facts/root_uses_default_external_fact_dir.rb
@@ -1,0 +1,47 @@
+test_name "Root default external facts directory (facts.d) is searched for facts"
+
+require 'facter/acceptance/user_fact_utils'
+extend Facter::Acceptance::UserFactUtils
+
+#
+# This test is intended to ensure that executable external facts placed into the
+# default facts.d directory are properly found and resolved by Facter when run
+# as root.
+#
+
+unix_content = <<EOM
+#!/bin/sh
+echo "external_fact=testvalue"
+EOM
+
+win_content = <<EOM
+@echo off
+echo external_fact=testvalue
+EOM
+
+agents.each do |agent|
+  os_version = on(agent, facter('kernelmajversion')).stdout.chomp.to_f
+  factsd = get_factsd_dir(agent['platform'], os_version)
+  ext = get_external_fact_script_extension(agent['platform'])
+
+  if agent['platform'] =~ /windows/
+    content = win_content
+  else
+    content = unix_content
+  end
+
+  step "Agent #{agent}: setup default external facts directory (facts.d)"
+  on(agent, "mkdir -p '#{factsd}'")
+
+  step "Agent #{agent}: create an executable external fact in default facts.d"
+  ext_fact = File.join(factsd, "external_fact#{ext}")
+  create_remote_file(agent, ext_fact, content)
+  on(agent, "chmod +x #{ext_fact}")
+
+  teardown do
+    on(agent, "rm -f '#{ext_fact}'")
+  end
+
+  step "Agent #{agent}: the external fact should resolve"
+  assert_equal("testvalue", fact_on(agent, "external_fact"))
+end
--- a/lib/schema/facter.yaml
+++ b/lib/schema/facter.yaml
@@ -305,6 +305,16 @@ ec2_userdata:
     caveats: |
         All platforms: `libfacter` must be built with `libcurl` support.
 
+env_windows_installdir:
+    type: string
+    description: Return the path of the directory in which Puppet was installed.
+    resolution: |
+      Windows: This fact is specific to the Windows MSI generated environment, and is
+        set using the `environment.bat` script that configures the runtime environment
+        for all Puppet executables. Please see [the original commit in the puppet_for_the_win repo](https://github.com/puppetlabs/puppet_for_the_win/commit/0cc32c1a09550c13d725b200d3c0cc17d93ec262) for more information.
+    caveats: |
+      This fact is specific to Windows, and will not resolve on any other platform.
+
 facterversion:
     type: string
     description: Return the version of facter.
--- a/acceptance/lib/facter/acceptance/user_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/user_fact_utils.rb
@@ -10,12 +10,12 @@ module Facter
       def get_user_fact_dir(platform, version)
         if platform =~ /windows/
           if version < 6.0
-            'C:/Documents and Settings/All Users/Application Data/PuppetLabs/facter/custom'
+            File.join('C:', 'Documents and Settings', 'All Users', 'Application Data', 'PuppetLabs', 'facter', 'custom')
           else
-            'C:/ProgramData/PuppetLabs/facter/custom'
+            File.join('C:', 'ProgramData', 'PuppetLabs', 'facter', 'custom')
           end
         else
-          '/opt/puppetlabs/facter/custom'
+          File.join('/', 'opt', 'puppetlabs', 'facter', 'custom')
         end
       end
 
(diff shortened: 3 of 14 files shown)


whopper:
I'd like to also add a test for the user external fact dir (`~/.puppetlabs/opt/facter/facts.d`) if possible. Does anyone know off the top of their head if Beaker provides an easy way to run commands as a user other than root? I haven't been able to find a precedent in any Puppet/Facter acceptance tests, and haven't seen much in Beaker's source yet.

acceptance/tests/external_facts/root_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,46 @@
...
+    content = win_content
+  else
+    content = unix_content
+  end
+
+  step "Agent #{agent}: setup default external facts directory (facts.d)"
+  on(agent, "mkdir -p '#{factsd}'")
+
+  step "Agent #{agent}: create an executable external fact in default facts.d"
+  ext_fact = "#{factsd}/external_fact#{ext}"
```
er0ck: should use File.join here, no?

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,53 @@
...
+  on(agent, "chown #{user_name}:#{user_name} /home/#{user_name}")
+
+  step "Agent #{agent}: setup default user external facts directory (facts.d)"
+  os_version = on(agent, facter('kernelmajversion')).stdout.chomp.to_f
+  factsd = get_non_root_factsd_dir(agent['platform'], os_version, user_name)
+  on(agent, "mkdir -p '#{factsd}'")
+
+  step "Agent #{agent}: create an executable external fact in default facts.d"
+  ext = get_external_fact_script_extension(agent['platform'])
+  ext_fact = "#{factsd}/external_fact#{ext}"
```
er0ck: File.join

acceptance/tests/external_facts/root_uses_default_external_fact_dir.rb
whopper: @er0ck good point. I'd based this off of some tests from Puppet, which sadly don't use File.join anywhere either... :(

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,53 @@
...
+  create_remote_file(agent, ext_fact, content)
+  on(agent, "chown #{user_name}:#{user_name} #{ext_fact}")
+  on(agent, "chmod +x #{ext_fact}")
+
+  teardown do
+    on(agent, "rm -f '#{ext_fact}'")
+  end
+
+  step "Agent #{agent}: the external fact should resolve"
+  output = agent.execute("su - #{user_name} -c 'facter external_fact'")
```
er0ck: this runs and passes on ubuntu.   running against windows...
why do we have to use agent.execute here rather than on(agent, "su - blah") ?

whopper: @er0ck so this is actually a workaround for the fact that Beaker doesn't support running commands as a non-root user (see https://tickets.puppetlabs.com/browse/BKR-168). I wanted to test that a user having external facts in their homedir would have those facts resolve correctly, but this seemed to be the only way I could get it to work through Beaker...

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,54 @@
...
+  if agent['platform'] =~ /windows/
+    content = win_content
+  else
+    content = unix_content
+  end
+
+  step "Agent #{agent}: create new user"
+  on(agent, "useradd #{user_name} --home /home/#{user_name}")
+  on(agent, "mkdir -p /home/#{user_name}")
+  on(agent, "chown #{user_name}:#{user_name} /home/#{user_name}")
```
joshcooper: Use puppet to do this:

`on(agent, puppet("resource user #{user_name} ensure=present managehome=true"))`

whopper: @joshcooper hm, but running Facter acceptance standalone means Puppet isn't present. I tried the Puppet route earlier and ran into that

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,54 @@
...
+  create_remote_file(agent, ext_fact, content)
+  on(agent, "chown #{user_name}:#{user_name} #{ext_fact}")
+  on(agent, "chmod +x #{ext_fact}")
+
+  teardown do
+    on(agent, "rm -f '#{ext_fact}'")
+  end
+
+  step "Agent #{agent}: the external fact should resolve"
+  on(agent, "su - #{user_name} -c 'facter external_fact'") do
```
joshcooper: Sadly this won't work on windows as cygwin doesn't support`su`. You could use `runas`, though you'll have to figure out the syntax in cygwin+ssh environment. Also `runas` will require you to specify a password (so you'll need to specify a password when the user account is created, and should likely be randomized each test run).

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
```diff
@@ -0,0 +1,53 @@
...
+
+agents.each do |agent|
+  if agent['platform'] =~ /windows/
+    content = win_content
+  else
+    content = unix_content
+  end
+
+  step "Agent #{agent}: create new user"
+  on(agent, "useradd #{user_name} --home /home/#{user_name}")
```
er0ck: useradd doesn't work in windows either.

whopper:
@joshcooper @er0ck maybe the best route would be to add the test for root's `facts.d` and wait for better Beaker non-root user support? As it stands, it sounds like we'll have to hack in some more workarounds that we'd probably just want to remove once that functionality is added into Beaker.

acceptance/tests/external_facts/user_uses_default_external_fact_dir.rb
joshcooper: Ah, I forgot this was facter...

whopper:
Updated to only push the root external facts test. The user test will probably be going into Puppet.

MikaelSmith:
:+1:

whopper:
Update: in its current state, running acceptance against this PR will be successful in both Windows and *nix.

lib/schema/facter.yaml
```diff
@@ -178,6 +178,12 @@ dhcp_servers:
             type: ip
             description: The DHCP server for the default interface.
 
+dir:
```
MikaelSmith: We should fix this on Windows. @Iristyle it comes in as a result of https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L20.

whopper:
Updated as per https://github.com/puppetlabs/puppet_for_the_win/pull/110, which removes the accidental `dir` fact.

Iristyle:
:+1: from what I know of the bits you're updating here.

er0ck:
:+1: 
 ```              - Test Case Summary for suite 'tests' -
       Total Suite Time: 9.14 seconds
      Average Test Time: 0.83 seconds
              Attempted: 11
                 Passed: 11
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 11
```

er0ck:
anyone know what's up with travis and appveyor here?  the tests work fine for me on windows2012r2 and ubuntu 14

whopper:
@er0ck Travis and appveyor were green until I pushed the update to document the env_windows_installdir fact. They should still be green once they're done running again.

PR #929, opened 2015-04-23
(maint) Remove lucid from build targets

melissa:
Ubuntu 10.04 (Lucid Lynx) goes EoL 2015-04-30, and as such, we no longer
want to be building packages for this platform.

https://lists.ubuntu.com/archives/ubuntu-announce/2015-March/000193.html


melissa:
NOTE: this file needs to be paired down since we will not be delivering standalone facter 3 packages. See https://github.com/puppetlabs/puppet/blob/master/ext/build_defaults.yaml for an example of what should happen. It probably should be cut down more than that...

PR #932, opened 2015-04-28
(maint) Remove debian stable

joshcooper:
Debian 8 jessie was released April 25th, 2015 and became "stable". As
part of this, system ruby was bumped to 2.1, and RbConfig['libdir'] no
longer points at the library location, which prevents native facter from
resolving ruby facts in a non-AIO configuration.

We will officially be adding jessie support to AIO at a later date, and
that change will be done in the AIO pipeline.


haus:
:+1:

PR #934, opened 2015-04-29
(FACT-975) Add ppc64le to processors discovery

sshnaidm:
Adding ppc64le architecture for processors discovery.
It's exactly as ppc64 behaviour.

### DIFF
--- a/lib/facter/util/processor.rb
+++ b/lib/facter/util/processor.rb
@@ -232,7 +232,7 @@ module Processor
           end
         end
 
-      when "ppc64"
+      when "ppc64", "ppc64le"
         File.readlines(cpuinfo).each do |l|
           if l =~ /processor\s+:\s+(\d+)/
             processor_num = $1.to_i
--- a/spec/fixtures/cpuinfo/ppc64le
+++ b/spec/fixtures/cpuinfo/ppc64le
@@ -1,0 +1,15 @@
+processor       : 0
+cpu             : POWER8E (raw), altivec supported
+clock           : 3690.000000MHz
+revision        : 2.1 (pvr 004b 0201)
+
+processor       : 1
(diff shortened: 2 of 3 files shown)


peterhuene:
Hi @sshnaidm.  Thanks for the contribution!  This looks good overall, but is missing a corresponding test to ensure the change is tested.  It looks like we have one for `ppc64` already in `spec/unit/processors/os_spec.rb`, so something along those lines.

peterhuene:
Hi @sshnaidm.  Just checking in to see if you've had any time this past week to add the missing test.  Also, could you sign the [Puppet Labs CLA](http://cla.puppetlabs.com/) when you get a chance?  Thanks!

sshnaidm:
Hi, I'm not so familiar with the stuff, I hope did it right.
Thanks

peterhuene:
Hi @sshnaidm.  Thanks for adding the test!  This looks good, but is missing `spec/fixtures/cpuinfo/ppc64le`.  If you could include this file with the expected contents (i.e. what `/proc/cpuinfo` would have for ppc64le), I think the spec should pass.  There's already one for `spec/fixtures/cpuinfo/ppc64` as a basis for what we expect there.  If you have access to a `ppc64le` machine, the actual contents of `/proc/cpuinfo` would be best.

Also, to run specs locally, you can do `bundle exec rake spec` in facter's source directory (you may need to install the bundle if you haven't already).

Thanks!

sshnaidm:
@peterhuene , thanks for explanation. I've added the spec fixture file.

peterhuene:
Hi @sshnaidm.  Thanks again for the contribution!  This has been merged into the Facter 2.x branch and should appear in the next release in the 2.x series.

PR #935, opened 2015-04-29
(FACT-949) Skip puppetlabs-release in AIO acceptance setup

whopper:
This commit skips installing the puppet-labs release package,
as it is no longer necessary in the AIO world.

In addition, this commit refactors the `install_repos_on`
acceptance helper method to curl the repo_config for the
specified SHA to the SUT's system location (i.e, /etc/yum.repos.d).
See
https://github.com/puppetlabs/puppet/commit/fa945fd799583a3e3daea40bcf94ab21fa362dee
for the equivalent commit in Puppet.


whopper:
@joshcooper pointed out that since puppet-agent currently still includes Ruby facter, `ci:test:aio` will fail as it can't find native Facter's binary. There are a few tickets in flight which will address this.

PR #937, opened 2015-05-06
(FACT-963) Remove pre-suite env setup for AIO

MikaelSmith:
The acceptance tests attempt to install the JSON gem everywhere for all
types of installations. AIO tests do not require them, and the other
testing types are no longer supported (by virtue of them simply not
working and we haven't figured out what to do with them).

Specifically testing is blocked for AIO on Debian 8 because it does not
contain the same libjson-gem that the tests try to install. Remove the
entire environment setup pre-suite step, as with AIO there's no need to
install git, ruby, or any gems.

### DIFF
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,7 +1,6 @@
 {
   :type => 'aio',
   :pre_suite => [
-    'setup/common/00_EnvSetup.rb',
     'setup/aio/pre-suite/010_Install.rb',
   ],
 }


MikaelSmith:
@joshcooper this addressed stable

joshcooper:
I get an acceptance fail in facter#stable unrelated to this PR I think:

```
$ bundle exec rake ci:test:aio SHA=0f7b130300710aa35013602b665b91109af831a6 CONFIG=config/nodes/ubuntu-1404-i386.yaml
...
Errored Tests Cases:
  Test Case tests/facter_json_output_validates.rb reported: #<JSON::Schema::ValidationError: The property '#/' contains additional properties ["disks", "dmi", "identity", "load_averages", "memory", "mountpoints", "networking", "ruby", "ssh"] outside of the schema when none are allowed in schema aac80585-b763-5f1a-a2d1-16678a8de957#>
  Test Case tests/ticket_7039_facter_multiple_facts_one_file.rb reported: #<Beaker::Host::CommandFailure: Host 'j6f96btpe6d76ym.delivery.puppetlabs.net' exited with 1 running:
 env FACTERLIB="/tmp/facter7039.1VJiL3" facter --puppet test_fact1
Last 10 lines of output were:
	about a system.

	If no queries are given, then all facts will be returned.

	Example Queries
	===============

	  facter kernel
	  facter networking.ip
	  facter processors.models.0>
```

MikaelSmith:
Which branch is that SHA from? That looks like it's running native Facter to me.

joshcooper:
oh, I think I used a puppet-agent SHA built from master pipeline! Let me try this again.

PR #939, opened 2015-05-06
(FACT-979) Update Beaker to 2.8

MikaelSmith:
Facter acceptance in AIO fails at
https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Master/view/Facter/job/platform_aio-facter_intn-sys_master/1/
because it's using Beaker 2.2. Update to using Beaker 2.8+ to get a
version of Beaker that supports AIO.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -1,14 +1,8 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-gem "beaker", "~> 2.2.0"
+gem "beaker", "~> 2.8"
 gem 'rake', "~> 10.1.0"
-
-# json-schema does not support windows, so use the 'ruby' platform to exclude it on windows
-platforms :ruby do
-  # json-schema uses multi_json, but chokes with multi_json 1.7.9, so prefer 1.7.7
-  gem "multi_json", "1.7.7", :require => false
-  gem "json-schema", "2.1.1", :require => false
-end
+gem "multi_json", "~> 1.8"
 
 if File.exists? "#{__FILE__}.local"
   eval(File.read("#{__FILE__}.local"), binding)


acceptance/Gemfile
```diff
@@ -1,15 +1,8 @@
...
+gem "beaker", "~> 2.8"
 gem 'rake', "~> 10.1.0"
 
-# json-schema does not support windows, so use the 'ruby' platform to exclude it on windows
-platforms :ruby do
-  # json-schema uses multi_json, but chokes with multi_json 1.7.9, so prefer 1.7.7
-  gem "multi_json", "1.7.7", :require => false
-  gem "json-schema", "2.1.1", :require => false
-end
-
```
joshcooper: I think we still need the json gems to be installed on the coordinator for test like `acceptance/tests/options.rb`

MikaelSmith:
Restored `multi_json`, which works on Windows as well. `json-schema` is no longer needed.

acceptance/Gemfile
```diff
@@ -1,14 +1,8 @@
...
+gem "beaker", "~> 2.8"
 gem 'rake', "~> 10.1.0"
-
-# json-schema does not support windows, so use the 'ruby' platform to exclude it on windows
-platforms :ruby do
-  # json-schema uses multi_json, but chokes with multi_json 1.7.9, so prefer 1.7.7
-  gem "multi_json", "1.7.7", :require => false
-  gem "json-schema", "2.1.1", :require => false
-end
+gem "multi_json", "~> 1.8"
```
MikaelSmith: I think `json` has been built in to Ruby since 1.9.3, so this is likely unnecessary. @haus backs me up on that.

kylog: Originally, I put in `multi_json` because `json-schema` uses it (and we needed to pin `multi_json` because reasons). But we've dropped json-based schemas in facter, so I'd have thought this whole block would now be unneeded.

What's the windows dependency we now have on `multi_json`? I didn't follow that part of the comments.

MikaelSmith: Josh was saying that `tests/options/json.rb` uses `json` locally, so we should keep `multi_json`. I don't think it's really needed, but I was trying to get CI working again first.

PR #940, opened 2015-05-06
Merge 2.x to stable

MikaelSmith:
Stable and 2.x should still be kept in-sync until we release 3.


joshcooper:
looks good, waiting on travis

joshcooper:
I verified we are not testing lucid and f19 in the aio stable pipeline, so removing them from build targets should not cause failures.

PR #943, opened 2015-05-07
(FACT-981) Fix Facter API to allow for stubbing from Puppet module specs

peterhuene:
Enables Puppet module specs to stub parts of Facter's API.

This will allow Puppet modules to write tests against their custom facts.


CMakeLists.txt
```diff
@@ -150,7 +150,10 @@ add_subdirectory(lib)
 add_subdirectory(exe)
 
 # Add test executables for unit testing
-add_test(NAME "library\\ tests" COMMAND libfacter_test)
+add_test(NAME "libfacter\\ tests" COMMAND libfacter_test)
+if (RUBY_FOUND)
+    add_test(NAME "libfacter\\ specs" COMMAND bundle exec rspec -I ${CMAKE_BINARY_DIR}/lib WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/lib")
```
MikaelSmith: You may need to quote this. `libfacter_test` and `facter` resolve to targets, so they can identify the test path.

peterhuene: It doesn't work if the entire thing is quoted (CMake thinks it is the executable).  It finds bundler just fine on the path, but doesn't seem like it can execute it.

peterhuene:
@MikaelSmith Looks like the problem is that ctest finds bundler.bat, but leaves the extension off.  If I manually tell it `bundler.bat` it now at least executes (looking into why it can't find libfacter though).

peterhuene:
@MikaelSmith Okay, this should pass CI now.

MikaelSmith:
Yay!

CMakeLists.txt
```diff
@@ -150,7 +150,14 @@ add_subdirectory(lib)
 add_subdirectory(exe)
 
 # Add test executables for unit testing
-add_test(NAME "library\\ tests" COMMAND libfacter_test)
+add_test(NAME "libfacter\\ tests" COMMAND libfacter_test)
+if (RUBY_FOUND)
+    if (WIN32)
+        add_test(NAME "libfacter\\ specs" COMMAND bundle.bat exec rspec WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}/lib")
```
branan: Automatic Windows searching for `.bat` suffixes on commands doesn't work here? Is this a CMake issue?

README.md
```diff
@@ -146,6 +146,10 @@ For a debug build:
 Test
 ----
 
+If a ruby was found during configuration, execute the following command before running tests:
```
branan: I suspect we could create a `bundle install` target that the `test` target depends on, and avoid the need for a human being to do this step.

CMakeLists.txt
peterhuene: It is a CTest issue.  It finds the .bat file but can't execute it without extension.

branan:
Looks like git rename detection barfed on the spec test file, so I'm gonna need to take a bit more time to see what the actual test changes were there.

README.md
peterhuene: We could, but iirc the test target is like the install target, a bit special and difficult to make it depend on things.

MikaelSmith: There are suggestions at http://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests, but they're kind of clunky.

peterhuene: Yeah, I can live with a little Windows specific thing for now rather than creating a new target or doing some other tomfoolery.

branan: yeah, it's probably not worth any real effort to screw around with.

We might want to consider moving to the `make check` pattern that lets us add other steps/deps before the test target (this is my preferred solution longer-term), but probably not worth it for this PR

MikaelSmith:
You can pull down the code and use `it diff master:lib/spec/unit/cfacter_spec.rb HEAD:lib/spec/unit/facter_spec.rb` to diff locally.

lib/spec/unit/facter_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      Facter.reset
+      expect(Facter.search_path).to eq([])
+    end
+
+    it 'should set external search paths' do
+      Facter.search_external(['foo', 'bar', 'baz'])
+      expect(Facter.search_external_path).to eq(['foo', 'bar', 'baz'])
+    end
+
+    it 'should support stubbing for confine testing' do
```
MikaelSmith: From here on are new tests. Everything earlier are changes for rspec 3.

lib/spec/unit/facter_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+          'bar'
+        end
+      end
+      expect(Facter.value(:test)).to eq('bar')
+    end
+
+    it 'should allow stubbing on which and exec' do
+      Facter::Util::Resolution.expects(:which).with("foo").returns('/usr/bin/foo')
+      Facter::Util::Resolution.expects(:exec).with("foo").returns('bar')
+      expect(Facter::Util::Resolution.which('foo')).to eq('/usr/bin/foo')
```
MikaelSmith: Is there an expectation that stubbing `which` will change the behavior of `exec`?

MikaelSmith:
Other than my question about `which` and `exec`, :+1:

lib/spec/unit/facter_spec.rb
peterhuene: The modules that @hunner demonstrated did not make that assumption; it was mostly for stubbing the fact resolution's calls to `which` itself.  @hunner, do you know if module authors expect to stub `which` and have it affect `exec`?  It seems like that would be unlikely since they'd need to stub the `exec` call anyway to make it "do something correctly".

hunner: I think the stubbing of `which` was related to the way one fact I looked at was using the facter method... custom facts tend to use `Facter::Util::Resolution` stuff (mostly `exec`, and sometimes `which`) because it work well, which means those are what get stubbed in tests.

`which` and `exec` would both be stubbed if the fact is calling them directly, as @peterhuene pointed out, to "do something correctly."

MikaelSmith: :+1:

MikaelSmith:
@branan any more review before I merge?

PR #944, opened 2015-05-12
(FACT-958) Add tests for platform dependent facts

whopper:
This commit adds acceptance tests to verify the resolution of
facts which depend on the platform they are being resolved on.

Facts tested include: os, processors, ruby, networking, identity, kernel

### DIFF
--- a/acceptance/tests/facts/debian.rb
+++ b/acceptance/tests/facts/debian.rb
@@ -1,0 +1,98 @@
+test_name "Facts should resolve as expected in Debian 6 and 7"
+
+#
+# This test is intended to ensure that facts specific to an OS configuration
+# resolve as expected in Debian 6 and 7.
+#
+# Facts tested: os, processors, networking, identity, kernel
+#
+
+confine :to, :platform => /debian-squeeze|debian-wheezy/
+
+agents.each do |agent|
+  os_version = agent['platform'] =~ /squeeze/ ? '6' : '7'
+  if agent['platform'] =~ /x86_64/
+    os_arch     = 'amd64'
+    os_hardware = 'x86_64'
+  else
+    os_arch     = 'i386'
+    os_hardware = 'i686'
+  end
+
+  step "Ensure the OS fact resolves as expected"
+  expected_os = {
+                  'os.architecture'         => os_arch,
+                  'os.distro.codename'      => os_version == '6' ? 'squeeze' : 'wheezy',
+                  'os.distro.description'   => /Debian GNU\/Linux #{os_version}\.\d/,
+                  'os.distro.id'            => 'Debian',
+                  'os.distro.release.full'  => /#{os_version}\.\d/,
+                  'os.distro.release.major' => os_version,
+                  'os.distro.release.minor' => /\d/,
+                  'os.family'               => 'Debian',
+                  'os.hardware'             => os_hardware,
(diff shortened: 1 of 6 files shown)


whopper:
This is a first pass on testing platform specific facts. I've run these through on just about every platform we have configs for with success.

peterhuene:
Nice work :+1:

acceptance/tests/facts/debian.rb
```diff
@@ -0,0 +1,119 @@
...
+                          'processors.physicalcount' => /[1-9]/,
+                          'processors.isa'           => /unknown|#{os_hardware}/,
+                          'processors.models'        => /"Intel\(R\).*"/
+                        }
+
+  expected_processors.each do |fact, value|
+    assert_match(value, fact_on(agent, fact))
+  end
+
+  if os_version == 7
```
whopper: Some of the older platforms (like Debian 6) have an older Ruby by default, and so the Ruby fact doesn't resolve.

acceptance/tests/facts/windows.rb
```diff
@@ -0,0 +1,95 @@
+test_name "Facts should resolve as expected in Windows 2012R2"
+
+#
+# This test is intended to ensure that facts specific to an OS configuration
+# resolve as expected in Windows 2012R2.
+#
+# Facts tested: os, processors, ruby, networking, identity, kernel
+#
+
+confine :to, :platform => "windows-2012"
```
MikaelSmith: Why only 2012?

acceptance/tests/facts/el.rb
```diff
@@ -0,0 +1,112 @@
+test_name "Facts should resolve as expected in EL 6 and 7"
```
MikaelSmith: We still support EL5.

acceptance/tests/facts/windows.rb
whopper: @MikaelSmith should have mentioned - I'm running out the door and ran out of time to get other windows versions up - but they're coming (tomorrow) :)

acceptance/tests/facts/el.rb
whopper: Ah, missed EL5. Should be easy to add in here

acceptance/tests/facts/windows.rb
MikaelSmith: Ah, thanks.

whopper: That being said, I'll put this back into Doing. Was mostly just putting it up for some early review since most of the platforms are present

whopper:
Added testing of el5, win2003R2 and win2008R2. Is it sufficient to test the *r2 Windows platforms?

MikaelSmith:
I don't see enough commonality I'd pull out.

MikaelSmith:
@melissa what OS are we adding soon that this should support?

MikaelSmith:
It might be worth breaking the Ruby facts out into their own test, that can do something different for AIO, or detect the Ruby version on the system and use that.

whopper:
@MikaelSmith Added an AIO-only ruby test. It'll be skipped for anything else.

MikaelSmith:
:+1: You could use `[skip ci]` to avoid the wait.

whopper:
CI skipped!

PR #946, opened 2015-05-14
(FACT-981) Execute spec tests only if bundler is installed.

peterhuene:
This commit only runs spec tests as part of the test target if bundler
can be found on the PATH.

This fixes attempting to run spec tests on packaging machines where the
gem isn't present.


MikaelSmith:
:+1: :+1: :+1:

peterhuene:
Verified specs ran on AppVeyor.

PR #947, opened 2015-05-14
(FACT-96) Deprecate `facter --puppet`

joshcooper:
The `--puppet` option in facter introduces a cyclic dependency between
the two projects, and adds complexity as facter needs to know how to
initialize puppet.

This commit adds a deprecation warning when executing `facter --puppet`
and in viewing facter help. The alternative is to run `puppet facts find`.
It already handles puppet-specific facts, e.g. puppetversion,
custom facts in `$vardir/lib/facter`, and external facts in
`$vardir/facts.d`.

Note the deprecation warning is not triggered if code loads facter as a
library and calls `load_puppet` directly. Trying to account for that
case seems like more trouble than it's worth.

### DIFF
--- a/lib/facter/application.rb
+++ b/lib/facter/application.rb
@@ -160,7 +160,10 @@ OPTIONS
                 "Enable timing.") { |v| Facter.timing(1) }
         opts.on("-p",
                 "--puppet",
-                "Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.") { |v| load_puppet }
+                "Deprecated: Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.") do |v|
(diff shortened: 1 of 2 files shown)


kylog:
In discussion on the Jira ticket, I raised the concern that issuing the deprecation per command-line invocation might get spammy, especially since we know of sites putting `facter --puppet` in, say, cron jobs (e.g. PE does this). @nfagerlund had the idea of displaying the deprecation *only* when running help, but no per-command line invocation.

Personally I'd still favor that approach. Thoughts?

joshcooper:
@kylog ah, missed that comment in the ticket. I'll update the PR. Also should this be targeted at 2.x (and merged up to stable)?

joshcooper:
closing to retarget 2.x

PR #948, opened 2015-05-15
(FACT-96) Deprecate `facter --puppet`

joshcooper:
The `--puppet` option in facter introduces a cyclic dependency between
the two projects, and adds complexity, as facter needs to know how to
initialize puppet.

This commit causes `facter [-h|--help]` to mention that the `--puppet`
option is deprecated. It does not emit a deprecation warning when
running `facter --puppet` because that command is run often on every
agent. It also adds tests that the warning is printed.

The recommended way to retrieve puppet-specific facts is to run
`puppet facts find`, which includes `puppetversion`, and pluginsync'ed
custom and external facts.

### DIFF
--- a/lib/facter/application.rb
+++ b/lib/facter/application.rb
@@ -160,7 +160,9 @@ OPTIONS
                 "Enable timing.") { |v| Facter.timing(1) }
         opts.on("-p",
                 "--puppet",
-                "Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.") { |v| load_puppet }
+                "(Deprecated: use `puppet facts` instead) Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.") do |v|
+          load_puppet
+        end
 
         opts.on_tail("-v",
                      "--version",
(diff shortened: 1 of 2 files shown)


joshcooper:
Supercedes PR #947

joshcooper:
ping @kylog @nfagerlund 

```
$ bundle exec facter -h
...
OPTIONS
-------
    -y, --yaml                       Emit facts in YAML format.
    -j, --json                       Emit facts in JSON format.
        --plaintext                  Emit facts in plaintext format.
        --trace                      Enable backtraces.
        --external-dir DIR           The directory to use for external facts.
        --no-external-dir            Turn off external facts.
    -d, --debug                      Enable debugging.
    -t, --timing                     Enable timing.
    -p, --puppet                     (Deprecated: use `puppet facts` instead) Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.
```

joshcooper:
Note to reviewer, please merge up to facter#stable only.

PR #949, opened 2015-05-18
(FACT-1006) Add Cisco NXOS support

ScottGarman:
Patch authored by Mike Wiebe <mwiebe@cisco.com>

### DIFF
--- a/lib/facter/operatingsystem/ciscolinux.rb
+++ b/lib/facter/operatingsystem/ciscolinux.rb
@@ -1,0 +1,34 @@
+require 'facter/util/ciscoos'
+require 'facter/operatingsystem/linux'
+
+module Facter
+  module Operatingsystem
+    class CiscoLinux < Linux
+
+      def initialize(file)
+        @cisco_info = Facter::Util::CiscoOS.cisco_release(file)
+      end
+
+      def get_operatingsystem
+        @operatingsystem ||= @cisco_info['ID']
+        @operatingsystem
+      end
+
+      def get_osfamily
+        @osfamily ||= @cisco_info['ID_LIKE']
+        @osfamily
+      end
+
+      def get_operatingsystemrelease
+        @operatingsystemrelease ||= @cisco_info['VERSION']
+        @operatingsystemrelease
+      end
+
+      def get_operatingsystemmajorrelease 
+        if operatingsystemrelease = get_operatingsystemrelease
+          operatingsystemrelease.split("(").first
+        end
+      end
+    end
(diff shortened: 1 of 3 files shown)


ScottGarman:
Cisco tells me they have signed the CLA already, btw.

lib/facter/util/ciscoos.rb
```diff
@@ -0,0 +1,22 @@
+module Facter
+  module Util
+    module CiscoOS 
+
+      # @file follows the same format as os-release
+      # @see http://www.freedesktop.org/software/systemd/man/os-release.html
+      def self.cisco_release(file)
```
peterhuene: Do we need a cisco-specific implementation for reading a release file?  We read release files for other operating systems, so is there something specific to this implementation that we need for this to work?

peterhuene:
@ScottGarman Thanks for the contribution!  I think we're on board to merging this into the 2.x branch, so could you retarget this PR there?  

Generally this looks good but we'd like to see some tests added with a fixture for detecting the OS properly.  We'd like to see that added before we can merge it.

I'll file a ticket to have this ported to Facter 3.  Thanks!

lib/facter/util/ciscoos.rb
ScottGarman: @peterhuene is there a standard for this I should point them to? I'll note also that the release file for NXOS is multiline - here's an example:

ID=nexus
ID_LIKE=wrlinux
NAME=Nexus
HOME_URL=http://www.cisco.com
BUILD_ID=TBD
CISCO_RELEASE_INFO=/etc/os-release

lib/facter/operatingsystem/ciscolinux.rb
```diff
@@ -0,0 +1,34 @@
...
+        @operatingsystemrelease ||= @cisco_info['VERSION']
+        @operatingsystemrelease
+      end
+
+      def get_operatingsystemmajorrelease 
+        if operatingsystemrelease = get_operatingsystemrelease
+          operatingsystemrelease.split("(").first
+        end
+      end
+    end
```
MikaelSmith: Omitting the definition of get_operatingsystemminorrelease lets this fall-through to the Linux implementation at https://github.com/puppetlabs/facter/blob/2.x/lib/facter/operatingsystem/linux.rb#L103. If the version string contains a `.`, that implementation will return a non-nil value. I'm not sure if that's the intent here.

peterhuene:
Closing now that a reference implementation is in place.

PR #950, opened 2015-05-18
(FACT-628) Move LDOM virtual implementation to virtual.rb

branan:
Previously, the implementation of the `virtual` fact that evaluated
LDOM info was stored in `ldom.rb`. This would cause the value of
`virtual` to be different depending on whether or not ldom.rb was
loaded.

This moves the implementation of the virtual fact to virtual.rb, and
weights it appropriately within the set of implementations there.

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -90,6 +90,17 @@ Facter.add("virtual") do
 end
 
 Facter.add("virtual") do
+  confine :kernel => 'SunOS'
+  confine :hardwareisa => 'sparc'
+  confine :ldom_domainrole_control => 'false'
+  has_weight 20 # Weight this before the default solaris implementation
+  setcode do
+    Facter.value(:ldom_domainrole_impl)
+  end
+end
+
+
+Facter.add("virtual") do
   confine :kernel => 'HP-UX'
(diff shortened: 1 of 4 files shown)


branan:
I've tested this on SOL 10 LDOM and on my linux machine. I want to play around with other virt systems on Solaris before we merge to make sure it's all weighted correctly and is behving as expected.

branan:
working on specs

kylog:
Makes sense.

branan:
I'm pretty unhappy with all the stubbing I had to do in order to get this to run as expected in specs. Does this look normal/correct to folks with more experience writing facter tests, or is it an indication that something could be weird somewhere?

lib/facter/virtual.rb
```diff
@@ -90,6 +90,17 @@
 end
 
 Facter.add("virtual") do
+  confine :kernel => 'SunOS'
```
MikaelSmith: Was this necessary, or just re-organization to put it with other definitions of virtual?

MikaelSmith:
No better ideas for the spec test.

kylog:
Yah, :+1: 

Ruby facter spec tests are a mock fest.

PR #951, opened 2015-05-19
(maint) Update for GCC 4.9.2

MikaelSmith:
- Fix signed/unsigned comparisons
- Update Windows build dependencies

Also updates other build packages.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,13 +3,13 @@ clone_depth: 10
 install:
   - git submodule update --init --recursive
 
-  - choco install mingw-w64 -y -Version 4.8.3 -source https://www.myget.org/F/puppetlabs
+  - choco install mingw-w64 -y -Version 4.9.2 -source https://www.myget.org/F/puppetlabs
   - SET PATH=C:\Ruby21-x64\bin;C:\tools\mingw64\bin;%PATH%
 
-  - ps: wget 'https://s3.amazonaws.com/kylo-pl-bucket/boost_1_55_0-x86_64_mingw-w64_4.8.3_posix_seh.7z' -OutFile "$pwd\boost.7z"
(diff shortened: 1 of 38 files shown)


peterhuene:
o.O with regards to those warnings; every one is a literal that can promoted to unsigned without any possibility of an incorrect comparison.  Also, I was building with GCC 4.9 on Arch for quite some time; is this warning coming only from Windows builds?

MikaelSmith:
I was only testing on Windows, but I agree it seems overly pedantic.

MikaelSmith:
I'll try a Linux build with 4.9.2 to compare.

MikaelSmith:
I could also disable the warning.

peterhuene:
Seems like a useful warning generally (especially when not comparing to positive signed literals), so I'm hesitant to disable it.  I can live with fixing up the warnings; curious as to why I wasn't seeing it on Linux though; perhaps it was changed in a minor release?

MikaelSmith:
@stahnma should I relabel this [RE-4468](https://tickets.puppetlabs.com/browse/RE-4468)?

peterhuene:
Should I merge or wait for response from @stahnma?

kylog:
Please don't change the first token. The jira/git reconciliation script only knows about maint/docs/packaging and the project name (here FACT). So if there's some request to include an out-of-project jira reference, just put it after the first token.

Also AFAIK RE doesn't do any automated harvesting of RE ticket #s from git commits.

MikaelSmith:
@haus isn't completely against it

kylog:
That clarified nothing for me :)

MikaelSmith:
I think wait for a response. I may rework this into two PRs, one for Boost/git/ruby, another for GCC and yaml-cpp.

MikaelSmith:
Closed in favor of #954 until RelEng is ready to advance to GCC 4.9.2.

PR #952, opened 2015-05-21
(FACT-892) Change "POSIX" load average resolver to "glib".

peterhuene:
The base load average resolver implementation is currently building as
part of the POSIX resolvers, but it calls getloadavg, which is not a
POSIX function.  It is actually a glib function.

This breaks the build on systems where "getloadavg" isn't part of the
standard library implementation, such as AIX.

This moves the resolver into a "glib" namespace and adds the resolver to
instances where a conforming glib is present (OSX, BSD, Linux, and
Solaris).

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -96,7 +96,6 @@ if (UNIX)
         "src/facts/posix/collection.cc"
         "src/facts/posix/identity_resolver.cc"
         "src/facts/posix/kernel_resolver.cc"
-        "src/facts/posix/load_average_resolver.cc"
         "src/facts/posix/networking_resolver.cc"
         "src/facts/posix/operatingsystem_resolver.cc"
         "src/facts/posix/processor_resolver.cc"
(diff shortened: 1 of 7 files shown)


branan:
This looks sane, and I should be able to rebase my AIX work on top of it with no problems. Waiting on tests

peterhuene:
I've also tested it on Solaris.  I didn't have a non-OSX BSD to test it out on, but from the documentation it should "work".

branan:
@peterhuene to be clear, "implement load_average on solaris" is definitely a ticket on its own now, as well as for AIX?

peterhuene:
Which ticket?  I didn't see anything linked in FACT-892.

branan:
@peterhuene hmm, @whopper thought the resolver was busted on Solaris as well as on AIX. Is this possibly a solaris 10 vs solaris 11 thing?

peterhuene:
The build was busted on Solaris previously because it was building the "POSIX" resolver, but didn't have it in the collection.  This fixes the latter.

branan:
Ah, that makes more sense. Cool

PR #953, opened 2015-05-21
Initial AIX support with kernel/operatingsystem facts

branan:
This is the beginning of some AIX stuff.

This still includes a bit of a hacky workaround for dlopen flags on AIX, so probably shouldn't be merged just yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,11 @@ if (MINGW)
     # MinGW crashes with large pre-compiled headers; ours definitely exceeds the limit.
     # See http://stackoverflow.com/questions/10841306/cc1plus-exe-crash-when-using-large-precompiled-header-file
     set(PRECOMPILED_HEADERS FALSE)
+elseif("${CMAKE_SYSTEM_NAME}" MATCHES "AIX")
+    # On AIX, precompiled headers are causing crazy crazy failures
+    # with things that need to be extern "C". This seems to clear it
+    # all up.
+    set(PRECOMPILED_HEADERS FALSE)
 else()
     set(PRECOMPILED_HEADERS TRUE)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,11 @@
 cmake_minimum_required(VERSION 2.8.12)
 project(FACTER)
 
+# Set this early, so it's available. AIX gets weird, man.
+if("${CMAKE_SYSTEM_NAME}" MATCHES "AIX")
+    set(AIX TRUE)
+endif()
+
 if (NOT CMAKE_BUILD_TYPE)
     message(STATUS "Defaulting to a release build.")
     set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." FORCE)
@@ -32,7 +37,7 @@ if (MINGW)
     # MinGW crashes with large pre-compiled headers; ours definitely exceeds the limit.
     # See http://stackoverflow.com/questions/10841306/cc1plus-exe-crash-when-using-large-precompiled-header-file
     set(PRECOMPILED_HEADERS FALSE)
-elseif("${CMAKE_SYSTEM_NAME}" MATCHES "AIX")
+elseif(AIX)
     # On AIX, precompiled headers are causing crazy crazy failures
     # with things that need to be extern "C". This seems to clear it
     # all up.
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -133,9 +133,10 @@ if (WIN32)
 endif()
 
 # Set the platform-specific sources
-if ("${CMAKE_SYSTEM_NAME}" MATCHES "AIX")
+if (AIX)
     set(LIBFACTER_PLATFORM_SOURCES
         "src/facts/aix/collection.cc"
+        "src/facts/aix/operating_system_resolver.cc"
     )
 elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
     set(LIBFACTER_PLATFORM_SOURCES
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -133,7 +133,11 @@ if (WIN32)
 endif()
 
 # Set the platform-specific sources
-if ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
+if ("${CMAKE_SYSTEM_NAME}" MATCHES "AIX")
+    set(LIBFACTER_PLATFORM_SOURCES
+        "src/facts/aix/collection.cc"
+    )
+elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
     set(LIBFACTER_PLATFORM_SOURCES
         "src/facts/bsd/filesystem_resolver.cc"
         "src/facts/bsd/networking_resolver.cc"
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -95,7 +95,6 @@ if (UNIX)
         "src/execution/posix/execution.cc"
         "src/facts/posix/collection.cc"
         "src/facts/posix/identity_resolver.cc"
-        "src/facts/posix/kernel_resolver.cc"
         "src/facts/posix/networking_resolver.cc"
         "src/facts/posix/operatingsystem_resolver.cc"
         "src/facts/posix/processor_resolver.cc"
@@ -110,6 +109,13 @@ if (UNIX)
     )
     if (OPENSSL_FOUND)
         set(LIBFACTER_STANDARD_SOURCES ${LIBFACTER_STANDARD_SOURCES} "src/util/posix/scoped_bio.cc")
+    endif()
+
+    if (NOT AIX)
+        set(LIBFACTER_STANDARD_SOURCES
+            ${LIBFACTER_STANDARD_SOURCES}
+            "src/facts/posix/kernel_resolver.cc"
+        )
     endif()
 
     set(POSIX_LIBRARIES pthread dl)
@@ -136,6 +142,7 @@ endif()
 if (AIX)
     set(LIBFACTER_PLATFORM_SOURCES
         "src/facts/aix/collection.cc"
+        "src/facts/aix/kernel_resolver.cc"
         "src/facts/aix/operating_system_resolver.cc"
     )
 elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin")
--- a/lib/inc/internal/facts/aix/kernel_resolver.hpp
+++ b/lib/inc/internal/facts/aix/kernel_resolver.hpp
@@ -1,17 +1,17 @@
 /**
  * @file
- * Declares the aix operating system fact resolver.
+ * Declares the AIX kernel fact resolver.
  */
 #pragma once
 
-#include "../posix/operating_system_resolver.hpp"
+#include "../resolvers/kernel_resolver.hpp"
 
 namespace facter { namespace facts { namespace aix {
 
     /**
-     * Responsible for resolving operating system facts.
+     * Responsible for resolving kernel facts.
      */
-    struct operating_system_resolver : posix::operating_system_resolver
+    struct kernel_resolver : resolvers::kernel_resolver
     {
      protected:
         /**
@@ -21,4 +21,5 @@ namespace facter { namespace facts { namespace aix {
          */
         virtual data collect_data(collection& facts) override;
     };
+
 }}}  // namespace facter::facts::aix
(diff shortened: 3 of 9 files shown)


lib/src/facts/aix/kernel_resolver.cc
```diff
@@ -0,0 +1,25 @@
...
+        data result;
+        struct utsname name;
+        if (uname(&name) == -1) {
+            LOG_WARNING("uname failed: %1% (%2%): kernel facts are unavailable.", strerror(errno), errno);
+            return result;
+        }
+
+        result.name = name.sysname;
+        result.release = name.release;
+        result.version = name.version;
```
branan: This only differs from the POSIX resolver by one line. Thoughts on making that an ifdef instead of having an entirely new file just for AIX?

peterhuene: A one line `#ifdef` is fine by me; it doesn't sound like that would grow into #ifdef hell.

MikaelSmith:
Lint is unhappy.

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -0,0 +1,68 @@
...
+
+    operating_system_resolver::data operating_system_resolver::collect_data(collection& facts)
+    {
+        // Default to the base implementation
+        auto result = posix::operating_system_resolver::collect_data(facts);
+        result.architecture = getattr("proc0", "type");
+        result.hardware = getattr("sys0", "modelname");
+
+        // Get OS release stuff
+        static boost::regex regexp_release("(\\d+)\\.(\\d+\\.\\d+\\.\\d+)");
```
whopper: @branan did we decide to just report the `6.1` from the `6.1.7.15` found in `aix_release.level`? If so, we can probably just do something like:

    static boost::regex regexp_release("(\\d+\\.\\d+)\\.\\d+\\.\\d+");
        file::each_line("/usr/lpp/bos/aix_release.level", [&](string& line) {
            string full_release;
            if (re_search(line, regexp_release, &full_release)) {
                result.release = full_release;
                return false;
            }
            return true;
        });
        return result;

No need to concatenate `major + '.' + minor`, as those get parsed in the `parse_release` function below.

CMakeLists.txt
```diff
@@ -32,6 +37,11 @@ if (MINGW)
     # MinGW crashes with large pre-compiled headers; ours definitely exceeds the limit.
     # See http://stackoverflow.com/questions/10841306/cc1plus-exe-crash-when-using-large-precompiled-header-file
     set(PRECOMPILED_HEADERS FALSE)
+elseif(AIX)
+    # On AIX, precompiled headers are causing crazy crazy failures
+    # with things that need to be extern "C". This seems to clear it
+    # all up.
+    set(PRECOMPILED_HEADERS FALSE)
```
MikaelSmith: No complaints here, pre-compiled headers are a hack.

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -0,0 +1,68 @@
...
+
+using namespace std;
+using namespace facter::util;
+namespace execution = facter::execution;
+
+string getattr(string object, string field)
+{
+    string result;
+
+    // To get architecture, can/should we use getconf KERNEL_BITMODE, getconf HARDWARE_BITMODE?
```
MikaelSmith: Get any answers to this?

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -0,0 +1,68 @@
...
+    execution::each_line(
+        "/usr/sbin/lsattr", {"-El", object, "-a", field},
+        [&](string& line) {
+            if (!line.empty()) {
+                vector<string> tokens;
+                istringstream iss(line);
+                // istream_iterator is a great way to tokenize your life
+                copy(istream_iterator<string>(iss),
```
MikaelSmith: Is this something we should use more generally, and put in a utility? I think other places we just use something like `boost::split(tokens, line, boost::is_space(), boost::token_compress_on)`.

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -0,0 +1,68 @@
...
+using namespace facter::util;
+namespace execution = facter::execution;
+
+string getattr(string object, string field)
+{
+    string result;
+
+    // To get architecture, can/should we use getconf KERNEL_BITMODE, getconf HARDWARE_BITMODE?
+    execution::each_line(
+        "/usr/sbin/lsattr", {"-El", object, "-a", field},
```
MikaelSmith: Is it possible to invoke this once to get all the data we need? Shelling out is something we try to minimize.

lib/src/facts/aix/operating_system_resolver.cc
MikaelSmith: And even if not, why do you use a regex that splits it up, just to rejoin them?

lib/src/util/posix/dynamic_library.cc
```diff
@@ -41,7 +41,11 @@ namespace facter { namespace util {
         close();
 
         // Don't actually perform a load to determine if it is already loaded
+#ifdef _AIX
+        _handle = dlopen(name.c_str(), RTLD_LAZY);
```
MikaelSmith: Why was this necessary?

lib/src/facts/aix/operating_system_resolver.cc
whopper: Heh, the rejoining bit was copied straight out of the Solaris resolver where we had to do that because some versions needed `major + '_u' + minor` and others needed `major + '.' + minor`. Looks like we just forgot to update this to get rid of that.

lib/src/util/posix/dynamic_library.cc
kylog: https://tickets.puppetlabs.com/browse/FACT-891 has backstory.

lib/src/facts/aix/operating_system_resolver.cc
whopper: @MikaelSmith I think the answer is no, looking at what `lsattr` and `getconf` return:
    getconf KERNEL_BITMODE
    64

    getconf HARDWARE_BITMODe
    64

    lsattr -El proc0 -a type
    type PowerPC_POWER7 Processor type False

    lsattr -El sys0 -a modelname
    modelname IBM,8231-E1D Machine name False

Although I can't entirely recall how we're determining the actual architecture with lsattr. If we're reporting POWERPC..., it isn't consistent with other platforms that just return `x86_64` or `i386`

lib/src/util/posix/dynamic_library.cc
MikaelSmith: Thanks, hadn't made the connection.

lib/src/facts/aix/operating_system_resolver.cc
whopper: It looks like it might not be possible, based on the `lsattr` man page. The `-E` option displays effective values for a specific 'Customized Device' object class (specified with `-l`). Since we're checking both `proc0` and `sys0`, I'm not sure if we can get both to return in one call.

MikaelSmith: Ok.

lib/src/util/posix/dynamic_library.cc
whopper: I believe this was @branan's workaround for FACT-891, and without this, Facter won't compile on AIX. He mentioned earlier that he was debating whether to leave this hack in until we fix FACT-891, so we don't push a commit that doesn't actually build in AIX.

MikaelSmith: Sounds fine to me; I'd probably prefer a comment explicitly linking it to FACT-891.

branan: Absolutely. The status of this hack is the main reason this is marked do
not merge.

I'll move it to its own commit and comment that its a dirty hack

lib/src/facts/aix/operating_system_resolver.cc
branan: I think this is consistent-ish. If the goal of the architecture fact is to
be able to say "what binaries can I drop on this system", I think this is a
reasonable value.

Of course, we don't return feature info (such as SSE support) on x86, so
maybe " power " is enough here

lib/src/util/posix/dynamic_library.cc
kylog: +1 on a discrete commit for this workaround. But if it's accompanied by an explanatory comment (something about "Workaround pending resolution of FACT-891"), this seems reasonable enough. 

And then if we end up living with some of these `ifdef`'s long-term for whatever reason, it would be nice to segregate them in some platform driven headers.

lib/src/facts/aix/operating_system_resolver.cc
branan: We should probably do a quick error-check on the output to ensure we got the number of tokens we expected.

I used the istream_iterator because that's how I'm used to tokenizing. boost::split is probably better. I'll fix this up.

branan:
Just pushed up a big commit to refactor the kernel/os facts back into values similar to what Facter 2 returned. It's kind of messy, I might try to refactor a bit before this gets merged.

branan:
I think this is ready for final review now

lib/CMakeLists.txt
```diff
@@ -112,6 +111,13 @@ if (UNIX)
         set(LIBFACTER_STANDARD_SOURCES ${LIBFACTER_STANDARD_SOURCES} "src/util/posix/scoped_bio.cc")
     endif()
 
+    if (NOT AIX)
```
MikaelSmith: So this still works on Solaris and BSD, just not AIX.

lib/inc/internal/facts/aix/kernel_resolver.hpp
```diff
@@ -0,0 +1,25 @@
+/**
+ * @file
+ * Declares the POSIX kernel fact resolver.
```
MikaelSmith: Description needs an update.

lib/src/facts/aix/kernel_resolver.cc
```diff
@@ -0,0 +1,70 @@
...
+        vector<string> versions;
+
+        // We reimplement part of oslevel here. We parse the rml cache
+        // and use that to determine the full AIX system version. If
+        // the rml cache doesn't exist, we just call out to oslevel to
+        // regenerate it (since we're gonna be slow in that case
+        // anyway).
+        bool success;
+        bool tried_oslevel = false;
+        do {
```
MikaelSmith: This is just 3 steps:
* Try to read .oslevel.rml.cache
* If that fails, run oslevel
* Try reading .oslevel.rml.cache again

It seems simpler to just have a function that reads .oslevel.rml.cache and just have the 3 steps explicitly. The loop complicates things.

lib/CMakeLists.txt
branan: Yeah. AIX has its own implementation of the kernel resolver that uses `oslevel` instead of uname output :/

lib/src/facts/aix/kernel_resolver.cc
```diff
@@ -0,0 +1,70 @@
...
+                std::tie(oslevel_success, out, err) = execute("/usr/bin/oslevel", {"-s"}, 0, { execution_options::trim_output, execution_options::redirect_stderr_to_stdout, execution_options::merge_environment });
+                if (!oslevel_success) {
+                    LOG_WARNING("oslevel failed: %1%: kernel facts are unavailable", out);
+                    return result;
+                }
+                tried_oslevel = true;
+            }
+        } while (!success);
+
+        sort(versions.rbegin(), versions.rend());
```
MikaelSmith: If we just want the largest number, we could keep track of that when reading the file rather than storing all of them and sorting.

lib/src/facts/aix/operating_system_resolver.cc
MikaelSmith: Should still update the comment.

lib/src/facts/aix/operating_system_resolver.cc
MikaelSmith: `boost::split`

lib/src/facts/aix/kernel_resolver.cc
branan: Yeah. I think I originally had this doing multiple retries and decided that was stupid :)

lib/src/facts/aix/kernel_resolver.cc
branan: This code is from Whopper, and I didn't dig into it in too much depth. It's probably a pretty faithful recreation of the bash, which doesn't have such niceties. I'll massage it a bit more

branan:
@MikaelSmith pulled RML parsing out to a new method and fixed it to not iterate twice.

MikaelSmith:
:+1:

PR #955, opened 2015-05-26
(FACT-1002) Add tests for custom facts in FACTERLIB and LOAD_PATH

whopper:
This commit adds two new custom fact acceptance tests. The first
ensures that facts in a directory specified by the FACTERLIB
environment variable are resolved, and the second ensures that
facts in a Ruby $LOAD_PATH/facter directory are resolved.

### DIFF
--- a/acceptance/tests/options/custom_facts.rb
+++ b/acceptance/tests/options/custom_facts.rb
@@ -1,5 +1,8 @@
 test_name "custom fact commandline options (--no-custom-facts and --custom-dir)"
 
+require 'puppet/acceptance/common_utils'
+extend Puppet::Acceptance::CommandUtils
+
 require 'facter/acceptance/user_fact_utils'
 extend Facter::Acceptance::UserFactUtils
 
@@ -9,6 +12,9 @@ extend Facter::Acceptance::UserFactUtils
 # custom fact load path will not execute when the `--no-custom-facts` option is passed.
 # The second step checks that a custom fact in a directory specified by the `--custom-dir`
 # option is found by Facter and resolved.
+#
+# The second set of tests are intended to ensure that custom facts located in FACTERLIB
+# or $LOAD_PATH directories are resolved.
 #
 
 content = <<EOM
@@ -23,15 +29,11 @@ agents.each do |agent|
 
   custom_dir = get_user_fact_dir(agent['platform'], on(agent, facter('kernelmajversion')).stdout.chomp.to_f)
 
-  step "Agent #{agent}: create custom fact directory and executable custom fact"
+  step "Agent #{agent}: create custom fact directory and custom fact"
   on(agent, "mkdir -p '#{custom_dir}'")
   custom_fact = File.join(custom_dir, 'custom_fact.rb')
   create_remote_file(agent, custom_fact, content)
-  on(agent, "chmod +x '#{custom_fact}'")
 
-  teardown do
-    on(agent, "rm -f '#{custom_fact}'")
-  end
 
   step "--no-custom-facts option should disable custom facts"
   on(agent, facter("--no-custom-facts custom_fact")) do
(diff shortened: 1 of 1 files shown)


acceptance/tests/options/custom_facts.rb
```diff
@@ -42,4 +45,32 @@
...
+  facterlib_dir = agent.tmpdir('arbitrary_dir')
+  custom_fact = File.join(facterlib_dir, 'custom_fact.rb')
+  create_remote_file(agent, custom_fact, content)
+  on(agent, "chmod +x '#{custom_fact}'")
+
+  on(agent, facter('custom_fact', :environment => { 'FACTERLIB' => facterlib_dir })) do
+    assert_equal("testvalue", stdout.chomp, "Custom fact output does not match expected output")
+  end
+
+  on(agent, "rm -rf '#{facterlib_dir}'")
```
er0ck: suggest deleting #{custom_fact} here as well so that we know we're resolving the correct custom fact in the next step/test.

acceptance/tests/options/custom_facts.rb
```diff
@@ -42,4 +45,32 @@
...
+  on(agent, "rm -f '#{custom_fact}'")
+
+  step "Agent #{agent}: ensure custom facts in $FACTERLIB resolve"
+  facterlib_dir = agent.tmpdir('arbitrary_dir')
+  custom_fact = File.join(facterlib_dir, 'custom_fact.rb')
+  create_remote_file(agent, custom_fact, content)
+  on(agent, "chmod +x '#{custom_fact}'")
+
+  on(agent, facter('custom_fact', :environment => { 'FACTERLIB' => facterlib_dir })) do
+    assert_equal("testvalue", stdout.chomp, "Custom fact output does not match expected output")
```
er0ck: should differentiate the helper string here so it's easier to find which test failed.

acceptance/tests/options/custom_facts.rb
```diff
@@ -42,4 +45,32 @@
...
+  step "Agent #{agent}: ensure custom facts in $LOAD_PATH resolve"
+  on(agent, "#{ruby_command(agent)} -e 'puts $LOAD_PATH[0]'")
+  load_path_facter_dir = File.join(stdout.chomp, 'facter')
+  on(agent, "mkdir -p \"#{load_path_facter_dir}\"")
+  custom_fact = File.join(load_path_facter_dir, 'custom_fact.rb')
+  create_remote_file(agent, custom_fact, content)
+  on(agent, "chmod +x '#{custom_fact}'")
+
+  on(agent, facter("custom_fact")) do
+    assert_equal("testvalue", stdout.chomp, "Custom fact output does not match expected output")
```
er0ck: differentiate this assertion output string also.

acceptance/tests/options/custom_facts.rb
```diff
@@ -42,4 +45,32 @@
...
+  end
+
+  on(agent, "rm -rf '#{load_path_facter_dir}'")
```
MikaelSmith: Are there any circumstances where LOAD_PATH[0]/facter would already exist, and be something we don't want to delete?

acceptance/tests/options/custom_facts.rb
```diff
@@ -29,9 +35,6 @@
   create_remote_file(agent, custom_fact, content)
   on(agent, "chmod +x '#{custom_fact}'")
```
MikaelSmith: Do custom facts need to be executable? We're not running them, we're loading them in Ruby.

acceptance/tests/options/custom_facts.rb
whopper: @er0ck ah, I decided to actually just recursively blow away the entire `facterlib_dir` which contains #{custom_fact}, as we don't need it for anything else in any of these tests.

acceptance/tests/options/custom_facts.rb
whopper: Doh, good point!

MikaelSmith:
Also use `[skip ci]` if you update so we don't need to wait on CI.

acceptance/tests/options/custom_facts.rb
whopper: Right, force of habit from external facts. Removing the `chmod`s :)

acceptance/tests/options/custom_facts.rb
whopper: @MikaelSmith hm, good point. Currently it won't hurt anything, and I'd hope we don't have a test depending on it in the future that doesn't create it itself.

I've looked at trying to just append another directory to the $LOAD_PATH of the agent without much success. It *seems* like `ruby -I path` should do it, but it hangs indefinitely...

MikaelSmith:
:+1:

acceptance/tests/options/custom_facts.rb
MikaelSmith: Can probably live with that.

PR #956, opened 2015-05-26
(FACT-1018) Use Ruby from CMake for tests

MikaelSmith:
Prior to this commit, tests would fail if cmake was invoked with Ruby
on the PATH, but `make test` wasn't.

Fix by finding the binary related to the Ruby environment custom facts
execute.


MikaelSmith:
This is another reason to create a `make check` target, but I don't expect us to add more tests that expect a Ruby executable.

haus:
:+1:

PR #957, opened 2015-05-27
(FACT-982) Add --no-ruby command-line option

whopper:
This commit adds a new command-line option: --no-ruby. This
option disables Ruby from being loaded, and thus disables
the Ruby loading warning when no Ruby is present. In addition,
this option disables custom facts and built-in facts which
require Ruby.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -123,6 +123,7 @@ int main(int argc, char **argv)
             ("no-color", "Disables color output.")
             ("no-custom-facts", "Disables custom facts.")
             ("no-external-facts", "Disables external facts.")
+            ("no-ruby", "Disables loading Ruby, facts requiring Ruby, and custom facts.")
             ("trace", "Enable backtraces for custom facts.")
             ("verbose", "Enable verbose (info) output.")
             ("version,v", "Print the version and exit.")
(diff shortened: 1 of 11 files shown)


whopper:
This is my first pass at disabling all things ruby with a `--no-ruby` option. This approach feels a little hackish, but the best way to do this seems to be to not add the ruby resolver into the common collection if the option is specified. Is there a better way to propagate the option into the ruby resolver than passing it between each call?

exe/facter.cc
```diff
@@ -123,6 +123,7 @@ int main(int argc, char **argv)
             ("no-color", "Disables color output.")
             ("no-custom-facts", "Disables custom facts.")
             ("no-external-facts", "Disables external facts.")
+            ("no-ruby", "Disables loading Ruby and facts requiring Ruby.")
```
peterhuene: Can this mention that it implies `--no-custom-facts`?

man/man8/facter.8
```diff
@@ -41,6 +41,7 @@ If no queries are given, then all facts will be returned\.
       \fB\-\-no-color\fR                   Disables color output\.
       \fB\-\-no-custom-fact\fR             Disables custom facts\.
       \fB\-\-no-external-facts\fR          Disables external facts\.
+      \fB\-\-no-ruby\fR                    Disables loading Ruby and facts requiring Ruby\.
```
peterhuene: Same comment as the above option documentation.

exe/facter.cc
```diff
@@ -202,7 +206,12 @@ int main(int argc, char **argv)
         log_command_line(argc, argv);
 
         // Initialize Ruby in main
-        bool ruby = facter::ruby::initialize(vm.count("trace") == 1);
+        bool ruby;
+        if (vm.count("no-ruby")) {
+            ruby = false;
+        } else {
+            ruby = facter::ruby::initialize(vm.count("trace") == 1);
```
peterhuene: Can these few lines be expressed as simply as `ruby = (vm.count("no-ruby") == 0) && facter::ruby::initialize(vm.count("trace") == 1);`?

peterhuene:
Other than the above, :+1:

whopper:
@peterhuene good call on the logic simplification there :). Updated the docs too - how's `"Disables loading Ruby, facts requiring Ruby, and custom facts."` sound?

peterhuene:
:+1:

PR #958, opened 2015-05-28
(FACT-976) Report correct OS for Cisco NXOS

MikaelSmith:
Native facter port of https://github.com/puppetlabs/facter/pull/949.

Major design decisions:
* I steered away from was extending `resolvers::operating_system_resolver` for specific distros, as I didn't want to move deciding which to instantiate to where `operating_system_resolver` is created.
* I decided not to make the implementation object (`os_linux` and children) a member of `linux::operating_system_resolver` because I didn't see any major benefit for keeping the overloaded `parse_release` and `determine_os_family` methods. That functionality was easily pulled into `collect_data` to be handled earlier.
* The distro release parsing in `resolvers::operating_system_resolver` should likely be moved into the Linux implementation, similar to the move for os release. I haven't done so because it's not needed for this work.

### DIFF
--- a/lib/inc/internal/facts/linux/os_cisco.hpp
+++ b/lib/inc/internal/facts/linux/os_cisco.hpp
@@ -52,6 +52,18 @@ namespace facter { namespace facts { namespace linux {
             auto val = _release_info.find("VERSION");
             return (val != _release_info.end()) ? val->second : std::string();
         }
+
+        /**
+         * Parses the release version string to return the major version.
+         * @param name Unused.
+         * @param release Unused.
+         * @return Returns a tuple of the major version and an empty string.
+         */
+        virtual std::tuple<std::string, std::string> parse_release(std::string const& name, std::string const& release) const override
+        {
+            auto pos = release.find('(');
+            return std::make_tuple(release.substr(0, pos), std::string());
+        }
     };
 
 }}}  // namespace facter::facts::linux
--- a/lib/inc/internal/facts/linux/os_cisco.hpp
+++ b/lib/inc/internal/facts/linux/os_cisco.hpp
@@ -1,0 +1,57 @@
+/**
+ * @file
+ * Declares the Cisco Linux operating system query helper.
+ */
+#pragma once
+
+#include <internal/facts/linux/os_linux.hpp>
+
+namespace facter { namespace facts { namespace linux {
+
+    /**
+     * Responsible for determining the name/family/release of Cisco operating systems.
+     */
+    struct os_cisco : os_linux
+    {
+        /**
+         * Constructs the os_cisco and reads a release file to gather relevant details.
+         * @param file The release file to read for OS data.
+         */
+        os_cisco(std::string const& file) : os_linux({"ID", "ID_LIKE", "VERSION"}, file) {}
+
+        /**
+         * Finds ID from the release file contents and returns it as the name.
+         * @param distro_id Unused.
+         * @return Returns the release name.
+         */
+        virtual std::string get_name(std::string const& distro_id) const override
+        {
+            auto val = _release_info.find("ID");
+            return (val != _release_info.end()) ? val->second : std::string();
+        }
+
+        /**
+         * Finds ID_LIKE from the release file contents and returns it as the family.
+         * @param name Unused.
+         * @return Returns the release family.
+         */
+        virtual std::string get_family(std::string const& name) const override
+        {
+            auto val = _release_info.find("ID_LIKE");
+            return (val != _release_info.end()) ? val->second : std::string();
+        }
+
+        /**
+         * Finds VERSION from the release file contents and returns it as the release.
+         * @param name Unused.
+         * @param distro_release Unused.
+         * @return Returns the release version.
+         */
+        virtual std::string get_release(std::string const& name, std::string const& distro_release) const override
+        {
+            auto val = _release_info.find("VERSION");
+            return (val != _release_info.end()) ? val->second : std::string();
+        }
+    };
+
+}}}  // namespace facter::facts::linux
--- a/lib/inc/internal/facts/linux/os_linux.hpp
+++ b/lib/inc/internal/facts/linux/os_linux.hpp
@@ -7,6 +7,8 @@
 #include <internal/facts/linux/release_file.hpp>
 #include <set>
 #include <map>
+#include <tuple>
+#include <string>
 
 namespace facter { namespace facts { namespace linux {
 
@@ -43,6 +45,14 @@ namespace facter { namespace facts { namespace linux {
          * @return Returns the release version.
          */
         virtual std::string get_release(std::string const& name, std::string const& distro_release) const;
+
+        /**
+         * Parses the release version string to return the major version.
+         * @param name The release name determined using get_name.
+         * @param release The release version determined using get_release.
+         * @return Returns a tuple of the major and minor versions.
+         */
+        virtual std::tuple<std::string, std::string> parse_release(std::string const& name, std::string const& release) const;
 
         /**
          * Parses a file containing key-value pairs separated by an equal (=) sign, one pair per line.
--- a/lib/inc/internal/facts/linux/os_linux.hpp
+++ b/lib/inc/internal/facts/linux/os_linux.hpp
@@ -1,0 +1,62 @@
+/**
+ * @file
+ * Declares the generic Linux operating system query helper.
+ */
+#pragma once
+
+#include <internal/facts/linux/release_file.hpp>
+#include <set>
+#include <map>
+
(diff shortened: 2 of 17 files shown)


kylog:
Working on facter poisons everyone's spelling of refactor ... :)

MikaelSmith:
@peterhuene I could use some review on this. I'm taking the approach of separating out OS resolution so we can have distribution-specific implementations (one of the abstractions Ruby Facter had) to handle #949, where OS name/family/release all come from a release file.

MikaelSmith:
Yes it does.

peterhuene:
I'll look it over.  This is basically the oldest code in libfacter, so I'm happy it's being reworked.

MikaelSmith:
It may be worth a more drastic rework; for now I kept it limited to Linux.

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -21,82 +22,186 @@ namespace bs = boost::system;
 
 namespace facter { namespace facts { namespace linux {
 
-    operating_system_resolver::data operating_system_resolver::collect_data(collection& facts)
+    // Return contents of the os-release file
+    // http://www.freedesktop.org/software/systemd/man/os-release.html
+    static map<string, string> key_value_file(string file)
```
peterhuene: I know these release files are generally trivially small, but I'm not a big fan of buffering a file's contents when reading through each line is really what we are doing.  Is there a way the base `os_linux` type could perform this operation and pass the key-value pairs to the derived type after each one is read? That way each derived type can make use of only the keys that they care about too.

Perhaps make this a base member function you can pass a lambda to that gets passed the key and value and then each derived type can process their relevant keys?  Thoughts?

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -21,82 +22,186 @@ namespace bs = boost::system;
...
-             result.name == os::kfreebsd ||
-             result.name == os::ubuntu)) {
-            result.architecture = "amd64";
-        } else if (re_search(result.architecture, boost::regex("i[3456]86|pentium"))) {
-            // For 32-bit, use "x86" for Gentoo and "i386" for everyone else
-            if (result.name == os::gentoo) {
-                result.architecture = "x86";
-            } else {
-                result.architecture = "i386";
+    static os_linux get_os()
```
peterhuene: Does this work?  We're returning an `os_linux` and not a pointer/reference type.

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -21,82 +22,186 @@ namespace bs = boost::system;
...
+            return os::oracle_enterprise_linux;
+        }
+        return {};
+    }
 
-        return result;
+    static string check_redhat_linux()
+    {
+        bs::error_code ec;
+        if (is_regular_file(release_file::redhat, ec)) {
```
peterhuene: I know this is just the old code, but I don't like how this implementation adhered a little too closely to the ruby implementation.  This reads the entire file in and just looks for any mention of these things anywhere in the file.  I presume there are particular keys we should be checking?  Just something to note for possible reimplementation.

peterhuene:
I'm :+1: for where this implementation is going.  Cleaning up this spaghetti code that resulted from effectively porting the old resolution logic from the ruby implementation is sorely needed (the more elegant "OS hierarchy" that exists in 2.x came after libfacter implemented these facts; it resulted from "backporting" the os structured fact to ruby iirc).

lib/src/facts/linux/operating_system_resolver.cc
MikaelSmith: I'm ... not sure. I can use unique_ptr instead.

lib/src/facts/linux/operating_system_resolver.cc
MikaelSmith: Possible, but I'd have to read the file twice: once to determine check if it's Cumulus, CoreOS, NXOS, or another Linux variant, and again to read the relevant data for that OS.

However, most variants only read it once, and twice isn't too bad.

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -123,7 +228,43 @@ namespace facter { namespace facts { namespace linux {
         return check_other_linux();
     }
 
-    string operating_system_resolver::get_release(string const& name, string const& distro_release)
+    string os_linux::get_family(string const& name)
```
MikaelSmith: I don't yet call get_family anywhere.

lib/src/facts/linux/operating_system_resolver.cc
MikaelSmith: I'm not really sure what the file format looks like. Quick search only reveals https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/5/html/Tuning_and_Optimizing_Red_Hat_Enterprise_Linux_for_Oracle_9i_and_10g_Databases/sect-Oracle_9i_and_10g_Tuning_Guide-Installing_Required_Software_Packages-Verifying_and_Updating_the_redhat_release_File.html.

MikaelSmith:
I think this works and I've addressed initial feedback. Going to work on the NXOS implementation.

MikaelSmith:
This works, but I need to decide how to rearrange release parsing.

MikaelSmith:
I believe this is done. @peterhuene @whopper you spent the most time on this functionality.

MikaelSmith:
Working on the test failures.

MikaelSmith:
I'm very happy we now have some acceptance tests checking the OS fact.

lib/inc/internal/facts/linux/os_cisco.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * @file
+ * Declares the Cisco Linux operating system query helper.
+ */
+#pragma once
+
+#include <internal/facts/linux/os_linux.hpp>
+
+namespace facter { namespace facts { namespace linux {
+
+    struct os_cisco : os_linux
```
peterhuene: I forget, will this generate doxygen warnings for missing comments? I can't remember if we were limiting docs to the public headers or not.

lib/inc/internal/facts/linux/os_linux.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <map>
+#include <tuple>
+#include <string>
+
+namespace facter { namespace facts { namespace linux {
+
+    struct os_linux
+    {
+        os_linux(std::set<std::string> items = {}, std::string file = release_file::os);
+        virtual std::string get_name(std::string const& distro_id) const;
```
peterhuene: I assume passing distro_id is for future use when we refactor the remaining logic (doesn't seem to be used currently)?

peterhuene: No wait, take that back.  It is used.

lib/inc/internal/facts/linux/os_cisco.hpp
MikaelSmith: They'll need documentation, I'll add it later this afternoon.

lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
```diff
@@ -25,6 +25,14 @@ namespace facter { namespace facts { namespace resolvers {
          */
         virtual void resolve(collection& facts) override;
 
+        /**
+         * Parses the major and minor OS release versions for Linux distros.
+         * @param name The name of the OS.
+         * @param release The release to parse.
+         * @return Returns a tuple of major and minor release versions.
+         */
+        static std::tuple<std::string, std::string> parse_distro(std::string const& name, std::string const& release);
```
peterhuene: Why is this in the base resolver instead of in the Linux resolver?

peterhuene: Never mind, this makes sense to do since it's parsing the "linux-specific" part of the structured fact.

lib/inc/internal/facts/linux/os_cisco.hpp
peterhuene: :+1:

peterhuene:
I'm :+1:.  This will hopefully form the basis for further cleanup of the OS resolver implementation (previous direct ruby Facter resolution port resulting in spaghetti code makes peter something something).

peterhuene:
Should we squash some of these commits?

MikaelSmith:
I'll fix up doxygen and squash commits in an hour or two.

MikaelSmith:
Fixed doxygen warnings and merged.

ScottGarman:
Nicely done, @MikaelSmith! :+1:

PR #963, opened 2015-06-01
(FACT-1021) Allow custom facts to `require 'facter'`.

peterhuene:
Some existing custom facts will `require 'facter'` even though it is
unnecessary.  A previous code cleanup removed some code that allowed
this to work when custom facts are evaluated from command line facter
(note: it still worked in Puppet because puppet requires facter).

This commit puts the previous workaround back so that custom facts that
`require 'facter'` can still work from the command line.


peterhuene:
Investigating AppVeyor failure.

MikaelSmith:
:+1:

PR #964, opened 2015-06-02
Master to stable

MikaelSmith:
Result of `git merge --ff-only upstream/master`. Prior to this, I did a `git merge upstream/stable -s ours` into `master`.


MikaelSmith:
pushed to stable

PR #965, opened 2015-06-05
(FACT-965) Ensure fact resolution occurs under JNI lock for JRuby.

peterhuene:
Currently, fact resolution occurs upon the first call to
`com.puppetlabs.Facter.lookup`. Because native libraries are shared between
Java threads and fact collections are not thread-safe, this is inherently unsafe.

This moves the resolution of the facts into loading of the library,
which is protected by a JVM mutex.  Puppet Server will now load the facter
JNI shim class at startup, causing all fact resolution to occur only
once at startup.  Because the JNI shim only supports reading a fact's
value, subsequent calls to `com.puppetlabs.Facter.lookup` do not need to
be protected as the collection is considered to be immutable after
loading of the native library.

### DIFF
--- a/lib/inc/facter/facts/collection.hpp
+++ b/lib/inc/facter/facts/collection.hpp
@@ -158,6 +158,20 @@ namespace facter { namespace facts {
         }
 
         /**
+        * Gets a fact value by name without resolving the fact.
+        * @tparam T The expected type of the value.
+        * @param name The name of the fact to get the value of.
+        * @return Returns a pointer to the fact value or nullptr if the fact is not resolved or the value is not the expected type.
+        */
+        template <typename T = value>
+        T const* get_resolved(std::string const& name) const
+        {
+            // Lookup the fact without resolving
+            auto it = _facts.find(name);
+            return dynamic_cast<T const*>(it == _facts.end() ? nullptr : it->second.get());
+        }
+
+        /**
          * Gets a fact value by name
          * @param name The name of the fact to get the value of.
          * @return Returns a pointer to the fact value or nullptr if the fact is not in the fact collection.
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Can we make collection queries a `const` operation?

peterhuene:
Unfortunately we can't, because get is a mutable operation on a lazy-populated fact collection (i.e. may trigger fact resolution).  Alternatively, I could add a new member that gets a fact and does not attempt to resolve any facts.

MikaelSmith:
I was leaning towards using a const reference and adding a new const member for getting facts, but it's only used one place. I guess a comment mentioning the restriction would be about as useful.

peterhuene:
I'll see what I can do to make this better so that we don't unintentionally mutate after initialization.

peterhuene:
@MikaelSmith Updated.

lib/inc/facter/facts/collection.hpp
```diff
@@ -158,6 +158,20 @@ namespace facter { namespace facts {
         }
 
         /**
+        * Gets a fact value by name without resolving the fact.
+        * @tparam T The expected type of the value.
+        * @param name The name of the fact to get the value of.
+        * @return Returns a pointer to the fact value or nullptr if the fact is not resolved or the value is not the expected type.
```
MikaelSmith: There's no dynamic cast, so will this be true?

peterhuene: Fixed.

MikaelSmith:
This should go to stable?

peterhuene:
Will retarget; didn't realize that we had already done this for facter (although we're still version 2.9 in master? :/).

MikaelSmith:
Might have to bump version numbers on both branches.

PR #967, opened 2015-06-05
(maint) Bump version to 3.0.0

MikaelSmith:


peterhuene:
Have we fixed all the jenkins jobs where having 3.0 (hence why it was changed back to 2.9 on master iirc) so that this is not a problem anymore?

branan:
@peterhuene the commit message for setting it to 2.9 doesn't mention any issues, but @MikaelSmith was the one that wrote it

MikaelSmith:
I don't see any issues. I don't remember having a good reason then either.

kylog:
Can you bump lib/Doxyfile also?

kylog:
Other than that :+1:

PR #969, opened 2015-06-08
(FACT-1020) Add augeasversion fact

MikaelSmith:
Runs `augtool --version` and parses the version.


lib/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ set(LIBFACTER_COMMON_SOURCES
     "src/facts/external/yaml_resolver.cc"
     "src/facts/map_value.cc"
     "src/facts/resolver.cc"
+    "src/facts/resolvers/augeasversion_resolver.cc"
```
peterhuene: Can we name this the "augeas resolver" that populates any augeas-related facts?  That way it's more generic in case in the future we want it to report on more than just the version.

lib/inc/internal/facts/resolvers/augeasversion_resolver.hpp
```diff
@@ -0,0 +1,35 @@
...
+#pragma once
+
+#include <facter/facts/resolver.hpp>
+
+namespace facter { namespace facts { namespace resolvers {
+
+    /**
+    * Responsible for resolving augeasversion fact, when an augeas runtime is available.
+    */
+    struct augeasversion_resolver : resolver
```
peterhuene: Same comment as above.

lib/src/facts/resolvers/augeasversion_resolver.cc
```diff
@@ -0,0 +1,59 @@
...
+            {
+                fact::augeasversion
+            })
+    {
+    }
+
+    string augeasversion_resolver::get_version()
+    {
+        string augtool = [] {
+#ifdef FACTER__PATH
```
peterhuene: Is this double underscored?

lib/CMakeLists.txt
MikaelSmith: I guess I can add a augeas > version structured fact as well.

lib/src/facts/resolvers/augeasversion_resolver.cc
```diff
@@ -0,0 +1,59 @@
...
+            "augeasversion",
+            {
+                fact::augeasversion
+            })
+    {
+    }
+
+    string augeasversion_resolver::get_version()
+    {
+        string augtool = [] {
```
peterhuene: The lambda adds a little extra confusion for no real benefit, I think.  Could we just assign to `augtool` inside of the `#ifdef` and outside of it simply assign to `"augtool"` if empty?

lib/CMakeLists.txt
peterhuene: Yeah, that might be useful to have the "major, minor, full" etc properties we do for other "version" facts.

peterhuene:
Other than the above comments, :+1:

lib/src/facts/resolvers/augeasversion_resolver.cc
MikaelSmith: This is a pattern I used in https://github.com/puppetlabs/facter/blob/master/lib/src/facts/linux/virtualization_resolver.cc#L105.

lib/src/facts/resolvers/augeasversion_resolver.cc
MikaelSmith: Looks like it, will fix.

MikaelSmith:
I was somewhat avoid adding an {{augeas}} structured fact until we had a reason for it.

peterhuene:
:+1:

peterhuene:
Coveralls seems lagged, but coverage appeared to have increased anyway.  Merging.

PR #971, opened 2015-06-10
(maint) add PE platform configs

justinstoller:
I copied the platform configs added to Puppet in puppetlabs/puppet#4016 into here. I can target this at master if that would be better.

===begin commit message===

Prior to this there were no acceptance test host config files for PE
only platforms. The distinction between PE and FOSS is largely going
away (save for where the failures are reported).

This commit adds the necessary host configs to test Facter against all
the platforms we will distribute puppet-agent on in the Shallow Gravy
timeframe.

### DIFF
--- a/acceptance/config/nodes/centos-4-i386.yaml
+++ b/acceptance/config/nodes/centos-4-i386.yaml
@@ -1,0 +1,19 @@
+---
+HOSTS:
+  master:
+    roles:
+    - master
+    platform: el-7-x86_64
+    hypervisor: vcloud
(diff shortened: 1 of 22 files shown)


MikaelSmith:
If they're needed for Shallow Gravy CI, they should go into stable.

justinstoller:
Doh! Took my maint/stable/sg-preview branch, wrote a description about how I could target it at master if that'd be better, and then opened the PR against master anyways....

Opened against stable at https://github.com/puppetlabs/facter/pull/973

PR #974, opened 2015-06-11
(FACT-704) Fix docker detection for systemd slices

bodgit:
Just ran into this whilst trying to run puppet inside a docker container on top of CentOS 7 and noticed that the `virtual` fact was mis-detecting the underlying VMware VM rather than the docker container.

I've changed the logic to match `/docker` somewhere in the string rather than `/docker/` at the beginning of the string.

I've added a sample `/proc/1/cgroup` fixture and updated the tests.

### DIFF
--- a/lib/facter/util/virtual.rb
+++ b/lib/facter/util/virtual.rb
@@ -162,7 +162,7 @@ module Facter::Util::Virtual
     path = Pathname.new('/proc/1/cgroup')
     return false unless path.readable?
     begin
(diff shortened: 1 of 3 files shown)


peterhuene:
Hi @bodgit.  Thanks for the contribution!  The fix looks good.  Note to reviewers: we should port this fix to 3.x too.

peterhuene:
Hah!  That's what I get for reading through my GitHub notification email from oldest to newest.  You've already beat me to the punch with a C++ PR.  Awesome :+1:

PR #975, opened 2015-06-11
(FACT-704) Fix docker detection for systemd slices

bodgit:
Similar to #974 this adds the same fix to cfacter. I'm not at all familiar with C++ or Boost so I've just compile tested the change but looking at the Boost API docs it should be the right "algorithm". I couldn't also see any obvious test fixtures I could update akin to the original Ruby-based facter.

### DIFF
--- a/lib/src/facts/linux/virtualization_resolver.cc
+++ b/lib/src/facts/linux/virtualization_resolver.cc
@@ -78,11 +78,11 @@ namespace facter { namespace facts { namespace linux {
             if (parts.size() < 3) {
                 return true;
             }
-            if (boost::starts_with(parts[2], boost::as_literal("/docker/"))) {
+            if (boost::contains(parts[2], boost::as_literal("/docker"))) {
                 value = vm::docker;
                 return false;
             }
-            if (boost::starts_with(parts[2], boost::as_literal("/lxc/"))) {
+            if (boost::contains(parts[2], boost::as_literal("/lxc"))) {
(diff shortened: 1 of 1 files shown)


peterhuene:
Hi @bodgit.  Thanks for the contribution (again)!  This fix looks good.  Ping to @kylog with regards to which branch this belongs in.

kylog:
Thanks @bodgit !

@peterhuene targeting master for this PR makes sense. We're pretty much in "lockdown" on stable atm.

bodgit:
I tested the change on the Ruby-based facter easily enough. For cfacter I initially only compile-tested it on my Mac, it looks like I made the same code change as the one you suggested.

It appears non-trivial to build on CentOS 7 where I first spotted the problem due to a lack of suitable dependencies but I've just created a Fedora 20 container and I've confirmed that it reports `virtual` as `vmware` without my change (due to it being docker on top of a VMware VM) but correctly changes to `docker` with my change.

peterhuene:
Hi @bodgit.  I've merged into master.  If you get a chance, would you mind updating FACT-704 with any instructions for our benefit of how best to reproduce the bug so our QA can verify the fix? Was this just broken with any systemd-based distro running under docker or was there something on systemd we needed to enable to reproduce?  Thanks again for the fix and making the effort to verify it!

bodgit:
Done, ticket updated. Hopefully that's sufficient detail for you.

PR #976, opened 2015-06-11
(FACT-1042) Allow time for EC2 to respond

MikaelSmith:
Without this change, EC2 will not always respond to the metadata request
before the http session times out. Extend the session timeout to 5
seconds, and add a connection timeout of 200 ms to ensure Facter doesn't
spend too long waiting on the http request if not in EC2. Update the
error and add more detailed trace info in case EC2 is very slow to
respond.

Also extend the GCE session timeout to 5 seconds, and set a connection
timeout of 1 second to avoid long waits if GCE is unavailable.

### DIFF
--- a/lib/src/facts/resolvers/ec2_resolver.cc
+++ b/lib/src/facts/resolvers/ec2_resolver.cc
@@ -33,11 +33,14 @@ namespace facter { namespace facts { namespace resolvers {
 #ifdef USE_CURL
     static const char* EC2_METADATA_ROOT_URL = "http://169.254.169.254/latest/meta-data/";
     static const char* EC2_USERDATA_ROOT_URL = "http://169.254.169.254/latest/user-data/";
+    static const unsigned int EC2_CONNECTION_TIMEOUT = 200;
+    static const unsigned int EC2_SESSION_TIMEOUT = 5000;
 
     void query_metadata_value(client& cli, map_value& value, string const& url, string const& name)
     {
         request req(url + name);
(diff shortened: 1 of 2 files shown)


MikaelSmith:
I did a test on EC2 last night with something like this. Trying again, and checking GCE, using a new build.

lib/src/facts/resolvers/ec2_resolver.cc
```diff
@@ -33,11 +33,14 @@ namespace facter { namespace facts { namespace resolvers {
 #ifdef USE_CURL
     static const char* EC2_METADATA_ROOT_URL = "http://169.254.169.254/latest/meta-data/";
     static const char* EC2_USERDATA_ROOT_URL = "http://169.254.169.254/latest/user-data/";
+    static unsigned int EC2_CONNECTION_TIMEOUT = 200;
```
peterhuene: Can we add `const` to these two?

lib/src/facts/resolvers/gce_resolver.cc
```diff
@@ -219,7 +219,8 @@ namespace facter { namespace facts { namespace resolvers {
         try
         {
             request req("http://metadata/computeMetadata/v1beta1/?recursive=true&alt=json");
-            req.timeout(1000);
+            req.connection_timeout(1000);
```
peterhuene: Can we add similar constants here?

peterhuene:
Other than above, :+1:

PR #977, opened 2015-06-11
(FACT-1016) Prevent unbounded growth of fact resolutions.

peterhuene:
A bug in Puppet exposed an unbounded memory growth if the same custom
fact file is repeatedly loaded.  The reason is that every execution
results in a new resolution being added to the fact.  While this itself
is desired behavior, we should limit the number of resolutions to some
arbitrary size so that memory doesn't grow unbounded.

### DIFF
--- a/lib/tests/ruby/ruby.cc
+++ b/lib/tests/ruby/ruby.cc
@@ -592,4 +592,19 @@ SCENARIO("custom facts written in Ruby") {
             REQUIRE(ruby_value_to_string(facts.get<ruby_value>("foo")) == "\"bar\"");
         }
     }
+    GIVEN("a fact that has 100 resolutions") {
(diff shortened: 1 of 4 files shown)


lib/tests/ruby/ruby.cc
```diff
@@ -592,4 +592,13 @@ SCENARIO("custom facts written in Ruby") {
             REQUIRE(ruby_value_to_string(facts.get<ruby_value>("foo")) == "\"bar\"");
         }
     }
+    GIVEN("a fact that has too many resolutions") {
```
MikaelSmith: Can you add a test case for exactly 100? Satisfies my need for bounds testing.

peterhuene: Will do.

peterhuene:
@MikaelSmith Updated.

PR #978, opened 2015-06-12
(maint) Build curl on Windows

MikaelSmith:
Build libcurl on Windows. We support static linking to simplify
dependencies, particularly because we don't need SSL enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,8 @@ endif()
 
 option(YAMLCPP_STATIC "Use yaml-cpp's static libraries" OFF)
 
+option(CURL_STATIC "Use curl's static libraries" OFF)
+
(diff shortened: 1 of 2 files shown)


MikaelSmith:
I need to build and push a 32-bit curl, and verify everything works.

MikaelSmith:
Pushed a 32-bit build of curl, looks like it works now.

MikaelSmith:
Manually tested a build and ran it on Server 2012 R2 under EC2; facts look good.

haus:
:+1:

haus:
I have questions about the future of PREFIX_PATH used here, but it doesn't need to block merging this.

MikaelSmith:
@peterhuene mind taking a look and merging?

PR #979, opened 2015-06-15
(FACT-765)  Constrain windows `id` gid resolution

reidmv:
Windows platforms should not attempt to resolve the gid fact if an "id" executable exists in the path. The gid fact should not return a value on Windows.

### DIFF
--- a/lib/facter/gid.rb
+++ b/lib/facter/gid.rb
@@ -10,7 +10,7 @@
 
 Facter.add(:gid) do
   confine do
-    Facter::Core::Execution.which('id') && Facter.value(:kernel) != "SunOS"
+    Facter::Core::Execution.which('id') && !["SunOS", "windows"].include?(Facter.value(:kernel))
   end
   setcode { Facter::Core::Execution.exec('id -ng') }
 end
--- a/spec/unit/gid_spec.rb
+++ b/spec/unit/gid_spec.rb
@@ -39,4 +39,13 @@ describe "gid fact" do
       Facter.fact(:gid).value.should == nil
     end
   end
+
+  describe "on windows systems" do
+    it "is not supported" do
+      Facter.fact(:kernel).stubs(:value).returns("windows")
+      Facter::Core::Execution.expects(:which).with('id').returns(true)
+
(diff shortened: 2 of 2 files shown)


ferventcoder:
What does this resolve to now?

reidmv:
On a stock Windows system, nothing. Windows doesn't usually have an "id" executable in the path so the confines in every resolution are sufficient to prevent Facter from trying to resolve it

Per the Jira ticket though, if a user does have any kind of executable named "id" in their path the default resolution will be considered valid, Facter will launch "id" and try to use whatever output it has for its resolution. In the specific instance for which the ticket was filed "id.exe" was a GUI app that Facter ended up launching, and then waiting (forever) for a value to be returned from it.

ferventcoder:
Or in the case of my system -

```sh
c:\> where.exe id
C:\tools\GnuWin\bin\id.exe
C:\Program Files (x86)\Git\bin\id.exe
```

ferventcoder:
@peterhuene @whopper @MikaelSmith This looks good to me. Is there any followup for compiled Facter?

ferventcoder:
Even better - on Windows maybe it should return the group? Or would that cause possible issues?

reidmv:
I'm targeting a .z level of impact with this PR, just trying to correctly constrain the existing id-based resolution.

It could make a kind of sense to return the group on Windows. It's transforming a Windows value into kind of a linux-ism fact which is a little weird but that doesn't mean it wouldn't make sense. However, that would be a new resolution and a behavior change warranting more serious thought than I've put into this so far. New ticket?

MikaelSmith:
No followup for compiled facter. This looks fine for a .z release.

ferventcoder:
@reidmv yup, makes sense.

ferventcoder:
I'm +1 on these changes as soon as Travis CI agrees we can roll it forward.

ferventcoder:
So anyone depending on the gid fact should have handling built-in for cases where it doesn't return anything, and probably already handling cases for that with Solaris, and likely not using it if they are on Windows. So this seems safe for a .z release to me.

PR #980, opened 2015-06-15
(WIP)(FACT-984, FACT-983, FACT-1044) Update kernel resolver and acceptance tests for Debian 8 and Ubuntu 15.04

whopper:
The first of these commits refactors version parsing in the kernel resolver by creating a new parsing utility function to be used to return major and minor versions via `collect data.` In addition, this commit adds a new Linux kernel resolver which uses a new regex to parse kernel version. This became necessary when `uname` output changed slightly in Debian 8, no longer including the patch level for the kernel.

The other two commits update two fact-based acceptance tests for Debian 8 and Ubuntu 15.04.

### DIFF
--- a/lib/inc/internal/util/version_parsing.hpp
+++ b/lib/inc/internal/util/version_parsing.hpp
@@ -1,0 +1,26 @@
+/**
+ * @file
+ * Declares utility functions parsing version strings.
+ */
+#pragma once
+
+#include <string>
+#include <tuple>
+
+namespace facter { namespace util { namespace version_parsing {
+
+    /**
+     * Parses the major and minor kernel versions for all platforms except Linux.
+     * @param version The version to parse.
+     * @return Returns a tuple of major and minor versions.
+     */
+    std::tuple<std::string, std::string> parse_kernel_version(std::string const& version);
+
+    /**
+     * Parses the major and minor kernel versions for Linux platforms
+     * @param version The version to parse.
+     * @return Returns a tuple of major and minor versions.
+     */
+    std::tuple<std::string, std::string> parse_linux_kernel_version(std::string const& version);
+
+}}}  // namespace facter::util::version_parsing
--- a/lib/src/facts/resolvers/kernel_resolver.cc
+++ b/lib/src/facts/resolvers/kernel_resolver.cc
@@ -31,32 +31,15 @@ namespace facter { namespace facts { namespace resolvers {
             facts.add(fact::kernel_release, make_value<string_value>(move(data.release)));
         }
 
-        if (!data.version.empty()) {
-            string major, minor;
-            tie(major, minor) = parse_version(data.version);
-
-            if (!major.empty()) {
-                facts.add(fact::kernel_major_version, make_value<string_value>(move(major)));
-            }
-            if (!minor.empty()) {
-                // TODO: for use in a structured fact; no point adding a new flat fact for it
-            }
-
-            facts.add(fact::kernel_version, make_value<string_value>(move(data.version)));
+        if(!data.full_version.empty()) {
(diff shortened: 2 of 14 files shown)


lib/src/facts/resolvers/kernel_resolver.cc
```diff
@@ -48,13 +51,9 @@ namespace facter { namespace facts { namespace resolvers {
...
     tuple<string, string> kernel_resolver::parse_version(string const& version) const
     {
-        auto pos = version.find('.');
-        if (pos != string::npos) {
-            auto second = version.find('.', pos + 1);
-            if (second != string::npos) {
-                pos = second;
-            }
-            return make_tuple(version.substr(0, pos), version.substr(pos + 1));
+        string major, full;
```
peterhuene: Is `full` the right name for this variable? Perhaps `minor` or `patch`?

peterhuene:
Other than above, :+1:

whopper:
@peterhuene good call on the name - forgot to change it after the first pass. Updated!

MikaelSmith:
We test most platforms. Will Solaris and AIX work with the new `parse_version`? Have you tested Windows?

peterhuene:
@MikaelSmith Good catch.  I wasn't paying attention and thought this was being done on the Linux resolver :/  Yeah, this probably doesn't belong in the base and needs to be overridden in the Linux resolver.  I'd expect `X.Y` to parse into major `X` and major `Y` for other operating systems.

whopper:
@peterhuene @MikaelSmith ah right, I didn't think about other platforms doing regular X.Y major/minor versioning - I'll rework this in the the linux resolver.

MikaelSmith:
When I was reworking code I considered moving kernel_major and kernel_minor into the data passed back by collect_data, and pulling versions parsers into utilities. It might be time to do that.

kylog:
IMHO the concept of "major" is really not well or consistently used industry-wide. Each kernel and each OS rolls their own way.

So :+1: on not trying to handle it by inheritance but via utility functions. Although, if this gets sticky, I'd also be fine deferring the Do the Right Thing portion of this to another day :)

lib/inc/internal/util/version_parsing.hpp
```diff
@@ -0,0 +1,26 @@
...
+#include <tuple>
+
+namespace facter { namespace util { namespace version_parsing {
+
+    /**
+     * Parses the major and minor kernel versions for all platforms except Linux.
+     * @param version The version to parse.
+     * @return Returns a tuple of major and minor versions.
+     */
+    std::tuple<std::string, std::string> parse_kernel_version(std::string const& version);
```
whopper: Naming is hard! Any suggestions on names for this? This function does the parsing everything *except* Linux (and AIX)...

whopper:
The refactor is underway, and tomorrow I'll be doing testing to make sure it looks good everywhere.

I ended up following Michael's vision of breaking the parsing into utility functions and having each resolver use them as needed. Major, minor, and full version strings are then passed back to the top-level resolver and added to the collection as flat facts (except minor, which we currently don't do anything with).

lib/inc/internal/util/version_parsing.hpp
MikaelSmith: Since it's internal, I'm fine with this for now. If we end up with more varieties we'll rename it.

MikaelSmith:
:+1:

MikaelSmith:
Retarget at stable.

whopper:
In the interest of not breaking stuff on the eve of 3.0.0 (and simplicity), @MikaelSmith and I decided to do this the simple way for now. Closing this in favor of https://github.com/puppetlabs/facter/pull/982

PR #981, opened 2015-06-16
(maint) Switch from deprecated to new cmake vars

MikaelSmith:
Without this change, CLion doesn't put libs in the correct location on
Windows, so Facter won't run. It seems to ignore EXECUTABLE_OUTPUT_PATH
and LIBRARY_OUTPUT_PATH, which CMake notes are deprecated.

Switch to using new equivalents CMAKE_RUNTIME_OUTPUT_DIRECTORY and
CMAKE_LIBRARY_OUTPUT_DIRECTORY, so CLion builds work on Windows.


MikaelSmith:
@peterhuene

MikaelSmith:
This seems like something that should be applied to all projects, so thinking of moving it to Leatherman's cflags.

peterhuene:
:+1: On the change.  Should probably wait until 3.0 is released before merging this in and then up to master.  I agree that this may also belong in LM.

MikaelSmith:
I'm going to close this in favor of (maint) commits as part of https://github.com/puppetlabs/leatherman/pull/23 and FACT-1047.

PR #984, opened 2015-06-18
(FACT-964) Search archlibdir and sitearchlibdir for libruby

whopper:
Prior to this commit, Facter only searched `RbConfig['libdir']`
for libruby. However, the ruby2.1 package places libruby
under `RbConfig['archlibdir']` or `sitearchlibdir` on some platforms.
Thus, Facter wasn't able to find libruby on said systems. This commit
adds `archlibdir` and `sitearchlibdir` to the list of paths to check
for libruby.

### DIFF
--- a/lib/inc/internal/ruby/api.hpp
+++ b/lib/inc/internal/ruby/api.hpp
@@ -623,7 +623,6 @@ namespace facter {  namespace ruby {
         static std::unique_ptr<api> create();
         static facter::util::dynamic_library find_library();
         static facter::util::dynamic_library find_loaded_library();
-        static std::string libruby_configdir();
         static VALUE callback_thunk(VALUE parameter);
         static VALUE rescue_thunk(VALUE parameter, VALUE exception);
         static VALUE protect_thunk(VALUE parameter);
--- a/lib/src/ruby/api.cc
+++ b/lib/src/ruby/api.cc
@@ -487,7 +487,13 @@ namespace facter { namespace ruby {
 
         bool success;
         string output, none;
-        tie(success, output, none) = execute(ruby, { "-e", "print File.join(RbConfig::CONFIG['"+libruby_configdir()+"'], RbConfig::CONFIG['LIBRUBY_SO'])" });
+
+        tie(success, output, none) = execute(ruby, { "-e", "print(['libdir', 'archlibdir', 'sitearchlibdir', 'bindir'].find do |name|"
+                                                                  "dir = RbConfig::CONFIG[name];"
+                                                                  "next unless dir;"
+                                                                  "file = File.join(dir, RbConfig::CONFIG['LIBRUBY_SO']);"
+                                                                  "break file if File.exist? file;"
+                                                                  "false end)" });
         if (!success) {
             LOG_WARNING("ruby failed to run: %1%", output);
             return library;
--- a/lib/src/ruby/posix/api.cc
+++ b/lib/src/ruby/posix/api.cc
@@ -10,9 +10,4 @@ namespace facter { namespace ruby {
         return dynamic_library::find_by_symbol("ruby_init");
     }
 
-    string api::libruby_configdir()
-    {
-        return "libdir";
-    }
-
 }}  // namespace facter::ruby
--- a/lib/src/ruby/windows/api.cc
+++ b/lib/src/ruby/windows/api.cc
@@ -11,9 +11,4 @@ namespace facter { namespace ruby {
         return dynamic_library::find_by_pattern(libruby_pattern);
     }
 
-    string api::libruby_configdir()
-    {
-        return "bindir";
-    }
-
 }}  // namespace facter::ruby


whopper:
Would this warrant an acceptance test? We could probably run one in Debian 8 just to ensure `libruby` gets loaded from `archlibdir` / `sitearchlibdir`, but I'm not sure if that's overkill or not.

peterhuene:
Spawning ruby is quite expensive as things go (~40ms doesn't seem like much, but that can be 33% of facter's wall clock time on some platforms).  Is there anyway to get this evaluated correctly with a single ruby execution?

lib/inc/internal/ruby/api.hpp
```diff
@@ -623,7 +623,7 @@ namespace facter {  namespace ruby {
         static std::unique_ptr<api> create();
         static facter::util::dynamic_library find_library();
         static facter::util::dynamic_library find_loaded_library();
-        static std::string libruby_configdir();
+        static std::vector<std::string> libruby_configdir();
```
peterhuene: Should probably be called something like `config_variables` or some such now.

whopper:
@peterhuene good point. Let me see what I can do for getting all the info in a single ruby run - I think I may have a few ideas.

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,47 @@ namespace facter { namespace ruby {
         }
         LOG_DEBUG("ruby was found at \"%1%\".", ruby);
 
-        bool success;
-        string output, none;
-        tie(success, output, none) = execute(ruby, { "-e", "print File.join(RbConfig::CONFIG['"+libruby_configdir()+"'], RbConfig::CONFIG['LIBRUBY_SO'])" });
+        bool success, loaded = false;
+        string ruby_cmd_args, ruby_output, libruby_file, libruby_full_path, none;
+        tie(success, libruby_file, none) = execute(ruby, { "-e", "print RbConfig::CONFIG['LIBRUBY_SO']"});
```
whopper: @peterhuene updated with a first pass of reducing ruby calls. I was able to query for all three libruby dirs in one go, but I had to do one other call to get `LIBRUBY_SO`. Previously we were executing File.join(RbConfig::CONFIG($dirname), RbConfig::CONFIG('LIBRUBY_SO')), but I ran into a situation where ruby returned nil for 'archlibdir' (on OSX), causing File.join to error out, even though there were other directories to check.

We could get `LIBRUBY_SO` in the same call as the three libruby directories, but we'd have to do some additional string parsing to save it and remove it from the list so we don't accidentally try and check an obviously invalid path. Does that sound like something we'd want to do here?

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,47 @@ namespace facter { namespace ruby {
...
-        if (!exists(output, ec) || is_directory(output, ec)) {
-            LOG_DEBUG("ruby library \"%1%\" was not found: ensure ruby was built with the --enable-shared configuration option.", output);
+        vector<string> lib_ruby_dirs = libruby_config_variables();
+        // Build the list of dirs to query for so we only have to execute ruby once
+        for (auto it = lib_ruby_dirs.begin(); it != lib_ruby_dirs.end(); ++it) {
+            ruby_cmd_args += "'"+*it+"',";
+        }
+
+        lib_ruby_dirs.clear();
+        tie(success, ruby_output, none) = execute(ruby, { "-e", "print RbConfig::CONFIG.values_at("+ruby_cmd_args+")" });
```
whopper: We can grab multiple keys from the `CONFIG` hash using `values_at()`

whopper:
Looking into the Windows failure

lib/src/ruby/api.cc
MikaelSmith: I'd rather do string parsing than multiple calls to the ruby executable.

peterhuene: How about:

```ruby
print(['libdir', 'archlibdir', "sitearchlibdir"].find do |name| dir = RbConfig::CONFIG[name]; next unless dir; file = File.join(dir, RbConfig::CONFIG['LIBRUBY_SO']); break file if File.exist? file end)
```

Then have the `['libdir', 'archlibdir', "sitearchlibdir"]` part inserted as a string based on `libruby_config_variables` (perhaps just returning this array as a string?)

MikaelSmith: :+1:

peterhuene: Or don't worry about the formatting and just always use `['libdir', 'archlibdir', 'sitearchlibdir'. 'bindir']`?

whopper:
@peterhuene @MikaelSmith updated to use Peter's fancy Ruby. I'm still using `libruby_config_variables`, but I'd be happy to get rid of it if you don't think it's necessary.

MikaelSmith:
It's still necessary to split out the Windows and POSIX variables.

peterhuene:
I don't see the point in keeping it, to be honest.  Most rubies we'll get a hit with `libdir` and then the next two for other platforms.  I don't think any of those locations would be picked up for Windows, which is why just defaulting to bindir last should be fine, imo.

lib/src/ruby/api.cc
```diff
@@ -8,6 +8,7 @@
 #include <facter/execution/execution.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/range/algorithm/remove_if.hpp>
```
MikaelSmith: `remove_if` is no longer used.

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,32 @@ namespace facter { namespace ruby {
         }
         LOG_DEBUG("ruby was found at \"%1%\".", ruby);
 
-        bool success;
-        string output, none;
-        tie(success, output, none) = execute(ruby, { "-e", "print File.join(RbConfig::CONFIG['"+libruby_configdir()+"'], RbConfig::CONFIG['LIBRUBY_SO'])" });
+        bool success, loaded = false;
```
MikaelSmith: The loaded variable is no longer needed since you got rid of the loop.

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,32 @@ namespace facter { namespace ruby {
...
+
+        tie(success, ruby_output, none) = execute(ruby, { "-e", "print(["+possible_lib_ruby_dirs+"].find do |name| dir = RbConfig::CONFIG[name]; next unless dir; file = File.join(dir, RbConfig::CONFIG['LIBRUBY_SO']); break file if File.exist? file; false end)" });
         if (!success) {
-            LOG_WARNING("ruby failed to run: %1%", output);
+            LOG_WARNING("ruby failed to run: %1%", ruby_output);
             return library;
         }
 
         boost::system::error_code ec;
-        if (!exists(output, ec) || is_directory(output, ec)) {
```
peterhuene: I think this code can be restored and remove the `loaded` flag (perhaps change `ruby_output` back to `output` now).

peterhuene:
@MikaelSmith other than potentially checking three additional places where libruby doesn't exist on windows, what would be the downside of having all four entries always in the array and removing `libruby_config_variables`?

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,32 @@ namespace facter { namespace ruby {
...
 
-        bool success;
-        string output, none;
-        tie(success, output, none) = execute(ruby, { "-e", "print File.join(RbConfig::CONFIG['"+libruby_configdir()+"'], RbConfig::CONFIG['LIBRUBY_SO'])" });
+        bool success, loaded = false;
+        string possible_lib_ruby_dirs, ruby_output, none;
+        vector<string> lib_ruby_dirs = libruby_config_variables();
+
+        // Build the list of directories to query for so we only have to execute ruby once
+        for (auto it = lib_ruby_dirs.begin(); it != lib_ruby_dirs.end(); ++it) {
```
MikaelSmith: Use
```
for (auto dir : lib_ruby_dirs) {
    possibly_lib_ruby_dirs += "'"+dir+"',";
}
```

You could also do `"'"+boost::join(lib_ruby_dirs, "', '")`.

lib/src/ruby/api.cc
```diff
@@ -485,21 +486,32 @@ namespace facter { namespace ruby {
...
+            possible_lib_ruby_dirs += "'"+*it+"',";
+        }
+
+        tie(success, ruby_output, none) = execute(ruby, { "-e", "print(["+possible_lib_ruby_dirs+"].find do |name| dir = RbConfig::CONFIG[name]; next unless dir; file = File.join(dir, RbConfig::CONFIG['LIBRUBY_SO']); break file if File.exist? file; false end)" });
```
MikaelSmith: You could split this into multiple lines using C-style string concatenation
```
"print(["+possible_lib_ruby_dirs+"].find do |name|"
"dir = RbConfig::CONFIG[name];"
...
```

MikaelSmith:
Ah, that would work. I'm fine with that.

On Fri, Jun 19, 2015 at 1:59 PM, Peter Huene <notifications@github.com>
wrote:

> I don't see the point in keeping it, to be honest. Most rubies we'll get a
> hit with libdir and then the next two for other platforms. I don't think
> any of those locations would be picked up for Windows, which is why just
> defaulting to bindir last should be fine, imo.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/puppetlabs/facter/pull/984#issuecomment-113642896>.
>



-- 
Michael Smith
Sr. Software Engineer, Puppet Labs

*PuppetConf 2015 <http://2015.puppetconf.com/> is coming to Portland,
Oregon! Join us October 5-9.*
*Register now to take advantage of the Early Adopter discount
<https://www.eventbrite.com/e/puppetconf-2015-october-5-9-tickets-13115894995?discount=EarlyAdopter>
*
*—**save $349!*

lib/src/ruby/posix/api.cc
```diff
@@ -1,4 +1,5 @@
 #include <internal/ruby/api.hpp>
+#include <vector>
```
MikaelSmith: `vector`'s no longer needed.

whopper:
@peterhuene @MikaelSmith updated once again and *wow* is this much simpler now... learned some new boost tricks even if we didn't end up using them :). I'm running some quick tests in Ubuntu and windows.

peterhuene:
:+1: @MikaelSmith can you merge when you're also :+1:?

peterhuene:
Do we need this ported to 3.0.1?

MikaelSmith:
It doesn't really affect released packages, just building from source.

Michael

PR #985, opened 2015-06-19
(FACT-1047) Add colored console output on Windows

MikaelSmith:
Ruby Facter used colored output in consoles for log messages. That didn't
make it into the initial native Facter implementation, so add it now.


ferventcoder:
Pretty.

PR #986, opened 2015-06-19
(FACT-954) Update build defaults for AIO targets

melissa:
With the switch to facter 3, we are no longer providing individual
packages for facter. Instead, facter will be delivered as a part of the
puppet-agent package. This commit updates the build_defaults used by the
packaging repo to comment out all build targets. We are, however,
keeping this file around in the off chance that community members are
using the packaging repo automation to build their own packages.


stahnma:
:+1:

PR #987, opened 2015-06-19
(docs) Improve docs template a bit

nfagerlund:
This PR adds type information, and lists the elements for map facts.


MikaelSmith:
Was this meant to go into stable for the release?

peterhuene:
@nfagerlund ^? I think the schema added the augeas fact since we split off stable, so we may need to cherry pick this to stable.

peterhuene:
Never mind, it looks like that change is in stable.  I believe there's nothing currently in master that isn't in stable wrt to the schema.

nfagerlund:
@peterhuene @MikaelSmith Awesome. Yeah, no, no need to have this in stable; I can generate facts from wherever and just pull the template in.

PR #991, opened 2015-06-26
(FACT-1075) Add flag to display legacy facts

MikaelSmith:
Facter 3 moved to only showing structured facts, hiding most flat fact
names used by Facter 2 in the default command-line output. This is
difficult for tools parsing the command-line output, particularly when
using json or yaml output.

Add a flag to enable displaying all legacy facts as part of command-line
output querying all facts.

### DIFF
--- a/lib/inc/facter/facts/collection.hpp
+++ b/lib/inc/facter/facts/collection.hpp
@@ -203,9 +203,10 @@ namespace facter { namespace facts {
          * @param stream The stream to write the facts to.
          * @param fmt The output format to use.
          * @param queries The set of queries to filter the output to. If empty, all facts will be output.
+         * @param show_legacy Show legacy facts when querying all facts.
          * @return Returns the stream being written to.
          */
-        std::ostream& write(std::ostream& stream, format fmt = format::hash, std::set<std::string> const& queries = std::set<std::string>());
+        std::ostream& write(std::ostream& stream, format fmt = format::hash, std::set<std::string> const& queries = std::set<std::string>(), bool show_legacy = false);
 
         /**
          * Resolves all facts in the collection.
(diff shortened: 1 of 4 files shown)


whopper:
:+1: this did end up being pretty simple! I verified that the option works as expected in both OSX and Ubuntu 14.04.

whopper:
@MikaelSmith oh one other thing - we'll want to add this option to the man page.

lib/inc/facter/facts/collection.hpp
```diff
@@ -212,6 +213,16 @@ namespace facter { namespace facts {
          */
         void resolve_facts();
 
+        /**
+         * Enables writing legacy facts when requesting all facts.
+         */
+        void show_legacy();
```
peterhuene: Personally I think this belongs as a parameter to `collection::write`.

man/man8/facter.8
```diff
@@ -35,6 +35,7 @@ If no queries are given, then all facts will be returned\.
       \fB\-\-external-dir\fR arg           A directory to use for external facts\.
       \fB\-\-help\fR                       Print help and usage information\.
 \fB\-j, [ \-\-json ]\fR                     Output facts in JSON format\.
+      \fB\-\-show-legacy\fR                Add legacy fact entries to output\.
```
peterhuene: Should this match the --help output?  i.e. `Show legacy facts when querying all facts.`

MikaelSmith: Yeah, that sounds clearer.

lib/inc/facter/facts/collection.hpp
MikaelSmith: Makes sense, I'll do that.

MikaelSmith:
Updated.

peterhuene:
:+1: will merge after CI passes.

PR #992, opened 2015-06-26
(FACT-1076) Extract file utility functions to Leatherman.

Magisus:
Extracts the file and directory utility classes to Leatherman and updates to be standardized across pegasus and cthun-agent.


lib/docs/facts.md
```diff
@@ -0,0 +1,1977 @@
+## `architecture`
```
MikaelSmith: I wouldn't expect this to be checked in. Where did it come from?

lib/src/execution/posix/execution.cc
```diff
@@ -26,6 +26,8 @@ extern char** environ;
 
 namespace facter { namespace execution {
 
+namespace lth_file = leatherman::file_util;
```
MikaelSmith: I'd indent by a tab since it's inside braces.

lib/docs/facts.md
Magisus: Yeah I wasn't exactly sure where it came from, but I also wasn't sure if it was supposed to be ignored or not. I'll add it.

lib/tests/util/directory.cc
```diff
@@ -1,88 +0,0 @@
-#include <catch.hpp>
```
MikaelSmith: I don't think these made it into leatherman. They probably should.

lib/tests/util/file.cc
```diff
@@ -1,76 +0,0 @@
-#include <catch.hpp>
```
MikaelSmith: These tests didn't make it into leatherman. They should.

lib/tests/util/directory.cc
Magisus: I deleted them because they do basically the same thing as the tests we did write. If you want redundancy/any extra cases they may cover, I'll add them. I didn't find them until after I'd already gone and written my own tests, so my bad.

MikaelSmith: Ah, I'll take a closer look at them Monday.

lib/tests/util/file.cc
Magisus: I went through and added some test cases based on these, mostly the ones that ensure a false callback will stop processing.

lib/docs/facts.md
MikaelSmith: and remove the file from this commit.

lib/tests/util/file.cc
MikaelSmith: Resolved in https://github.com/puppetlabs/leatherman/pull/38.

lib/tests/util/directory.cc
MikaelSmith: Resolved in https://github.com/puppetlabs/leatherman/pull/38.

MikaelSmith:
:+1:

PR #995, opened 2015-06-26
(FACT-1056) Ensure OSX OS fact resolves major and minor release values

whopper:
Commit e32ccc9 introduced a refactoring of the operatingsystem resolver,
but did not update the OSX specific resolver. This commit adds the
necessary logic to parse OSX full, major and minor releases to
populate the release key of the OS fact with.

### DIFF
--- a/lib/src/facts/osx/operating_system_resolver.cc
+++ b/lib/src/facts/osx/operating_system_resolver.cc
@@ -1,5 +1,8 @@
 #include <internal/facts/osx/operating_system_resolver.hpp>
+#include <facter/facts/collection.hpp>
 #include <facter/execution/execution.hpp>
+#include <facter/facts/fact.hpp>
+#include <facter/facts/scalar_value.hpp>
 #include <boost/algorithm/string.hpp>
 #include <string>
 
@@ -33,8 +36,21 @@ namespace facter { namespace facts { namespace osx {
                 result.osx.version = move(value);
             }
 
+            // Populate the major and minor version facts using kernel_release
+            auto release = facts.get<string_value>(fact::kernel_release);
+
+            if (!release->value().empty()) {
+                result.release = move(release->value());
+                auto pos = release->value().find('.');
+                if (pos != string::npos) {
+                    auto second = release->value().find('.', pos + 1);
+                    result.major = release->value().substr(0, pos);
+                    result.minor = release->value().substr(pos + 1, second - (pos + 1));
+                }
+            }
+
             // Continue only if we haven't populated the data
-            return result.osx.product.empty() || result.osx.build.empty() || result.osx.version.empty();
+            return result.osx.product.empty() || result.osx.build.empty() || result.osx.version.empty() || result.release.empty();
         });
 
         return result;


whopper:
Comparative output of Facter 2.4.4 and this PR:

Facter 2.4.4
```
{
  "os": {
    "name": "Darwin",
    "family": "Darwin",
    "release": {
      "major": "13",
      "minor": "4",
      "full": "13.4.0"
    }
  }
}
```

This commit
```
{
  architecture => "x86_64",
  family => "Darwin",
  hardware => "x86_64",
  macosx => {
    build => "13F34",
    product => "Mac OS X",
    version => {
      full => "10.9.5",
      major => "10.9",
      minor => "5"
    }
  },
  name => "Darwin",
  release => {
    full => "13.4.0",
    major => "13",
    minor => "4"
  }
}
```

whopper:
Note: if we merge https://github.com/puppetlabs/facter/pull/996, this will no longer be needed, as that PR has the same functionality.

lib/src/facts/osx/operating_system_resolver.cc
```diff
@@ -33,8 +36,21 @@ namespace facter { namespace facts { namespace osx {
...
             }
 
+            // Populate the major and minor version facts using kernel_release
+            auto release = facts.get<string_value>(fact::kernel_release);
+
+            if (!release->value().empty()) {
+                result.release = move(release->value());
+                auto pos = release->value().find('.');
+                if (pos != string::npos) {
+                    auto second = release->value().find('.', pos + 1);
```
peterhuene: Can we guard against the unlikely event that OSX changes to just `X.Y`

lib/src/facts/osx/operating_system_resolver.cc
```diff
@@ -33,8 +36,21 @@ namespace facter { namespace facts { namespace osx {
...
+                    result.major = release->value().substr(0, pos);
+                    result.minor = release->value().substr(pos + 1, second - (pos + 1));
```
peterhuene: Can we guard against the unlikely event that OSX changes to just X.Y version numbers?

lib/src/facts/osx/operating_system_resolver.cc
```diff
@@ -33,8 +36,21 @@ namespace facter { namespace facts { namespace osx {
                 result.osx.version = move(value);
             }
 
+            // Populate the major and minor version facts using kernel_release
+            auto release = facts.get<string_value>(fact::kernel_release);
```
peterhuene: Does this belong in the lambda?

peterhuene:
Looking over #996, I think the more general approach makes sense.

whopper:
@peterhuene sounds good - I'm going to close this in favor of Branan's PR.

lib/src/facts/osx/operating_system_resolver.cc
whopper: @peterhuene Doh, at first I thought I was going to be using sw_vers to get the version but then realized it had to come from the kernel facts, and just forgot to move this bit out of the lambda :P

PR #996, opened 2015-06-26
(FACT-1073) Add major/minor OS release facts for distros that lack them

branan:
Previously, if a given OS did not have a proper OS release, we would
use the kernel release. This release would never be split into
major/minor versions.

This now splits that kernel release into major/minor versions, as
Facter 2 did. For platforms that have their own os release
implementation, this will continue to behave as it did before.

Updating the test for this is a simple matter of removing major/minor
from the fixture data so that we can see it has been calculated by the
resolver.

### DIFF
--- a/lib/src/facts/resolvers/operating_system_resolver.cc
+++ b/lib/src/facts/resolvers/operating_system_resolver.cc
@@ -1,5 +1,6 @@
 #include <internal/facts/resolvers/operating_system_resolver.hpp>
 #include <internal/util/regex.hpp>
+#include <internal/util/versions.hpp>
 #include <facter/facts/scalar_value.hpp>
 #include <facter/facts/map_value.hpp>
 #include <facter/facts/collection.hpp>
@@ -7,6 +8,7 @@
 #include <facter/facts/os.hpp>
 
 using namespace std;
+using namespace facter::util;
 
 namespace facter { namespace facts { namespace resolvers {
 
@@ -56,6 +58,12 @@ namespace facter { namespace facts { namespace resolvers {
 
         if (!data.release.empty()) {
             auto value = make_value<map_value>();
+
+            // When we have no major or minor, do a 'trivial' parse of
+            // the release to try to get SOMETHING out of it.
+            if (data.minor.empty() && data.major.empty()) {
+                 std::tie(data.major, data.minor) = versions::major_minor(data.release);
+            }
 
             if (!data.major.empty()) {
                 facts.add(fact::operating_system_major_release, make_value<string_value>(data.major, true));
--- a/lib/inc/internal/util/versions.hpp
+++ b/lib/inc/internal/util/versions.hpp
@@ -1,0 +1,34 @@
+/**
+ * @file
+ * Defines helpers for parsing various version strings
+ */
+#pragma once
+
+#include <string>
+#include <tuple>
+
+namespace facter { namespace util { namespace versions {
+    /**
+     * Helper function for parsing X.Y from an arbitrary version
+     * string. If there is a .Z component, it will be ignored.
+     * @param version the version string to parse
+     * @return A tuple of <maj, min>
+     */
+    inline std::tuple<std::string, std::string> major_minor(const std::string& version)
+    {
+         std::string major, minor;
+         auto pos = version.find('.');
+         if (pos != std::string::npos) {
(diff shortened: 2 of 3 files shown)


whopper:
If we decide to take this more general approach, this change would also resolve https://github.com/puppetlabs/facter/pull/995, which implements this for OSX only.

branan:
This seemed like the best option - if we don't have major/minor by the time we've finished the rest of the processing, just split it out.

If this works for OSX as well, all the better :)

lib/src/facts/resolvers/operating_system_resolver.cc
```diff
@@ -57,6 +57,18 @@ namespace facter { namespace facts { namespace resolvers {
...
+            // When we have no major or minor, do a 'trivial' parse of
+            // the release to try to get SOMETHING out of it.
+            if(data.minor.empty() && data.major.empty()) {
+                 auto begin = data.release.begin();
+                 auto end = data.release.end();
+                 auto end_first = find(begin, end, '.');
+                 auto begin_second = next(end_first);
+                 auto end_second = find(begin_second, end, '.');
+                 data.major = std::string(begin, end_first);
+                 data.minor = std::string(begin_second, end_second);
```
whopper: @branan check out https://github.com/whopper/facter/blob/FACT-1056/osx_release/lib/src/facts/osx/operating_system_resolver.cc#L42-50

We've actually had this functionality before one of the OS refactors, and that was how it was originally implemented. Could be a bit cleaner.

branan: Ah, that uses indexes instead of iterators. I'm not particular dead-set on iterators here, so could definitely use that code instead of we think it's more readable (and honestly it probably is).

branan:
@whopper updated to use indices instead of iterators

branan:
Hmm, coverage decreased. I'll see about adding a new test to cover this logic.

peterhuene:
The coverage sometimes fluctuates by a very small percentage.  It's within the coveralls tolerance, though.

peterhuene:
I'm :+1: on this PR.  Not sure if @branan is going to add more test coverage, but I think it's good to go as-is.

branan:
@peterhuene I'm currently working on a test that covers this change specifically (having a release but no major/minor calculated from it). Even if the coverage is within tolerance I don't think this codepath is being tested with the current suite.

peterhuene:
@branan Good point; I made the incorrect assumption we already covered having a release without major and minor set, but clearly not if this was regressed.  You're right, we definitely need a test to cover this.

lib/src/facts/resolvers/operating_system_resolver.cc
```diff
@@ -57,6 +57,23 @@ namespace facter { namespace facts { namespace resolvers {
         if (!data.release.empty()) {
             auto value = make_value<map_value>();
 
+            // When we have no major or minor, do a 'trivial' parse of
```
MikaelSmith: I'd rather break this out as a separate method, such as parse_simver.

peterhuene: +1 to that.

branan: Word. We probably are doing similar "split on a period" logic in a bunch of places... We should at some point see where we can de-duplicate, but for now I'll add a method to utils and use it here.

lib/src/facts/resolvers/operating_system_resolver.cc
```diff
@@ -57,6 +57,23 @@ namespace facter { namespace facts { namespace resolvers {
...
+            // the release to try to get SOMETHING out of it.
+            if (data.minor.empty() && data.major.empty()) {
+                 auto pos = data.release.find('.');
+                 if (pos != string::npos) {
+                      auto second = data.release.find('.', pos+1);
+                      decltype(second) end;
```
MikaelSmith: What is this line doing?

lib/src/facts/resolvers/operating_system_resolver.cc
MikaelSmith: Similar logic exists in resolvers::kernel_resolver::parse_version.

branan: that is unfortunately different :(

PR #997, opened 2015-06-26
(FACT-867) Port xendomains fact

MikaelSmith:
Re-implements the xendomains fact in Facter 3.


peterhuene:
Looks like there are legitimate build failures in Travis.

MikaelSmith:
Manually verified that if the system looks like we expect a xen0 machine to look like, it should work.

lib/src/facts/posix/xen_resolver.cc
```diff
@@ -0,0 +1,45 @@
...
+
+        bs::error_code ec;
+        if (exists(xen_toolstack, ec) && !ec) {
+            bool success;
+            string output, error;
+            tie(success, output, error) = execute(xen_toolstack);
+            if (success) {
+                return output;
+            } else {
+                LOG_DEBUG("failure executing %1%: %2%", xen_toolstack, error);
```
peterhuene: If xen-toolstack fails, should we fallback to xl/xm or not?

MikaelSmith: I followed the original logic that said no. I'm open to argument, but not sure what the case is for having xen-toolstack installed.

lib/src/facts/resolvers/xen_resolver.cc
```diff
@@ -0,0 +1,81 @@
...
+            static boost::regex domain_header("^(Name|Domain-0)");
+            static boost::regex domain_entry("^([^\\s]*)\\s");
+            execution::each_line(command, {"list"}, [&](string& line) {
+                string domain;
+                if (!boost::regex_match(line, domain_header) && util::re_search(line, domain_entry, &domain)) {
+                    result.domains.emplace_back(move(domain));
+                }
+                return true;
+            }, [&](string& line) {
+                LOG_DEBUG("output on stderr running %1% list: %2%", command, line);
```
peterhuene: I think you can pass in `nullptr` (or omit the stderr callback as nullptr is the default) and it'll default to redirecting stderr to null when not debug level and automatically output the stderr lines (prefixed with `!!!` instead of `|`) when debug level.

lib/src/facts/posix/xen_resolver.cc
peterhuene: :+1: for parity.

lib/src/facts/resolvers/xen_resolver.cc
peterhuene: Said functionality: https://github.com/puppetlabs/facter/blob/master/lib/src/execution/execution.cc#L196

MikaelSmith: Ah right, forgot that was there.

peterhuene:
:+1: will merge after CI.

PR #1001, opened 2015-06-30
(maint) Use correct manpage path on OpenBSD

jasperla:
Non-base software gets installed with a prefix of `/usr/local`, however the manpage path (`man/`) differs from base (`share/man/`).


peterhuene:
@jasperla Thanks for the contribution!

jasperla:
Thanks a lot for the merges!

PR #1002, opened 2015-06-30
(maint) Do not enable pre-compiled headers on OpenBSD also

jasperla:


peterhuene:
Hi @jasperla.  Thanks for the contribution!  We may just remove cotire entirely; let me discuss this with @MikaelSmith.

MikaelSmith:
I'll merge this. We're likely going to remove cotire entirely, but I'm not sure it'll happen today.

PR #1005, opened 2015-07-01
(maint) Remove cotire

MikaelSmith:
Cotire has become more of a maintenance burden than useful. We've had to
turn off unity builds pretty much everywhere because making them work
was too complex. We keep having to disable precompiled headers because
they don't work on specific platforms.

Remove cotire as its not worth the hastle to give us precompiled headers
on a few platforms.


MikaelSmith:
On review, verify that Travis and AppVeyor build performance hasn't regressed significantly.

PR #1007, opened 2015-07-01
(FACT-891) Extend check to unbreak compilation on OpenBSD

jasperla:
It'd be great if this could be committed while the underlying issues as described in FACT-891 are being resolved.

### DIFF
--- a/lib/src/util/posix/dynamic_library.cc
+++ b/lib/src/util/posix/dynamic_library.cc
@@ -41,10 +41,10 @@ namespace facter { namespace util {
         close();
 
         // Don't actually perform a load to determine if it is already loaded
-#ifdef _AIX
+#if !defined(RTLD_NOLOAD)
         // HACK HACK HACK HACK HACK
 
-        // On AIX, RTLD_NOLOAD doesn't exist. This gets things to compile
+        // On AIX/OpenBSD, RTLD_NOLOAD doesn't exist. This gets things to compile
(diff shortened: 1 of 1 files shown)


lib/src/util/posix/dynamic_library.cc
```diff
@@ -41,10 +41,10 @@ namespace facter { namespace util {
         close();
 
         // Don't actually perform a load to determine if it is already loaded
-#ifdef _AIX
+#if defined(_AIX) || !defined(RTLD_NOLOAD)
```
peterhuene: Can probably remove the `_AIX` check now, as `RTLD_NOLOAD` won't be defined in either case.

I think we can probably remove the usage of `RTLD_NOLOAD` entirely, as we search for exported ruby symbols in POSIX and for loaded modules by name on Windows to find the existing library before attempting to load it.  This code existed before those checks were made.

jasperla:
Ok, for the moment I've adjusted the checks. Though your rationale for removing `RTLD_NOLOAD` entirely makes sense.

MikaelSmith:
:+1:

MikaelSmith:
Actually, @branan should this work for AIX?

branan:
This will also work on AIX, yes.

Obviously the issues with facter loading ruby (and vice-versa) won't be resolved, and this is going to behave badly in those situations.

PR #1008, opened 2015-07-02
(FACT-1083) Replace fork/execv with vfork/execve.

peterhuene:
To assist running facter from puppetserver where it is consuming a large
amount of memory in a constrained environment, we should use vfork over
fork so that the parent process' page tables aren't copied into the
child.  This allows facter to spawn additional processes from
puppetserver even when there isn't enough memory to commit to a forked
puppetserver process.

The downside of vfork is that it is a very unsafe system call.  It
creates a child process like fork does, but instead of copying the
parent process' page table, the child process shares the parent process'
address space entirely.  That means that the child can very easily
corrupt the parent.

Therefore, it is with great care that we control exactly what code the
child process runs.  It should not modify any program state, allocate
memory, throw exceptions, or do anything other than duping and closing
file descriptors before calling execve.  Also note that unlike fork,
vfork implementations suspend the parent until the child calls an exec
function or _exit.  This is so that the child can call functions and
append to the current frame's stack space (it must modify the stack
above the current stack pointer, though).

Note: we no longer need the hacky ruby::api::cleanup flag because the
child process now calls _exit when it fails to execve, which prevents
global destructors from running in the child process; therefore
preventing ruby from attempting to shutdown from the child process.

### DIFF
--- a/lib/src/execution/posix/execution.cc
+++ b/lib/src/execution/posix/execution.cc
@@ -67,6 +67,14 @@ namespace facter { namespace execution {
         command_timedout = true;
     }
 
+    static string format_error(string const& message = string(), int error = errno)
+    {
+        if (message.empty()) {
+            return (boost::format("%1% (%2%)") % strerror(error) % error).str();
+        }
+        return (boost::format("%1%: %2% (%3%).") % message % strerror(error) % error).str();
+    }
+
     string which(string const& file, vector<string> const& directories)
     {
         // If the file is already absolute, return it if it's executable
@@ -132,7 +140,7 @@ namespace facter { namespace execution {
             int result = select(max + 1, &set, nullptr, nullptr, timeout ? &read_timeout : nullptr);
             if (result == -1) {
                 if (errno != EINTR) {
-                    LOG_ERROR("select call failed: %1% (%2%).", strerror(errno), errno);
+                    LOG_ERROR(format_error("select call failed"));
                     throw execution_exception("failed to read child output.");
                 }
                 // Interrupted by signal
@@ -153,7 +161,7 @@ namespace facter { namespace execution {
                 auto count = read(pipe.descriptor, &pipe.buffer[0], pipe.buffer.size());
                 if (count < 0) {
                     if (errno != EINTR) {
-                        LOG_ERROR("%1% pipe read failed: %2% (%3%).", pipe.name, strerror(errno), errno);
+                        LOG_ERROR("%1% pipe read failed: %2%.", pipe.name, format_error());
                         throw execution_exception("failed to read child output.");
                     }
                     // Interrupted by signal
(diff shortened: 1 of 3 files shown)


peterhuene:
This diff actually looks worse than it is.  The parent code is entirely unchanged, it just moved down below the child code so that nothing unintentional gets inserted between the `vfork` call and calling `run_child`.

peterhuene:
Erm.  I forgot to update the fork failure error message.  Let me push one.

peterhuene:
Looks like a build failure on release for unused result.  Let me push up a quick fix.

lib/src/execution/posix/execution.cc
```diff
@@ -176,6 +178,131 @@ namespace facter { namespace execution {
...
+
+        // Execute the given program; this should not return if successful
+        execve(program, const_cast<char* const*>(argv), const_cast<char* const*>(envp));
+    }
+
+    // Helper function that turns a vector of strings into a vector of const cstr pointers
+    // This is used to pass arguments and environment to execve
+    static vector<char const*> to_exec_arg(vector<string> const* argument, string const* first = nullptr)
+    {
+        vector<char const*> result((argument ? argument->size() : 0) + (first ? 1 : 0) + 1 /* terminating null */);
```
MikaelSmith: For some reason I have a preference for: reserve, push_back first, transform with back_inserter, push_back null. But they do pretty much identical things.

lib/src/execution/posix/execution.cc
```diff
@@ -199,226 +326,142 @@ namespace facter { namespace execution {
         // Create the pipes for stdin/stdout redirection
         int pipes[2];
         if (::pipe(pipes) < 0) {
-            throw execution_exception("failed to allocate pipe for stdin redirection.");
+            throw execution_exception((boost::format("failed to allocate pipe for stdin redirection: %1% (%2%).") % strerror(errno) % errno).str());
```
MikaelSmith: If this is something we're going to do more, I like pulling it out into a helper function. I did so on Windows at https://github.com/puppetlabs/leatherman/blob/master/windows/src/system_error.cc.

MikaelSmith:
:+1: otherwise.

lib/src/execution/posix/execution.cc
MikaelSmith: It helps enforce a consistent format for reporting error numbers.

peterhuene: I'll see what I can do to scrub that in various places.

lib/src/execution/posix/execution.cc
peterhuene: I can probably clean this up.

peterhuene:
@MikaelSmith Pushed up some fixes.

PR #1010, opened 2015-07-02
(FACT-1085) Fix regressions in custom fact command execution from 2.x.

peterhuene:
This fixes a few regressions that were noticed when executing commands between 2.x and 3.x.

* 3.x was incorrectly redirecting stderr to stdout for command execution.
* 3.x was incorrectly returning empty string fact values when setcode with a command returned no output on stderr; it should not resolve the fact instead.
* 3.x was not properly handling non-existent commands such that they should not be executed, instead nil should be returned for `Facter::Core::Execution::exec` and `Facter::Core::Execution::execute` should raise a command not found error by default.

### DIFF
--- a/lib/src/execution/windows/execution.cc
+++ b/lib/src/execution/windows/execution.cc
@@ -78,6 +78,11 @@ namespace facter { namespace execution {
             return is_executable(p, &helper) ? p.string() : string();
         }
 
+        // On Windows, treat 'echo' as a command that can be found
+        if (file == "echo") {
+            return "echo";
+        }
+
         // Otherwise, check for an executable file under the given search paths
         for (auto const& dir : directories) {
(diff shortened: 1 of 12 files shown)


peterhuene:
Looking into the Windows failures.  Getting commands that run both on Windows and POSIX is a pain :/

lib/src/execution/windows/execution.cc
```diff
@@ -78,6 +78,11 @@ namespace facter { namespace execution {
             return is_executable(p, &helper) ? p.string() : string();
         }
 
+        // On Windows, treat 'echo' as a command that can be found
```
MikaelSmith: That's an entertaining special case.

MikaelSmith:
:+1:

PR #1011, opened 2015-07-02
(FACT-1084) Catch failures to convert address

MikaelSmith:
Without this change, Facter calls WSAAddressToStringW with an invalid
SOCKET_ADDRESS and throws an uncaught exception when the call fails. This
leads Facter to exit ungracefully.

Ensure calls to WSAAddressToStringW catch any exceptions thrown as a
result of errors, so we simply fail to resolve the address.

Also extend allowed IP address length on Windows to account for null
terminator, and skip resolving dhcp if not enabled.

### DIFF
--- a/lib/src/facts/windows/networking_resolver.cc
+++ b/lib/src/facts/windows/networking_resolver.cc
@@ -207,11 +207,28 @@ namespace facter { namespace facts { namespace windows {
             // Only supported on platforms after Windows Server 2003.
             if (pCurAddr->Flags & IP_ADAPTER_DHCP_ENABLED && pCurAddr->Length >= sizeof(IP_ADAPTER_ADDRESSES_LH)) {
                 auto adapter = reinterpret_cast<IP_ADAPTER_ADDRESSES_LH&>(*pCurAddr);
-                net_interface.dhcp_server = winsock.saddress_to_string(adapter.Dhcpv4Server);
+                if (adapter.Flags & IP_ADAPTER_DHCP_ENABLED) {
+                    try {
+                        net_interface.dhcp_server = winsock.saddress_to_string(adapter.Dhcpv4Server);
+                    } catch (wsa_exception &e) {
+                        LOG_DEBUG("failed to retrieve dhcp v4 server address for %1%: %2%", net_interface.name, e.what());
+                    }
+                }
             }
 
             for (auto it = pCurAddr->FirstUnicastAddress; it; it = it->Next) {
-                string addr = winsock.saddress_to_string(it->Address);
+                string addr;
+                try {
+                    addr = winsock.saddress_to_string(it->Address);
+                } catch (wsa_exception &e) {
+                    string iptype =
+                        (it->Address.lpSockaddr->sa_family == AF_INET) ? " v4"
+                        : (it->Address.lpSockaddr->sa_family == AF_INET6) ? " v6"
+                        : "";
+                    LOG_DEBUG("failed to retrieve ip%1% address for %2%: %3%",
+                        iptype, net_interface.name, e.what());
+                }
+
                 if (addr.empty()) {
                     continue;
                 }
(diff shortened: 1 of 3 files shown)


MikaelSmith:
@Iristyle @cyberious

lib/src/facts/windows/networking_resolver.cc
```diff
@@ -207,11 +207,28 @@ namespace facter { namespace facts { namespace windows {
...
             for (auto it = pCurAddr->FirstUnicastAddress; it; it = it->Next) {
-                string addr = winsock.saddress_to_string(it->Address);
+                string addr;
+                try {
+                    addr = winsock.saddress_to_string(it->Address);
+                } catch (wsa_exception &e) {
+                    string iptype =
+                        (it->Address.lpSockaddr->sa_family == AF_INET) ? "v4"
+                        : (it->Address.lpSockaddr->sa_family == AF_INET6) ? "v6"
+                        : "";
```
peterhuene: nit picky, but this will add an extra space to the message, e.g. `failed to retrieve ip  address for...`.  Perhaps add a trailing space to the above strings and remove the trailing space in the format message?

lib/src/facts/windows/networking_resolver.cc
```diff
@@ -207,11 +207,28 @@ namespace facter { namespace facts { namespace windows {
...
+                    }
+                }
             }
 
             for (auto it = pCurAddr->FirstUnicastAddress; it; it = it->Next) {
-                string addr = winsock.saddress_to_string(it->Address);
+                string addr;
+                try {
+                    addr = winsock.saddress_to_string(it->Address);
+                } catch (wsa_exception &e) {
```
peterhuene: So there's no good way to check that the address translation is valid before attempting it?  I dislike having to handle exceptions in normal use cases, but if there's no other option, c'est la vie.

peterhuene:
I'm :+1: waiting on review from the Windows team.

lib/src/facts/windows/networking_resolver.cc
MikaelSmith: I can check if it->Address.lpSockaddr is null, since that's what's causing the failure in this case.

lib/src/facts/windows/networking_resolver.cc
MikaelSmith: Yeah, I'll clean it up.

lib/src/facts/windows/networking_resolver.cc
MikaelSmith: Could also drop using exceptions and return an empty string in case of error, but then you potentially lose the Windows error code.

MikaelSmith: I updated line 210 to check that the Dhcpv4Server sockaddr is initialized. I think the unicast addresses should always be initialized according to documentation, so this still seems like a reasonable exception case.

lib/src/util/windows/wsa.cc
```diff
@@ -40,8 +40,12 @@ namespace facter { namespace util { namespace windows {
...
     string wsa::saddress_to_string(SOCKET_ADDRESS const& addr) const
     {
-        DWORD size = INET6_ADDRSTRLEN;
-        wchar_t buffer[INET6_ADDRSTRLEN];
+        if (!addr.lpSockaddr) {
+            return {};
+        }
+
+        DWORD size = INET6_ADDRSTRLEN+1;
+        wchar_t buffer[INET6_ADDRSTRLEN+1];
```
Iristyle: FWIW, it looks like [GetNameInfoW](https://msdn.microsoft.com/en-us/library/windows/desktop/ms738531\(v=vs.85\).aspx) could provide you with more detail about potential errors if you want to do more validation of the address.

MikaelSmith: I'm tempted not to, because I think we're unlikely to run into this error as long as the sockaddr is initialized. It's a pretty simple structure.

peterhuene:
:+1: re: recent changes.

peterhuene:
@Iristyle If you're :+1:, I'll merge.

Iristyle:
@peterhuene the build that @MikaelSmith passed off to me worked great, so I'm :+1:

PR #1013, opened 2015-07-06
(FACT-950) Un-vendor leatherman

branan:
submodules--

This should be paired with the related updates to our puppet-agent
build script and to Leatherman itself to make everything build
properly in CI. We should also discuss the workflow for
building/installing Leatherman locally for Facter to find.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "vendor/leatherman"]
-	path = vendor/leatherman
-	url = https://github.com/puppetlabs/leatherman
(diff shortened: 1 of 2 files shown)


branan:
It looks like the Travis situation can be handled with travis Github releases - it will only push up on a tag, which is perfect for us. I'm not sure what the appveyor step would be.

So the updates to this PR will be to the travis build script to unpack the latest release from leatherman.

Leatherman travis will need to be updated to generate a zip of an installed leatherman PREFIX

branan:
I'm closing this out for now until we're ready to actually deal with this properly

PR #1012, opened 2015-07-02
(FACT-1074) Update external facts tests for default facts.d / custom facts.d interaction

whopper:
This commit adds two new tests to our suite of external facts tests. The
first ensures that given a user specified external facts directory with
an external fact, both it and any facts in the default facts.d dir will
be resolved. The second ensures that given an external fact in a user
specified directory which has the same name as a fact in the default
facts.d directory, the former will win and be resolved.

PR #1014, opened 2015-07-06
(FACT-1074) Update external facts tests for default facts.d / custom facts.d interaction

whopper:
This commit adds two new tests to our suite of external facts tests. The
first ensures that given a user specified external facts directory with
an external fact, both it and any facts in the default facts.d dir will
be resolved. The second ensures that given an external fact in a user
specified directory which has the same name as a fact in the default
facts.d directory, the former will win and be resolved.


whopper:
Retargeted version of https://github.com/puppetlabs/facter/pull/1012

@MikaelSmith ^

MikaelSmith:
:+1:

PR #1016, opened 2015-07-06
(maint) Fix tests on Windows Server 2003 (again)

whopper:
Windows Server 2003 still uses the 'Documents and Settings' directory as
part of the user's home directory. The acceptance update for FACT-1074
did not take the need for quotes into account. This commit updates paths to include the quotes.

[skip ci]


MikaelSmith:
I'm pretty sure the `rm -f` at line 93 is ignoring the fact it's failing because `-f`. We should change that to individually quote file names.

whopper:
@MikaelSmith heh good call! Look sane now?

PR #1020, opened 2015-07-08
(FACT-1088) Update acceptance to validate Lucid

whopper:


acceptance/tests/facts/ubuntu.rb
```diff
@@ -28,12 +32,12 @@
     os_kernel  = '3.19'
   end
 
-  if agent['platform'] =~ /x86_64/
+  if agent['platform'] =~ /x86_64|amd64/
     os_arch     = 'amd64'
-    os_hardware = 'x86_64'
+    os_hardware = /x86_64|unknown/
```
MikaelSmith: Hardware is `unknown` on `lucid`? I'd rather break this out than have regex `|`.

whopper:
@MikaelSmith indeed- `uname -p` returns `unknown`, same as a few Debian versions. I'll update to break that out for Lucid.

PR #1021, opened 2015-07-08
(FACT-1090) Query Ruby `$LOAD_PATH` when used

MikaelSmith:
Facter 3.0 included a regression where running Facter in some Puppet
commands would not include the correct Ruby `$LOAD_PATH`, resulting in
missing facts. This happens because Facter caches the `$LOAD_PATH` when
initialized, and only resets it when `Facter.reset` is called.

Fix the regression by updating Facter to query `$LOAD_PATH` when facts
are loaded instead of caching on initialization.

### DIFF
--- a/lib/inc/internal/ruby/api.hpp
+++ b/lib/inc/internal/ruby/api.hpp
@@ -387,7 +387,7 @@ namespace facter {  namespace ruby {
          * Gets the load path being used by Ruby.
          * @return Returns the load path being used by Ruby.
          */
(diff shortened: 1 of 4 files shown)


MikaelSmith:
@peterhuene got a little more complex, but I think this replicates Facter 2.x behavior.

MikaelSmith:
I think I can add some tests that hack around with Ruby's $LOAD_PATH in Facts (that's going to be ugly). I'll see how it goes.

PR #1023, opened 2015-07-08
WIP (FACT-1090) Query Ruby `$LOAD_PATH` when used

MikaelSmith:
Facter 3.0 included a regression where running Facter in some Puppet
commands would not include the correct Ruby `$LOAD_PATH`, resulting in
missing facts. This happens because Facter caches the `$LOAD_PATH` when
initialized, and only resets it when `Facter.reset` is called.

Fix the regression by updating Facter to query `$LOAD_PATH` when facts
are loaded instead of caching on initialization.

### DIFF
--- a/lib/inc/internal/ruby/api.hpp
+++ b/lib/inc/internal/ruby/api.hpp
@@ -387,7 +387,7 @@ namespace facter {  namespace ruby {
          * Gets the load path being used by Ruby.
          * @return Returns the load path being used by Ruby.
          */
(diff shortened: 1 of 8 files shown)


peterhuene:
I like this implementation more than the initial one.  If we don't go with fixing this in Puppet, I'm :+1: with this.

MikaelSmith:
Ticket has details; closing this in favor of a fix in Puppet.

PR #1024, opened 2015-07-08
(maint) Use TravisCI container-based builds

MikaelSmith:
Enable TravisCI container-based builds by avoiding (and explicitly
disabling) sudo and using the APT addon instead.

Normally pip install requires root access, which is incompatible with
TravisCI container-based builds. Use `pip install --user` instead to
install in user-space.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,45 +1,58 @@
 language: cpp
 compiler:
   - gcc
+sudo: false
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - boost-latest
+    packages:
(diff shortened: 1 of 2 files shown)


MikaelSmith:
Reminder: file a bug against cppcheck and push up a /tmp/userdir build of cppcheck-1.68

MikaelSmith:
This needs a fix in leatherman that LTH-14 is already adding, so blocked on that.

MikaelSmith:
Merged to master as 553a0a98be1461d74d714c8a3bd43d463737dfa5.

PR #1025, opened 2015-07-08
(maint) Install dmg on osx

joshcooper:
Install puppet-agent dmg on osx during aio acceptance, which depends on
beaker 2.16 or later.

Updates the expected ruby platform regex to allow for versions of the
form "x86_64-darwin14.0". Note 10.9 and up are only 64-bit, so we can
assume x86_64.

### DIFF
--- a/acceptance/Gemfile
+++ b/acceptance/Gemfile
@@ -1,6 +1,6 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
-gem "beaker", "~> 2.8"
+gem "beaker", "~> 2.16"
 gem 'rake', "~> 10.1.0"
 gem "multi_json", "~> 1.8"
 
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -33,11 +33,19 @@ PACKAGES = {
 install_packages_on(hosts, PACKAGES)
 
 agents.each do |agent|
-  if agent['platform'] =~ /windows/
+  case agent['platform']
+  when /windows/
     arch = agent[:ruby_arch] || 'x86'
     base_url = ENV['MSI_BASE_URL'] || "http://builds.puppetlabs.lan/puppet-agent/#{ENV['SHA']}/artifacts/windows"
     filename = ENV['MSI_FILENAME'] || "puppet-agent-#{arch}.msi"
 
     install_puppet_from_msi(agent, :url => "#{base_url}/#{filename}")
+  when /osx/
+    opts = {
+      :puppet_collection => 'PC1',
+      :puppet_agent_sha => ENV['SHA'],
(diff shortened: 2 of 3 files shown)


joshcooper:
This enables acceptance tests to run and all but one test consistently pass 10.10 and 10.9, e.g.

```
$ be rake ci:test:aio SHA=65e8a370d306a36660ad181a22c120911ab20e44 SUITE_VERSION=1.2.1.32.g65e8a37 CONFIG=config/nodes/osx-1010-x86_64.yaml
...
       Total Suite Time: 7.49 seconds
      Average Test Time: 0.31 seconds
              Attempted: 24
                 Passed: 20
                 Failed: 0
                Errored: 0
                Skipped: 4
                Pending: 0
                  Total: 24
```

However, the verify_facts tests sometimes fails with the uptime fact:

```
$ be rake ci:test:aio SHA=65e8a370d306a36660ad181a22c120911ab20e44 SUITE_VERSION=1.2.1.32.g65e8a37 CONFIG=config/nodes/osx-109-x86_64.yaml
...
system_uptime:
  days: 0
  hours: 7
  seconds: 26929
  uptime: 7:28 hours
timezone: PDT
virtual: vmware

qh5a6eyt44bp61l.delivery.puppetlabs.net (agent) executed in 0.41 seconds
#<RuntimeError: type of uptime did not match schema, 26940 is not a string>
```

MikaelSmith:
:+1:

joshcooper:
@MikaelSmith do you know what's up with uptime?

MikaelSmith:
What's up with uptime? According to http://docs.puppetlabs.com/facter/3.0/core_facts.html#systemuptime, each entry in system_uptime is the full uptime in floor(seconds), or in floor(hours), or in floor(days). They aren't combined to equal uptime.

MikaelSmith:
So SUITE_VERSION is now a requirement to run Mac acceptance? Is this considered temporary or long-term?

joshcooper:
SUITE_VERSION is currently required because we don't have an easy way to know the `git describe` used in the puppet-agent filename. Linux platforms just need the SHA to find the repo_config, which then knows the puppet-agent filename. On Windows, we create a hardlink, e.g. puppet-agent-x64.msi -> puppet-agent-<git describe>-x64.msi, avoiding the issue. Release will be doing something similar for platforms without package managers, so long term we can eliminate SUITE_VERSION, but that's going to be awhile.

The alternative is to have beaker download the build data yaml, and extract the relevant git describe info. But that requires beaker changes. Alternatively, we could implement that logic in puppet/facter/hiera, but I don't want to accrue more tech debt.

joshcooper:
@MikaelSmith about uptime, I was wondering whether you knew why the acceptance test *sometimes* fails?

MikaelSmith:
Oh, not sure why it would sometimes fail. If it doesn't think a 5-digit number is a string I'd expect it to always fail. It expects a string, which means uit parses the YAML output and then expects `<parsed_value>.is_a? String` to be true.

PR #1026, opened 2015-07-08
(FACT-1113) Move dlopen wrapper class to Leatherman

Magisus:
Updates facter to use Leatherman's dynamic library class.


lib/inc/internal/ruby/api.hpp
```diff
@@ -21,6 +21,8 @@ namespace facter { namespace facts {
 
 namespace facter {  namespace ruby {
 
+namespace lth_dl = leatherman::dynamic_library;
```
peterhuene: We shouldn't alias namespaces in headers if we can avoid it.  I agree that `leatherman::dynamic_library::dynamic_library` is a bit verbose and redundant, but we could alias it properly in the source file and only have to use the fully-qualified name a few times in the header.  If we do prefer to alias a namespace, I'd prefer something like `lm_dl`.  `lth` looks a little strange to me.

lib/src/ruby/posix/api.cc
```diff
@@ -1,13 +1,12 @@
 #include <internal/ruby/api.hpp>
 
 using namespace std;
-using namespace facter::util;
```
peterhuene: Can probably add a `using namespace leatherman::dynamic_library` here instead of qualifying it.

lib/src/ruby/windows/api.cc
```diff
@@ -1,14 +1,13 @@
 #include <internal/ruby/api.hpp>
 
 using namespace std;
-using namespace facter::util;
```
peterhuene: Same comment as above.

peterhuene:
Other than above comments, :+1:

lib/inc/internal/ruby/api.hpp
Magisus: I have been using lth_xx throughout, so for consistency's sake I will continue to use it for now (I'll remove it from headers though). If we really want to change it to lm_xx, that is a task for another time, because it should be done everywhere.

PR #1028, opened 2015-07-09
(PUP-4757) added :peaio rake task to test PE installed AIO agents on PE-only platforms

kevpl:


MikaelSmith:
This is all acceptance setup, so you could use `[skip ci]` in the commit to avoid waiting on Travis/AppVeyor.

MikaelSmith:
@kevpl are the master config changes really needed? We don't test master interaction in Facter accptance, and I just put up a #1029 to remove them all.

peterhuene:
I assume there isn't a corresponding FACT ticket?  We generally don't like referencing PUP tickets for facter if we can avoid it.

PR #1030, opened 2015-07-09
(maint) Relax validation of uptime related facts

joshcooper:
On OSX, the system profiler returns uptime as either '5 minutes' or
'1:23', where the latter is used if the system has been up for at least
an hour. The verify_facts acceptance test does a YAML.load of facter's
output, automatically converting '1:23' to 4980 (of type Fixnum), which
doesn't match the schema.

This commit relaxes the validation to allow uptime related facts to be
strings or fixnums. On OSX, this can happen with the
system_profiler/uptime fact and the legacy sp_uptime fact, so we regex
on the fact name.

Also outputs the value's type when we fail.

### DIFF
--- a/acceptance/tests/verify_facts.rb
+++ b/acceptance/tests/verify_facts.rb
@@ -29,7 +29,10 @@ def validate_fact(name, node, fact_value, hidden)
               # YAML-CPP seems to drop the decimal whenever it feels like it, so just match Int too.
               fact_value.is_a? Float or fact_value.is_a? Integer or fact_value.to_s =~ /^(\.inf|\.nan|-\.inf)$/
             when 'string'
-              fact_value.is_a? String
+              # YAML.load will automatically convert timestamp values to a Fixnum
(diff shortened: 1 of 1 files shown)


joshcooper:
```
$ be rake ci:test:aio SHA=65e8a370d306a36660ad181a22c120911ab20e44 SUITE_VERSION=1.2.1.32.g65e8a37 CONFIG=osx.yaml TESTS=tests/verify_facts.rb
...
sp_uptime: 3:50
...
system_profiler:
  ...
  uptime: 3:50
  username: System Administrator (root)
system_uptime:
  days: 0
  hours: 3
  seconds: 13829
  uptime: 3:50 hours
...
       Total Suite Time: 1.28 seconds
      Average Test Time: 1.28 seconds
              Attempted: 1
                 Passed: 1
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 1
```

PR #1032, opened 2015-07-10
(maint) Restore setting CMAKE_INSTALL_RPATH

MikaelSmith:
CMP0042 means CMAKE_MACOSX_RPATH no longer needs to be set, but we still
need to set the CMAKE_INSTALL_RPATH based on the install prefix if not
going to a system library. Removing it causes Facter to fail to find
libfacter when running from the puppet-agent install location, so
restore setting CMAKE_INSTALL_RPATH.


peterhuene:
:+1:

PR #1033, opened 2015-07-10
(FACT-1111) Bring back `facter -p`

branan:
The beast rises from beyond the grave to once again terrorize the
local villagers.

This brings back our circular dependency on Puppet. When the
`--puppet` or `-p` arguments are passed to Facter, it will attempt to
load Puppet's ruby code in order to find additional facts that may
have been pluginsunc to the agent.

### DIFF
--- a/acceptance/tests/runs_puppet_facts.rb
+++ b/acceptance/tests/runs_puppet_facts.rb
@@ -1,0 +1,30 @@
+test_name "facter -p loads facts from puppet"
+
+agents.each do |agent|
+  external_dir = agent.puppet['pluginfactdest']
+  external_file = "#{external_dir}/external.txt"
+  custom_dir = agent.puppet['plugindest']
+  custom_dir += "/facter"
+  custom_file = "#{custom_dir}/custom.rb"
+
+  teardown do
+    on agent, "rm -f #{external_file} #{custom_file}"
+  end
+
+  step "Agent #{agent}: create external fact"
+  agent.mkdir_p(external_dir)
+  create_remote_file(agent, external_file, "external=external")
+
+  step "Agent #{agent}: create custom fact"
+  agent.mkdir_p(custom_dir)
+  create_remote_file(agent, custom_file, "Facter.add(:custom) { setcode { 'custom' } }")
+
+  step "Agent #{agent}: verify facts"
+  on(agent, facter("-p external")) do
+    assert_equal("external", stdout.chomp)
+  end
+
+  on(agent, facter("-p custom")) do
+    assert_equal("custom", stdout.chomp)
+  end
+end
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -125,6 +125,7 @@ int main(int argc, char **argv)
             ("no-custom-facts", "Disables custom facts.")
             ("no-external-facts", "Disables external facts.")
             ("no-ruby", "Disables loading Ruby, facts requiring Ruby, and custom facts.")
+            ("puppet,p", "(Deprecated: use `puppet facts` instead) Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.")
             ("trace", "Enable backtraces for custom facts.")
             ("verbose", "Enable verbose (info) output.")
             ("version,v", "Print the version and exit.")
@@ -173,7 +174,13 @@ int main(int argc, char **argv)
                 throw po::error("debug, verbose, and log-level options conflict: please specify only one.");
             }
             if (vm.count("no-ruby") && vm.count("custom-dir")) {
-              throw po::error("no-ruby and custom-dir options conflict: please specify only one.");
+                throw po::error("no-ruby and custom-dir options conflict: please specify only one.");
+            }
+            if (vm.count("puppet") && vm.count("no-custom-facts")) {
+                throw po::error("puppet and no-custom-facts options conflict: please specify only one.");
+            }
+            if (vm.count("puppet") && vm.count("no-ruby")) {
+                throw po::error("puppet and no-ruby options conflict: please specify only one.");
             }
         }
         catch (exception& ex) {
@@ -243,7 +250,7 @@ int main(int argc, char **argv)
         facts.add_environment_facts();
 
         if (ruby && !vm.count("no-custom-facts")) {
-            facter::ruby::load_custom_facts(facts, custom_directories);
+            facter::ruby::load_custom_facts(facts, vm.count("puppet"), custom_directories);
         }
 
         // Output the facts
--- a/lib/src/ruby/ruby.cc
+++ b/lib/src/ruby/ruby.cc
@@ -1,12 +1,21 @@
 #include <facter/ruby/ruby.hpp>
+#include <facter/logging/logging.hpp>
 #include <internal/ruby/api.hpp>
 #include <internal/ruby/module.hpp>
 
 using namespace std;
 using namespace facter::facts;
 
-namespace facter { namespace ruby {
+static const char load_puppet[] =
+"require 'puppet'\n"
+"Puppet.initialize_settings\n"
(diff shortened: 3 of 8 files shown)


branan:
```
[branan@branan-thinkpad build]$ bin/facter myfact

[branan@branan-thinkpad build]$ bin/facter -p myfact
2015-07-10 10:47:32.987521 WARN  puppetlabs.facter - Could not load puppet; some facts may be unavailable: cannot load such file -- puppet

[branan@branan-thinkpad build]$ RUBYLIB=/home/branan/proj/pl/puppet/lib bin/facter -p myfact
foobar
```

branan:
This needs a test at the acceptance layer still. Working on that now.

lib/src/ruby/ruby.cc
```diff
@@ -1,11 +1,39 @@
 #include <facter/ruby/ruby.hpp>
+#include <facter/logging/logging.hpp>
 #include <internal/ruby/api.hpp>
 #include <internal/ruby/module.hpp>
 
 using namespace std;
 using namespace facter::facts;
 
+static const string load_puppet = "\
```
branan: I could try to map this to the appropriate rb API constructs, but that is super gross and probably more fragile than just using `rb_eval_string` here

whopper:
@branan I think `facter -p` will conflict with `--no-custom-facts` and `--no-ruby` - should we add a message to https://github.com/branan/facter/blob/bug/stable/fact-1111-facter-p/exe/facter.cc#L160 ?

MikaelSmith:
Oh yeah, using those together should definitely error.

branan:
@whopper @MikaelSmith oh, yup. On it.

lib/src/ruby/ruby.cc
```diff
@@ -18,10 +46,21 @@ namespace facter { namespace ruby {
         return true;
     }
 
-    void load_custom_facts(collection& facts, vector<string> const& paths)
+    void load_custom_facts(collection& facts, bool initialize_puppet, vector<string> const& paths)
     {
         module mod(facts, paths);
+        if (initialize_puppet) {
+            try {
+                eval(load_puppet);
```
MikaelSmith: Have you checked that this works? You're setting `$LOAD_PATH` after `initialize_search_paths` is called from the `module` constructor. You may need to do a `ruby_reset` as well to pick up custom facts in `libdir`.

lib/src/ruby/ruby.cc
```diff
@@ -1,11 +1,39 @@
...
 
 using namespace std;
 using namespace facter::facts;
 
+static const string load_puppet = "\
+require 'puppet'\n\
+Puppet.initialize_settings\n\
+unless $LOAD_PATH.include?(Puppet[:libdir])\n\
+    $LOAD_PATH << Puppet[:libdir]\n\
+end";
```
peterhuene: We still need to fix up `pluginfactsdest`, no?

lib/src/ruby/ruby.cc
peterhuene: Might as well just keep it a `Facter.reset` followed by a `Facter.external_search_path(Puppet[:pluginfactsdest]) if ...` in the ruby code snippet.

lib/src/ruby/ruby.cc
joshcooper: That is both amazing and terrifying at the same time!

lib/src/ruby/ruby.cc
peterhuene: This could just be a `const char*` since we're only passing it directly to Ruby anyway.

lib/src/ruby/ruby.cc
MikaelSmith: That also makes sense.

peterhuene:
Other than noted above, :+1:

daenney:
So a feature is being introduced which is immediately marked as deprecated? It can't be that hard to teach people to use `puppet facts` instead...

peterhuene:
@daenney It's technically not a new feature, just a reinstatement of a previously deprecated feature :) The help message matches what we have in Facter 2.x.  We're fully restoring the previous functionality with a "TODO: completely rethink this in the future".

daenney:
@peterhuene Ah, I see. Sadness though.

kylog:
@daenney many tears have been shed on this one. 

It's hard to strike the balance between moving briskly into the future and easing transitions from the past. Not just here but in many places.

Sigh.

exe/facter.cc
```diff
@@ -173,7 +174,13 @@ int main(int argc, char **argv)
                 throw po::error("debug, verbose, and log-level options conflict: please specify only one.");
             }
             if (vm.count("no-ruby") && vm.count("custom-dir")) {
-              throw po::error("no-ruby and custom-dir options conflict: please specify only one.");
+                throw po::error("no-ruby and custom-dir options conflict: please specify only one.");
+            }
+            if (vm.count("puppet") && vm.count("no-custom-facts")) {
+                throw po::error("puppet and no-custom-facts conflict: please specify only one.");
```
whopper: Might want to make this `puppet and no-custom-facts options conflict...`

exe/facter.cc
```diff
@@ -173,7 +174,13 @@ int main(int argc, char **argv)
...
+            }
+            if (vm.count("puppet") && vm.count("no-ruby")) {
+                throw po::error("puppet and no-ruby conflict: please specify only one.");
```
whopper: Same as above

exe/facter.cc
branan: Word. easy enough.

branan:
@peterhuene addressed your issues.

branan:
* I've confirmed I can load facts from `plugindest` and `pluginfactdest` when using -p but not without.
* I've confirmed that adding -p does not break loading of external facts from `~/.facter/facts.d`
* I've confirmed that adding -p does not break loading of custom facts from the path provided by `--custom-dir`

There are probably a few more combinations of arguments I could teset, but I'm pretty confident this is behaving correctly

MikaelSmith:
Acceptance tests coming in a separate commit?

branan:
@MikaelSmith same PR, new commit. Yup

peterhuene:
:+1:

MikaelSmith:
:+1:

branan:
Test is pusehd up

acceptance/tests/runs_puppet_facts.rb
```diff
@@ -0,0 +1,30 @@
+test_name "facter -p loads facts from puppet"
```
MikaelSmith: Is there a good way to guard on finding puppet? I can envision test environments where we only build Facter to run acceptance.

acceptance/tests/runs_puppet_facts.rb
```diff
@@ -0,0 +1,30 @@
...
+    on agent, "rm -f #{external_file} #{custom_file}"
+  end
+
+  step "Agent #{agent}: create external fact"
+  agent.mkdir_p(external_dir)
+  create_remote_file(agent, external_file, "external=external")
+
+  step "Agent #{agent}: create custom fact"
+  agent.mkdir_p(custom_dir)
+  create_remote_file(agent, custom_file, "Facter.add(:custom) do setcode do 'custom' end end")
```
MikaelSmith: Single lines are nicer with `{}`.

MikaelSmith:
:+1:

acceptance/tests/runs_puppet_facts.rb
whopper: `skip_test "Could not find Puppet" if on(agent, "which puppet") == nil` ?

MikaelSmith: I'm ok merging this and leaving that for later, when we actually have a workflow setup.

PR #1034, opened 2015-07-10
(maint) Restore libfacter binary compatibility

MikaelSmith:
Restores libfacter binary compatibility with the 3.0.0 and 3.0.1
releases.


MikaelSmith:
Since this is only called from exe/facter.cc so far, I chose to prioritize being able to remove the old function while preserving source compatibility over a shorter call signature for showing legacy facts.

PR #1035, opened 2015-07-10
(FACT-1081) Explicitly use make_tuple to prevent any ambiguities

jasperla:
This allows building with gcc 4.9.2 on OpenBSD, including a separate commit to make it more clear what those changes are.


peterhuene:
:+1:

peterhuene:
Verified that this change does not affect the OSX build, as they share the BSD filesystem resolver.

peterhuene:
@jasperla Thanks again for putting up a PR for this and getting things going under FreeBSD!

jasperla:
@peterhuene you mean OpenBSD? :)

peterhuene:
Whoops, my fingers like typing the word "free" :)

PR #1039, opened 2015-07-11
(FACT-1081) Add proper guards for OpenBSD

jasperla:


kylog:
Btw @jasperla, do you have any ideas on how to get CI for PRs on OpenBSD? That would be very nice to ensure we don't regress of course. But I've looked around a little bit and not seen anything. 

My only idea would be to use TravisCI and load an OpenBSD vm, but that seems onerous, and in any event, TravisCI has a timeout IIRC and we might run afoul of that timeout.

jasperla:
@kylog I've updated the guards, all should be good now.

As to CI, I don't know of any service offering support for OpenBSD (or any other BSD) for that matter.
As a side note, it seems it does support OS X, so that could also be used too to expand CI testing.

jasperla:
Betterer? :)

MikaelSmith:
Where did `noatime` come from?

jasperla:
@MikaelSmith the future! (it was part of an upcoming PR); i've zapped it now.

MikaelSmith:
Ok; it should definitely be a separate commit.

jasperla:
That's why I mentioned it's part of an upcoming PR (hint: FACT-1118).

jasperla:
Thank you!

PR #1040, opened 2015-07-11
(FACT-1116) Unbreak bsd::networking_resolver and implement mtu fact

jasperla:
This unbreaks building the `bsd/networking_resolver.cc` and also implements some missing functionality like the link address and link mtu.

### DIFF
--- a/lib/inc/internal/facts/bsd/networking_resolver.hpp
+++ b/lib/inc/internal/facts/bsd/networking_resolver.hpp
@@ -29,7 +29,21 @@ namespace facter { namespace facts { namespace bsd {
          * @param data The data pointer from the link layer interface.
          * @return Returns The MTU of the interface.
          */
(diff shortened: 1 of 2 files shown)


jasperla:
Ugh, given that Linux also tries to compile `bsd/networking_resolver.cc`, would it perhaps makes sense to move OpenBSD specific bits into `openbsd/networking_resolver.cc` instead?

peterhuene:
Yeah, unfortunately the glibc function we use on linux, `getifaddrs` was originally implemented in BSD, so even though it doesn't make much sense logically for the linux networking resolver to inherit from the generic BSD one, it does because some implementation can be shared between the BSDs and Linux.

I would not be opposed to having an openbsd/freebsd specific implementation if needed.

jasperla:
@peterhuene Alright, I will move the OpenBSD-specific bits to a separate file/folder then. This won't be before Tuesday as I'll be on travel in the meantime.

jasperla:
Closing this in favor of #1041

PR #1041, opened 2015-07-13
(FACT-1116) Implement networking facts for OpenBSD

jasperla:
Continuation of #1040 as the first move of OpenBSD bits into a subdir.
When this (along with #1039) has been merged I can submit a PR to actually build this on OpenBSD :)

### DIFF
--- a/lib/inc/internal/facts/openbsd/networking_resolver.hpp
+++ b/lib/inc/internal/facts/openbsd/networking_resolver.hpp
@@ -1,0 +1,35 @@
+/**
+ * @file
+ * Declares the OpenBSD networking fact resolver.
+ */
+#pragma once
+
+#include "../bsd/networking_resolver.hpp"
+#include <map>
+#include <ifaddrs.h>
+
+namespace facter { namespace facts { namespace openbsd {
+
+    /**
+     * Responsible for resolving networking facts.
+     */
+    struct networking_resolver : bsd::networking_resolver
+    {
+     protected:
+        /**
+         * Gets the MTU of the link layer data.
+         * @param interface The name of the link layer interface.
+         * @param data The data pointer from the link layer interface.
+         * @return Returns The MTU of the interface.
+         */
+        virtual boost::optional<uint64_t> get_link_mtu(std::string const& interface, void* data) const override;
+
+        /**
+         * Determines if the given sock address is a link layer address.
+         * @param addr The socket address to check.
+         * @returns Returns true if the socket address is a link layer address or false if it is not.
+         */
+        virtual bool is_link_address(sockaddr const* addr) const override;
+    };
+
+}}}  // namespace facter::facts::openbsd
--- a/lib/src/facts/openbsd/collection.cc
+++ b/lib/src/facts/openbsd/collection.cc
@@ -1,0 +1,29 @@
+#include <facter/facts/collection.hpp>
+#include <internal/facts/posix/kernel_resolver.hpp>
+#include <internal/facts/resolvers/operating_system_resolver.hpp>
+#include <internal/facts/bsd/uptime_resolver.hpp>
+#include <internal/facts/openbsd/networking_resolver.hpp>
+#include <internal/facts/bsd/filesystem_resolver.hpp>
+#include <internal/facts/posix/ssh_resolver.hpp>
+#include <internal/facts/posix/identity_resolver.hpp>
+#include <internal/facts/posix/timezone_resolver.hpp>
+#include <internal/facts/glib/load_average_resolver.hpp>
+
+using namespace std;
+
+namespace facter { namespace facts {
+
+    void collection::add_platform_facts()
+    {
+        add(make_shared<posix::kernel_resolver>());
+        add(make_shared<resolvers::operating_system_resolver>());
+        add(make_shared<bsd::uptime_resolver>());
+        add(make_shared<bsd::filesystem_resolver>());
+        add(make_shared<posix::ssh_resolver>());
+        add(make_shared<posix::identity_resolver>());
+        add(make_shared<posix::timezone_resolver>());
+        add(make_shared<glib::load_average_resolver>());
+        add(make_shared<openbsd::networking_resolver>());
+    }
+
+}}  // namespace facter::facts
--- a/lib/src/facts/openbsd/networking_resolver.cc
+++ b/lib/src/facts/openbsd/networking_resolver.cc
@@ -1,0 +1,43 @@
+#include <internal/facts/openbsd/networking_resolver.hpp>
+#include <internal/util/bsd/scoped_ifaddrs.hpp>
+#include <facter/execution/execution.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <boost/algorithm/string.hpp>
+#include <sys/sockio.h>
+#include <sys/ioctl.h>
+#include <net/if.h>
+#include <net/if_dl.h>
+#include <netinet/in.h>
+
+using namespace std;
+using namespace facter::util;
+using namespace facter::util::bsd;
+using namespace facter::execution;
+
+namespace facter { namespace facts { namespace openbsd {
+
+    bool networking_resolver::is_link_address(sockaddr const* addr) const
+    {
+        return addr && addr->sa_family == AF_LINK;
+    }
+
+    boost::optional<uint64_t> networking_resolver::get_link_mtu(string const& interface, void* data) const
+    {
+        ifreq ifr;
+        memset(&ifr, 0, sizeof(ifr));
+        strncpy(ifr.ifr_name, interface.c_str(), sizeof(ifr.ifr_name));
+        int s = socket(AF_INET, SOCK_DGRAM, 0);
+        if (s < 0) {
+            LOG_WARNING("socket failed: %1% (%2%): interface MTU fact is unavailable for interface %3%.", strerror(errno), errno, interface);
+            return boost::none;
+        }
+
+        if (ioctl(s, SIOCGIFMTU, &ifr) == -1) {
+            LOG_WARNING("ioctl failed: %1% (%2%): interface MTU fact is unavailable for interface %3%.", strerror(errno), errno, interface);
+            return boost::none;
+        }
+
+        return ifr.ifr_mtu;
+    }
+
+}}}  // namespace facter::facts::openbsdbsd


lib/src/facts/openbsd/collection.cc
```diff
@@ -0,0 +1,29 @@
...
+    {
+        add(make_shared<posix::kernel_resolver>());
+        add(make_shared<resolvers::operating_system_resolver>());
+        add(make_shared<bsd::uptime_resolver>());
+	add(make_shared<bsd::filesystem_resolver>());
+        add(make_shared<posix::ssh_resolver>());
+        add(make_shared<posix::identity_resolver>());
+        add(make_shared<posix::timezone_resolver>());
+        add(make_shared<glib::load_average_resolver>());
+	add(make_shared<openbsd::networking_resolver>());
```
daenney: Indentation appears to be a bit off.

lib/inc/internal/facts/openbsd/networking_resolver.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "../posix/networking_resolver.hpp"
+#include <map>
+#include <ifaddrs.h>
+
+namespace facter { namespace facts { namespace openbsd {
+
+    /**
+     * Responsible for resolving networking facts.
+     */
+    struct networking_resolver : posix::networking_resolver
```
peterhuene: Is there a reason we can't derive from the bsd resolver here?

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,185 @@
...
+#include <netinet/in.h>
+
+using namespace std;
+using namespace facter::util;
+using namespace facter::util::bsd;
+using namespace facter::execution;
+
+namespace facter { namespace facts { namespace openbsd {
+
+    networking_resolver::data networking_resolver::collect_data(collection& facts)
```
peterhuene: Does this differ from the `bsd::networking_resolver` implementation?

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,185 @@
...
+
+        if (ioctl(s, SIOCGIFMTU, &ifr) == -1) {
+            LOG_WARNING("ioctl failed: %1% (%2%): interface MTU fact is unavailable for interface %3%.", strerror(errno), errno, interface);
+            return boost::none;
+        }
+
+        return ifr.ifr_mtu;
+    }
+
+    void networking_resolver::populate_address(interface& iface, ifaddrs const* addr) const
```
peterhuene: Same comment as above.

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,185 @@
...
+
+        if (!address) {
+            // Unsupported address
+            return;
+        }
+
+        *address = address_to_string(addr->ifa_addr);
+    }
+
+    void networking_resolver::populate_network(interface& iface, ifaddrs const* addr) const
```
peterhuene: Same comment as above.

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,185 @@
...
+            // Check to see if the data already exists; interfaces can have multiple addresses of the same type
+            if (!iface.netmask.v6.empty()) {
+                return;
+            }
+            iface.netmask.v6 = address_to_string(addr->ifa_netmask);
+            iface.network.v6 = address_to_string(addr->ifa_addr, addr->ifa_netmask);
+        }
+    }
+
+    void networking_resolver::populate_mtu(interface& iface, ifaddrs const* addr) const
```
peterhuene: Same comment as above.

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,185 @@
...
+    {
+        // The MTU exists on link addresses
+        if (!is_link_address(addr->ifa_addr) || !addr->ifa_data) {
+            return;
+        }
+
+        iface.mtu = get_link_mtu(addr->ifa_name, addr->ifa_data);
+    }
+
+    string networking_resolver::get_primary_interface() const
```
peterhuene: Same comment as above.

lib/src/facts/openbsd/networking_resolver.cc
jasperla: It does, it currently doesn't include the dhcp server-related facts.

peterhuene: If the derived type implements `find_dhcp_servers` to return an empty map and `find_dhcp_server` to return an empty string, you shouldn't need to re-implement `collect_data` as it will have the same effect, right?

jasperla: I think that would be acceptable, yes.

It seems that `populate_address()` is a private function so now trying to figure out how to clean this up and get it working again.

jasperla:
Ok, with this code I end up with:
```
/usr/obj/ports/facter-3.0.1/build-amd64/lib/libfacter.so.3.0: undefined reference to `facter::facts::bsd::networking_resolver::get_link_mtu(std::string const&, void*) const'
/usr/obj/ports/facter-3.0.1/build-amd64/lib/libfacter.so.3.0: undefined reference to `facter::facts::bsd::networking_resolver::get_link_address_bytes(sockaddr const*) const'
/usr/obj/ports/facter-3.0.1/build-amd64/lib/libfacter.so.3.0: undefined reference to `facter::facts::bsd::networking_resolver::is_link_address(sockaddr const*) const'
```
which are exactly the three methods implemented in `openbsd/networking_resolver.cc` (my cmakefile does have `src/facts/openbsd/networking_resolver.cc` and `src/facts/bsd/networking_resolver.cc`).

peterhuene:
The CMakeLists.txt wasn't included in the commit, though, so I can't see what was changed.

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,69 @@
...
+            return boost::none;
+        }
+
+        return ifr.ifr_mtu;
+    }
+
+    map<string, string> networking_resolver::find_dhcp_servers() const
+    {
+        /* XXX: Unimplemented. */
+        return {{"",""}};
```
peterhuene: This returns a map with one entry, `""` to `""`.  Returning `map<string, string>()` is preferred as it is empty.

peterhuene:
Other than the above and solving the link issues, this looks good.

jasperla:
I didn't include the CMakeLists bit as that would be a separate one to hookup OpenBSD proper. However I'm using this right now:

```
elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "OpenBSD")
    set(LIBFACTER_PLATFORM_SOURCES
        "src/facts/openbsd/collection.cc"
        "src/facts/glib/load_average_resolver.cc"
        "src/facts/bsd/filesystem_resolver.cc"
        "src/facts/bsd/uptime_resolver.cc"
        "src/facts/bsd/networking_resolver.cc"
        "src/facts/openbsd/networking_resolver.cc"
        "src/util/bsd/scoped_ifaddrs.cc"
        "src/facts/openbsd/processor_resolver.cc"
        "src/facts/openbsd/memory_resolver.cc"
    )
```

lib/src/facts/openbsd/networking_resolver.cc
jasperla: That works on OpenBSD, which is the only platform of interest for this particular file.

jasperla:
I could use a little pointer on the undefined reference though.

lib/src/facts/openbsd/collection.cc
```diff
@@ -0,0 +1,29 @@
...
+    {
+        add(make_shared<posix::kernel_resolver>());
+        add(make_shared<resolvers::operating_system_resolver>());
+        add(make_shared<bsd::uptime_resolver>());
+        add(make_shared<bsd::filesystem_resolver>());
+        add(make_shared<posix::ssh_resolver>());
+        add(make_shared<posix::identity_resolver>());
+        add(make_shared<posix::timezone_resolver>());
+        add(make_shared<glib::load_average_resolver>());
+        add(make_shared<bsd::networking_resolver>());
```
peterhuene: The link error is a result of this line, I believe.  This should be `openbsd::networking_resolver` and not `bsd::networking_resolver`.

lib/inc/internal/facts/openbsd/networking_resolver.hpp
```diff
@@ -0,0 +1,55 @@
...
+    struct networking_resolver : bsd::networking_resolver
+    {
+     protected:
+        /**
+         * Gets the MTU of the link layer data.
+         * @param interface The name of the link layer interface.
+         * @param data The data pointer from the link layer interface.
+         * @return Returns The MTU of the interface.
+         */
+        virtual boost::optional<uint64_t> get_link_mtu(std::string const& interface, void* data) const;
```
peterhuene: We should have an `override` declaration on this member, since it overrides the base member.

lib/src/facts/openbsd/collection.cc
jasperla: That would result in:
```
lib/CMakeFiles/libfactersrc.dir/src/facts/bsd/networking_resolver.cc.o:(.data.rel.ro._ZTVN6facter5facts3bsd19networking_resolverE[_ZTVN6facter5facts3bsd19networking_resolverE]+0x30): undefined reference to `facter::facts::bsd::networking_resolver::is_link_address(sockaddr const*) const'
lib/CMakeFiles/libfactersrc.dir/src/facts/bsd/networking_resolver.cc.o:(.data.rel.ro._ZTVN6facter5facts3bsd19networking_resolverE[_ZTVN6facter5facts3bsd19networking_resolverE]+0x38): undefined reference to `facter::facts::bsd::networking_resolver::get_link_address_bytes(sockaddr const*) const'
lib/CMakeFiles/libfactersrc.dir/src/facts/bsd/networking_resolver.cc.o:(.data.rel.ro._ZTVN6facter5facts3bsd19networking_resolverE[_ZTVN6facter5facts3bsd19networking_resolverE]+0x40): undefined reference to `facter::facts::bsd::networking_resolver::get_link_mtu(std::string const&, void*) const'
```

peterhuene: All three of those are pure virtual members, so the implementation in the `openbsd::networking_resolver` should be the one dispatched to :/

lib/inc/internal/facts/openbsd/networking_resolver.hpp
jasperla: I've added the missing `override` now.

jasperla:
Ahum...all is good actually. Somehow a stray patch on `bsd/networking_resolver.hpp` crept in. When travis is happy this can be merged.

peterhuene:
We're still adding the `bsd::networking_resolver` to the collection instead of the `openbsd::networking_resolver` in collection.cc with your latest push, though.

Also, I think we should probably push up a base CMakeLists.txt change so that people can test this on OpenBSD.  Is it possible to separate out what you have to get a minimal CMakeLists.txt change to get this building with some rudimentary facts?

jasperla:
Yes, when this one and #1039 are merged all bits should be in place to allow building facter on OpenBSD. That's also when I'll submit the `CMakeLists.txt` patch.

After that I've got the processors facts lined up and then it would be great if a release could be cut for facter :-)

MikaelSmith:
A release will be forthcoming, but maybe not that quickly. We also have plans to finish AIX and Solaris support for Facter 3.1.

jasperla:
Alright; for now this one can be merged, right?

MikaelSmith:
I'm alright merging this as well, although I'm not sure why you're putting it up as a separate PR from enabling everything. This could be a PR with multiple commits enabling OpenBSD facts.

lib/src/facts/openbsd/networking_resolver.cc
```diff
@@ -0,0 +1,69 @@
...
+            return nullptr;
+        }
+        sockaddr_dl const* link_addr = reinterpret_cast<sockaddr_dl const*>(addr);
+        if (link_addr->sdl_alen != 6) {
+            return nullptr;
+        }
+        return reinterpret_cast<uint8_t const*>(LLADDR(link_addr));
+    }
+
+    boost::optional<uint64_t> networking_resolver::get_link_mtu(string const& interface, void* data) const
```
MikaelSmith: This appears to do exactly the same thing that linux::networking_resolver::get_link_mtu does, which is what the BSD version uses. Why was it needed?

MikaelSmith: Unless you can't call `close` on it for some reason?

MikaelSmith: I take that back, the relationship is reversed.

MikaelSmith: `get_link_mtu` implementation will now be shared between `linux` and `openbsd`; `get_link_address_bytes` and `is_link_address` will be shared between `osx` and `openbsd`. Might be worth pushing the shared implementations up and letting the outliers override them, rather than copying implementations around.

Currently we never instantiate the `bsd::networking_resolver` directly anywhere, we inherit from it for Linux and Mac OS X. I'd vote for pushing the implementations necessary to run on OpenBSD up to BSD, and let Linux and Mac deviate where necessary. That would mean pushing up the Linux implementation for `get_link_mtu`, and the Mac implementations for every other function being overridden here.

kylog: Plus one on not copying implementations around.

Sidenote: It would help guide this decision if we knew what the FreeBSD implementation would need.

jasperla:
I've moved `is_link_address()` and `get_link_address_bytes()` to the bsd resolver, but kept `get_link_mtu()` where it is. This is because all three implementations are different after all.

Also, the OpenBSD resolver can re-use the dhcp functions from the bsd resolver as it actually does work (earlier leases files had too restrictive permissions which caused them not to be read, which I misinterpreted as the actual functions not working).

peterhuene:
The reason `is_link_address` and `get_link_address_bytes` were not implemented in the BSD provider was because Linux, while sharing most of the BSD implementation, has its own way of access link-level information (AF_PACKET).

jasperla:
Ok, but now that OpenBSD entered the stage it is the right approach to move it the bsd provider and let Linux override it, right?

peterhuene:
Linux does override it, but it has to also compile the base implementation.  Provided the base implementation compiles on Linux, :+1:

MikaelSmith:
Good point on compilation, I wasn't thinking about that. And there's Travis failing. This abstraction needs a bit of work... it would still be nice to share Mac and OpenBSD implementations.

peterhuene:
Let's move ahead with this overriding in the FreeBSD provider and file a ticket for refactoring the base BSD provider into a "most everything we support but not Solaris or AIX!" type?

jasperla:
I'm working on getting the Linux compile fixed. However I'd very much prefer to get the OSX/OpenBSD sharing worked out after this has been merged so it's less of convoluted patch to reason about.

jasperla:
Ugh, so `get_link_address_bytes()` cannot be moved up to the bsd resolver either w/o breaking Linux. I think we're back to my original approach and revisit the refactoring afterwards. I'll update this PR again in a few.

jasperla:
Ok, so now everything builds again and OpenBSD still works. So +1 on merging now and figuring out how to rework the inheritance/overriding afterwards :)

MikaelSmith:
I'm ok with that.

peterhuene:
@jasperla Thanks again for the contribution!  Hopefully we can get all of your FreeBSD work into 3.1.

jasperla:
*cough* OpenBSD *cough*

Thank you though ;-)

peterhuene:
Sigh.  At this point I wouldn't fault you for thinking I'm antagonizing you :) Sorry about that.  I swear I'll get it right next time :(

PR #1044, opened 2015-07-13
(FACT-1118) Add 'noatime' option to mountpoint fact

jasperla:


MikaelSmith:
Can you add a short description of why this is needed in the commit message?

jasperla:
Like this? It's not strictly needed for anything, but since it's supported on the operating system level, we might as well display it to the user too, no?

MikaelSmith:
@peterhuene why did we leave it off previously?

peterhuene:
Appears to be an oversight.

MikaelSmith:
:+1: for merging then.

peterhuene:
It seems we're missing `MNT_CPROTECT`, `MNT_QUARANTINE`, `MNT_IGNORE_OWNERSHIP`, `MNT_NOUSERXATTR`, and `MNT_MULTILABEL` on OSX too.

jasperla:
@peterhuene could you please handle that? I don't have an OSX machine around to test that; though I could cook up the patch if you like.

peterhuene:
I assume that the above are all OSX specific?

jasperla:
@peterhuene it certainly seems like that; at least I cannot find them on OpenBSD or FreeBSD.

MikaelSmith:
I'll file a ticket for the OSX specific flags.

PR #1048, opened 2015-07-13
(maint) Remove orphaned text fixture

MikaelSmith:
A test fixture exists that's no longer removed. Remove it.

### DIFF
--- a/lib/tests/fixtures/util/multiline_file.txt
+++ b/lib/tests/fixtures/util/multiline_file.txt
@@ -1,3 +1,0 @@
-this is a line of text
-this is another line of text
-this is a third line of text


MikaelSmith:
@peterhuene remember anything important about this? Looks like it was for a test that we removed.

peterhuene:
Provided the fixture was copied to leatherman when ported, this does look like it should be removed here.

MikaelSmith:
The tests were slightly rewritten during the port, so they generate the file rather than use a fixture. Equivalent test is at https://github.com/puppetlabs/leatherman/blob/master/file_util/tests/file_utils_test.cc#L161-L180 and https://github.com/puppetlabs/leatherman/blob/master/file_util/tests/file_utils_test.cc#L140-L149. /cc @Magisus

PR #1049, opened 2015-07-14
(FACT-1120) fix build on FreeBSD

Magisus:
Removes linking to dl entirely, since library loading has been moved to Leatherman.


Magisus:
Going to make this change off master at https://github.com/puppetlabs/facter/pull/1050.

PR #1051, opened 2015-07-14
(FACT-1119) Add acceptance test for Mac OS X facts

MikaelSmith:
Adds an acceptance test for core Mac OS X facts on 10.9 and 10.10.

[skip ci]

### DIFF
--- a/acceptance/tests/facts/macosx.rb
+++ b/acceptance/tests/facts/macosx.rb
@@ -1,0 +1,97 @@
+test_name "Facts should resolve as expected in Mac OS X 10.9 and 10.10"
+
+#
+# This test is intended to ensure that facts specific to an OS configuration
+# resolve as expected in Mac OS X 10.9 and 10.10
+#
+# Facts tested: os, processors, networking, identity, kernel
+#
+
+confine :to, :platform => /osx-mavericks|osx-yosemite/
+
+agents.each do |agent|
+  if agent['platform'] =~ /osx-mavericks/
+    os_version = '10.9'
+    kernel_major = '13'
+  else
+    os_version = '10.10'
+    kernel_major = '14'
+  end
+
+  step "Ensure the OS fact resolves as expected"
+
+  expected_os = {
+                  'os.architecture'         => 'x86_64',
+                  'os.family'               => 'Darwin',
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Verified using `rake ci:test:aio SHA=8a07e1a5c98ca3a8ddeb9c55d820ade9e1fffa67 CONFIG=config/nodes/osx-109-x86_64.yaml TEST=tests/facts/macosx.rb SUITE_VERSION=1.2.1.49.g8a07e1a` and `osx-1010-x86_64.yaml`.

acceptance/tests/facts/macosx.rb
```diff
@@ -0,0 +1,97 @@
...
+  expected_processors.each do |fact, value|
+    assert_match(value, fact_on(agent, fact))
+  end
+
+  step "Ensure the Networking fact resolves with reasonable values for at least one interface"
+
+  expected_networking = {
+                          "networking.dhcp"     => /10\.\d+\.\d+\.\d+/,
+                          "networking.ip"       => /10\.\d+\.\d+\.\d+/,
+                          "networking.ip6"      => /[a-z0-9]+:+/,
```
kylog: These can be `[a-f0-9]` instead (i.e. the alpha range should only be a-f not a-z), but that's not specific to this test.

acceptance/tests/facts/macosx.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  step "Ensure the OS fact resolves as expected"
+
+  expected_os = {
+                  'os.architecture'         => 'x86_64',
+                  'os.family'               => 'Darwin',
+                  'os.hardware'             => 'x86_64',
+                  'os.name'                 => 'Darwin',
+                  'os.macosx.build'         => /#{kernel_major}[A-F]\d+/,
```
kylog: I didn't find a specification, but I think that letter doesn't have to be in the "hex range" of A-F. E.g. see some of the build versions here: https://support.apple.com/en-us/HT204319

acceptance/tests/facts/macosx.rb
MikaelSmith: Done for other files at https://github.com/puppetlabs/facter/pull/1052

PR #1053, opened 2015-07-14
(FACT-1121) Move execution class and associated utilities to Leatherman

Magisus:
Moves the execution class to Leatherman, along with its utilities. Functions in the public API are converted to wrappers which call through to Leatherman.


lib/inc/facter/execution/execution.hpp
```diff
@@ -11,10 +11,16 @@
...
+#include <leatherman/util/option_set.hpp>
+#include <leatherman/execution/execution.hpp>
+#include <leatherman/execution/execution.hpp>
+
+namespace lth_util = leatherman::util;
+namespace lth_exec = leatherman::execution;
+
 namespace facter { namespace execution {
 
     /**
```
Magisus: What is the best way to handle all of these exceptions and whatnot? They currently appear both here and in Leatherman.

lib/inc/facter/util/environment.hpp
```diff
@@ -1,70 +0,0 @@
-/**
```
Magisus: In the process of putting this back (along with its .cc files) since it was in inc/facter

Magisus:
I think scope_exit is one of those nasty cases we were thinking about yesterday, where the class is exposed in facter but needed to be ported to leatherman. Are we just going to keep a duplicate for now, or how should we handle this?

Magisus:
Also, @MikaelSmith how would you like me to incorporate https://github.com/puppetlabs/facter/pull/1024? I'm a little unclear on the process.

MikaelSmith:
I would try a `git cherry-pick 06b9cad733dfb98e5f9c3c7707bf06f1f21ddc6f` into your branch.

lib/inc/facter/execution/execution.hpp
MikaelSmith: We preferably don't want to refer to `leatherman` in the inc/facter header files at all, which means we'll need a lot of interface classes. @peterhuene any thoughts on alternatives?

lib/src/ruby/api.cc
```diff
@@ -3,8 +3,8 @@
 #include <facter/facts/scalar_value.hpp>
 #include <facter/facts/map_value.hpp>
 #include <facter/facts/array_value.hpp>
-#include <facter/util/environment.hpp>
```
MikaelSmith: It would've simplified your life before... I wouldn't change the references to anything in <facter/...>, as those APIs still need to remain and work the same.

lib/inc/facter/execution/execution.hpp
peterhuene: I hadn't really thought about it.  Could perhaps remove them from facter's public API and just make users use leatherman?  No one uses this stuff yet, so now would be the time to do that.

lib/src/ruby/api.cc
MikaelSmith: Decision reversed; the Facter API will henceforth depend on Leatherman.

MikaelSmith:
:+1:

MikaelSmith:
This is a big one, could probably use a second review: @peterhuene @branan @kylog

lib/src/execution/windows/execution.cc
```diff
@@ -1,578 +0,0 @@
...
-                return p.string();
-            }
-        }
-        return {};
-    }
-
-    // Create a pipe, throwing if there's an error. Returns {read, write} handles.
-    static tuple<scoped_resource<HANDLE>, scoped_resource<HANDLE>> CreatePipeThrow(DWORD read_mode = 0, DWORD write_mode = 0)
-    {
-        static LONG counter = 0;
```
MikaelSmith: Hmm, @peterhuene this seems like it could have a problem if we statically link leatherman.execution multiple times in the same process.

peterhuene: Yes, that would be a problem.  Hopefully we won't do this :/ Leatherman should probably become a proper shared library at that point.

MikaelSmith: We could create the name based off a timestamp, like GetTickCount64. Probably still have potential for identical numbers if running in parallel.

MikaelSmith: I'll file a ticket against Leatherman.

MikaelSmith: I forget why this has to be a named pipe :/

PR #1054, opened 2015-07-14
(FACT-1123) Retry locale setup with LANG=C when we fail

branan:
Previously, if we failed to setup a locale (for any reason) we would fail hard. With this PR, we now clear the locale-related environment and try again, printing a warning message to the user.

If this second attempt fails, we'll still fail hard. We have no logging if our locale setup has failed, so there's not a lot we can do.

This also fixes a bug in the ruby module where it was initializing logging incorrectly.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -264,6 +264,9 @@ int main(int argc, char **argv)
         bool show_legacy = vm.count("show-legacy");
         facts.write(boost::nowide::cout, fmt, queries, show_legacy);
         boost::nowide::cout << endl;
+    } catch (locale_error const& e) {
+        boost::nowide::cerr << "failed to initialize logging system due to a locale error: " << e.what() << "\n" << endl;
+        return 2;  // special error code to indicate we failed harder than normal
     } catch (exception& ex) {
         log(level::fatal, "unhandled exception: %1%", ex.what());
     }
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -1,13 +1,40 @@
 #include <facter/logging/logging.hpp>
+#include <facter/util/environment.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <leatherman/locale/locale.hpp>
 #include <boost/filesystem.hpp>
 
 using namespace std;
+using namespace facter::util;
 namespace lm = leatherman::logging;
 
-namespace facter { namespace logging {
+static void setup_logging_internal(ostream& os)
+{
+    // Initialize boost filesystem's locale to a UTF-8 default.
+    // Logging gets setup the same way via the default 2nd argument.
+#if !defined(__sun) || !defined(__GNUC__)
+    // Locale support in GCC on Solaris is busted, so skip it.
+    boost::filesystem::path::imbue(leatherman::locale::get_locale());
+#endif
+    lm::setup_logging(os);
+}
 
+static const char* lc_vars[] = {
+    "LC_CTYPE",
+    "LC_NUMERIC",
+    "LC_TIME",
+    "LC_COLLATE",
+    "LC_MONETARY",
+    "LC_MESSAGES",
+    "LC_PAPER",
+    "LC_ADDRESS",
+    "LC_TELEPHONE",
+    "LC_MEASUREMENT",
+    "LC_IDENTIFICATION",
+    "LC_ALL"
+};
+
+namespace facter { namespace logging {
     istream& operator>>(istream& in, level& lvl)
     {
         lm::log_level lm_level;
@@ -24,13 +51,16 @@ namespace facter { namespace logging {
 
     void setup_logging(ostream& os)
     {
-        // Initialize boost filesystem's locale to a UTF-8 default.
-        // Logging gets setup the same way via the default 2nd argument.
-#if !defined(__sun) || !defined(__GNUC__)
-        // Locale support in GCC on Solaris is busted, so skip it.
-        boost::filesystem::path::imbue(leatherman::locale::get_locale());
-#endif
-        lm::setup_logging(os);
+        try {
+            setup_logging_internal(os);
+        } catch (exception const&) {
+            for (auto var : lc_vars) {
+                environment::clear(var);
+            }
+            environment::set("LANG", "C");
+            setup_logging_internal(os);
+            log(level::warning, "Locale environment variables were bad; continuing with LANG=C");
+        }
     }
 
     void set_level(level lvl)
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -58,8 +58,19 @@ namespace facter { namespace logging {
                 environment::clear(var);
             }
             environment::set("LANG", "C");
-            setup_logging_internal(os);
-            log(level::warning, "Locale environment variables were bad; continuing with LANG=C");
+            try {
+                setup_logging_internal(os);
+            } catch (exception const& e) {
+                // If we fail again even with a clean environment, we
+                // need to signal to our consumer that things went
+                // sideways.
+                //
+                // Since logging is busted, we raise an exception that
+                // signals to the consumer that a special action must
+                // be taken to alert the user.
+                throw locale_error(e.what());
+            }
+            log(level::warning, "locale environment variables were bad; continuing with LANG=C");
         }
     }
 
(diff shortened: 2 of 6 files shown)


lib/src/ruby/module.cc
```diff
@@ -4,6 +4,7 @@
 #include <internal/ruby/confine.hpp>
 #include <internal/ruby/simple_resolution.hpp>
 #include <facter/facts/collection.hpp>
+#include <facter/logging/logging.hpp>
```
MikaelSmith: You can remove `#include <leatherman/logging/logging.hpp>` and `using namespace leatherman::logging` below.

lib/src/logging/logging.cc
```diff
@@ -3,11 +3,38 @@
...
+
 using namespace std;
 namespace lm = leatherman::logging;
 
-namespace facter { namespace logging {
+static void setup_logging_internal(ostream& os)
+{
+    // Initialize boost filesystem's locale to a UTF-8 default.
+    // Logging gets setup the same way via the default 2nd argument.
+#if !defined(__sun) || !defined(__GNUC__)
```
MikaelSmith: Don't want to try the same approach on Sun?

lib/src/logging/logging.cc
```diff
@@ -3,11 +3,38 @@
 #include <leatherman/locale/locale.hpp>
 #include <boost/filesystem.hpp>
 
+#include <stdlib.h>
```
MikaelSmith: Use `cstdlib`.

lib/src/logging/logging.cc
```diff
@@ -24,13 +51,16 @@ namespace facter { namespace logging {
...
-#if !defined(__sun) || !defined(__GNUC__)
-        // Locale support in GCC on Solaris is busted, so skip it.
-        boost::filesystem::path::imbue(leatherman::locale::get_locale());
-#endif
-        lm::setup_logging(os);
+        try {
+            setup_logging_internal(os);
+        } catch (...) {
+            for (auto var : lc_vars) {
+                unsetenv(var);
```
MikaelSmith: No raw unsetenv, look at https://github.com/puppetlabs/facter/blob/stable/lib/inc/facter/util/environment.hpp.

branan: Word. I was looking at the appveyor failure and realized Windows was awful. Glad we already have helpers

lib/src/logging/logging.cc
branan: I think it is a different issue, but it is worth a test run

lib/src/logging/logging.cc
branan: Probably not needed if using the facter environment helpers

lib/src/ruby/module.cc
branan: Nope. I tried that and everything blew up. This does not play nice with just having facter/logging included.

MikaelSmith: Probably all the log_level -> level changes.

lib/src/logging/logging.cc
```diff
@@ -24,13 +51,16 @@ namespace facter { namespace logging {
...
-#endif
-        lm::setup_logging(os);
+        try {
+            setup_logging_internal(os);
+        } catch (...) {
+            for (auto var : lc_vars) {
+                environment::clear(var);
+            }
+            environment::set("LANG", "C");
+            setup_logging_internal(os);
```
MikaelSmith: Should we catch again and exit gracefully if it still throws an exception?

lib/src/logging/logging.cc
```diff
@@ -24,13 +51,16 @@ namespace facter { namespace logging {
...
-        // Initialize boost filesystem's locale to a UTF-8 default.
-        // Logging gets setup the same way via the default 2nd argument.
-#if !defined(__sun) || !defined(__GNUC__)
-        // Locale support in GCC on Solaris is busted, so skip it.
-        boost::filesystem::path::imbue(leatherman::locale::get_locale());
-#endif
-        lm::setup_logging(os);
+        try {
+            setup_logging_internal(os);
+        } catch (...) {
```
peterhuene: `...` catches SEH on Windows, potentially masking things like access violations.  This should probably be `exception const&` to catch any C++ exceptions (although it'd be nice if it was more specific).

lib/src/logging/logging.cc
peterhuene: Seconded.

branan: This will get caught by our normal "catch and log from main" mechanism which... actually can't log if we can't set this up. 

I think the right thing to do here is to raise a special exception that can be caught by main and cause an abort() (after writing to stderr in lieu of logging), OR by the ruby bindings to throw a LoadError or something like that on the `require facter` that brought us here.

MikaelSmith: :+1:

lib/src/ruby/module.cc
branan: Ah, is that the secret sauce here? I'll try sedding that and see if that gets it all working

lib/src/logging/logging.cc
branan: I think it just throws a RuntimeError, which isn't THAT much more specific.

lib/src/ruby/module.cc
branan: OK, just tried this and it's the LOG_WHATEVER macros that are causing the problem. AFAICT those do come from the leatherman header, so we still need it here.

MikaelSmith: Oh, I forgot we made facter/logging so minimal.

branan:
This is updated with the exception propagation discussed earlier. We will now print an error message to stderr if locale setup fails in the facter executable, or raise a LoadError from ruby if it happens when loaded as a module.

exe/facter.cc
```diff
@@ -264,6 +264,9 @@ int main(int argc, char **argv)
         bool show_legacy = vm.count("show-legacy");
         facts.write(boost::nowide::cout, fmt, queries, show_legacy);
         boost::nowide::cout << endl;
+    } catch (LocaleError const& e) {
```
MikaelSmith: Still using snake_case for classes in Facter.

lib/src/logging/logging.cc
```diff
@@ -24,13 +51,27 @@ namespace facter { namespace logging {
...
+                // If we fail again even with a clean environment, we
+                // need to signal to our consumer that things went
+                // sideways.
+                //
+                // Since logging is busted, we raise an exception that
+                // signals to the consumer that a special action must
+                // be taken to alert the user.
+                throw locale_error(e.what());
+            }
+            log(level::warning, "Locale environment variables were bad; continuing with LANG=C");
```
MikaelSmith: We tend toward lower-case log messages.

exe/facter.cc
```diff
@@ -264,6 +264,9 @@ int main(int argc, char **argv)
         bool show_legacy = vm.count("show-legacy");
         facts.write(boost::nowide::cout, fmt, queries, show_legacy);
         boost::nowide::cout << endl;
+    } catch (locale_error const& e) {
+        cerr << "Failed to initialize logging system due to a locale error: " << e.what() << endl;
```
MikaelSmith: Can we use colorize at this point? I'd follow what line 187 is doing.

lib/src/ruby/module.cc
```diff
@@ -117,8 +125,13 @@ extern "C" {
         }
         ruby->initialize();
 
-        // Create the context
-        facter::ruby::require_context::create();
+        // If logging init failed, we'll raise a load error
+        // here. Otherwise we Create the context
+        if (logging_init_failed) {
+            ruby->rb_raise(*ruby->rb_eLoadError, "Could not initialize facter due to a locale error: %s", logging_init_error_msg.c_str());
```
MikaelSmith: Tend to lower-case Ruby errors as well.

exe/facter.cc
branan: I'm not sure if we can use nowide at this point, since our locale is potentially hosed. colorize ought to work, though

branan: The colorize flag is set very last thing in setup_l;ogging. Since that fails when locale is bad, we don't get around to enabling colorized output.

I don't think it's worth trying to rejigger colorization initialization for what is likely to be an extremely rare code path. The normal warning code path when we only fail once is still colorized properly

MikaelSmith: Ok.

branan:
Updated with lower-case error messages

lib/src/logging/logging.cc
MikaelSmith: And here?

branan:
@MikaelSmith bah. Fixed.

PR #1056, opened 2015-07-14
(FACT-1125) Include facter bin in PATH on windows

jpartlow:
Facter was not loading on Windows because the path to the libfacter.so
is not present and Facter could not load itself.  This was affecting MCO
on windows in particular when trying to run the puppet agent plugin
which requires Puppet, which requires Facter.  The work around here is
to prepend the windows PATH environment variable with the directory
libfacter.so is located in.  We're looking it up relative to the
facter.rb file being executed, and the expectation is that the directory
structure on Windows should remain stable.

### DIFF
--- a/lib/facter.rb.in
+++ b/lib/facter.rb.in
@@ -27,6 +27,13 @@ module Facter
   else
     # Simply require libfacter.so; this will define all of the Facter API
     begin
+      # This is a cmake pre-processor check.  On *nix it will end up '' == '1'
+      # On windows, where we want the changes it will be '1' == '1'
+      if '${WIN32}' == '1'
+        facter_root = File.expand_path("#{File.dirname(__FILE__)}/..")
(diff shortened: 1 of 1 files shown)


MikaelSmith:
:+1: I've verified this works on Windows when running `require 'facter'` in `irb -I <path_to_facter_lib>` without FACTERDIR set.

joshcooper:
:+1:

MikaelSmith:
@peterhuene Had to add PATH, so this is the least dangerous change we could come up with for a patch release. Later may want to rework it a bit.

joshcooper:
If `ruby.exe`/`rubyw.exe` and `libfacter.so` were in the same directory, we wouldn't need to modify `PATH`, but they're not currently on windows.

Iristyle:
:+1:

PR #1057, opened 2015-07-15
(FACT-1126) Fix timezone fact on Windows to use wide API.

peterhuene:
Fixing the timezone fact on Windows to use wcsftime instead of strftime so
that localized timezone strings are not improperly converted to UTF-8.


MikaelSmith:
:+1:

peterhuene:
Let's wait on merging this until we confirm it's for 3.0.2.

MikaelSmith:
@kylog?

PR #1058, opened 2015-07-16
(MAINT) Fix FACTER_RUBY_PATH path in cmake message

jasperla:
When passing `-DFACTER_RUBY` it should show an empty path as `FACTERPATH_PATH` is not defined.


MikaelSmith:
This looks right, although your description isn't clear to me.

jasperla:
What I mean is that I build facter on OpenBSD with `-DFACTER_RUBY="/usr/local/lib/libruby21.so"`.

Thanks for the merge.

PR #1059, opened 2015-07-16
(FACT-1129) Correctly setup Ruby load path to unbreak loading rbconfig

jasperla:
This allows the `ruby.sitedir` fact to correctly be resolved on OpenBSD.

### DIFF
--- a/lib/inc/internal/ruby/api.hpp
+++ b/lib/inc/internal/ruby/api.hpp
@@ -624,6 +624,7 @@ namespace facter {  namespace ruby {
         void (* const ruby_init)();
         void* (* const ruby_options)(int, char**);
         int (* const ruby_cleanup)(volatile int);
+        void (* const ruby_init_loadpath)();
 
         static std::unique_ptr<api> create();
 
--- a/lib/src/ruby/api.cc
+++ b/lib/src/ruby/api.cc
@@ -96,6 +96,7 @@ namespace facter { namespace ruby {
         LOAD_SYMBOL(ruby_init),
         LOAD_SYMBOL(ruby_options),
         LOAD_SYMBOL(ruby_cleanup),
+        LOAD_SYMBOL(ruby_init_loadpath),
         _library(move(library))
     {
     }
@@ -155,6 +156,8 @@ namespace facter { namespace ruby {
         } else {
             ruby_init();
         }
+
+        ruby_init_loadpath();
 
         LOG_INFO("using ruby version %1% to resolve custom facts.", to_string(rb_const_get(*rb_cObject, rb_intern("RUBY_VERSION"))));
 


peterhuene:
This fix should not be needed if we fix the "first load" detection logic in `dynamic_library` which we're skipping right now for FreeBSD.  `ruby_options` is called when Facter knows it was the one that loaded libruby into the process.  This in turn initializes the load path, along with a few other important things, like setting up encoding, etc.

I think we need to figure out a proper fix to the `dynamic_library` issue rather than always calling `ruby_init_loadpath`, especially when it is already initialized (if it is, this will duplicate paths in the load path array).

branan:
I'm in agreement with @peterhuene.

I think the right thing to do is to `dlopen(NULL)` and try to `dlsym` a ruby symbol. If it exists, ruby is loaded (and presumably initialized). We can continue to use that handle for our purposes and load our ruby symbols.

If we can't get a ruby symbol, we can then proceed to `dlopen` libruby.so explicitly.

Since this means that the "check for first load" logic will now need to be aware of what library it's loading, we can either move that logic out of the dynamic library layer and into the ruby layer, or probably just add a parameter for the sentinal symbol.

peterhuene:
So we already do a pre-emptive check to see if anything exposes a `ruby_init` symbol in the API.  If not, we fall back to loading the library, which has an additional check to see if the library is already in the process (depends on `RTLD_NOLOAD`).  A minimal hack would be to assume first load following the call to dlopen (assuming success) https://github.com/puppetlabs/leatherman/blob/fc93461c0641c84f41a2e451ffb918d91943d3bb/dynamic_library/src/posix/dynamic_library.cc#L51.

The RTLD_NOLOAD check is likely already redundant from the `ruby_init` symbol search performed by libfacter, so we could potentially just rip it out and make `load` always assume it's the first load.  It makes `dynamic_library` less generic though.

peterhuene:
I believe this issue to be [FACT-891](https://tickets.puppetlabs.com/browse/FACT-891).  @branan and I have discussed the fix for AIX which should fix this issue for the *BSDs (or any platform that doesn't support `RTLD_NOLOAD`).  There's a bug in `dynamic_library::load` right now that's always treating libraries loaded when `RTLD_NOLOAD` isn't available as not being the first load. 

@branan and I think that simply changing the semantics of `dynamic_library::load` to always treat it as a first load and simply remove any dependency on `RTLD_NOLOAD`.  Facter uses `find_by_symbol` and `find_by_name` before attempting to load the library, so if it is already loaded, those functions will find the library and treat it as already loaded; otherwise, `load` should be called which should always treat it as not having been previously loaded.

peterhuene:
@jasperla I'm going to close this PR for now.  @branan will be tackling FACT-891 in the coming sprint, so this should be addressed soon.  Thanks!

jasperla:
Thanks for the update; if there's anything I could test on OpenBSD, please let me know.

branan:
I'll definitely ping you when I have the PR open so you can verify OpenBSD behavior

PR #1060, opened 2015-07-16
(FACT-1130) Install facter.rb into vendorlibdir instead of vendordir

jasperla:


MikaelSmith:
@haus how does this affect our package builds?

MikaelSmith:
This seems like it would be fine, but we'd prefer to make it configurable.

haus:
@MikaelSmith it will load fine, but look out of place and all alone in the package. I agree, making it configurable would be ideal here.

MikaelSmith:
Having a flag like RUBY_LIB_INSTALL (setup  similar to https://github.com/puppetlabs/facter/blob/master/CMakeLists.txt#L18-L19) that defaults to vendorlibdir if not set should work.

jasperla:
Here we go :)

MikaelSmith:
:+1:

PR #1061, opened 2015-07-16
(FACT-1132) dmi resolver for OpenBSD

jasperla:
Ok, while the issues are being ironed out in the networking resolver, let's add the dmi resolver which also adds the infrastructure to actually build OpenBSD facts.

### DIFF
--- a/lib/src/facts/openbsd/dmi_resolver.cc
+++ b/lib/src/facts/openbsd/dmi_resolver.cc
@@ -1,0 +1,40 @@
+#include <internal/facts/openbsd/dmi_resolver.hpp>
+#include <leatherman/logging/logging.hpp>
+
+#include <sys/sysctl.h>
+
+using namespace std;
+
+namespace facter { namespace facts { namespace openbsd {
+
+    dmi_resolver::data dmi_resolver::collect_data(collection& facts)
+    {
+        data result;
+        result.bios_vendor = sysctl_lookup(HW_VENDOR);
+        result.uuid = sysctl_lookup(HW_UUID);
+        result.serial_number = sysctl_lookup(HW_SERIALNO);
+        result.product_name = sysctl_lookup(HW_PRODUCT);
+        result.bios_version = sysctl_lookup(HW_VERSION);
+
+        return result;
+    }
+
+    string dmi_resolver::sysctl_lookup(int mib_2)
+    {
+        int mib[2];
+        size_t len;
+        char value[BUFSIZ];
+
+        mib[0] = CTL_HW;
+        mib[1] = mib_2;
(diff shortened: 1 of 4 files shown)


peterhuene:
@jasperla Looks like this needs to be rebased with merging the previous PR.

jasperla:
Yup, I've rebased it now.

lib/src/facts/openbsd/dmi_resolver.cc
```diff
@@ -0,0 +1,40 @@
...
+    {
+        int mib[2];
+        size_t len;
+        char value[BUFSIZ];
+
+        mib[0] = CTL_HW;
+        mib[1] = mib_2;
+        len = sizeof(value) - 1;
+
+        if (sysctl(mib, 2, &value, &len, nullptr, 0) == -1) {
```
peterhuene: Is this guaranteed to copy a null character into the buffer?  Otherwise, if it copies `len` bytes without a null, `value[len]` will be uninitialized.

jasperla: `value` is null-terminated if that's what you mean.

peterhuene: More that if `sysctl` copies `len` bytes in without a null character (i.e. expecting the caller to null terminate), it won't be null terminated because `value` was not initialized.  Assuming it doesn't null terminate, either we should be doing `char value[BUFSIZ] = {};` before the call or `value[len] = 0;` after the call.  If it does, then there's no issue.

jasperla: Then there's no issue :)

peterhuene: :+1:

lib/src/facts/openbsd/dmi_resolver.cc
```diff
@@ -0,0 +1,40 @@
...
+        result.bios_vendor = sysctl_lookup(HW_VENDOR);
+        result.uuid = sysctl_lookup(HW_UUID);
+        result.serial_number = sysctl_lookup(HW_SERIALNO);
+        result.product_name = sysctl_lookup(HW_PRODUCT);
+        result.bios_version = sysctl_lookup(HW_VERSION);
+
+        return result;
+    }
+
+    string dmi_resolver::sysctl_lookup(uint64_t mib_2)
```
peterhuene: SHould `mib_2` be int?  It assigns to one below.

jasperla: That works too; I've updated the diff.

MikaelSmith:
:+1:

PR #1062, opened 2015-07-16
(FACT-1114) updated :peaio build location

kevpl:
before, :peaio was set to build from the pe_promoted repo.
this was a misunderstanding of the context of the pipeline
that this task is used in. this commit fixes that so that
it gets puppet-agent from the builds server.

[skip ci]

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -77,11 +77,6 @@ def beaker_test(mode = :packages, options = {})
     final_options[:install] << "#{build_giturl('facter')}##{sha}"
   end
 
-  if mode == :peaio
-    final_options[:puppet_agent_sha] = sha
-    final_options[:puppet_agent_version] = sha
-  end
-
   options_file = 'merged_options.rb'
   File.open(options_file, 'w') do |merged|
     merged.puts <<-EOS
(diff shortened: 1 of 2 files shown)


kevpl:
I was able to get the [job](https://jenkins.puppetlabs.com/view/PE%20AIO/job/platform_aio-suite_intn-sys_stable-pe-facter/21/) to pass using this branch.

The ubuntu failures are due to an unrelated issue that I'm working on, I've fixed it for this [job](https://jenkins.puppetlabs.com/view/PE%20AIO/job/platform_aio-suite_intn-sys_stable-pe-facter/22/) to be sure, but haven't submitted the PR yet.

MikaelSmith:
Can we now use nightly build SHAs?

MikaelSmith:
And does the `:peaio` rake task do anything special now?

kevpl:
The special thing for the `:peaio` task is that it installs on pe-only platforms.

Other projects do need this update, I'm working on that this morning.

With the nightly SHAs question, if you're referring to me pulling the `puppet_agent_sha` & `puppet_agent_version` explicit setting, that's because when I moved build resources, they needed to be set separately.  I'm setting them via environment variables in the Jenkins job.  

[This pipeline](https://jenkins.puppetlabs.com/view/PE%20AIO/) is where the job using this is located, and it works with the SHA there.  I'm not sure what qualifies as a nightly build SHA, but I don't see any reason why you shouldn't be able to use any SHA, as long as it has a build on the builds server.

PR #1064, opened 2015-07-16
(FACT-1133) Fix unnecessary object allocation from Facter API.

peterhuene:
When resolving a value for a built-in fact, the `_resolved` flag never
gets set, so the cached Ruby object for the built-in fact is thrown
away and a new object is allocated.  This is unnecessary and just
creates garbage.

The fix is to always set the flag to true when we're done resolving the
fact.

### DIFF
--- a/lib/src/ruby/fact.cc
+++ b/lib/src/ruby/fact.cc
@@ -107,14 +107,12 @@ namespace facter { namespace ruby {
 
                 // Set the value to what was resolved
                 _value = value;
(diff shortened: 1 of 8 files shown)


peterhuene:
DO NOT MERGE: this is not meant for the 3.0.2 release.

MikaelSmith:
@peterhuene all clear to merge now?

MikaelSmith:
:+1:

PR #1065, opened 2015-07-17
(FACT-1134) Fix Facter operatingsystem resolution

MikaelSmith:
Facter 3's operatingsystem fact resolution rearranged the order release
files were searched for so that some RedHat distributions would resolve
to RedHat instead of more specific names, for example OracleLinux.

Restore original ordering with respect to searching specific
distributions before RedHat and SUSE.

### DIFF
--- a/lib/src/facts/linux/os_linux.cc
+++ b/lib/src/facts/linux/os_linux.cc
@@ -118,19 +118,19 @@ namespace facter { namespace facts { namespace linux {
     static string check_other_linux()
     {
         static vector<tuple<string, string>> const files {
-            make_tuple(string(release_file::openwrt),        string(os::openwrt)),
+            make_tuple(string(release_file::arista_eos),     string(os::arista_eos)),
             make_tuple(string(release_file::gentoo),         string(os::gentoo)),
+            make_tuple(string(release_file::mageia),         string(os::mageia)),
             make_tuple(string(release_file::mandriva),       string(os::mandriva)),
             make_tuple(string(release_file::mandrake),       string(os::mandrake)),
             make_tuple(string(release_file::meego),          string(os::meego)),
             make_tuple(string(release_file::archlinux),      string(os::archlinux)),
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Original search order https://github.com/puppetlabs/facter/blob/2.x/lib/facter/operatingsystem/linux.rb#L241-L286

lib/src/facts/linux/os_linux.cc
```diff
@@ -144,12 +144,16 @@ namespace facter { namespace facts { namespace linux {
...
-        // Check for Debian next
+        // Check for Debian first; this happens after AristaEOS in Facter 2.x
+        // but that platform is not a Debian so shouldn't matter.
         auto value = check_debian_linux(distro_id);
         if (!value.empty()) {
             return value;
         }
 
+        // Check for specialized distributions next
+        return check_other_linux();
```
peterhuene: Early return?

MikaelSmith:
Bah, Amazon's also broken.

MikaelSmith:
Amazon is a RedHat, so this should also fix Amazon's reported OS.

MikaelSmith:
```
[root@x5nhcn0p0c027pk ~]# cat /etc/oracle-release
Oracle Linux Server release 7.0
[root@x5nhcn0p0c027pk ~]# /opt/puppetlabs/bin/facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "OracleLinux",
  release => {
    full => "7.0",
    major => "7",
    minor => "0"
  },
  selinux => {
    config_mode => "enforcing",
    current_mode => "enforcing",
    enabled => true,
    enforced => true,
    policy_version => "28"
  }
}
```

MikaelSmith:
Compared to Facter 2.4.4 on the same machine
`{"name"=>"OracleLinux", "family"=>"RedHat", "release"=>{"major"=>"7", "minor"=>"0", "full"=>"7.0"}}`

PR #1067, opened 2015-07-21
(FACT-1136) Move Ruby API to Leatherman

Magisus:
Updates Facter to use the Ruby API that was moved to Leatherman. Facter public API now depends on Leatherman.


Magisus:
Still waiting on https://github.com/puppetlabs/leatherman/pull/55 to be merged.

lib/src/ruby/module.cc
```diff
@@ -113,7 +116,7 @@ extern "C" {
         set_level(log_level::warning);
 
         // Initialize ruby
-        auto ruby = facter::ruby::api::instance();
+        auto ruby = api::instance();
```
MikaelSmith: We need to set `api::ruby_lib_location`, doing it before this call should be sufficient, based on FACTER_RUBY if defined.

MikaelSmith: We need to set `api::ruby_lib_location` at line 138 below in `module::module`, based on FACTER_RUBY if defined.

It's not needed here because Init_libfacter is run from inside Ruby, but this is the closest spot to line 138.

lib/src/ruby/module.cc
```diff
@@ -327,6 +330,47 @@ namespace facter { namespace ruby {
         return ruby.to_native<fact>(fact_self)->value();
     }
 
+    VALUE module::to_ruby(value const* val) const
```
MikaelSmith: This could be a static method; it doesn't use anything stored in `module`.

MikaelSmith: On the other hand, it's better not to call `api::instance` before a `module`'s been created, so let's leave this an instance method.

lib/src/ruby/module.cc
MikaelSmith: I vaguely lean towards statically initializing `api::ruby_lib_location`, but I think static initialization of module.cc contents vs `api::ruby_lib_location` may be non-deterministic.

MikaelSmith:
:+1: squash it

PR #1071, opened 2015-07-22
(maint) Simplify Travis scripts and add check for commit summaries

MikaelSmith:
Separates Doxygen generation as its own step, and simplifies the script
based on puppetlabs/cpp-project-template. Only installs files that are
necessary for the particular build task being performed.


scripts/travis_target.sh
```diff
@@ -1,85 +1,75 @@
...
+          "\t\t(FACT-<digits>) # this is most commond and should be a ticket at tickets.puppetlabs.com\n" \
+          "\t\t(doc)\n" \
+          "\t\t(maint)\n" \
+          "\t\t(packaging)\n" \
+          "\n\tThis test for the commit summary is case-insensitive.\n\n\n"
+      exit 1
     fi
+  done
+  exit 0
+elif [ ${TRAVIS_TARGET} == DEBUG ]; then
```
peterhuene: Shouldn't this be RELEASE?  The release build in Travis built debug with these changes.

MikaelSmith: Oh right, oops.

MikaelSmith: Fixed by swapping cmake invocations.

MikaelSmith:
Rebased for the doxygen changes, and made some fixes to the regex (it used to still check for pup-, \d doesn't appear to be supported in Bash regex) and added "Revert".

scripts/travis_target.sh
```diff
@@ -1,84 +1,72 @@
...
-        make DESTDIR=$USERDIR install
-        if [ $? -ne 0 ]; then
-            echo "install failed."
-            exit 1
-        fi
-    fi
-}
+  if [ ${TRAVIS_TARGET} == DEBUG ]; then
+    # Ignore coveralls failures, keep service success uncoupled
+    coveralls --gcov gcov-4.8 --gcov-options '\-lp' -e vendor >/dev/null || true
```
MikaelSmith: Moving the build dir to root includes files that aren't compiled on this platform. It's a more complete view of what's tested in Travis, but doesn't include testing from Windows.

MikaelSmith:
Updated with fix to commit title regex based on https://github.com/puppetlabs/hiera/pull/284.

PR #1072, opened 2015-07-23
(maint) Confirm doxygen issues no warnings

MikaelSmith:
Changes the doxygen test to use quiet output and scrape for any
warnings, as WARN_LOGFILE doesn't appear to work.


MikaelSmith:
@peterhuene do you know off-hand how we should fix the `\ref` warning on README.md?

peterhuene:
I couldn't figure out a way to do so earlier.  Doesn't seem like any escaping of it was working.

PR #1073, opened 2015-07-23
(FACT-1143) Add processor resolver for OpenBSD

jasperla:


lib/src/facts/openbsd/processor_resolver.cc
```diff
@@ -0,0 +1,67 @@
...
+    processor_resolver::data processor_resolver::collect_data(collection& facts)
+    {
+        auto result = posix::processor_resolver::collect_data(facts);
+        size_t len;
+        int mib[2];
+        mib[0] = CTL_HW;
+
+        // Get the physical count of processors
+        len = sizeof(result.logical_count);
+        mib[1] = HW_NCPUFOUND;
```
kylog: The comment says `physical` but the variable is called `logical_count` but then the debug message says `physical`.

So I went to http://www.openbsd.org/cgi-bin/man.cgi/OpenBSD-current/man3/sysctl.3?query=sysctl&sec=3 to see how `HW_NCPUFOUND` and `HW_NCPU` are defined:

```
HW_NCPU
The number of CPUs being used.
HW_NCPUFOUND
The number of CPUs found.
```

And from that I'm not clear at all that used/found maps to physical/logical or v-v.

And then I found this thread: http://comments.gmane.org/gmane.os.openbsd.misc/181238, where Theo de Raadt's response to a question about determining the number of cores ("logical" processors in our nomenclature) is "You look at dmesg, and then at your processor manual."

So ... I'm confused as to whether we can retrieve logical processor count via sysctl, and if so, how.

jasperla: That was a pasto in the code; `HW_NCPUFOUND` should be the physical number of processors (it's the number of processors/cores actually present in the machine). `HW_NCPU` should the number of logical CPUs (it's the number of processors/cores actually in use).
Does this mapping make sense?

kylog: Okay thanks for fixing the copy pasta. If it works, that's great. The naming/documentation for the constants may just be unintuitive to me. Thanks!

MikaelSmith:
Mind adding in the output from the resolver?

jasperla:
```
{
  count => 4,
  isa => "amd64",
  models => [
    "Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz",
    "Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz",
    "Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz",
    "Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz"
  ],
  physicalcount => 4,
  speed => "2.50 GHz"
}
```

kylog:
For comparision, here's a snippet from my OS X laptop:

```
{
  count => 4,
  isa => "i386",
  models => [
    "Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz",
    "Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz",
    "Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz",
    "Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz"
  ],
  physicalcount => 2,
  speed => "2.60 GHz"
}
```

I'm pretty sure both these processors are dual-core. Note that physicalcount != count on mine, but that physicalcount == count in the OpenBSD snippet above. That plus the man page for sysctl I mentioned above makes me wonder if it's even possible to retrieve this info on OpenBSD.

What does facter 2 report?

jasperla:
Facter 2 reports:
```
{"count"=>4, "physicalcount"=>4, "speed"=>"2.5 GHz"}
```

There's indeed no way in OpenBSD to differentiate between physical dies and cores.

peterhuene:
:+1: @kylog any other comments before we merge?

kylog:
Since `HW_NCPUFOUND` doesn't actually equate to `physical`, I'd say we just query `sysctl` with `HW_NCPU` for `count`.

And then re `physicalcount`, either a) not report `physicalcount` (accurate) or b) just assign `physicalcount` to be the same as `count` (backward compatible). I'd favor (a).

peterhuene:
I'd also favor option a, although we need to change the base resolver to either not report 0 counts or explicitly treat < 0 as "don't report".

kylog:
Mild leaning to go the route where -1 means "don't report". Seems like something we could establish as standard semantics for other facts where, maybe, 0 is meaningful.

jasperla:
I'm alright with option a, provided it can report -1. Having a `physicalcount` of 0 could have odd implications.
So that means I'll wait with incorporating your feedback until the base resolver has been adjusted?

peterhuene:
I can throw up a related-PR that changes the base behavior if you'd like or feel free to amend this commit to update the initialization of `logical_count` and `physical_count` to -1 in the relevant constructor (currently initialized to 0) and only add the facts if >= 0.

kylog:
I put up #1095, which changes the base behavior (using 0 as don't report rather than -1 however).

kylog:
@jasperla #1095 is merged now. So now it looks like this needs a rebase and you can remove the `HW_NCPUFOUND` code we talked about above. Thanks for all your work on this one!

jasperla:
Should be good now.

peterhuene:
One more rebase needed :/

jasperla:
Should be good now.

MikaelSmith:
Thanks for all your work on this!

jasperla:
Thanks again for merging! :)

PR #1074, opened 2015-07-23
(FACT-1144) Add memory resolver for OpenBSD

jasperla:
Currently without `swap` facts while those are being sorted out.

### DIFF
--- a/lib/src/facts/openbsd/memory_resolver.cc
+++ b/lib/src/facts/openbsd/memory_resolver.cc
@@ -1,0 +1,43 @@
+#include <internal/facts/openbsd/memory_resolver.hpp>
+#include <facter/execution/execution.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <sys/types.h>
+#include <sys/param.h>
+#include <sys/mount.h>
+#include <sys/sysctl.h>
+#include <sys/swap.h>
+#include <unistd.h>
+
+using namespace std;
+using namespace facter::execution;
+using namespace facter::util;
+
+namespace facter { namespace facts { namespace openbsd {
+
+    memory_resolver::data memory_resolver::collect_data(collection& facts)
+    {
+        data result;
+
+        // Get the system page size
+        int pagesize_mib[] = { CTL_HW, HW_PAGESIZE};
+        int page_size = 0;
+        size_t len = sizeof(page_size);
+        if (sysctl(pagesize_mib, 2, &page_size, &len, nullptr, 0) == -1) {
+            LOG_DEBUG("sysctl failed: %1% (%2%): system page size is unknown.", strerror(errno), errno);
+        } else {
+            int uvmexp_mib[] = { CTL_VM, VM_UVMEXP };
+            struct uvmexp uvmexp;
+            len = sizeof(uvmexp);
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Mind commenting with the output from the resolver?

jasperla:
```
{
  system => {
    available => "1.47 GiB",
    available_bytes => 1578127360,
    capacity => "59.23%",
    total => "3.61 GiB",
    total_bytes => 3870953472,
    used => "2.14 GiB",
    used_bytes => 2292826112
  }
}
```

lib/src/facts/openbsd/memory_resolver.cc
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+                int uvmexp_mib[] = { CTL_VM, VM_UVMEXP };
+                struct uvmexp uvmexp;
+                len = sizeof(uvmexp);
+                if (sysctl(uvmexp_mib, 2, &uvmexp, &len, nullptr, 0) == -1) {
+                    LOG_DEBUG("sysctl uvmexp failed: %1% (%2%): free memory is not available.", strerror(errno), errno);
+                }
+
+                // Should we account for the buffer cache?
+                result.mem_total = (uint64_t) uvmexp.npages << uvmexp.pageshift;
```
peterhuene: We generally prefer c++-style casts over c-style casts (for various reasons).  `result.mem_total = static_cast<uint64_t>(uvmexp.npages << uvmexp.pageshift);` should do it.  Same comment for the line below.

lib/src/facts/openbsd/memory_resolver.cc
```diff
@@ -0,0 +1,44 @@
...
+using namespace std;
+using namespace facter::execution;
+using namespace facter::util;
+
+namespace facter { namespace facts { namespace openbsd {
+
+    memory_resolver::data memory_resolver::collect_data(collection& facts)
+    {
+        data result;
+        size_t len;
```
peterhuene: Can we move this declaration to its assignment?  I prefer not to declare variables apart from where they are initialized unless I have to.

peterhuene:
Other than above minor comments, this looks good :+1:

jasperla:
I've fixed up the casts and declaration.

lib/src/facts/openbsd/memory_resolver.cc
```diff
@@ -0,0 +1,43 @@
...
+        data result;
+
+        // Get the system page size
+        int pagesize_mib[] = { CTL_HW, HW_PAGESIZE};
+        int page_size = 0;
+        size_t len = sizeof(page_size);
+        if (sysctl(pagesize_mib, 2, &page_size, &len, nullptr, 0) == -1) {
+            LOG_DEBUG("sysctl failed: %1% (%2%): system page size is unknown.", strerror(errno), errno);
+        } else {
+                int uvmexp_mib[] = { CTL_VM, VM_UVMEXP };
```
peterhuene: Looks like this block is unintentionally indented more than needed.

jasperla:
That's also fixed now.

HAIL9000:
@peterhuene, could you take another look now that it's been updated?

peterhuene:
@jasperla Thanks again for the contribution!

PR #1075, opened 2015-07-23
(FACT-1145) Add virtualization resolver for OpenBSD

jasperla:
This is currently implemented only by looking up `fact::product_name`, it could be expanded in the future to use other tools too. Currently only tested on KVM.

Note however that `get_product_name_vm` is a verbatim copy of Linux' resolver. Should it be moved higher up in the hierarchy so it can be shared?

### DIFF
--- a/lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
+++ b/lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
@@ -39,6 +39,14 @@ namespace facter { namespace facts { namespace resolvers {
          * @return Returns true if the hypervisor is virtual or false if it is physical.
          */
         virtual bool is_virtual(std::string const& hypervisor);
+
+        /**
+         * Gets the product name which is matched against a list of known
+         * hypervisors.
+         * @param facts The fact collection that is resolving facts.
+         * @return Returns the hypervisor product name if matched.
+         */
+        static std::string get_product_name_vm(collection& facts);
     };
 
 }}}  // namespace facter::facts::resolvers
--- a/lib/src/facts/openbsd/virtualization_resolver.cc
+++ b/lib/src/facts/openbsd/virtualization_resolver.cc
@@ -1,0 +1,22 @@
+#include <internal/facts/openbsd/virtualization_resolver.hpp>
+#include <facter/facts/scalar_value.hpp>
+#include <facter/facts/collection.hpp>
+#include <facter/facts/fact.hpp>
+#include <facter/facts/vm.hpp>
+#include <facter/execution/execution.hpp>
+#include <boost/algorithm/string.hpp>
+
+using namespace std;
+using namespace facter::facts;
+using namespace facter::execution;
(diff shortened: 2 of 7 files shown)


MikaelSmith:
Mind commenting with the output from the resolver?

lib/src/facts/openbsd/virtualization_resolver.cc
```diff
@@ -0,0 +1,51 @@
...
+    string virtualization_resolver::get_hypervisor(collection& facts)
+    {
+        string value = get_product_name_vm(facts);
+
+        return value;
+    }
+
+    string virtualization_resolver::get_product_name_vm(collection& facts)
+    {
+        static vector<tuple<string, string>> vms = {
```
MikaelSmith: This looks pulled straight from the Linux provider. We should really pull this out into a helper.

jasperla: That's actually what I wrote in my original comment :) What would be the preferred location to move this to?

jasperla:
On a physical system:

```
virtual => physical
is_virtual => false
```

KVM guest:
```
dmi => {
  bios => {
    vendor => "Red Hat",
    version => "RHEL 7.0.0 PC (i440FX + PIIX, 1996)"
  },
  product => {
    name => "KVM",
    uuid => "dbcf944d-a908-4f60-b68a-529a231294e9"
  }
}
is_virtual => true
virtual => kvm 
```

lib/src/facts/openbsd/virtualization_resolver.cc
peterhuene: This could probably be moved up to the base resolver as a utility function as it's just converting product names to vm names.

peterhuene:
@jasperla Looks like this needs to be rebased now.

jasperla:
I've rebased and moved `get_product_name_vm()` into the base resolver.

lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
```diff
@@ -39,6 +39,14 @@ namespace facter { namespace facts { namespace resolvers {
...
         virtual bool is_virtual(std::string const& hypervisor);
+
+        /**
+	 * Gets the product name which is matched against a list of known
+         * hypervisors.
+         * @param facts The fact collection that is resolving facts.
+	 * @return Returns the hypervisor product name if matched.
+	 */
+        static std::string get_product_name_vm(collection& facts);
     };
```
kylog: Lines 44, 47, 48 are indented inconsistently.

kylog:
Other than that one whitespace thing, :+1: .

jasperla:
Thanks, whitespace has been mopped up.

peterhuene:
@jasperla This has been merged to master and should appear in Facter 3.1.  Thanks for the contribution as always!

jasperla:
Cheers!

PR #1077, opened 2015-07-24
(FACT-1140) Fix decoding DMI data containing non-printable ASCII.

peterhuene:
On Linux, we directly read DMI from data that is provided from the kernel.
On some hardware the bytes read are not printable ASCII, resulting in
invalid UTF-8 sequences when interpreted as UTF-8.

The dmidecode utility converts any non-printable character to a '.'
character.  This fix mimics the behavior of dmidecode so that invalid DMI
data is not converted to invalid UTF-8.

### DIFF
--- a/lib/src/facts/linux/dmi_resolver.cc
+++ b/lib/src/facts/linux/dmi_resolver.cc
@@ -45,6 +45,14 @@ namespace facter { namespace facts { namespace linux {
         }
 
         boost::trim(value);
+
+        // Replace any non-printable ASCII characters with '.'
+        // This mimics the behavior of dmidecode
+        for (auto& c : value) {
+            if (c < 32 || c == 127) {
+                c = '.';
+            }
+        }
         return value;
     }
 


peterhuene:
This can't be easily mocked up by the unit test framework because the resolver directly reads from sysfs files to populate these fact values.

I did test this by temporarily altering the resolver to read from a file on disk that contained non-printable characters and correctly received the '.' replacement characters as expected.

It's also not going to be easy to test without hardware that exhibits the invalid characters in its DMI table :/

lib/src/facts/linux/dmi_resolver.cc
```diff
@@ -44,6 +44,14 @@ namespace facter { namespace facts { namespace linux {
             return {};
         }
 
+        // Replace any non-printable ASCII characters with '.'
+        // This mimics the behavior of dmidecode
+        for (auto& c : value) {
+            if (c < 32 || c == 127) {
```
kylog: Should that be `c >= 127`?

peterhuene: char is signed.

peterhuene: The trim call needs to happen before this though, otherwise trailing newlines won't be trimmed (instead converted to '.').  Pushing up fix.

MikaelSmith:
We'd have to add a file-system abstraction that we can override for testing. Not a route we've gone down yet.

MikaelSmith:
:+1:

lib/src/facts/linux/dmi_resolver.cc
kylog: Ah right.

MikaelSmith:
Should this go on stable?

peterhuene:
Yes, yes it should.  Grr I hate GitHub UI.

PR #1079, opened 2015-07-27
(FACT-1114) added puppet-agent installation for PE-only platforms

kevpl:
After speaking with @joshcooper, I realized that I was going down the wrong path with the `:peaio` rake task.  

This commit reverts that old work, and simply adds support for installing puppet-agent on PE-only platforms (using Beaker's `install_puppet_agent_dev_repo_on` method).


joshcooper:
Tested on ubuntu 10.04 and sles 12, looks good, just waiting on update to commit message.

```
$ be rake ci:test:aio SHA=9bc81ceeac7971e61d76d9578bf50f1678490e45 CONFIG=config/nodes/sles-12-x86_64.yaml SUITE_VERSION=1.2.1.73.g9bc81ce
```

```
$ be rake ci:test:aio SHA=9bc81ceeac7971e61d76d9578bf50f1678490e45 CONFIG=config/nodes/ubuntu-1004-x86_64.yaml SUITE_VERSION=1.2.1.73.g9bc81ce
```

PR #1081, opened 2015-07-27
(FACT-1147, FACT-1150) Fix network interfaces with multiple addresses.

peterhuene:
Network interfaces with multiple assigned addresses were not being
handled properly.  Facter was considering the last address enumerated to
be the address of the interface.  This may not be correct as the order
may have a non-link-local address first and a link-local address later.
This would result in facter favoring the link-local address which is not
desired (FACT-1150).

Additionally, if the interface had only link-local addresses, the last
address was being selected rather than the first (FACT-1147).

The fix is for the resolvers to return an array of bound addresses for
each network interface. The base resolver now selects what it considers
to be the address of the interface by finding the first address that is
not link-local. If all addresses are link-local, then the first bound
address is used.

Additionally, if network interfaces have more than one bound address,
a new `bindings` (or `bindings6` for IPv6) element will appear on the
interface containing an array of structures containing the address,
network, and netmask for each binding.  This will allow users to see all
addresses on the interface.

This fix also includes returning a "primary" element in the "networking"
fact which is set to the name of the primary interface.  Users can use
this to quickly determine the interface that facter considers to be the
interface used for the top-level networking facts.

### DIFF
--- a/acceptance/tests/facts/debian.rb
+++ b/acceptance/tests/facts/debian.rb
@@ -83,6 +83,23 @@ agents.each do |agent|
     assert_match(value, fact_on(agent, fact))
   end
 
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  assert_not_equal("", primary_interface)
+
+  step "Ensure bindings for the primary networking interface are present."
+  expected_bindings = {
+                        "networking.interfaces.#{primary_interface}.bindings.0.address" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.netmask" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.network" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings6.0.address" => /[a-f0-9:]+/,
+                        "networking.interfaces.#{primary_interface}.bindings6.0.netmask" => /[a-f0-9:]+/,
+                        "networking.interfaces.#{primary_interface}.bindings6.0.network" => /[a-f0-9:]+/
+                      }
+  expected_bindings.each do |fact, value|
+    assert_match(value, fact_on(agent, fact))
+  end
+
   step "Ensure the identity fact resolves as expected"
   expected_identity = {
                         'identity.gid'   => '0',
--- a/acceptance/tests/facts/windows.rb
+++ b/acceptance/tests/facts/windows.rb
@@ -61,6 +61,20 @@ agents.each do |agent|
     assert_match(value, fact_on(agent, fact))
   end
 
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  assert_not_equal("", primary_interface)
+
+  step "Ensure bindings for the primary networking interface are present."
+  expected_bindings = {
+                        "networking.interfaces.#{primary_interface}.bindings.0.address" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.netmask" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.network" => /\d+\.\d+\.\d+\.\d+/
+                      }
+  expected_bindings.each do |fact, value|
+    assert_match(value, fact_on(agent, fact))
+  end
+
   step "Ensure the identity fact resolves as expected"
   expected_identity = {
                         'identity.user'   => /.*\\cyg_server/,
--- a/lib/src/facts/bsd/networking_resolver.cc
+++ b/lib/src/facts/bsd/networking_resolver.cc
@@ -62,7 +62,8 @@ namespace facter { namespace facts { namespace bsd {
                     }
 
                     string ip = address_to_string(addr->ifa_addr, addr->ifa_netmask);
-                    if (!boost::starts_with(ip, "127.") && ip != "::1" && !boost::starts_with(ip, "fe80")) {
+                    if ((addr->ifa_addr->sa_family == AF_INET && !ignored_ipv4_address(ip)) ||
+                        (addr->ifa_addr->sa_family == AF_INET6 && !ignored_ipv6_address(ip))) {
                         data.primary_interface = name;
                         break;
                     }
@@ -74,8 +75,7 @@ namespace facter { namespace facts { namespace bsd {
 
             // Walk the addresses of this interface and populate the data
             for (; it != interface_map.end() && it->first == name; ++it) {
-                populate_address(iface, it->second);
-                populate_network(iface, it->second);
+                populate_binding(iface, it->second);
                 populate_mtu(iface, it->second);
             }
 
@@ -92,48 +92,33 @@ namespace facter { namespace facts { namespace bsd {
         return data;
     }
 
-    void networking_resolver::populate_address(interface& iface, ifaddrs const* addr) const
+    void networking_resolver::populate_binding(interface& iface, ifaddrs const* addr) const
     {
-        string* address = nullptr;
+        // If the address is a link address, populate the MAC
+        if (is_link_address(addr->ifa_addr)) {
+            iface.macaddress = address_to_string(addr->ifa_addr);
+            return;
+        }
+
+        // Populate the correct bindings list
+        vector<binding>* bindings = nullptr;
         if (addr->ifa_addr->sa_family == AF_INET) {
-            address = &iface.address.v4;
+            bindings = &iface.ipv4_bindings;
         } else if (addr->ifa_addr->sa_family == AF_INET6) {
-            address = &iface.address.v6;
-        } else if (is_link_address(addr->ifa_addr)) {
-            address = &iface.macaddress;
+            bindings = &iface.ipv6_bindings;
         }
 
-        if (!address) {
-            // Unsupported address
+        if (!bindings) {
             return;
         }
 
-        *address = address_to_string(addr->ifa_addr);
-    }
-
-    void networking_resolver::populate_network(interface& iface, ifaddrs const* addr) const
-    {
-        // Limit these facts to IPv4 and IPv6 with a netmask address
-        if ((addr->ifa_addr->sa_family != AF_INET &&
-             addr->ifa_addr->sa_family != AF_INET6) || !addr->ifa_netmask) {
-            return;
+        binding b;
+        b.address = address_to_string(addr->ifa_addr);
+        if (addr->ifa_netmask) {
+            b.netmask = address_to_string(addr->ifa_netmask);
+            b.network = address_to_string(addr->ifa_addr, addr->ifa_netmask);
         }
-
-        if (addr->ifa_addr->sa_family == AF_INET) {
-            // Check to see if the data already exists; interfaces can have multiple addresses of the same type
-            if (!iface.netmask.v4.empty()) {
-                return;
-            }
-            iface.netmask.v4 = address_to_string(addr->ifa_netmask);
-            iface.network.v4 = address_to_string(addr->ifa_addr, addr->ifa_netmask);
-        } else {
-            // Check to see if the data already exists; interfaces can have multiple addresses of the same type
-            if (!iface.netmask.v6.empty()) {
-                return;
-            }
-            iface.netmask.v6 = address_to_string(addr->ifa_netmask);
-            iface.network.v6 = address_to_string(addr->ifa_addr, addr->ifa_netmask);
-        }
+        bindings->emplace_back(std::move(b));
     }
 
     void networking_resolver::populate_mtu(interface& iface, ifaddrs const* addr) const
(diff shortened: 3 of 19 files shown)


peterhuene:
Marking WIP as I should fix FACT-1147 as well.

daenney:
:+1:

MikaelSmith:
Did you take a look at Windows at all?

lib/src/facts/bsd/networking_resolver.cc
```diff
@@ -94,21 +95,12 @@ namespace facter { namespace facts { namespace bsd {
...
-
-        if (!address) {
-            // Unsupported address
+        if (is_link_address(addr->ifa_addr)) {
+            iface.macaddress = address_to_string(addr->ifa_addr);
             return;
         }
 
-        *address = address_to_string(addr->ifa_addr);
+        assign_address(iface.address, addr->ifa_addr);
```
MikaelSmith: So we appear to expect a method named assign_address to be defined? Should we add that as an abstract method to the bsd::networking_resolver?

MikaelSmith: Oh nevermind, looks like the relationship is reversed from what I was thinking.

lib/src/facts/posix/networking_resolver.cc
```diff
@@ -120,4 +120,32 @@ namespace facter { namespace facts { namespace posix {
...
     }
 
+    void networking_resolver::assign_address(ipaddress& ip, sockaddr const* address, sockaddr const* mask) const
+    {
+        if (!address) {
+           return;
+        }
+
+        string* assigned = nullptr;
+        bool (*ignored)(string const&) = nullptr;
```
MikaelSmith: Is there a good reason to use std::function over a raw function pointer?

lib/src/facts/posix/networking_resolver.cc
```diff
@@ -120,4 +120,32 @@ namespace facter { namespace facts { namespace posix {
...
+            ignored = ignored_ipv4_address;
+        } else if (address->sa_family == AF_INET6) {
+            assigned = &ip.v6;
+            ignored = ignored_ipv6_address;
+        } else {
+            return;
+        }
+
+        // If the current address is not ignored or there is existing ignored address, do not assign
+        if (!ignored(*assigned) || !assigned->empty()) {
```
MikaelSmith: This logic is confusing me. Is the `!assigned-empty()` redundant?

lib/inc/internal/facts/posix/networking_resolver.hpp
```diff
@@ -43,6 +43,14 @@ namespace facter { namespace facts { namespace posix {
...
          */
         virtual data collect_data(collection& facts) override;
+
+        /**
+         * Assigns an IP address provided the existing address is not ignored.
+         * @param ip The ipaddress to assign to.
+         * @param addr The socket address to convert to a string.
+         * @param mask The mask to apply to the address.
+         */
+        void assign_address(ipaddress& ip, sockaddr const* address, sockaddr const* mask = nullptr) const;
```
MikaelSmith: I'd be tempted to add a unit test for assign_address.

lib/src/facts/posix/networking_resolver.cc
peterhuene: If `ignored` returns true, then it could be a non-empty ignored address.  This second check ensures that we do not overwrite an existing non-empty ignored address with another one (this is FACT-1147).

MikaelSmith: So checking ignored seems unnecessary then. You can just return if assigned is not empty.

lib/src/facts/posix/networking_resolver.cc
peterhuene: Not really.  I generally just use raw pointers when pointing at one of (potentially) many objects to manipulate based on some conditional logic.  I have no preference either way and can easily change that to a `std::function` for readability.

lib/inc/internal/facts/posix/networking_resolver.hpp
peterhuene: I can do so on Monday.

peterhuene:
I didn't take a look at Windows.  Does Windows support multiple addresses for a single protocol?

lib/src/facts/posix/networking_resolver.cc
MikaelSmith: Did you mean if to check the new address in one of those?

MikaelSmith:
I'm not sure, @Iristyle might know. The data structure we walk (https://github.com/puppetlabs/facter/blob/master/lib/src/facts/windows/networking_resolver.cc#L220) can define multiple addresses, but I haven't found anything saying whether there can be only one IPv4 and one IPv6 address. We currently take the last non-empty address.

A cursory look at manually configuring the networking stack in Windows suggests you can assign multiple addresses, so I'll do some quick tests.

lib/src/facts/posix/networking_resolver.cc
peterhuene: Checking for just not assigned will prevent a non-link-local address from overwriting a link-local address, though

Iristyle:
Yes, Windows NICs can be multihomed as well.  AFAIK, DHCP servers can hand out multiple addresses.  For certain, with DHCP off you can manually configure as many IPs for IPv4 or IPv6 as you want.

screen shot 2015-07-31 at 8 20 39 pm

FWIW, it seems to me like Facter should return *all* addresses rather than picking one, since multi-homing is pretty common.  I know given Facters current structure that looks like this (irrelevant bits removed)::

```ruby
networking => {
  dhcp => "192.168.1.1",
  interfaces => {
    Local Area Connection 2 => {
      dhcp => "192.168.1.1",
      ip => "192.168.1.209",
      ip6 => "fe80::cee:e7de:2ae0:f4c1%15",
      netmask => "255.255.255.0",
      network => "192.168.1.0",
      network6 => "fe80::%15"
    }
```

It seems like we could maintain backward compat and add a few new arrays like:

```ruby
networking => {
  dhcp => "192.168.1.1",
  interfaces => {
    Local Area Connection 2 => {
      ips => [
          {
              ip => "192.168.1.209",
              netmask => "255.255.255.0",
              network => "192.168.1.0",
          },
          {
              ip => "192.168.1.210",
              netmask => "255.255.255.0",
              network => "192.168.1.0",
          },
      ]
      ip6s => []
    }
```

Maybe you guys have already thought about this?

peterhuene:
Yeah, I think we'll go with adding all of the addresses in a new backwards-compatible field.  The existing fields will be for the first non-link-local encountered.

peterhuene:
How about this: if an interface has more than one binding for a particular protocol, then we add `bindings` for IPv4 and `bindings6` for IPv6, like so:

```
...
lo0 => {
      bindings6 => [
        {
          address => "::1",
          netmask => "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
          network => "::1"
        },
        {
          address => "fe80::1",
          network => "fe80::"
        }
      ],
      ip => "127.0.0.1",
      ip6 => "::1",
      mtu => 16384,
      netmask => "255.0.0.0",
      netmask6 => "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
      network => "127.0.0.0",
      network6 => "::1"
    },
...
```

In the above, we don't show `bindings` because there's only one bound address for IPv4 and that information is already present in `ip`, `netmask`, and `network`.

`ip`, `netmask`, and `network` (and the corresponding IPv6 fields) are populated from the first non-link-local address found or the very first binding if all bindings are link-local.

Thoughts / comments on this approach?

peterhuene:
@daenney @MikaelSmith @Iristyle ^

daenney:
That seems solid to me.

peterhuene:
Whoops, accidentally rebased off of master. Pushing fix.

Iristyle:
Yeah @peterhuene that looks like pretty much the same thing I put together aside from changing `ips` to `bindings` and `ip` to `address` - so :+1: on that part.

As far as not omitting a redundant array, as a consumer I might prefer an always-there `bindings` array, even when it duplicates what's present in `ip`, `netmask` and `network`.  Otherwise, I have to add some extra branching logic.

MikaelSmith:
@peterhuene Would hiding the bindings arrays if they have a single element? Not sure hiding works outside of top-level facts.

peterhuene:
Hiding is only respected at the top-level I believe.

peterhuene:
I'm fine with having it always present.

peterhuene:
@Iristyle @MikaelSmith thoughts on names?  I choose "bindings" because it was somewhat generic and I thought `ips` and `ip6s` could be confused with `ip` and `ip6` respectively.

MikaelSmith:
`bindings` seems fine to me, `ips` also seems odd with netmask going in there.

kylog:
:+1: on `bindings` and `bindings6`. Naming is hard.

Iristyle:
:+1: on `bindings` and `bindings6` - I agree they're better than `ips` and `ip6s.

MikaelSmith:
This affects all the platform acceptance tests. Should we update them to include the bindings arrays?

MikaelSmith:
Otherwise :+1:

peterhuene:
I'll update the acceptance tests to query the bindings.

peterhuene:
So none of the acceptance tests appear to directly look at any values on an individual interface, only what is under the `networking` fact directly (i.e. values for whatever is the primary interface).  Thoughts on how to approach this?  We'd need to hard code interface names into the tests afaik.

MikaelSmith:
Hard-coding interface names wouldn't work... those tests are generally ensuring values exist and match regexes. We could ensure bindings is a non-empty array with values matching regexes, but I'm not sure that would do anything more than the schema tests do. I think we could probably leave the acceptance tests alone.

peterhuene:
Given that this added a "networking.primary" attribute, perhaps we could assert it to be non-empty and then use it to verify the bindings on that particular interface, no hard coding required.

MikaelSmith:
:+1:

peterhuene:
@MikaelSmith Added updates to the acceptance tests, but I haven't had a chance to test them yet.

acceptance/tests/facts/debian.rb
```diff
@@ -83,6 +83,23 @@
...
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  assert_not_equal("", primary_interface)
+
+  step "Ensure bindings for the primary networking interface are present."
+  expected_bindings = {
+                        "networking.interfaces.#{primary_interface}.bindings.0.address" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.netmask" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings.0.network" => /\d+\.\d+\.\d+\.\d+/,
+                        "networking.interfaces.#{primary_interface}.bindings6.0.address" => /[a-z0-9:]+/,
```
kylog: I believe these are all hex values, so the range only needs to be `a-f` not `a-z`.

MikaelSmith:
I'm ok breaking things and fixing them after (creating an AIO for acceptance is still hard).

acceptance/tests/facts/debian.rb
MikaelSmith: Yup, could be narrower.

peterhuene: Yeah, copy pasta from the above verification, basically.  I can narrow.

MikaelSmith:
Oh, I thought I'd fixed it. Should a new bindings fact go into stable? Or better question, @kylog is there any reason to believe we'll do a 3.0.3, rather than just jumping to 3.1?

peterhuene:
That's a good question as I forgot that this was going into stable.  I'm open to backing out only the changes to the schema (I'd keep the internal structure of how we're populating things now, just change the base resolver to not output the "bindings" attributes) so that we fix this without altering the schema; would be a simple follow up commit in master to enable.

MikaelSmith:
Probably makes sense to do that.

kylog:
Re versions, 3.0.3 was a bit of a placeholder in case there was an outcry around any particularly virulent bugs. But there hasn't been an outcry so I think we'll probably end up just doing a 3.1.0.

peterhuene:
In that case, I've pushed up slightly stricter regexes in the acceptance change.

daenney:
Out of curiosity; looking at the spec they're basically duplicates of each other except for Windows. Isn't there a way to mark tests to be run on platforms X, Y, Z but not W?

MikaelSmith:
@daenney @peterhuene That would definitely reduce the duplication, we did something similar for testing the Ruby meta facts in https://github.com/puppetlabs/facter/blob/master/acceptance/tests/facts/ruby.rb.

acceptance/tests/facts/windows.rb
```diff
@@ -61,6 +61,20 @@
     assert_match(value, fact_on(agent, fact))
   end
 
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  assert_not_equal("", primary_interface)
+
+  step "Ensure bindings for the primary networking interface are present."
+  expected_bindings = {
+                        "networking.interfaces.#{primary_interface}.bindings.0.address" => /\d+\.\d+\.\d+\.\d+/,
```
MikaelSmith: We're not getting IPv6 back on Windows?

peterhuene: I assume it's this way to support windows 2003.

MikaelSmith: Oh, right :(

PR #1082, opened 2015-07-28
(FACT-1136) Move Ruby API to Leatherman

Magisus:
Changes to exception throwing and handling to make error reporting more clear and debugging easier.

### DIFF
--- a/lib/src/facts/resolvers/ruby_resolver.cc
+++ b/lib/src/facts/resolvers/ruby_resolver.cc
@@ -80,14 +80,14 @@ namespace facter { namespace facts { namespace resolvers {
     {
         data rb_data;
 
-        auto const* ruby = api::instance();
-        if (!ruby || !ruby->initialized()) {
+        auto const& ruby = api::instance();
+        if (!ruby.initialized()) {
             return rb_data;
         }
 
-        rb_data.platform = get_platform(*ruby);
-        rb_data.sitedir = get_sitedir(*ruby);
-        rb_data.version = get_version(*ruby);
+        rb_data.platform = get_platform(ruby);
+        rb_data.sitedir = get_sitedir(ruby);
+        rb_data.version = get_version(ruby);
 
         return rb_data;
     }
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -48,7 +48,7 @@ namespace facter { namespace ruby {
             // Ruby doesn't have a proper way of notifying extensions that the VM is shutting down
             // The easiest way to get notified is to have a global data object that never gets collected
             // until the VM shuts down
-            auto const& ruby = *api::instance();
+            auto const& ruby = api::instance();
             _canary = ruby.rb_data_object_alloc(*ruby.rb_cObject, this, nullptr, cleanup);
             ruby.rb_gc_register_address(&_canary);
             ruby.register_data_object(_canary);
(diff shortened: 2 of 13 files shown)


lib/src/facts/resolvers/ruby_resolver.cc
```diff
@@ -80,14 +80,14 @@ namespace facter { namespace facts { namespace resolvers {
     {
         data rb_data;
 
-        auto const* ruby = api::instance();
-        if (!ruby || !ruby->initialized()) {
```
MikaelSmith: We don't expect api::instance to throw an exception here, since Ruby should already have been initialized.

lib/src/ruby/module.cc
```diff
@@ -116,11 +116,12 @@ extern "C" {
...
 
         // Initialize ruby
-        auto ruby = api::instance();
-        if (!ruby) {
-            return;
+        try {
+            auto& ruby = api::instance();
+            ruby.initialize();
+        } catch (runtime_error ex) {
+            LOG_WARNING("%1%: facts requiring Ruby will not be resolved.", ex.what());
```
MikaelSmith: If an exception is thrown, we should return rather than continuing.

lib/src/ruby/module.cc
```diff
@@ -135,14 +136,7 @@ namespace facter { namespace ruby {
         _collection(facts),
         _loaded_all(false)
     {
-#ifdef FACTER_RUBY
```
MikaelSmith: This was the wrong place - and caused the failure in Jenkins CI - because Ruby is always first initialized in facter::ruby::initialize.

lib/src/ruby/module.cc
```diff
@@ -155,8 +149,7 @@ namespace facter { namespace ruby {
         ruby.rb_gc_register_address(&_on_message_block);
 
         // Install a logging message handler
-        on_message([this](log_level level, string const& message) {
-            auto const& ruby = *api::instance();
```
MikaelSmith: @peterhuene why didn't we inherit ruby from the parent scope? Looks safe enough to me.

lib/src/ruby/module.cc
```diff
@@ -304,13 +298,11 @@ namespace facter { namespace ruby {
 
     void module::clear_facts(bool clear_collection)
     {
-        auto ruby = api::instance();
+        auto const& ruby = api::instance();
```
MikaelSmith: This is not a static method, so I don't see any way an exception would be thrown here. So skipping the try/catch seems safe.

lib/src/ruby/ruby.cc
```diff
@@ -17,24 +18,29 @@ static const char load_puppet[] =
...
-            return false;
+#ifdef FACTER_RUBY
+        api::ruby_lib_location = FACTER_RUBY;
+#endif
+        try {
+            auto& ruby = api::instance();
+            ruby.initialize();
+            ruby.include_stack_trace(include_stack_trace);
+        } catch (runtime_error ex) {
+            LOG_WARNING("%1%: facts requiring Ruby will not be resolved.", ex.what());
```
MikaelSmith: If we get an exception, we should return false.

lib/src/ruby/module.cc
MikaelSmith: Based on my thought-process there, we should still call api::instance within the lambda, just in-case.

MikaelSmith:
:+1: once it passes CI.

MikaelSmith:
Needs commits reformatted (and probably squashed).

lib/src/ruby/module.cc
peterhuene: The catch by value is incorrect.

MikaelSmith: Oh yeah, catches should be by reference.

lib/src/ruby/module.cc
peterhuene: Sorry, missed this mention.  Of the top of my head I can't think of a reason why, but this should be fine to do as we currently have it.

PR #1083, opened 2015-07-28
(FACT-1128) Alert user if we can't manage job groups

branan:
Previously, if Facter on Windows was running in a job group that did
not allow detaching child processes, we would fail with an "Access
Denied" error when trying to attach our child processes to our own job
group.

This now checks early in the execute function for whether we are
running as part of such a job group, and alerts the user with an
appropriate error message.

### DIFF
--- a/lib/src/execution/windows/execution.cc
+++ b/lib/src/execution/windows/execution.cc
@@ -358,6 +358,22 @@ namespace facter { namespace execution {
         option_set<execution_options> const& options,
         uint32_t timeout)
     {
+        // Since we use a job object in the windows world, we want to
+        // be sure we're not in a job object, or at least able to
+        // break our processes out if we are in one.
+        BOOL in_job;
+        bool use_job_object = true;
+        if (!IsProcessInJob(GetCurrentProcess(), nullptr, &in_job)) {
+            throw execution_exception("could not determine if facter is running in a job object");
+        }
+        if (in_job) {
+            JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+            if (!QueryInformationJobObject(nullptr, JobObjectBasicLimitInformation, &limits, sizeof(limits), nullptr)
+                || !(limits.LimitFlags & JOB_OBJECT_LIMIT_BREAKAWAY_OK)) {  // short-circuits if QueryInformationJobObject fails
+                use_job_object = false;
+            }
+        }
+
         // Search for the executable
         string executable = which(file);
         log_execution(executable.empty() ? file : executable, arguments);
@@ -477,7 +493,7 @@ namespace facter { namespace execution {
             NULL,           /* Don't allow child process to inherit process handle */
             NULL,           /* Don't allow child process to inherit thread handle */
             TRUE,           /* Inherit handles from the calling process for communication */
-            CREATE_NO_WINDOW,
+            use_job_object ? CREATE_NO_WINDOW | CREATE_BREAKAWAY_FROM_JOB : CREATE_NO_WINDOW,
             options[execution_options::merge_environment] ? NULL : modified_environ.data(),
             NULL,           /* Use existing current directory */
             &startupInfo,   /* STARTUPINFO for child process */
@@ -497,21 +513,28 @@ namespace facter { namespace execution {
 
         // Use a Job Object to group any child processes spawned by the CreateProcess invocation, so we can
         // easily stop them in case of a timeout.
-        scoped_resource<HANDLE> hJob(CreateJobObjectW(nullptr, nullptr), CloseHandle);
-        if (hJob == NULL) {
-            LOG_ERROR("failed to create job object: %1%.", system_error());
-            throw execution_exception("failed to create job object.");
-        } else if (!AssignProcessToJobObject(hJob, hProcess)) {
-            LOG_ERROR("failed to associate process with job object: %1%.", system_error());
-            throw execution_exception("failed to associate process with job object.");
+        scoped_resource<HANDLE> hJob;
+        if (use_job_object) {
+            hJob = scoped_resource<HANDLE>(CreateJobObjectW(nullptr, nullptr), CloseHandle);
+            if (hJob == NULL) {
+                LOG_ERROR("failed to create job object: %1%.", system_error());
+                throw execution_exception("failed to create job object.");
+            } else if (!AssignProcessToJobObject(hJob, hProcess)) {
+                LOG_ERROR("failed to associate process with job object: %1%.", system_error());
+                throw execution_exception("failed to associate process with job object.");
+            }
         }
 
         bool terminate = true;
         scope_exit reaper([&]() {
             if (terminate) {
                 // Terminate the process on an exception
-                if (!TerminateJobObject(hJob, -1)) {
-                    LOG_ERROR("failed to terminate process: %1%.", system_error());
+                if (use_job_object) {
+                    if (!TerminateJobObject(hJob, -1)) {
+                        LOG_ERROR("failed to terminate process: %1%.", system_error());
+                    }
+                } else {
+                    LOG_WARNING("could not terminate process %1% because a job object could not be used.", procInfo.dwProcessId);
                 }
             }
         });


branan:
My windows dev environment is completely hosed at this point. I can confirm this code builds, but it won't link. I'm waiting on appveyor to see what happens here.

lib/src/execution/windows/execution.cc
```diff
@@ -358,6 +358,19 @@ namespace facter { namespace execution {
...
     {
+        // Since we use a job object in the windows world, we want to
+        // be sure we're not in a job object, or at least able to
+        // break our processes out if we are in one.
+        BOOL in_job = FALSE;
+        IsProcessInJob(GetCurrentProcess(), nullptr, &in_job);
+        if (in_job) {
+            JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+            QueryInformationJobObject(nullptr, JobObjectBasicLimitInformation, &limits, sizeof(limits), nullptr);
+            if (!(limits.LimitFlags & JOB_OBJECT_LIMIT_BREAKAWAY_OK)) {
```
peterhuene: Instead of errorring, could we simply not assign the process to a job and skip any attempt to terminate it?

branan: We could, but that would break timeouts. I'd kind of rather fail hard instead of trying to stumble forward, especially since this should be pretty rare in the real world.

peterhuene: Well, timeouts will still work, just that the child process can't be terminated (something we could certainly log a warning for if it occurs; e.g. "cannot terminate process <ID> because a break-away job could not be created" or some such).

This effectively improves the error message but still prevents facter from running when part of a restrictive job.  Would that mean the transient CI failure we were seeing as a result of this problem still occur or are we expecting tests/acceptance to start ignoring when it fails with this error message?

lib/src/execution/windows/execution.cc
```diff
@@ -358,6 +358,19 @@ namespace facter { namespace execution {
...
         uint32_t timeout)
     {
+        // Since we use a job object in the windows world, we want to
+        // be sure we're not in a job object, or at least able to
+        // break our processes out if we are in one.
+        BOOL in_job = FALSE;
+        IsProcessInJob(GetCurrentProcess(), nullptr, &in_job);
+        if (in_job) {
+            JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+            QueryInformationJobObject(nullptr, JobObjectBasicLimitInformation, &limits, sizeof(limits), nullptr);
```
peterhuene: On a side note, there's no initialization of `limits` and no error handling here if `QueryInformationJobObject` fails.

lib/src/execution/windows/execution.cc
peterhuene: Here's what I'd propose:

* Query to see if the current process is either not part of a job or part of a job and that `JOB_OBJECT_LIMIT_BREAKAWAY_OK` is set on the job.  If so, set flags to `CREATE_NO_WINDOW | CREATE_BREAKAWAY_FROM_JOB`.  Otherwise, just `CREATE_NO_WINDOW`.
* If `CREATE_BREAKAWAY_FROM_JOB` is set in the creation flags, create a new job and store in `hJob` and associate the child process with the job.  Otherwise, leave `hJob` initialized to simply `scoped_resource<HANDLE>(nullptr)`.
* The call to TerminateProcess should only occur if `hJob` is not nullptr.

Thoughts?

branan: @peterhuene that sounds reasonable, yeah.

branan:
@peterhuene This now does what we discussed - if it can't use a job object it will warn the user (but only if it needs to actually terminate the process)

MikaelSmith:
:+1:

lib/src/execution/windows/execution.cc
```diff
@@ -358,6 +358,24 @@ namespace facter { namespace execution {
         option_set<execution_options> const& options,
         uint32_t timeout)
     {
+        // Since we use a job object in the windows world, we want to
+        // be sure we're not in a job object, or at least able to
+        // break our processes out if we are in one.
+        BOOL in_job;
+        bool use_job_object = true;
+        if (!IsProcessInJob(GetCurrentProcess(), nullptr, &in_job)) {
+            throw execution_exception("could not determine if facter is in a object");
```
peterhuene: Did the message intend to say `could not determine if facter is is running in a job` or some such?  Do we know when `IsProcessInJob` can fail?

lib/src/execution/windows/execution.cc
```diff
@@ -358,6 +358,24 @@ namespace facter { namespace execution {
...
+        }
+        if (in_job) {
+            JOBOBJECT_BASIC_LIMIT_INFORMATION limits;
+            if (!QueryInformationJobObject(nullptr, JobObjectBasicLimitInformation, &limits, sizeof(limits), nullptr)) {
+                throw execution_exception("could not query the job object we are running under");
```
peterhuene: Could this call fail because the job doesn't have `JOB_OBJECT_QUERY` set?  Perhaps if it does fail, instead of throwing, we just assume we shouldn't create a job?

lib/src/execution/windows/execution.cc
```diff
@@ -497,21 +515,28 @@ namespace facter { namespace execution {
...
         // Use a Job Object to group any child processes spawned by the CreateProcess invocation, so we can
         // easily stop them in case of a timeout.
-        scoped_resource<HANDLE> hJob(CreateJobObjectW(nullptr, nullptr), CloseHandle);
-        if (hJob == NULL) {
-            LOG_ERROR("failed to create job object: %1%.", system_error());
-            throw execution_exception("failed to create job object.");
-        } else if (!AssignProcessToJobObject(hJob, hProcess)) {
-            LOG_ERROR("failed to associate process with job object: %1%.", system_error());
-            throw execution_exception("failed to associate process with job object.");
+        scoped_resource<HANDLE> hJob(nullptr, CloseHandle);
```
peterhuene: This should probably be `scoped_resource<HANDLE> hJob;` to prevent an unwanted call to `CloseHandle` with nullptr.

lib/src/execution/windows/execution.cc
```diff
@@ -497,21 +515,28 @@ namespace facter { namespace execution {
...
             if (terminate) {
                 // Terminate the process on an exception
-                if (!TerminateJobObject(hJob, -1)) {
-                    LOG_ERROR("failed to terminate process: %1%.", system_error());
+                if (use_job_object) {
+                    if (!TerminateJobObject(hJob, -1)) {
+                        LOG_ERROR("failed to terminate process: %1%.", system_error());
+                    }
+                } else {
+                    LOG_WARNING("tried to terminate child process, but we can't use a job object");
```
peterhuene: Perhaps this should be `LOG_WARNING("cannot terminate child process %1% because a job could not be created.", procInfo.dwProcessId);` or some such.  It's nice to inform the user of the process id in case they need to track it down.

peterhuene:
Other than above, :+1:

lib/src/execution/windows/execution.cc
branan: IsProcessInJob should only fail if we don't have permissions to query our own process. I don't know if this is possible.

I can clean up the error, but it seems a VERY unlikely codepath, and all we'd get from system_error is a useless "access denied"

lib/src/execution/windows/execution.cc
branan: That's reasonable

lib/src/execution/windows/execution.cc
peterhuene: :+1:

lib/src/execution/windows/execution.cc
```diff
@@ -497,21 +513,28 @@ namespace facter { namespace execution {
...
             if (terminate) {
                 // Terminate the process on an exception
-                if (!TerminateJobObject(hJob, -1)) {
-                    LOG_ERROR("failed to terminate process: %1%.", system_error());
+                if (use_job_object) {
+                    if (!TerminateJobObject(hJob, -1)) {
+                        LOG_ERROR("failed to terminate process: %1%.", system_error());
+                    }
+                } else {
+                    LOG_WARNING("could not terminate process: %1% because a job object could not be used.", procInfo.dwProcessId);
```
peterhuene: Nit pick: this reads a bit strange.  Perhaps `could not terminate process %1%: a job object could not be used.`?

peterhuene:
:+1:

lib/src/execution/windows/execution.cc
branan: whoops, I shuffled the phrasing a few times there and left a spurious colon

branan:
@peterhuene all the tests have passed. I'm gonna push up the typo fix, but if you're ready to hit the button we don't need to wait through it again

MikaelSmith:
I think this looks good now :+1:

PR #1084, opened 2015-07-29
Merge stable up to master

branan:
Stable and Master both contained separate changes to error handling in
the ruby module. This attempts to merge those two different error
paths.

This also includes a submodule bump for leatherman, since some of the
stable changes were on components that are part of leatherman in
master.

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -3,6 +3,7 @@
 #include <internal/ruby/confine.hpp>
 #include <internal/ruby/simple_resolution.hpp>
 #include <facter/facts/collection.hpp>
+#include <facter/logging/logging.hpp>
 #include <facter/version.h>
 #include <facter/export.h>
 #include <leatherman/util/environment.hpp>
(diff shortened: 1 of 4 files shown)


lib/src/ruby/module.cc
```diff
@@ -112,20 +113,47 @@ extern "C" {
      */
     void LIBFACTER_EXPORT Init_libfacter()
```
branan: At this point this function looks almost nothing like the version in stable or in master. It needs proper review

lib/src/ruby/module.cc
```diff
@@ -112,20 +113,47 @@ extern "C" {
...
+            logging_init_failed = true;
+            logging_init_error_msg = e.what();
+        }
 
         // Initialize ruby
+        api* ruby = nullptr;
         try {
-            auto& ruby = api::instance();
-            ruby.initialize();
+            ruby = &api::instance();
```
MikaelSmith: Why'd you choose to change to this, rather than keeping everything in the try block?

branan: It makes it clearer which parts of the code can actually throw an exception and makes it clearer at which points we'd be potentially failing out.

MikaelSmith: Going to let @peterhuene have a look at this, he spent more time on Ruby init than I. However, it seems highly unlikely api::instance would fail when initializing when loaded by Ruby, so this all looks reasonable.

branan: My understanding is "should not be possible unless somehow someone who's not ruby calls Init_libfacter"... in which case honestly they deserve what's coming to them ;)

MikaelSmith:
Lint is unhappy.

lib/src/ruby/module.cc
peterhuene: :+1:

PR #1085, opened 2015-07-30
(FACT-1152) Use /sys/block in the linux FS resolver

haus:
Previously /sys/class/block was used in the linux FS resolver. This is
not available on older linuxes such as el4/el5/sles10, which meant that
the size on partitions there would be returned as 0. /sys/block exists
in both older linux and newer linux, so this commit updates the linux
size function to read /sys/block instead of /sys/class/block.
The /sys/block varation requires deriving the device name for the
partition, which requires boost::regex.

### DIFF
--- a/lib/src/facts/linux/filesystem_resolver.cc
+++ b/lib/src/facts/linux/filesystem_resolver.cc
@@ -5,6 +5,7 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/lexical_cast.hpp>
+#include <boost/regex.hpp>
 #include <mntent.h>
 #include <sys/vfs.h>
 #include <set>
@@ -21,6 +22,8 @@ using namespace boost::filesystem;
 
 using boost::lexical_cast;
 using boost::bad_lexical_cast;
+
+namespace bs = boost::system;
 
 namespace facter { namespace facts { namespace linux {
 
@@ -152,7 +155,23 @@ namespace facter { namespace facts { namespace linux {
             }
 
             // Populate the size (the size is given in 512 byte blocks)
-            string blocks = file::read((path("/sys/class/block") / path(part.name).filename() / "/size").string());
+            // Check for /sys/class/block. Fall back to /sys/block if not present.
+            bs::error_code ec;
+            string blocks;
+            auto part_path = path(part.name).filename();
+            if (exists("/sys/class/block", ec) && !ec) {
+              blocks = file::read((path("/sys/class/block") / part_path / "/size").string());
+            } else {
+              boost::regex expression("\\d*(\\D+)\\d*");
+              boost::cmatch what;
+              if (boost::regex_match(part_path.string().c_str(), what, expression)) {
+                blocks = file::read((path("/sys/block") / what[1].str() / part_path / "/size").string());
+              } else {
+                LOG_WARNING("could not derive device name from partition '%1%'", part.name);
+              }
+            }
+            ec.clear();
+
             boost::trim(blocks);
             if (!blocks.empty()) {
                 try {


lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -152,7 +155,23 @@ namespace facter { namespace facts { namespace linux {
...
+            // Check for /sys/class/block. Fall back to /sys/block if not present.
+            bs::error_code ec;
+            string blocks;
+            auto part_path = path(part.name).filename();
+            if (exists("/sys/class/block", ec) && !ec) {
+              blocks = file::read((path("/sys/class/block") / part_path / "/size").string());
+            } else {
+              boost::regex expression("(\\D+)\\d*");
+              boost::cmatch what;
+              if (boost::regex_match(part_path.string().c_str(), what, expression)) {
```
MikaelSmith: https://github.com/puppetlabs/leatherman/blob/master/util/inc/leatherman/util/regex.hpp#L70 has a re_search helper that might make this a little simpler. You can do things like

```
string device;
if (re_search(path_part.string(), boost::regex("(\\D+)\\d*"), device)) ...
```

MikaelSmith: Also, do you really mean anything that's not `[0-9]`?

haus: I borrowed the regex directly from ruby facter at https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/partitions/linux.rb#L66

haus: But I'm open to other options.

MikaelSmith: Ah, that works.

haus: Looks like the re_search helper isn't in the vendored version of leatherman here (stable branch).

MikaelSmith: Ah right, it's in `internal/util/regex.hpp`.

MikaelSmith:
Was there any more work planned on this? I thought I saw discussion about some things that it's still missing.

haus:
@MikaelSmith I'm not sure. There is still something weird going on...here is ruby facter on el5

```
facter partitions
{"sdb1"=>{"size"=>"16777184"}, "sda1"=>{"label"=>"/boot", "size"=>"208782", "filesystem"=>"ext3", "uuid"=>"ccee0da7-8190-4ede-ad69-4714afbc98a6", "mount"=>"/boot"}, "sda2"=>{"size"=>"16563015"}}
```

and cfacter

```
{
  /dev/VolGroup00/LogVol00 => {
    filesystem => "ext3",
    size => "0 bytes",
    size_bytes => 0,
    uuid => "eff0db7e-020b-409f-94af-c758112f745b"
  },
  /dev/VolGroup00/LogVol01 => {
    filesystem => "swap",
    size => "0 bytes",
    size_bytes => 0
  },
  /dev/cdrom => {
    filesystem => "iso9660",
    label => "VMware Tools",
    size => "0 bytes",
    size_bytes => 0
  },
  /dev/hdc => {
    filesystem => "iso9660",
    label => "CentOS",
    size => "0 bytes",
    size_bytes => 0
  },
  /dev/mapper/VolGroup00-LogVol00 => {
    filesystem => "ext3",
    mount => "/",
    size => "0 bytes",
    size_bytes => 0,
    uuid => "eff0db7e-020b-409f-94af-c758112f745b"
  },
  /dev/mapper/VolGroup00-LogVol01 => {
    filesystem => "swap",
    size => "0 bytes",
    size_bytes => 0
  },
  /dev/sda1 => {
    filesystem => "ext3",
    label => "/boot",
    mount => "/boot",
    size => "101.94 MiB",
    size_bytes => 106896384,
    uuid => "ccee0da7-8190-4ede-ad69-4714afbc98a6"
  }
}
```

Ruby facter gets the listing of partitions differently than cfacter, and it included unmounted partitions. I'm not sure which is the correct behavior here.

peterhuene:
I think the 2.x behavior is correct with respect to unmounted partitions.  We should be displaying them.

MikaelSmith:
What changed in your new commit?

haus:
@MikaelSmith the regex was changed to include optional leading digits. i still need to see if they are valid device/partition names, but it shouldn't break anything.

MikaelSmith:
Shouldn't `facter partitions` still work?

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -152,7 +155,23 @@ namespace facter { namespace facts { namespace linux {
             }
 
             // Populate the size (the size is given in 512 byte blocks)
-            string blocks = file::read((path("/sys/class/block") / path(part.name).filename() / "/size").string());
+            // Check for /sys/class/block. Fall back to /sys/block if not present.
+            bs::error_code ec;
+            string blocks;
+            auto part_path = path(part.name).filename();
+            if (exists("/sys/class/block", ec) && !ec) {
+              blocks = file::read((path("/sys/class/block") / part_path / "/size").string());
```
MikaelSmith: Indentation doesn't follow convention. Also, I'm mostly seeing that blkid doesn't find any devices (in containers, i.e. Travis and Docker).

MikaelSmith:
blkid doesn't seem to find blocks on containers (Docker, and thus Travis).

As opposed to Facter 2.4.4 output in a container:
```
$ facter partitions
{"sda1"=>{"size"=>"38973690", "mount"=>"/etc/resolv.conf"}, "sda2"=>{"size"=>"1975932"}}
```

peterhuene:
I'm with @MikaelSmith on just rewriting the partitions resolving code.

To fix the partitions fact we should:

* scan /sys/block for entries containing a 'device' link (like Facter 2.x).
* map various symlinks under /dev/disk/by-XXX to populate `label`, `partlabel`, `uuid`, and `partuuid` (unlike Facter 2.x, that relies on blkid for some of this).
* only use libblkid to detect the filesystem type on unmounted disks; mounted disks should favor the information in the `mountpoints` fact that was already retrieved. 

Also, the mountpoints fact needs to take into acount `/dev/root` (should translate to whatever `root=` in `/proc/cmdline`.

haus:
This all sounds sane. I'm not sure I'll have bandwidth to pick that up or not.

peterhuene:
I can take it over.

peterhuene:
Closing this as I'll be putting up a new PR to fix this and the reliance on libblkid.

PR #1086, opened 2015-07-30
(FACT-1123) Adding acceptance test for facter with invalid locale

shaigy:
This commit adds an acceptance test to make sure facter is not crashing
when locale is set to an invalid value

### DIFF
--- a/acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
+++ b/acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
@@ -1,0 +1,9 @@
+test_name "ticket 1123  facter should not crash with invalid locale setting"
+confine :except, :platform => 'windows'
+agents.each do |host|
+  step "set an invalid value for locale and run facter"
+  result = on host, 'LANG=ABCD facter facterversion'
+  fail_test "facter did not warn about the locale " unless
+    result.stderr.include? 'WARN'
+end
+


MikaelSmith:
You could use `[skip ci]` on this.

acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
```diff
@@ -0,0 +1,7 @@
+test_name "ticket 1123  facter should not crash with invalid locale setting"
+confine :except, :platform => 'windows'
+agents.each do |host|
+  step "set an invalid value for locale and run facter"
+  on host,( "LANG=ABCD facter facterversion"), :acceptable_exit_codes => [ 0 , 2 ]
```
MikaelSmith: Why is 2 an acceptable exit code? We really don't expect that to happen in this test, and not sure how to cause it.

Also, it'd be nice to confirm that something is printed to stdout, and nothing to stderr.

shaigy: Hmm...didn't know facter exit code was different from puppet exit code. 
I will also add a check for "WARN" in the stdout and null in stderr

shaigy:
@MikaelSmith Restricted the acceptable exit code to 0. 
The warning message is actually printed out to stderr, not stdout and I added a check for text "WARN" in the stderr. 
There is not anything specific that we can check for in stdout.

acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
```diff
@@ -0,0 +1,9 @@
+test_name "ticket 1123  facter should not crash with invalid locale setting"
+confine :except, :platform => 'windows'
+agents.each do |host|
+  step "set an invalid value for locale and run facter"
+  result = on host, 'LANG=ABCD facter facterversion'
+  fail_test "facter did not warn about the locale " unless
+    result.stderr.include? 'WARN'
```
branan: I'd prefer a more precise check, so that we don't accidentally pass if something causes an unrelated warning in the test environment.

How about looking for "locale environment variables were bad"

MikaelSmith: @shaigy's out, so I'll re-open a new PR with those fixes.

MikaelSmith:
Opened a new PR against stable with @branan's concerns addressed: https://github.com/puppetlabs/facter/pull/1088

PR #1089, opened 2015-08-04
(maint) Update LIBFACTER_PLATFORM_LIBRARIES for SunOS

haus:
nanosleep, sched_yield and clock_gettime are required for building
libfacter_test. These libraries live in librt on solaris, so this commit
adds rt to the LIBFACTER_PLATFORM_LIBRARIES for SunOS.


MikaelSmith:
:+1:

PR #1090, opened 2015-08-05
(FACT-1155) Fix libfacter unit tests under Solaris.

peterhuene:


MikaelSmith:
Has this been tested on Solaris?

haus:
I can try this on my build this morning.

haus:
Oops...I'm building off of master currently. Let me see if I can just build stable instead easily.

haus:
```
===============================================================================
All tests passed (10549 assertions in 66 test cases)
```

peterhuene:
:+1: thanks @haus!

MikaelSmith:
Way more complicated, but looks fair to me. Is there the possibility of a significant performance hit? Would we want to limit this to only Solaris?

peterhuene:
I don't think this will hurt performance much; the groups do end up getting cached.  Given that Solaris may not be the only platform affected (AIX?), I'd prefer to just use `stat` over #ifdef'ing.

PR #1092, opened 2015-08-05
(FACT-1137) Use dmidecode if Linux kernel does not support DMI.

peterhuene:
Linux kernels older than 2.6.23 do not support exposing DMI information
via sysfs.  To support those kernels, we should fallback to parsing the
output of the dmidecode utility (when root).

### DIFF
--- a/lib/src/facts/linux/dmi_resolver.cc
+++ b/lib/src/facts/linux/dmi_resolver.cc
@@ -1,5 +1,7 @@
 #include <internal/facts/linux/dmi_resolver.hpp>
+#include <internal/util/regex.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <facter/execution/execution.hpp>
 #include <facter/util/file.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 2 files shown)


peterhuene:
Verified on a Centos 5.11 VM:

```
vagrant@localhost debug]$ sudo bin/facter dmi
{
  bios => {
    release_date => "12/01/2006",
    vendor => "innotek GmbH",
    version => "VirtualBox"
  },
  board => {
    asset_tag => "Not Specified",
    manufacturer => "Oracle Corporation",
    product => "VirtualBox",
    serial_number => "0"
  },
  chassis => {
    asset_tag => "Not Specified",
    type => "Other"
  },
  manufacturer => "innotek GmbH",
  product => {
    name => "VirtualBox",
    serial_number => "0",
    uuid => "FOOBAR"
  }
}
```

lib/src/facts/linux/dmi_resolver.cc
```diff
@@ -56,4 +71,124 @@ namespace facter { namespace facts { namespace linux {
         return value;
     }
 
+    void dmi_resolver::parse_dmidecode_output(data& result, string& line, int& dmi_type)
```
MikaelSmith: I think it would make sense to test this function; that helps document our expectation of the output.

peterhuene: I'll add some tests.

peterhuene:
Closing this while I work on the tests.

PR #1094, opened 2015-08-08
(maint) refactor acceptance rakefile and options files

er0ck:
This change contains many changes to the acceptance Rakefile to put most
of the options back in the options files and use new beaker features for
creating preserved_hosts files.
It removes some of the functionality in parsing log files, but this was
all broken by beaker changes anyway. If people need this we can add it back in at that
time.

[skip ci]

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -1,30 +1,19 @@
 require 'rake/clean'
 require 'pp'
 require 'yaml'
-$LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), 'lib'))
-require 'puppet/acceptance/git_utils'
-extend Puppet::Acceptance::GitUtils
 
 ONE_DAY_IN_SECS = 24 * 60 * 60
-REPO_CONFIGS_DIR = "repo-configs"
-CLEAN.include('*.tar', REPO_CONFIGS_DIR, 'merged_options.rb')
+PRESERVED_HOSTS_FILENAME = 'hosts_preserved.yml'
+SUTS_FILENAME = 'sut.log'
+CLEAN.include('*.tar', 'merged_options.rb')
 
 module HarnessOptions
 
   DEFAULTS = {
-    :type => 'git',
-    :helper => ['lib/helper.rb'],
     :tests  => ['tests'],
-    :log_level => 'debug',
-    :color => false,
-    :root_keys => true,
     :ssh => {
       :keys => ["~/.ssh/id_rsa-acceptance"],
     },
-    :xml => true,
-    :timesync => false,
-    :repo_proxy => true,
-    :add_el_extras => true,
     :preserve_hosts => 'onfail',
   }
 
@@ -36,7 +25,7 @@ module HarnessOptions
     end
 
     def get_options(file_path)
-      puts file_path
+      puts "Attempting to merge config file: #{file_path}"
       if File.exists? file_path
         options = eval(File.read(file_path), binding)
       else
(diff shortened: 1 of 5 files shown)


MikaelSmith:
I just tried `be rake ci:test:aio SHA=1.2.2 TEST=tests/no_errors_on_stderr.rb CONFIG=config/nodes/redhat-6-x86_64.yaml` and it didn't work with this PR.

er0ck:
thanks @MikaelSmith.

didn't realize i had removed that.  probably should be deprecated to match beaker's option deprecation

MikaelSmith:
:+1: this looks ok to me.

MikaelSmith:
I'd vote for some more feedback before merging.

er0ck:
yes, please.  @joshcooper if you get a chance, some perspective on this would be great.  i'm going to PR something similar against puppet and hiera after this goes through.  
later we should probably fix the git method for facter (pull source and build it on the SUT)

MikaelSmith:
Needs a rebase now.

er0ck:
@MikaelSmith done.

MikaelSmith:
Needs a rebase, and @joshcooper did you have any input?

acceptance/Rakefile
```diff
@@ -85,88 +96,40 @@ def beaker_test(mode = :packages, options = {})
...
+      Dir.glob("#{dir}/*").each do |date_dir|
+        preserved_config_path = "#{date_dir}/#{SUTS_FILENAME}"
+        if preserved_config = File.exists?(preserved_config_path)
+          directory = File.new(date_dir)
+          yesterday = Time.now - secs_ago.to_i
+          if directory.ctime > yesterday
+            File.open(preserved_config_path, 'r') do |file|
+              if file.ctime > yesterday
+                file.each_line do |line|
+                  matchdata = /(\w+\.delivery\.puppetlabs\.net) \(.*?\)$/.match(line.encode!('UTF-8', 'UTF-8', :invalid => :replace))
```
joshcooper: I would expect the hostname to be `US_ASCII`, can we drop the `encode`?

acceptance/setup/aio/pre-suite/010_Install.rb
```diff
@@ -7,7 +7,7 @@
 test_name "Install Packages"
 
 step "Install repositories on target machines..." do
-  sha = ENV['SHA']
+  sha = ENV['VERSION'] || ENV['SHA']
```
joshcooper: I'm wondering if this will have an implications in CI? If the job defines `VERSION` environment variable to be something other than `SHA`, then acceptance will likely fail. Also metaquestion, is `VERSION` the preferred way of specifying the "thing" to test?

acceptance/Rakefile
er0ck: inherited... i forget if this even works.  
i can certainly remove the `encode`.

most of this functionality can/should move to beaker/vmpooler API.
which leads me to investigating a rakefile gem for at least puppet/facter/hiera, and i'd love to jettison the acceptance Rakefile altogether as in other projects.... but i digress

acceptance/setup/aio/pre-suite/010_Install.rb
er0ck: hmmm. i hadn't noticed that.  VERSION is used in certain cases and seems more uniform than SHA which is not always what the value 'is'.  
in many jobs these days i think this is also called GIT_SHA.
argh...  attempts to move to uniformity are always punished

er0ck: @joshcooper updated, not sure why it's not out-dated.

er0ck:
@MikaelSmith done.

branan:
@er0ck @MikaelSmith It's been six weeks with no activity. Should this be rebased/resurected post-agent-1.3.0?

er0ck:
yeah... i'll refactor this and address josh's concerns this week

er0ck:
@MikaelSmith rebased and ready to go.

MikaelSmith:
:+1: on deleting code. I'll try it out and make sure some workflows still work.

MikaelSmith:
This looks ok. I say merge it and see how yells.

On a side note, maybe we should remove the `git` and `package` targets unless we expect to make them work.

er0ck:
good idea removing the package options for sure.  i don't think we have packages?
the git option we should probably remove under YAGNE rules.  can add it back and make it work if we want it.

kylog:
Agreed on all counts. Let it rip and see what falls out. Including +1 on removing git/package, since they're  bitrotted/obsolete at this point.

er0ck:
FYI, passed in CI after the merge.

PR #1095, opened 2015-08-09
(maint) Treat zero processor counts as missing facts

kylog:
Previously, the processor resolver would faithfully report a value
of 0 for the logical or physical processor counts. However, that is
counterintuitive (since clearly no system ever actually has zero
processors) and is inconsistent with handling of other zero values
both in the processor facts (e.g. a speed of zero is treated as a
missing fact) and with other facts (e.g. zero network interfaces
is treated as a missing fact).

So this simply treats a value of zero for the logical and physical
counts as a missing fact.

This is of real-world interest for platforms, like OpenBSD, which
don't have a way of reporting the physical processor count.

### DIFF
--- a/lib/src/facts/resolvers/processor_resolver.cc
+++ b/lib/src/facts/resolvers/processor_resolver.cc
@@ -37,11 +37,15 @@ namespace facter { namespace facts { namespace resolvers {
             cpus->add("isa", make_value<string_value>(move(data.isa)));
         }
 
-        facts.add(fact::processor_count, make_value<integer_value>(data.logical_count, true));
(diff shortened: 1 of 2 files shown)


kylog:
Once I looked at this a bit more, I changed my mind relative to my comment at https://github.com/puppetlabs/facter/pull/1073#issuecomment-125306596, and this PR goes with "0 means don't report the fact" as opposed to introducing the notion of supporting reported processor counts of 0, and using -1 as the "don't report" value.

jasperla:
+1, I'll rebase #1073 when this has been committed.

PR #1099, opened 2015-08-12
(FACT-1159) Fix cross-compile where Ruby is absent

MikaelSmith:
Incorporates a fix in Leatherman to remove a requirement on Ruby, and
changes Facter to generate Ruby-related template files even when Ruby is
absent, so they can be packaged up and used in environments where it is.

Unit tests will still fail if Ruby is missing, adding verification that
everything Ruby-related still works.


MikaelSmith:
Waiting on the Leatherman fix to be merged.

MikaelSmith:
Leatherman changes merged.

PR #1101, opened 2015-08-12
(FACT-1152) Fix Linux partitions fact to enumerate devices like 2.x. 

peterhuene:
Facter 2.x looked through the /sys/block structure to detect partitions
to populate the fact.  3.x changed this behavior to use libblkid to
enumerate the devices.

Using libblkid to enumerate the devices is problematic as probing
requires root access and if no cache file is present, normal users won't
the partitions populated correctly.

This changes the probing to the 2.x with additional support for
displaying mapped and loop devices.

Now libblkid is only used to query the device attributes, meaning if
libblkid support isn't enabled, the partitions will just show their size
and mount points.

This also fixes the size detection for mapped devices as the opening of
the size file was being done with the mapping name and not the
underlying block device name.  That resulted in some partitions showing
as having 0 bytes.  Now the correct name is used under /sys/block.

This also adds a "backing_file" attribute for partitions that are loop
devices.

### DIFF
--- a/lib/inc/facter/util/directory.hpp
+++ b/lib/inc/facter/util/directory.hpp
@@ -17,19 +17,19 @@ namespace facter { namespace util {
     {
         /**
          * Enumerates the files that match the given pattern in the given directory.
-         * @param path The directory path to search for the files.
+         * @param directory The directory to search for the files.
          * @param callback The callback to invoke when a matching file is found.
          * @param pattern The pattern to filter the file names by.  If empty, all files are passed.
          */
-        static void each_file(std::string const& path, std::function<bool(std::string const&)> callback, std::string const& pattern = {});
+        static void each_file(std::string const& directory, std::function<bool(std::string const&)> const& callback, std::string const& pattern = {});
 
         /**
          * Enumerates the subdirectories in the given directory.
-         * @param path The directory path to search for the subdirectories.
+         * @param directory The directory to search for the subdirectories.
          * @param callback The callback to invoke when a matching subdirectory is found.
          * @param pattern The pattern to filter the subdirectory names by.  If empty, all subdirectories are passed.
          */
-        static void each_subdirectory(std::string const& path, std::function<bool(std::string const&)> callback, std::string const& pattern = {});
+        static void each_subdirectory(std::string const& directory, std::function<bool(std::string const&)> const& callback, std::string const& pattern = {});
     };
 
 }}  // namespace facter::util
--- a/lib/inc/internal/facts/linux/filesystem_resolver.hpp
+++ b/lib/inc/internal/facts/linux/filesystem_resolver.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "../resolvers/filesystem_resolver.hpp"
+#include <map>
 
 namespace facter { namespace facts { namespace linux {
 
@@ -25,6 +26,7 @@ namespace facter { namespace facts { namespace linux {
         void collect_mountpoint_data(data& result);
         void collect_filesystem_data(data& result);
         void collect_partition_data(data& result);
+        void populate_partition_attributes(partition& part, std::string const& device_directory, void* cache, std::map<std::string, std::string> const& mountpoints);
     };
 
 }}}  // namespace facter::facts::linux
--- a/lib/src/facts/linux/filesystem_resolver.cc
+++ b/lib/src/facts/linux/filesystem_resolver.cc
@@ -1,6 +1,8 @@
 #include <internal/facts/linux/filesystem_resolver.hpp>
 #include <internal/util/scoped_file.hpp>
+#include <internal/util/regex.hpp>
 #include <facter/util/file.hpp>
+#include <facter/util/directory.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/algorithm/string.hpp>
 #include <boost/filesystem.hpp>
@@ -18,6 +20,7 @@ using namespace std;
 using namespace facter::facts;
 using namespace facter::util;
 using namespace boost::filesystem;
+namespace sys = boost::system;
 
 using boost::lexical_cast;
 using boost::bad_lexical_cast;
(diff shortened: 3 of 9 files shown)


lib/inc/facter/util/directory.hpp
```diff
@@ -17,19 +17,19 @@ namespace facter { namespace util {
     {
         /**
          * Enumerates the files that match the given pattern in the given directory.
-         * @param path The directory path to search for the files.
+         * @param directory The directory to search for the files.
          * @param callback The callback to invoke when a matching file is found.
          * @param pattern The pattern to filter the file names by.  If empty, all files are passed.
          */
-        static void each_file(std::string const& path, std::function<bool(std::string const&)> callback, std::string const& pattern = {});
+        static void each_file(std::string const& path, std::function<bool(std::string const&)> const& callback, std::string const& pattern = {});
```
MikaelSmith: You updated the docs but not the variable itself.

lib/inc/internal/facts/linux/filesystem_resolver.hpp
```diff
@@ -25,6 +26,7 @@ namespace facter { namespace facts { namespace linux {
         void collect_mountpoint_data(data& result);
         void collect_filesystem_data(data& result);
         void collect_partition_data(data& result);
+        void populate_partition_attributes(partition& part, std::string const& device_directory, void* cache, std::map<std::string, std::string> const& mountpoints);
```
MikaelSmith: Why pass in partition rather than return it?

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -41,17 +43,44 @@ namespace facter { namespace facts { namespace linux {
...
         while (mntent *ptr = getmntent_r(file, &entry, buffer, sizeof(buffer))) {
+            string device = ptr->mnt_fsname;
+
             // Skip over anything that doesn't map to a device
-            if (!boost::starts_with(ptr->mnt_fsname, "/dev/")) {
+            if (!boost::starts_with(device, "/dev/")) {
                 continue;
             }
 
+            // If the "root" device, lookup the actual device from the kernel options
```
MikaelSmith: What system exposed this?

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -41,17 +43,44 @@ namespace facter { namespace facts { namespace linux {
...
+            // This is done because not all systems symlink /dev/root
+            if (device == "/dev/root") {
+                if (root_device.empty()) {
+                    file::each_line("/proc/cmdline", [&](string& line) {
+                        auto begin = line.find("root=");
```
MikaelSmith: re_search would've made this pretty succinct. I wonder what the overhead would be.

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -83,97 +112,139 @@ namespace facter { namespace facts { namespace linux {
...
-        // Populate a map of device -> mountpoint
-        map<string, string> device_mountpoints;
-        for (auto const& point : result.mountpoints) {
-            device_mountpoints.insert(make_pair(point.device, point.name));
+#ifdef USE_BLKID
+        blkid_cache actual = nullptr;
+        if (blkid_get_cache(&actual, nullptr) == 0) {
+            cache = actual;
+        } else {
+            LOG_DEBUG("blkid_get_cache failed: partition attributes are not unavailable.");
```
MikaelSmith: A double negative?

lib/inc/internal/facts/linux/filesystem_resolver.hpp
MikaelSmith: Ah, you make use of values in the partition.

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -83,97 +112,139 @@ namespace facter { namespace facts { namespace linux {
...
+        const int block_size = 512;
 
-            result.partitions.emplace_back(move(part));
+        // Read the size
+        string blocks = file::read(device_directory + "/size");
+        boost::trim(blocks);
+        if (!blocks.empty()) {
+            try {
+                part.size = lexical_cast<uint64_t>(blocks) * block_size;
+            } catch (bad_lexical_cast& ex) {
```
MikaelSmith: LOG_DEBUG if size doesn't contain what we expect?

lib/src/facts/linux/filesystem_resolver.cc
peterhuene: Not much; I can fix.

lib/src/facts/linux/filesystem_resolver.cc
peterhuene: Looks like I edited the string at a later time and didn't remove the `un`.  Will fix.

lib/src/facts/linux/filesystem_resolver.cc
peterhuene: Will add.

lib/src/facts/linux/filesystem_resolver.cc
MikaelSmith: Not line 131 has the same typo.

lib/src/facts/linux/filesystem_resolver.cc
peterhuene: iirc, an arch system.

lib/inc/facter/util/directory.hpp
peterhuene: Weird, this passed doxygen warning?

MikaelSmith: I don't think I fixed that up on stable, only master.

lib/inc/internal/facts/linux/filesystem_resolver.hpp
peterhuene: Yeah, it's really due to device mapper devices; it needs to set the name before populating the attributes.

lib/inc/facter/util/directory.hpp
peterhuene: Oh right, good catch then.

lib/src/facts/linux/filesystem_resolver.cc
peterhuene: Added.

MikaelSmith:
:+1:

PR #1102, opened 2015-08-13
(FACT-1137) Use dmidecode if Linux kernel does not support DMI.

peterhuene:
Linux kernels older than 2.6.23 do not support exposing DMI information
via sysfs.  To support those kernels, we should fallback to parsing the
output of the dmidecode utility (when root).

### DIFF
--- a/lib/src/facts/linux/dmi_resolver.cc
+++ b/lib/src/facts/linux/dmi_resolver.cc
@@ -1,5 +1,7 @@
 #include <internal/facts/linux/dmi_resolver.hpp>
+#include <internal/util/regex.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <facter/execution/execution.hpp>
 #include <facter/util/file.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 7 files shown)


peterhuene:
@MikaelSmith Added some dmidecode tests.

lib/src/facts/linux/dmi_resolver.cc
```diff
@@ -14,22 +16,155 @@ namespace facter { namespace facts { namespace linux {
...
+        // Get the value and trim it
+        string value = line.substr(it->size());
+        boost::trim(value);
+
+        // Calculate the index into the header vector
+        size_t index = it - headers.begin();
+
+        // Assign to the appropriate member
+        string* member = nullptr;
+        switch (dmi_type) {
```
MikaelSmith: This is not my favorite switch statement, but the alternative would create a vector of sections of pairs of string and address of a variable in result. I can't decide which is worse.

peterhuene: I know, parsing this is...complex.  I was having difficulty balancing performance and maintainability.

peterhuene:
I'm going to push up a change to the "alternative" dmidecode output so that we properly hit the alternative for "chassis type".

lib/src/facts/linux/dmi_resolver.cc
MikaelSmith: I'd be tempted to create the specific section you need after you know dmi_type, and make it pairs of strings and addresses in result. I'm not sure how much making it static is important.

peterhuene: It is called on each line of output, could have moved these structures out of the member.  It should be const though.

MikaelSmith: Oh right. Not going to belabor it then.

MikaelSmith:
:+1:

lib/src/facts/linux/dmi_resolver.cc
peterhuene: Plus, this is fallback code on really old kernels only ¯\_(ツ)_/¯

PR #1105, opened 2015-08-13
(FACT-1147) Fix networking resolver to ignore all IPv4 loopback addrs.

peterhuene:
Fixing the networking resolver to ignore all IPv4 loopback addresses
(127.0.0.0/8) instead of just 127.0.0.1.

Without this, multiple IPv4 loopbacks assigned to an interface result in
respecting latter ones if 127.0.0.1 comes first.  Instead, all should be
ignored and it should then consider the first to be primary.


MikaelSmith:
:+1:

PR #1107, opened 2015-08-14
(FACT-1141) Make libfacter.so suitable for dynamically loading

branan:
On AIX, a library has to be built specially so that it can be
dynamically loaded. Since we want libfacter.so to be loadable by the
ruby process, we need to make it dynamically loadable.

The necessary LDFLAGS were copied directly from how Ruby builds its
own plugins.


branan:
Since nobody can confirm this on their own, here's my verification:

```
[0] [AIX] root@pe-aix-71-client:/home/facter/facter/build # /opt/puppetlabs/puppet/bin/irb 
irb(main):001:0> require './lib/libfacter.so'
2015-08-14 14:05:42.278554 WARN  puppetlabs.facter - locale environment variables were bad; continuing with LANG=C
=> true
irb(main):002:0> Facter.value("ruby")
=> {"platform"=>"powerpc-aix7.1.2.0", "sitedir"=>"/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0", "version"=>"2.1.6"}
```

PR #1108, opened 2015-08-17
(maint) Restrict Ruby requirement

MikaelSmith:
Previously if Ruby was not found while building, facter.rb would not be
installed, and links for libfacter.bundle to libfacter.so wouldn't be
created.

Isolate the Ruby dependency to only be needed to find RUBY_VENDORDIR,
and only if RUBY_LIB_INSTALL is not specified.


haus:
This allows my cross-compile to succeed in laying down facter.rb. YAY!

PR #1109, opened 2015-08-17
(FACT-1152) Remove dependency on /sys/block/<device>/<part>/partition.

peterhuene:
Currently facter is using the existence of a partition file underneath
each partition directory to respect the directory as a partition in the
partitions fact.  Unfortunately this file only exists on kernels more
recent than the one used in Centos 5 and SLES 10.

Instead of checking the file, this change reverts back to the behavior
in 2.x which relied on naming conventions in the path; notably the
partition directory's file name starts with the device directory's
file name.


MikaelSmith:
Yuck.

peterhuene:
Indeed.  I grow weary of supporting ancient kernels :/

MikaelSmith:
:+1:

PR #1112, opened 2015-08-19
(FACT-970) Implement processor resolver for AIX

branan:
This uses the AIX [object data manager][odm] to query the system for all available processors and hardware threads.

Since I'm not sure if other resolvers will use the ODM, it currently uses the APIs directly. If we see more use of it in other resolvers we should consider abstracting things, as there is some serious weirdness with this API.

[odm]: https://www-01.ibm.com/support/knowledgecenter/ssw_aix_71/com.ibm.aix.genprogc/odm.htm

### DIFF
--- a/lib/src/facts/aix/processor_resolver.cc
+++ b/lib/src/facts/aix/processor_resolver.cc
@@ -1,0 +1,143 @@
+#include <internal/facts/aix/processor_resolver.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <leatherman/util/scope_exit.hpp>
+
+#include <cstdint>
+#include <stdexcept>
+#include <odmi.h>
+#include <sys/cfgodm.h>
+
+using namespace std;
+
+// unlike just about every other error code to string function in the
+// world, odm_err_msg doesn't just return its pointer - it wants a
+// char** argument to put it in. This wraps that ugly API in a more
+// strerror()-like interface for ease of use.
+static const char* odm_error_string() {
+    static char* msg;
+    int result = odm_err_msg(odmerrno, &msg);
+    if (result < 0) {
+        return "failed to retrieve ODM error message";
+    } else {
+        return msg;
+    }
+}
+
+// ODM return value handling is kind of gross, so this saves us from
+// it a bit.  Even though it nominally returns a pointer, it can have
+// the value "-1" on error. In that case, we want to get the error
+// string and throw an exception. This handles all of that in all the
+// places we need it.
+#define check_odm_return(var, msg) if (reinterpret_cast<intptr_t>(var) == -1) { throw runtime_error((boost::format(msg ": %1%") % odm_error_string()).str()); }
+
+struct physical_processor
+{
+     string type;
+     long long frequency;
+     int smt_threads;
+     bool smt_enabled;
+};
+
+namespace facter { namespace facts { namespace aix {
+
+    processor_resolver::data processor_resolver::collect_data(collection& facts)
+    {
+        auto result = posix::processor_resolver::collect_data(facts);
+
+        // On AIX, we query the object data manager (odm) for
+        // processor information. This is a semi-hierarchical
+        // datastore of all the information about the system. For
+        // processors, we need to go through three links:
+        //
+        // 1. We query for all "predefined devices" with the
+        // "processor" class. I don't know if it's actually possible
+        // to get more than one result here - on the Puppet LPARs
+        // there is only one.
+        //
+        // 2. For each predefined device, we query the "custom
+        // devices". These represent the actual processors in the
+        // machine.
+        //
+        // 3. For each custom device, we query its attributes. These
+        // are things like frequency, type, and SMT information.
+
+        if (odm_initialize() < 0) {
+            throw runtime_error((boost::format("failed to initialize ODM: %1%") % odm_error_string()).str());
+        }
+        leatherman::util::scope_exit cleanup(odm_terminate);
+
+        CLASS_SYMBOL PdDv_class = odm_mount_class(const_cast<char*>("PdDv"));
+        CLASS_SYMBOL CuDv_class = odm_mount_class(const_cast<char*>("CuDv"));
+        CLASS_SYMBOL CuAt_class = odm_mount_class(const_cast<char*>("CuAt"));
+        check_odm_return(PdDv_class, "couldn't mount PdDv ODM class");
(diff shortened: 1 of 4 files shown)


branan:
This is kinda gross and c-like because I'm dealing with a kinda-gross C API.

I've tried to restrict c-style casts to places where I was getting warnings about `const char*` conversions (bad function signatures on IBMs part), and places where I'm converting a `void*` return to a concrete type.

I could have used const_cast and static_cast, but they add a lot of noise to what are already kind of messy lines, and the c-style casts here are fairly idiomatic when dealing with this sort of API.

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -0,0 +1,117 @@
...
+    processor_resolver::data processor_resolver::collect_data(collection& facts)
+    {
+        auto result = posix::processor_resolver::collect_data(facts);
+
+        if (odm_initialize() < 0) {
+            throw runtime_error((boost::format("failed to initialize ODM: %1%") % odm_error_string()).str());
+        }
+        leatherman::util::scope_exit(odm_terminate);
+
+        CLASS_SYMBOL PdDv_class = odm_mount_class((char *)"PdDv");
```
peterhuene: C++-style casts are preferred over C-style casts.  `const_cast<char*>("PdDv")` is a clearer indication that this terrible, terrible function isn't const-correct (at least one hopes it doesn't invoke UB by writing to these strings).

branan:
Of course, lint still hates those casts. :(

lib/src/facts/aix/processor_resolver.cc
branan: I commented globally about my decision to use c-style casts when dealing with c-APIs, but the linter is grumpy enough I'm just gonna go ahead and swap them out

branan:
@peterhuene alright, updated with C++ casts. Also pushed up a bunch of comments.

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -0,0 +1,143 @@
...
+                if (attribute->attribute == string("frequency")) {
+                    proc.frequency = stoll(attribute->value);
+                } else if (attribute->attribute == string("type")) {
+                    proc.type = attribute->value;
+                } else if (attribute->attribute == string("smt_threads")) {
+                    proc.smt_threads = stoi(attribute->value);
+                } else if (attribute->attribute == string("smt_enabled")) {
+                    proc.smt_enabled = (attribute->value == string("true"));
+                } else {
+                    LOG_INFO("don't know what to do with processor attribute %1%", attribute->attribute)
```
branan: This is info because it's really not warning-level (so what if there's more info than we know what to do with?) But is the sort of thing a user might care about if they don't see facter reporting information they expect. Info seemed better than debug here, since debug is a lot of noise for users to comb through.

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -0,0 +1,143 @@
...
+                free(attribute);
+                attribute = static_cast<CuAt*>(odm_get_next(CuAt_class, nullptr));
+            }
+
+            result.physical_count++;
+
+            if (result.speed == 0) {
+                result.speed = proc.frequency;
+            } else if (result.speed != proc.frequency) {
+                LOG_WARNING("mismatched processor frequencies found; facter will only report one of them");
```
branan: How do we handle this on other platforms? Do we just assume all processors are identical and not even bother checking?

peterhuene: That's correct.  The places where we do this the platform supports querying the "frequency of the CPU" based on its product information, not the current frequency of any core / unit.  I have never encountered a platform where SMP is supported with heterogeneous CPUs, anyway.

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            if (proc.smt_enabled) {
```
branan: I'm unsure if `smt_threads` will report as 1 (or even be available) if `smt_enabled==false`. All of our AIX cores have SMT. This seems like the safest option, though.

branan:
OK, apparently I've introduced a behavior change compared to Facter 2 on AIX here.

On Linux, in both Facter 2 and Facter 3, we will report `processorcount` as the total number of hardware *threads* (including any provided by SMT/hyperthreading). On AIX with Facter 2, we reported only the total number of physical cores.

This code matches the linux behavior and reports the total number of hardware threads.

I'm generally in favor of consistency in the information Facter reports across platforms, but in theory Facter 3 is supposed to be Facter-2-compatible. Anyone have strong feelings here?

peterhuene:
Sounds like a bug fix to me :)

branan:
I'll be sure to mention it in the release notes

lib/src/facts/aix/processor_resolver.cc
branan: I have no idea if AIX supports heterogenous processors or not, but it's an easy enough check/warning to leave in just in case. If a user reports it we can figure out how to represent it properly

branan:
I think I'm done tinkering with this now

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -0,0 +1,141 @@
...
+        // devices". These represent the actual processors in the
+        // machine.
+        //
+        // 3. For each custom device, we query its attributes. These
+        // are things like frequency, type, and SMT information.
+
+        if (odm_initialize() < 0) {
+            throw runtime_error((boost::format("failed to initialize ODM: %1%") % odm_error_string()).str());
+        }
+        leatherman::util::scope_exit(odm_terminate);
```
peterhuene: Just caught this: doesn't do what you want.  It declares a default constructed `scope_exit` named `odm_terminate`.  I think you want `leatherman::util::scope_exit cleanup{odm_terminate};`.

branan: Good catch. Shouldn't this trip up some "result of expression not used" warning or lint somewhere? :(

peterhuene: Sadly it has a non-trivial destructor, so likely the compiler couldn't want you about `odm_terminate` not being used.

peterhuene:
:+1: after `scope_exit` declaration is fixed.

branan:
@peterhuene fixed up the scope_exit

PR #1113, opened 2015-08-19
(FACT-1158) Fix expanding commands with spaces in the path.

peterhuene:
On Windows, expanding commands like "git" which are on the path under the
program files directory that contains a space, resulted in the command not
being properly quoted.  The result was a failure to execute the command
because the command shell was not properly given a quoted path.

Two fixes are needed: execution::expand_command should properly quote the
expanded command if it contains a space and Windows execution should
respect an option to "preserve" the arguments as given because the Ruby
execution functions pass the command lines themselves (user's
responsibility to quote arguments).

### DIFF
--- a/lib/src/execution/execution.cc
+++ b/lib/src/execution/execution.cc
@@ -97,10 +97,11 @@ namespace facter { namespace execution {
             return {};
         }
 
-        string quote = result[0] == '"' || result[0] == '\'' ? string(1, result[0]) : "";
+        bool quoted = result[0] == '"' || result[0] == '\'';
+
         string file;
         string remainder;
-        if (!quote.empty()) {
+        if (quoted) {
             // Look for the ending quote for the command
             auto pos = result.find(result[0], 1);
             if (pos == string::npos) {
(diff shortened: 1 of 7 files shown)


peterhuene:
Hold off on merging this as I investigate something.

MikaelSmith:
Test case?

peterhuene:
I'll see what I can do.

peterhuene:
@MikaelSmith Added a test for Windows via the Ruby API (effectively the same use case as the bug).

peterhuene:
Looking into failure

peterhuene:
Looks like what was passing on my machine wasn't passing AppVeyor because I failed to reload the cached search path.  Test fixed.

lib/src/execution/execution.cc
```diff
@@ -124,7 +125,16 @@ namespace facter { namespace execution {
         if (file.empty()) {
             return {};
         }
-        return quote + file + remainder;
+
+        // If originally quoted but the expanded file has a space, quote it
```
branan: I think you mean "originally unquoted"?

peterhuene: I did.  Fixing.

peterhuene:
@branan Fixed.

PR #1114, opened 2015-08-20
(FACT-1169) Strip trailing dots from domain

MikaelSmith:
When we read domain name from /etc/resolv.conf, it can validly contain a
trailing dot in the name. We don't want to report that as part of the
name, so strip it at the end if found.

### DIFF
--- a/lib/src/facts/resolvers/networking_resolver.cc
+++ b/lib/src/facts/resolvers/networking_resolver.cc
@@ -47,6 +47,10 @@ namespace facter { namespace facts { namespace resolvers {
     {
         auto data = collect_data(facts);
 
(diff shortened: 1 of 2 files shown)


kylog:
Should this be done generic to all network resolvers? Or do we believe that it's only when reading from /etc/resolv.conf that this can happen?

MikaelSmith:
resolv.conf is the only place I'm aware of for it to be valid. However Facter 2.x didn't limit it, and we could probably safely strip everywhere.

peterhuene:
It's technically part of any domain name to signify it's "fully-qualified-ness", so I'd expect it could be coming from any source.  I'm fine with stripping it in the base resolver before any use.

MikaelSmith:
Makes it easier to test, so I've done that.

PR #1115, opened 2015-08-20
(DO NOT MERGE) (RFC) Implement a C++ wrapper around the AIX ODM iteration API

branan:
This is a wrapper around the low-level querying and iteration of the AIX ODM API. Further wrappers could be built on top of this to wrap the concepts used in higher-level groups of class (for example, the {Predefined,Custom}{Device,Attribute} tables used in the configuration system)

This attempts to enforce the rules required by the ODM API itself, but unfortunately is not thread-safe, since std::mutex does not appear to be implemented for AIX in GCC 4.8

The examples directory is added just so I had a good place to dump an example of how this is intended to be used.

### DIFF
--- a/lib/inc/internal/util/aix/odm.hpp
+++ b/lib/inc/internal/util/aix/odm.hpp
@@ -1,0 +1,263 @@
+#include <odmi.h>
+
+#include <cstdint>
+#include <memory>
+#include <string>
+
+namespace facter { namespace util { namespace aix {
+
+    // This is a singleton class that handles initialization and
+    // termination of the ODM library. As long as any other class holds a
+    // reference to this one, ODM will remain initialized.
+    //
+    // This class also provides static helpers for interacting with the ODM
+    class odm {
+    public:
+         typedef std::shared_ptr<odm> ptr;
+         static ptr get() {
+             static std::weak_ptr<odm> self;
+
+             auto result = self.lock();
+             if(!result) {
+                 result = ptr(new odm);
+                 self = result;
+             }
+             return result;
+         }
+
+         odm(const odm&) = delete;
+         odm& operator=(const odm&) = delete;
+
+         ~odm() {
+             odm_terminate();
+         }
+
+         // unlike just about every other error code to string function in the
+         // world, odm_err_msg doesn't just return its pointer - it wants a
+         // char** argument to put it in. This wraps that ugly API in a more
+         // strerror()-like interface for ease of use.
+         static const char* error_string() {
+             static char* msg;
+             int result = odm_err_msg(odmerrno, &msg);
+             if (result < 0) {
+                 return "failed to retrieve ODM error message";
+             } else {
+                 return msg;
+             }
+         }
+
+
+
+    private:
+         odm() {
+             if (odm_initialize() < 0) {
+                 throw std::runtime_error(odm::error_string());
+             }
+         }
+    };
+
+    // This templatized class represents an open ODM class (which you
+    // should think of like a table in a traditional database).
+    //
+    // Each class is a singleton, since it is not possible to access a
+    // single ODM class concurrently.
+    //
+    // calling odm_class<T>::get() will actually return a proxy object,
+    // rather than the shared_ptr for the singleton directly. This proxy
+    // object allows using value semantics instead of pointer
+    // semantics. This is mostly for the sake of range-based for loops,
+    // since otherwise a dereference would be necessary.
+    //
+    // The proxy object implements the basic begin/end iterator interface,
+    // as well as a .query() method, which returns a special odm-query
+    // object (tied to the odm_class<T> singleton) which also implements
+    // the begin/end iterator interface.
+    template <typename T>
+    class odm_class : public std::enable_shared_from_this<odm_class<T>> {
+    public:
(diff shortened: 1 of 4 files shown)


branan:
Travis does not like `(rfc)` in a commit message. Whatevs, not like we'd merge it in that state anyway

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+namespace facter { namespace util { namespace aix {
+
+    // This is a singleton class that handles initialization and
+    // termination of the ODM library. As long as any other class holds a
+    // reference to this one, ODM will remain initialized.
+    //
+    // This class also provides static helpers for interacting with the ODM
+    class odm {
+    public:
+         typedef std::shared_ptr<odm> ptr;
```
MikaelSmith: Could use using instead of typedef here.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+         static ptr get() {
+             static std::weak_ptr<odm> self;
```
MikaelSmith: This is kind of fun.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+         }
+
+         odm(const odm&) = delete;
+         odm& operator=(const odm&) = delete;
+
+         ~odm() {
+             odm_terminate();
+         }
+
+        // unlike just about every other error code to string function in the
```
MikaelSmith: Spacing?

branan: I need to redo all the comments are proper doc comments anyway

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+        // world, odm_err_msg doesn't just return its pointer - it wants a
+        // char** argument to put it in. This wraps that ugly API in a more
+        // strerror()-like interface for ease of use.
+         static const char* error_string() {
+             static char* msg;
```
MikaelSmith: So repeated calls to error_string will overwrite the value. I suppose that's fine, if you want to keep a copy put it in a std::string.

lib/inc/internal/util/aix/odm.hpp
branan: This is basically just implementing the lazy-singleton pattern, but on top of a weak ptr so it deallocates when it loses all references (instead of a true static member variable of the type we return). It seemed like the best way to handle "I can only have one of this thing, but I still want RAII-like initialization and cleanup"

MikaelSmith: It's probably a pattern we should use more.

lib/inc/internal/util/aix/odm.hpp
branan: I... actually don't think that needs to be static. We pass a pointer-to-pointer to `odm_err_msg` so are getting back a pointer to something that AIX has allocated (either statically or during library initialization). I'm not sure what inspired me to make that static

lib/inc/internal/util/aix/odm.hpp
branan: It's not thread-safe in this form, because I don't have std::mutex on AIX and didn't want to mess with low-level locking primitives. With a bit of cleanup we could probably turn it into a mixin (ala `enable_shared_from_this`) - something like `class my_thing : public lazy_shared_singleton<my_thing>`

lib/inc/internal/util/aix/odm.hpp
branan: yup. Haven't quite re-trained myself on using vs typedefs since C++11

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+         }
+    };
+
+    // This templatized class represents an open ODM class (which you
+    // should think of like a table in a traditional database).
+    //
+    // Each class is a singleton, since it is not possible to access a
+    // single ODM class concurrently.
+    //
+    // calling odm_class<T>::get() will actually return a proxy object,
```
branan: s/get/open/. I renamed the method without fixing this comment. weeeee

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,263 @@
...
+    // rather than the shared_ptr for the singleton directly. This proxy
+    // object allows using value semantics instead of pointer
```
branan: "semantics" probably isn't the right term here. Maybe just "syntax" is best

MikaelSmith:
:+1: I'm curious if we could make this an abstract pattern we can use with other system calls where we need to iterate over a common resource. It wouldn't be trivial to abstract out.

branan:
This is only a couple hundred lines of code. I suspect the complexity of trying to abstract out all the little differences in how iteration APIs behave would be worse than the small amount of duplication between multiple implementations of the pattern

branan:
This probably also needs a const-correctness pass

PR #1118, opened 2015-08-21
(FACT-1164) Uninitialize Ruby before dll unload

MikaelSmith:
When unloading libruby.dll on Windows loaded by libfacter.dll, OLE server
disconnects seem to hang. This appears to be due to some ordering issue in
how libruby.dll and libfacter.dll are unloaded.

Add a function to explicitly uninitialize Ruby from Facter that is used by
facter.exe to ensure Ruby shutdown occurs before atexit handlers trigger.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -2,6 +2,7 @@
 #include <facter/logging/logging.hpp>
 #include <facter/facts/collection.hpp>
 #include <facter/ruby/ruby.hpp>
+#include <facter/util/scope_exit.hpp>
 #include <boost/algorithm/string.hpp>
 // Note the caveats in nowide::cout/cerr; they're not synchronized with stdio.
 // Thus they can't be relied on to flush before program exit.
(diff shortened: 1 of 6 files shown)


MikaelSmith:
This could be done as a unit test, but appears to only happen on Windows and only when using libfacter.dll. So this is probably best addressed with an acceptance test that connects to a server using win32ole.

exe/facter.cc
```diff
@@ -264,6 +264,10 @@ int main(int argc, char **argv)
         bool show_legacy = vm.count("show-legacy");
         facts.write(boost::nowide::cout, fmt, queries, show_legacy);
         boost::nowide::cout << endl;
+
+        if (ruby) {
```
branan: we need to use `leatherman::util::scope_exit()` when we initialize ruby isntead of doing a conditional cleanup here, to protect against exceptions.

MikaelSmith:
Fixed.

branan:
It's a bummer we had to backport `scope_exit` :/

I guess the plan is just to remove it when we merge up?

MikaelSmith:
It was there, I was looking in the wrong place.

branan:
Ah, word. I should double-check the code before I whine next time

branan:
I'm :+1: on this

peterhuene:
:+1: as well.

PR #1119, opened 2015-08-24
(MAINT) Cache output of df/mount and speed up resolution about 15-times

jasperla:
Are PRs for the ruby facter still accepted? If so, here's one that reduces the resolution time of the `partitions` facter dramatically. On my i5 it goes down from 96ms to about 6ms. There's no need to run `mount` or `df` 3 times per partition if we can just cache the full output up front.

### DIFF
--- a/lib/facter/util/partitions.rb
+++ b/lib/facter/util/partitions.rb
@@ -44,4 +44,16 @@ module Facter::Util::Partitions
   def self.available?
     !self.list.empty?
   end
+
+  def self.flushable?
+    implementation.flushable?
+  end
+
+  def self.flushed?
+    implementation.flushed?
+  end
+
+  def self.flush!
+    implementation.flush!
+  end
 end
--- a/lib/facter/partitions.rb
+++ b/lib/facter/partitions.rb
@@ -36,4 +36,7 @@ Facter.add(:partitions) do
     end
     partitions
   end
+  on_flush do
+    Facter::Util::Partitions.flush! if Facter::Util::Partitions.flushable?
+  end
 end
--- a/lib/facter/util/partitions/linux.rb
+++ b/lib/facter/util/partitions/linux.rb
@@ -22,6 +22,10 @@ module Facter::Util::Partitions
       end
     end
 
+    def self.flushable?
+      false
+    end
+
     def self.uuid(partition)
       uuid = nil
       if File.exist?(DEVDISK_BY_UUID_DIRECTORY)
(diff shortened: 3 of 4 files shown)


kylog:
We do still take PRs for ruby facter for maintenance (not for feature additions).

My main question about this is whether you want to invalidate the cache ever? E.g. if mount points change during the lifetime of a daemonized puppet agent, I think this change would mean you'd miss those changes.

jasperla:
Good point; `#list` would actually be the right place to update the cache then, won't it?

branan:
@jasperla take a look at `lib/facter/util/solaris_zones.rb`. It shows how to handle caching and flushing (look for `do_flush`).

It appears that currently Puppet does not ever ask facter to clear its cache, but we're looking into that now.

jasperla:
@branan Thanks for the `on_flush` pointer. I've re-implemented this PR with it. However I'm not entirely sure how the Linux implementation should be handled. Or if there's a way to implement `flushable?` by checking if the implementation has a `flush!` method.

lib/facter/util/partitions.rb
```diff
@@ -44,4 +44,16 @@ def self.label(partition)
   def self.available?
     !self.list.empty?
   end
+
+  def self.flushable?
+    implementation.flushable?
+  end
+
+  def self.flushed?
```
peterhuene: This doesn't seem to be used.  What is the intention of it?

kylog:
Adding a link to https://tickets.puppetlabs.com/browse/PUP-5125. We'll need this to for puppet's use of facter to see the benefits of this PR.

lib/facter/util/partitions.rb
```diff
@@ -44,4 +44,16 @@ def self.label(partition)
   def self.available?
     !self.list.empty?
   end
+
+  def self.flushable?
```
peterhuene: Rather than having two different methods each implementation needs to implement, I think I'd prefer to see each implementation simply implement a `flush!` method and any implementations that don't need flushing simply do nothing.  It's less churn and harder for some theoretical new implementation (ignoring the fact 2.x is in maintenance mode) to accidentally implement one without the other.  Ideally a "base" implementation would implement the no-op flush, but it isn't worth refactoring this to make that happen.

peterhuene:
Hi @jasperla!  Just following up to see if you had any chance to review the feedback above.  I'm not sure what's going on with AppVeyor for this PR.  It may need to be rekicked.

jasperla:
Hi, I've not yet incorporated the feedback. So no need to kick AppVeyor yet :)

joshcooper:
Hi @jasperla, thank you for your contribution! Just wanted to see if you've had a chance to review feedback?

HAIL9000:
@jasperla, just checking in since it's been a little while. Are you still interested in working on this one? Thanks!

jasperla:
Hi, Let's close this PR. I've been too short on time to properly look after it.

PR #1120, opened 2015-08-24
(FACT-1167) Fix Solaris networking resolver not resolving primary interface.

peterhuene:
The Solaris networking resolver was missing the fallback logic to assign the
primary interface to the first interface that contains a valid address.

This logic existed in the BSD resolver, but it makes more sense to move it into
the base resolver itself so that the fallback logic is shared between all
platforms.


branan:
:+1: This looks sane to me, and the test for it looks reasonable.

We will probably end up having to do something similar for AIX, so I'm glad to see this is already abstracted out.

PR #1121, opened 2015-08-24
(maint) Switch AIX processor resolver over to the new ODM helper classes

branan:
This is a wrapper around the low-level querying and iteration of the AIX ODM API. Further wrappers could be built on top of this to wrap the concepts used in higher-level groups of class (for example, the {Predefined,Custom}{Device,Attribute} tables used in the configuration system). For now the low-level wrappers provide sufficient safety/abstraction for users.

This attempts to enforce the rules required by the ODM API itself, but unfortunately is not thread-safe, since std::mutex does not appear to be implemented for AIX in GCC 4.8

### DIFF
--- a/lib/src/facts/aix/processor_resolver.cc
+++ b/lib/src/facts/aix/processor_resolver.cc
@@ -1,4 +1,5 @@
 #include <internal/facts/aix/processor_resolver.hpp>
+#include <internal/util/aix/odm.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <leatherman/util/scope_exit.hpp>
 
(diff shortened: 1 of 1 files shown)


branan:
Gonna have to make a pass at getting doc comments on more things

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -61,61 +42,40 @@ namespace facter { namespace facts { namespace aix {
...
-            CuDv* processor = static_cast<CuDv*>(odm_get_first(CuDv_class, const_cast<char*>(query.c_str()), nullptr));
-            while (processor) {
-                check_odm_return(processor, "could not fetch processors from CuDv");
-                LOG_DEBUG("got a processor: %1%", processor->name);
-                processor_names.push_back(processor->name);
-                free(processor);
-                processor = static_cast<CuDv*>(odm_get_next(CuDv_class, nullptr));
+            auto cu_dv_query = odm_class<CuDv>::open("CuDv").query(query);
+            for (auto& cu_dv : cu_dv_query) {
+                LOG_DEBUG("Got a processor: %1%", cu_dv.name);
```
MikaelSmith: Should make casing consistent, so lower-case 'g'.

MikaelSmith:
I'm reviewing the helper in #1122

branan:
@MikaelSmith typo fixed and rebased on updated ODM commit from review in the other PR

branan:
Rebased onto master since #1122  is merged now

PR #1122, opened 2015-08-24
(FACT-971) Implement memory resolver for AIX

branan:
For memory, we just use the `vminfo` system call. For swap, we query the ODM for all devices marked as "paging", and use the swapqry system call to get information about them.

This has the same ODM implementation as #1121 (in fact the same commit). I'll be sure to rebase them as appropriate to cover any review of the ODM classes.

### DIFF
--- a/lib/inc/internal/util/aix/odm.hpp
+++ b/lib/inc/internal/util/aix/odm.hpp
@@ -1,0 +1,372 @@
+/**
+ * @file
+ * Declares an iterator interface for examining the AIX object data manager
+ */
+
+#include <odmi.h>
+
+#include <cstdint>
+#include <memory>
+#include <string>
+#include <utility>
+
+namespace facter { namespace util { namespace aix {
+
+    /**
+     * Singleton representing the ODM subsystem.
+     * The ODM subsystem needs to be explicitly initialized and terminated when
+     * it is used. This wraps that initialization/cleanup such that ODM is
+     * initialized when it is first used, and cleaned up when all consumers
+     * have released their references to it.
+     *
+     * some helper methods for interacting with the ODM are also defined here,
+     * since we can't have a `namespace odm` here.
+     */
+    class odm {
+    public:
+         /// shared_ptr to the odm reference
+         using ptr = std::shared_ptr<odm>;
+
+         /**
+          * Grab a reference to the ODM subsystem to keep it open.
+          * If no references currently exist, this will initialize the ODM
+          * subsystem.
+          * @return 
+          */
+         static ptr open() {
+             static std::weak_ptr<odm> self;
+
+             auto result = self.lock();
+             if (!result) {
+                 result = ptr(new odm);
+                 self = result;
+             }
+             return result;
+         }
+
+         /// deleted copy constructor
+         odm(const odm&) = delete;
+
+         /// deleted assignment operator
+         /// @return nothing
+         odm& operator=(const odm&) = delete;
+
+         /**
+          * Clean up the odm library when there are no more users
+          */
+         ~odm() {
+             odm_terminate();
+         }
+
+         /**
+          * Get the error string for an ODM error state.
+          * @return an error string owned by the ODM subsystem
+          */
+         static const char* error_string() {
+             char* msg;
+             int result = odm_err_msg(odmerrno, &msg);
+             if (result < 0) {
+                 return "failed to retrieve ODM error message";
+             } else {
+                 return msg;
+             }
+         }
+
+    private:
+         /**
+          * Initialize the ODM library
+          */
+         odm() {
+             if (odm_initialize() < 0) {
+                 throw std::runtime_error(odm::error_string());
+             }
+         }
+    };
+
+    /**
+     * This represents an ODM class as an iterable thing.  An ODM
+     * class can be throught of as a table in a typical database.
+     * Each templatized version of this has a singleton instance,
+     * which represents a process' handle to that table. Each process
+     * can open a single ODM class only once, and concurrent
+     * operations are not supported.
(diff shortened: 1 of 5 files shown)


lib/src/facts/aix/memory_resolver.cc
```diff
@@ -0,0 +1,47 @@
...
+
+        vminfo info;
+        auto res = vmgetinfo(&info, VMINFO, sizeof(info));
+        if (res < 0) {
+            throw system_error(errno, system_category());
+        }
+        result.mem_total = info.memsizepgs * 4096;
+        result.mem_free = info.numfrb * 4096;
+
+        auto cu_at_query = odm_class<CuAt>::open("CuAt").query("value=paging and attribute=type");
```
MikaelSmith: So `CuAt` comes from `cfgodm.h`, and has a data member named `name`?

lib/src/facts/aix/memory_resolver.cc
```diff
@@ -0,0 +1,47 @@
...
+    memory_resolver::data memory_resolver::collect_data(collection& facts)
+    {
+        data result;
+
+        vminfo info;
+        auto res = vmgetinfo(&info, VMINFO, sizeof(info));
+        if (res < 0) {
+            throw system_error(errno, system_category());
+        }
+        result.mem_total = info.memsizepgs * 4096;
```
MikaelSmith: Are these 4KiB sizes fixed, or could they depend on OS configuration?

branan: page size is fixed at 4K. no good doc link - it's in comments in the various headers :(

lib/src/facts/aix/memory_resolver.cc
branan: Yeah. the odm_class type just wraps an existing ODM struct. cfgodm provides the ODM structs used in the AIX system configuration, including CuAt.

lib/src/facts/aix/memory_resolver.cc
branan: I should add that IBM's documentation points to the comments in the headers as being authoritative, so I trust it's not going to change anytime soon

branan:
@MikaelSmith updated to use the PAGE_SIZE define from sys/limits.h

MikaelSmith:
:+1:

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,367 @@
...
+     * can open a single ODM class only once, and concurrent
+     * operations are not supported.
+     *
+     * @tparam T the struct that is stored in the ODM class.
+     */
+    template <typename T>
+    class odm_class : public std::enable_shared_from_this<odm_class<T>> {
+    public:
+        /// shared_ptr to an odm_class
+        typedef std::shared_ptr<odm_class<T>> ptr;
```
MikaelSmith: Could also switch this to `using ptr = ...`.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,367 @@
...
+                     owner->locked = true;
+                 } else {
+                     // In theory nobody should be constructing us with
+                     // null data and valid owner, but why take the risk?
+                     owner = ptr(nullptr);
+                 }
+             }
+
+        private:
+             T *data;
```
MikaelSmith: These should use `_` prefix/suffix.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,367 @@
...
+        protected:
+             /**
+              * Construct a query_proxy for an odm_class query
+              * @param q the query string
+              * @param o the odm_class that owns this query
+              */
+             query_proxy(std::string q, ptr o) : query(q), owner(o) {}
+
+        private:
+             std::string query;
```
MikaelSmith: Should use `_`.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,367 @@
...
+
+        protected:
+             /**
+              * Construct a proxy for an odm_class
+              * @param self The odm_class that we proxy for
+              */
+             proxy(ptr self) : self(self) {}
+
+        private:
+             ptr self;
```
MikaelSmith: Should use `_`.

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,367 @@
...
+
+    protected:
+        /// The ODM class that we are wrapping
+        CLASS_SYMBOL klass;
+
+        /// whether we are currently locked for iteration
+        bool locked;
+
+    private:
+        odm::ptr the_odm;
```
MikaelSmith: `_`? Also not sure on protected members, probably also `_`.

branan:
@MikaelSmith underscores everywhere.

Alsoadjusted to enable perfect forwarding of the string argument in query creation. We were relying on inlining to avoid several string copies there. I'm pretty sure that'd all get inlined in an optimized build, but it was easy enough to explicitly use `std::forward`

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+    protected:
+        /**
+         * @see proxy::query
+         * @param query the query string
+         * @return a query_proxy representing the provided query
+         */
+        template <typename Arg>
+       query_proxy query(Arg&& query) {
```
MikaelSmith: Indentation.

branan:
Indentation fixed

PR #1126, opened 2015-08-26
(FACT-1010) Fix innacurate uptime on Solaris zones

MikaelSmith:
On Solaris zones - besides the default `global` zone - uptime in Facter
3 was inaccurate compared to Facter 2. This seems to reflect a bug in
Solaris related to querying kstat for uptime.

If not in the `global` zone on Solaris, call `uptime` to get the uptime
rather than using `kstat`.

### DIFF
--- a/lib/inc/internal/facts/bsd/uptime_resolver.hpp
+++ b/lib/inc/internal/facts/bsd/uptime_resolver.hpp
@@ -16,9 +16,10 @@ namespace facter { namespace facts { namespace bsd {
      protected:
         /**
          * Gets the system uptime in seconds.
+         * @param facts The fact collection.
          * @return Returns the system uptime in seconds.
          */
-        virtual int64_t get_uptime() override;
+        virtual int64_t get_uptime(collection& facts) override;
     };
 
 }}}  // namespace facter::facts::bsd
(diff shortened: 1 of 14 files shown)


branan:
Is `kstat` sufficiently faster than `uptime` that it's worth having the conditional?
Alternatively: could we instead use whatever syscall `uptime` itself uses to query that information?

Basically it makes me grumpy that what should be a really simple resolver is so complex.

MikaelSmith:
Probably not sufficiently faster; we could make this much simpler and just use the posix resolver on Solaris.

MikaelSmith:
The implementation of uptime is at http://procps.cvs.sourceforge.net/viewvc/procps/procps/proc/sysinfo.c?view=markup; I'm not sure how portable reading from /proc/uptime is.

[`sysinfo` on Linux](https://github.com/puppetlabs/facter/commit/b8652e7c9c810e8812eea950a88e57c0c574d7f3) has an uptime field, but that doesn't seem supported on Solaris.

MikaelSmith:
I can explore a little more, but calling `uptime` isn't a bad option.

MikaelSmith:
@peterhuene any thoughts on other options?

peterhuene:
I don't think `kstat` was used for performance reasons, only that it was what commonly used in other implementations of facts so it was just used here as well.  I'm completely fine with using `uptime` or just the POSIX resolver if it works.

PR #1128, opened 2015-08-26
(maint) Include a versioned blkid so in the LIBARIES list

haus:
On el4, libblkid does not include a non-versioned so, and so cmake
prefers the .a file, which causes it to not dynamically link against
libuuid. This commit adds libblkid.so.1 to the library search list, so
that on el4 the versioned so can be found.


branan:
I think worst case we end up linked agains the versioned so directly on platforms other than EL4 which... shouldn't be a real problem? But is kind of gross

Given how much of a pain it is to do things based on a specific version fo a specific distro in CMake, this is probably OK :/

MikaelSmith:
Let's see how it goes.

PR #1131, opened 2015-08-27
(FACT-969) Implement networking resolver for AIX

branan:
The AIX networking stack forked from BSD a very long time ago, and
interactions with it are kind of gross.

This is based on what little documentation IBM provides, old BSD
mailing list posts, the source code of 15-year old network daemons,
and a little bit of reverse engineering. It's as close to sane as
we're going to get working with these APIs.

The big grossness with this is that the APIs for the stack are
basically "stick a bunch of structs back to back with size
fields". This requires us to walk the buffer ourselves, parsing out
whatever we find based on the type/size fields.

### DIFF
--- a/lib/src/facts/aix/networking_resolver.cc
+++ b/lib/src/facts/aix/networking_resolver.cc
@@ -1,0 +1,239 @@
+#include <internal/facts/aix/networking_resolver.hpp>
+#include <leatherman/execution/execution.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <boost/algorithm/string.hpp>
+
+#include <unordered_map>
+
+#include <inttypes.h>
+#include <net/if.h>
+#include <net/if_dl.h>
+#include <net/route.h>
+#include <netinet/in.h>
+#include <sys/ndd_var.h>
+#include <sys/kinfo.h>
+
+// This usage is recommended in several mailing lists, and is used at
+// least in Samba to query mac addresses. I saw some references to old
+// IBM publications also recommending its use, but couldn't actually
+// dig those pubs up.
+//
+// The structs it *returns* are in headers, though (specifically
+// mentioning usage of this function).
+//
+// This is all leftovers from ancient versions of the BSD networking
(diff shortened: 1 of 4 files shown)


branan:
I want to do a pass on the solaris, posix, and BSD resolvers to see if there are any other helpers I could be using to process these structures in AIX, but the overall shape of the code is done here.

lib/src/facts/aix/networking_resolver.cc
```diff
@@ -0,0 +1,231 @@
...
+            }
+        }
+
+        data.primary_interface = get_primary_interface();
+
+        return data;
+    }
+
+    networking_resolver::ndd_map networking_resolver::get_ndd_info() const {
+        int ksize = getkerninfo(KINFO_NDD, nullptr, nullptr, 0);
```
MikaelSmith: This information could change between calls. Other places where we deal with this we've tended to loop several times in case the number of devices increased between calls. We should probably do that here.

lib/src/facts/aix/networking_resolver.cc
```diff
@@ -0,0 +1,231 @@
...
+        ndd_map result;
+        for (auto& device : ndd) {
+            result[device.ndd_name] = device;
+            result[device.ndd_alias] = device;
+        }
+        return result;
+    }
+
+    std::vector<networking_resolver::interface> networking_resolver::get_interfaces() const {
+        int ksize = getkerninfo(KINFO_RT_IFLIST, nullptr, nullptr, 0);
```
MikaelSmith: Same comment here about looping to allow for system changes.

MikaelSmith: You could probably use a templated helper to wrap getkerninfo with retries.

MikaelSmith:
Everything else looks reasonable.

MikaelSmith:
Can you retarget this at stable when you update it?

branan:
closing this to retarget at stable

PR #1138, opened 2015-08-31
WIP (FACT-1180) Override Leatherman's logging namespace

MikaelSmith:
When libraries were moved to the Leatherman project, they gained their
own Leatherman-based logging namespaces. We'd rather use Facter's
namespace for all the output.

LTH-29 adds the ability to override logging namespaces used in
Leatherman. Use that to specify Facter's namespace for all logging
messages that used the Leatherman namespace; others, like execution
messages that use special prefixes for stdout/stderr from other
processes, will continue to use those prefixes.


MikaelSmith:
Waiting on https://github.com/puppetlabs/leatherman/pull/81 to be merged so I can update the submodule.

PR #1141, opened 2015-08-31
(FACT-918) Fix up Solaris SPARC virtual facts

MikaelSmith:
Reviewing the implementation here compared to the Facter 2.x
implementation, all the methods attempted on Solaris i386 appear to
apply. So re-enable those methods, as well as using the ldom facts to
identify when in an LDom.

### DIFF
--- a/lib/src/facts/solaris/virtualization_resolver.cc
+++ b/lib/src/facts/solaris/virtualization_resolver.cc
@@ -4,14 +4,12 @@
 #include <facter/facts/collection.hpp>
 #include <facter/facts/fact.hpp>
 #include <leatherman/execution/execution.hpp>
-#include <leatherman/util/regex.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/algorithm/string.hpp>
 #include <map>
 
 using namespace std;
 using namespace facter::facts;
-using namespace leatherman::util;
 using namespace leatherman::execution;
 
 namespace facter { namespace facts { namespace solaris {
@@ -38,28 +36,14 @@ namespace facter { namespace facts { namespace solaris {
 
         string guest_of;
 
-        static map<boost::regex, string> virtual_map = {
-            {boost::regex("Parallels"),  string(vm::parallels)},
-            {boost::regex("VM[wW]are"),  string(vm::vmware)},
-            {boost::regex("VirtualBox"), string(vm::virtualbox)},
-            {boost::regex("HVM domU"),   string(vm::xen_hardware)},
-            {boost::regex("KVM"),        string(vm::kvm)},
-            {boost::regex("oVirt Node"), string(vm::ovirt)}
-        };
-
         // Use the same timeout as in Facter 2.x
         const uint32_t timeout = 20;
         try {
             each_line(
                 "/usr/sbin/prtdiag",
                 [&](string& line) {
-                    for (auto const& it : virtual_map) {
-                        if (re_search(line, it.first)) {
-                            guest_of = it.second;
-                            return false;
-                        }
-                    }
-                    return true;
+                    guest_of = get_product_name_vm(line);
+                    return guest_of.empty();
                 },
                 nullptr,
                 timeout);
--- a/lib/src/facts/solaris/virtualization_resolver.cc
+++ b/lib/src/facts/solaris/virtualization_resolver.cc
@@ -18,6 +18,16 @@ namespace facter { namespace facts { namespace solaris {
 
     string virtualization_resolver::get_hypervisor(collection& facts)
     {
+        // If in an LDom, the LDom facts will resolve and we can use them to identify
+        // that we're in a virtual LDom environment. They should only resolve on SPARC.
+        auto ldom_domainrole_control = facts.get<string_value>("ldom_domainrole_control");
+        if (ldom_domainrole_control && ldom_domainrole_control->value() == "false") {
+            auto ldom_domainrole_impl = facts.get<string_value>("ldom_domainrole_impl");
+            if (ldom_domainrole_impl) {
+                return ldom_domainrole_impl->value();
+            }
+        }
+
         // works for both x86 & sparc.
         bool success;
         string output, none;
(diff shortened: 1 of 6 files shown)


MikaelSmith:
This depends on @whopper's work for validation.

lib/src/facts/solaris/virtualization_resolver.cc
```diff
@@ -18,6 +18,17 @@ namespace facter { namespace facts { namespace solaris {
 
     string virtualization_resolver::get_hypervisor(collection& facts)
     {
+        auto isa = facts.get<string_value>(fact::hardware_isa);
+        if (isa && isa->value() == "sparc") {
+            auto ldom_domainrole_control = facts.get<boolean_value>("ldom_domainrole_control");
```
whopper: Maybe note that "ldom_domainrole_control" and other ldom facts are based on an uncommitted interface and may change suddenly, which would break this logic.

MikaelSmith: That's pretty much the same with all of virtinfo. The most straight-forward way to get this without the `ldom` fact is to call `virtinfo -t` and regex it as below (the plain call to virtinfo doesn't work on Solaris 11 sparc).

MikaelSmith: It looks like ldom_domainrole_control will only be defined on sparc, so I can probably drop the hardware_isa checks.

MikaelSmith:
Looks ready now. Example output
```
bash-3.2# /opt/puppetlabs/bin/facter is_virtual virtual ldom os
is_virtual => true
ldom => {
  domainchassis => "BEL0819I98",
  domaincontrol => "opdx-a3-sun-01",
  domainname => "sol10",
  domainrole => {
    control => "false",
    impl => "LDoms",
    io => "false",
    root => "false",
    service => "false"
  },
  domainuuid => "93470f97-9d25-ead7-c8ef-ffe10ac42a67"
}
os => {
  architecture => "sun4v",
  family => "Solaris",
  hardware => "sun4v",
  name => "Solaris",
  release => {
    full => "10.0",
    major => "10",
    minor => "0"
  }
}
virtual => LDoms
```

MikaelSmith:
And works on Solaris 11 sparc as well
```
root@sol11:~# /opt/puppetlabs/bin/facter ldom os is_virtual virtual
is_virtual => true
ldom => {
  domainchassis => "BEL0819I98",
  domaincontrol => "opdx-a3-sun-01",
  domainname => "sol11",
  domainrole => {
    control => "false",
    impl => "LDoms",
    io => "false",
    root => "false",
    service => "false"
  },
  domainuuid => "61df57d3-a88c-c3fc-e96a-cad66094f2f1"
}
os => {
  architecture => "sun4v",
  family => "Solaris",
  hardware => "sun4v",
  name => "Solaris",
  release => {
    full => "11.2",
    major => "11",
    minor => "2"
  }
}
virtual => LDoms
```

MikaelSmith:
Checked Solaris 10 and 11 x86_64 still report correct virtual facts.

whopper:
Woo! :+1:

lib/src/facts/solaris/virtualization_resolver.cc
```diff
@@ -26,65 +36,37 @@ namespace facter { namespace facts { namespace solaris {
...
-        if (isa->value() == "i386") {
-            static map<boost::regex, string> virtual_map = {
-                {boost::regex("VMware"),     string(vm::vmware)},
-                {boost::regex("VirtualBox"), string(vm::virtualbox)},
-                {boost::regex("Parallels"),  string(vm::parallels)},
-                {boost::regex("KVM"),        string(vm::kvm)},
-                {boost::regex("HVM domU"),   string(vm::xen_hardware)},
-                {boost::regex("oVirt Node"), string(vm::ovirt)}
-            };
+        static map<boost::regex, string> virtual_map = {
```
peterhuene: Any chance we can reuse `facter::facts::resolvers::virtualization_resolver::get_product_name_vm` instead of having this mapping?

peterhuene: Based on the "VMware" regex, perhaps we need a mapping for `VMWare` there?

MikaelSmith: Could be; I was picking up differences that would be difficult to test from Facter 2.x. I'll look into re-using get_product_name_vm.

MikaelSmith: I'll have to make get_product_name_vm use regexes instead of string comparison.

peterhuene: Other than the VMware one, which one requires a regex?  The base implementation is just substring searching.

peterhuene: Although I'd like to see the base implementation take the product name by argument rather than looking into the fact collection for a particular fact :/

MikaelSmith: That's a change I have to make. I could skip using regex if we assume it'll always be VMware. That seems to work in practice.

peterhuene: What I mean is we could just add an entry for `VMWare` in this list so that it matches both without having to use regexes.  I believe that's the only one that had a legitimate use for using regular expressions in the first place.

MikaelSmith: Parallels also had some interesting regex matching at https://github.com/puppetlabs/facter/blob/2.x/lib/facter/virtual.rb#L159, not sure what that was about.

MikaelSmith: I think VMware has always been lower-case 'w', the original code was being overly defensive. Going to drop it.

MikaelSmith:
Building packages to test that Linux and Solaris still behave.

MikaelSmith:
Should be ready, passed in Solaris 10 i386 (the sparc ldom code didn't change) and RedHat 7.

peterhuene:
:+1: waiting on AppVeyor.

MikaelSmith:
AppVeyor's super slow today, but it's running.

PR #1147, opened 2015-09-02
(FACT-969) Implement networking resolver for AIX

branan:
The AIX networking stack forked from BSD a very long time ago, and
interactions with it are kind of gross.

This is based on what little documentation IBM provides, old BSD
mailing list posts, the source code of 15-year old network daemons,
and a little bit of reverse engineering. It's as close to sane as
we're going to get working with these APIs.

The big grossness with this is that the APIs for the stack are
basically "stick a bunch of structs back to back with size
fields". This requires us to walk the buffer ourselves, parsing out
whatever we find based on the type/size fields.

### DIFF
--- a/lib/src/facts/aix/networking_resolver.cc
+++ b/lib/src/facts/aix/networking_resolver.cc
@@ -1,0 +1,239 @@
+#include <internal/facts/aix/networking_resolver.hpp>
+#include <leatherman/execution/execution.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <boost/algorithm/string.hpp>
+
+#include <unordered_map>
+
+#include <inttypes.h>
+#include <net/if.h>
+#include <net/if_dl.h>
+#include <net/route.h>
+#include <netinet/in.h>
+#include <sys/ndd_var.h>
+#include <sys/kinfo.h>
+
+// This usage is recommended in several mailing lists, and is used at
+// least in Samba to query mac addresses. I saw some references to old
+// IBM publications also recommending its use, but couldn't actually
+// dig those pubs up.
+//
+// The structs it *returns* are in headers, though (specifically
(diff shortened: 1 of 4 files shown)


branan:
@MikaelSmith Updated to keep the `getkerninfo` calls and the size logic abstracted.

branan:
It looks like I'm getting bad MTU values from this (everything else looks fine). I'm looking into the MTU values now

lib/src/facts/aix/networking_resolver.cc
```diff
@@ -0,0 +1,239 @@
...
+// gross in a lot of ways, but is what we have to work with on AIX.
+//
+// There does not appear to be a different or documented way to get
+// this information, outside of shelling out to tools which provide
+// human- rather than machine-readable output.
+extern "C" int getkerninfo(int, char*, int*, int32long64_t);
+
+template <typename T>
+static std::vector<T> getkerninfo(int query) {
+    for (;;) {
```
MikaelSmith: We could bound it in case the size is ever increasing, but it'll error with out-of-memory anyway. The similar example I was thinking of is https://github.com/puppetlabs/facter/blob/master/lib/src/facts/windows/networking_resolver.cc#L154-L169.

MikaelSmith:
Doxygen error looks unrelated, probably from the merge. I'll look into it.

lib/src/facts/aix/networking_resolver.cc
branan: If it's ever increasing then the number of network addresses (or interfaces) in the system is also ever-increasing... I'm willing to call that rare enough that waiting for the malloc to fail is fine by me (if the system doesn't die first)

MikaelSmith:
#1149 was merged, you can rebase on it.

PR #1150, opened 2015-09-02
(FACT-1221) Add fallback for resolving total and free memory on SPARC

MikaelSmith:
On Solaris SPARC in our LDoms, kstat_read fails similar to FACT-909. Add a fallback to use when it fails, where we call kstat directly.

### DIFF
--- a/lib/src/facts/solaris/memory_resolver.cc
+++ b/lib/src/facts/solaris/memory_resolver.cc
@@ -1,8 +1,8 @@
 #include <internal/facts/solaris/memory_resolver.hpp>
 #include <internal/util/solaris/k_stat.hpp>
-#include <leatherman/file_util/file.hpp>
 #include <leatherman/execution/execution.hpp>
 #include <leatherman/logging/logging.hpp>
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This should be ready to go.

MikaelSmith:
ping @whopper

whopper:
:+1: going to build this on top of stable as a final pass in SPARC. I'll merge once I'm able to run it.

whopper:
:+1: Tested on Sol 10/11 x86 and SPARC - works everywhere.

```
{
  swap => {
    available => "2.00 GiB",
    available_bytes => 2147475456,
    capacity => "0%",
    total => "2.00 GiB",
    total_bytes => 2147475456,
    used => "0 bytes",
    used_bytes => 0
  },
  system => {
    available => "1.95 GiB",
    available_bytes => 2088837120,
    capacity => "65.67%",
    total => "5.67 GiB",
    total_bytes => 6084788224,
    used => "3.72 GiB",
    used_bytes => 3995951104
  }
}

```

PR #1151, opened 2015-09-03
(FACT-1222) Update OS release regex for SPARC Solaris 10

whopper:
SPARC and X86 Solaris systems contain slightly different
release strings in /etc/release. They differ by a single
character, where SPARC uses the letter 's' and X86 uses
the letter 'x'. This commit relaxes the regex used to grab
this string so as to be compatible with both architectures.


whopper:
Actually, retargeting this at stable.

PR #1152, opened 2015-09-03
(FACT-1222) Update OS release regex for SPARC Solaris 10

whopper:
SPARC and X86 Solaris systems contain slightly different
release strings in /etc/release. They differ by a single
character, where SPARC uses the letter 's' and X86 uses
the letter 'x'. This commit relaxes the regex used to grab
this string so as to be compatible with both architectures.

### DIFF
--- a/lib/src/facts/solaris/operating_system_resolver.cc
+++ b/lib/src/facts/solaris/operating_system_resolver.cc
@@ -54,7 +54,7 @@ namespace facter { namespace facts { namespace solaris {
          to be resolved further using the `pkg info kernel` command (TODO).
          */
 
-        static boost::regex regexp_s10("Solaris \\d+ \\d+/\\d+ s(\\d+)x_u(\\d+)wos_");
+        static boost::regex regexp_s10("Solaris \\d+ \\d+/\\d+ s(\\d+)[sx]?_u(\\d+)wos_");
         static boost::regex regexp_s11("Solaris (\\d+)[.](\\d+)");
         static boost::regex regexp_s11b("Solaris (\\d+) ");
         lth_file::each_line("/etc/release", [&](string& line) {


whopper:
Tested in Sol10 SPARC and x86:

Sol 10 SPARC:

```
-bash-3.2# /opt/puppetlabs/puppet/bin/facter os
{
  architecture => "sun4v",
  family => "Solaris",
  hardware => "sun4v",
  name => "Solaris",
  release => {
    full => "10_u10",
    major => "10",
    minor => "10"
  }
}
```

sol 10 x86:

```
-bash-3.2# /opt/puppetlabs/puppet/bin/facter os
-bash-3.2# /opt/puppetlabs/puppet/bin/facter os
{
  architecture => "i86pc",
  family => "Solaris",
  hardware => "i86pc",
  name => "Solaris",
  release => {
    full => "10_u11",
    major => "10",
    minor => "11"
  }
}

```

MikaelSmith:
What's it actually matching in /etc/release?

lib/src/facts/solaris/operating_system_resolver.cc
```diff
@@ -54,7 +54,7 @@ namespace facter { namespace facts { namespace solaris {
          to be resolved further using the `pkg info kernel` command (TODO).
          */
 
-        static boost::regex regexp_s10("Solaris \\d+ \\d+/\\d+ s(\\d+)x_u(\\d+)wos_");
+        static boost::regex regexp_s10("Solaris \\d+ \\d+/\\d+ s(\\d+)\\S_u(\\d+)wos_");
```
MikaelSmith: I lean towards being slightly more restrictive and following Facter 2's example with `[sx]` rather than `\\S`.

whopper:
The lines we're matching are:

x86:
    Oracle Solaris 10 1/13 s10x_u11wos_24a X86

SPARC:
    Oracle Solaris 10 8/11 s10s_u10wos_17b SPARC

Notice the 'x' and 's' which differ between the two.

PR #1154, opened 2015-09-03
(FACT-1185) Add Solaris 10/11 SPARC node defs

johnduarte:
This commit adds node definitions for solaris sparc 10 and 11
agents using the SPARC architecture.

### DIFF
--- a/acceptance/config/nodes/solaris-10-sparc.yaml
+++ b/acceptance/config/nodes/solaris-10-sparc.yaml
@@ -1,0 +1,6 @@
+---
+HOSTS:
+  solaris-10-sparc:
+    roles:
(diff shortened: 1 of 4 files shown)


MikaelSmith:
:+1: we'll want to add these to Puppet and Hiera as well.

joshcooper:
I noticed these files don't start with `---`, should theyl? Also they don't have the `.yaml` extension like we do with other files.

MikaelSmith:
I guess others start with `---`, might be nice to add it to the existing Solaris files as well.

johnduarte:
Solaris yaml files updated per comments from @joshcooper and @MikaelSmith

PR #1155, opened 2015-09-03
(FACT-969) Use netstat to query MTUs on AIX

branan:
The ndd info for MTUs is just plain wrong. Instead, we just shell out
to netstat. Netstat is fast enough, and returns accurate information
for all devices.


MikaelSmith:
:+1:

PR #1156, opened 2015-09-04
(maint) Enable uptime resolver on AIX

branan:
AIX can just use the normal Posix uptime resolver - we just never got
around to turning it on.


kylog:
:+1:

PR #1163, opened 2015-09-10
(FACT-1230) Disable attempting to use locales on AIX

branan:
Much as in solaris, the libstdc++ support for locales on AIX is... bad. We shouldn't even bother.


MikaelSmith:
:+1:

PR #1165, opened 2015-09-11
(maint) Add acceptance pre-suite to rm puppet on sparc

johnduarte:
This commit adds a beaker pre-suite to ensure that puppet packages
and components are removed when the platform is on sparc and the
hypervisor is none.

This ensures that the sparc testing platform does not have any
remaining puppet bits when it is not being managed by a provisioning
hypervisor.

### DIFF
--- a/acceptance/config/aio/options.rb
+++ b/acceptance/config/aio/options.rb
@@ -1,6 +1,7 @@
 {
   :type => 'aio',
   :pre_suite => [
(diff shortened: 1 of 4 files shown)


MikaelSmith:
:+1:

MikaelSmith:
We should just kill packages. I guess git is still useful.

PR #1168, opened 2015-09-11
(FACT-549) Implement serial number resolution for AIX

branan:
On other platforms, the serial number is queried via DMI. There is no DMI on AIX, so we have a custom resolver here to get that information.

### DIFF
--- a/lib/src/facts/aix/serial_number_resolver.cc
+++ b/lib/src/facts/aix/serial_number_resolver.cc
@@ -1,0 +1,43 @@
+#include <internal/facts/aix/serial_number_resolver.hpp>
+#include <internal/util/aix/odm.hpp>
+#include <facter/facts/collection.hpp>
+#include <facter/facts/fact.hpp>
+#include <facter/facts/scalar_value.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <leatherman/util/regex.hpp>
+#include <boost/regex.hpp>
+
+#include <sys/cfgodm.h>
+
+using namespace std;
+using namespace facter::util::aix;
+using namespace leatherman::util;
+
+namespace facter { namespace facts { namespace aix {
+
+    serial_number_resolver::serial_number_resolver() :
+        resolver(
+            "AIX serial number",
+            {
+                fact::serial_number,
+            })
+    {
+    }
+
+    void serial_number_resolver::resolve(collection& facts) {
+        auto cu_at_query = odm_class<CuAt>::open("CuAt").query("name=sys0 and attribute=systemid");
+        auto result = *cu_at_query.begin();
+
+        // the ODM returns a string of the form "IBM,XXSERIAL#". We
+        // need to strip the "IBM,XX" from the start ofthis, and keep
+        // only the 7 actual serial number bytes.
+        const auto regex = boost::regex("^IBM,\\d\\d(.+)");
+        string serial;
+        if (re_search(string(result.value), regex, &serial)) {
+             facts.add(fact::serial_number, make_value<string_value>(move(serial)));
+        } else {
+            LOG_DEBUG("Could not retrieve serial number: sys0 systemid did not match the expected format");
+        }
+    }
+
+}}}  // namespace facter::facts::aix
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Docs and lint need fixing up.

branan:
Woohoo, more boiler-plate docs :crying_cat_face: 

I get the need to have public interfaces documented, but for these internal "we need a subclass" files writing the boilerplate into every header isn't super helpful. Is it maybe worth evaluating which paths we run Doxygen on so we can get documentation warnings where it matters and not have a bunch of not-actually-helpful comments strewn about?

For now I'll fix this up to meet our current Doxygen rules.

MikaelSmith:
We could consider that, we could have a lower requirement for internal functions that doesn't go through doxygen. /cc @peterhuene

lib/src/facts/aix/serial_number_resolver.cc
```diff
@@ -0,0 +1,34 @@
...
+    {
+    }
+
+    void serial_number_resolver::resolve(collection& facts) {
+        auto cu_at_query = odm_class<CuAt>::open("CuAt").query("name=sys0 and attribute=systemid");
+        auto result = *cu_at_query.begin();
+
+        // the ODM returns a string of the form "IBM,XXSERIAL#". We
+        // need to strip the "IBM,XX" from the start ofthis, and keep
+        // only the 7 actual serial number bytes.
```
MikaelSmith: Could that number ever change?

lib/src/facts/aix/serial_number_resolver.cc
```diff
@@ -0,0 +1,34 @@
...
+using namespace facter::util::aix;
+
+namespace facter { namespace facts { namespace aix {
+
+    serial_number_resolver::serial_number_resolver() :
+        resolver(
+            "AIX serial number",
+            {
+                fact::serial_number,
+             })
```
peterhuene: Nitpick: extra space on this line.

lib/src/facts/aix/serial_number_resolver.cc
peterhuene: Agreed, I'd prefer us not to crash if the returned value is not what is expected.

MikaelSmith: A regex to match a number seems safer here.

branan: This was borrowed from an IBM whitepaper, but I have no idea what they consider "official/standard" vs. just information. I can update this to a regex.

lib/src/facts/aix/serial_number_resolver.cc
```diff
@@ -0,0 +1,39 @@
...
+    void serial_number_resolver::resolve(collection& facts) {
+        auto cu_at_query = odm_class<CuAt>::open("CuAt").query("name=sys0 and attribute=systemid");
+        auto result = *cu_at_query.begin();
+
+        // the ODM returns a string of the form "IBM,XXSERIAL#". We
+        // need to strip the "IBM,XX" from the start ofthis, and keep
+        // only the 7 actual serial number bytes.
+        const auto regex = boost::regex("^IBM,\\d\\d(.+)");
+        string serial;
+        re_search(string(result.value), regex, &serial);
```
MikaelSmith: It'd be nice to have a debug message if regex matching fails.

lib/src/facts/aix/serial_number_resolver.cc
```diff
@@ -0,0 +1,43 @@
...
+
+        // the ODM returns a string of the form "IBM,XXSERIAL#". We
+        // need to strip the "IBM,XX" from the start ofthis, and keep
+        // only the 7 actual serial number bytes.
+        const auto regex = boost::regex("^IBM,\\d\\d(.+)");
+        string serial;
+        if (re_search(string(result.value), regex, &serial)) {
+             facts.add(fact::serial_number, make_value<string_value>(move(serial)));
+        } else {
+            LOG_WARNING("Could not retreive serial number: sys0 systemid did not match the expected format");
```
MikaelSmith: typo, should be [retrieve](https://www.google.com/search?q=retrieve). Also this would tend to be LOG_DEBUG, as we tend not to warn if we can't resolve facts.

branan:
@MikaelSmith typo fixed, log level changed

PR #1169, opened 2015-09-14
(FACT-1226) Add facts-based acceptance test for AIX

whopper:
[skip-ci]


whopper:
Note that this will legitimately fail in AIX 5.3 on builds that don't contain the fix in https://tickets.puppetlabs.com/browse/FACT-1232.

MikaelSmith:
FACT-1232 was merged. Have you verified this now works with a new build?

whopper:
@MikaelSmith we'll have to wait for new 5.3 builds which will take a while. I put this up after seeing the output from Branan's change - this should now pass, and if it doesn't then something else is still broken.

PR #1172, opened 2015-09-14
(FACT-1235) Fix output of negative numbers returned by custom facts.

peterhuene:
rb_num2ulong was being called when outputting a Ruby value in JSON, YAML, or to
the console.  This resulted in an underflow when the value was negative and the
value was displayed as a large positive value.

The fix is to instead call rb_num2long which returns a signed version.


MikaelSmith:
:+1:

PR #1174, opened 2015-09-16
(maint) Bump for LTH-48 change to execute

MikaelSmith:
LTH-48 changed the return value of execution::execute, bump the
submodule and fix use of execute.


MikaelSmith:
LTH-49 doesn't introduce any changes to the existing interface, so that won't require any code changes, just another module bump.

PR #1175, opened 2015-09-18
(maint) Execute returns a struct instead of a tuple

joshcooper:
Update zone resolver to expect a struct to match commit e1516f1b9b
and LTH-48.

### DIFF
--- a/lib/src/facts/solaris/zone_resolver.cc
+++ b/lib/src/facts/solaris/zone_resolver.cc
@@ -14,7 +14,10 @@ namespace facter { namespace facts { namespace solaris {
     zone_resolver::data zone_resolver::collect_data(collection& facts)
     {
         data result;
-        result.current_zone_name = get<1>(execute("/bin/zonename"));
+        auto exec = execute("/bin/zonename");
+        if (exec.success) {
+          result.current_zone_name = exec.output;
+        }
 
         static boost::regex zone_pattern("(\\d+):([^:]*):([^:]*):([^:]*):([^:]*):([^:]*):([^:]*)");
 


lib/src/facts/solaris/zone_resolver.cc
```diff
@@ -14,7 +14,10 @@ namespace facter { namespace facts { namespace solaris {
     zone_resolver::data zone_resolver::collect_data(collection& facts)
     {
         data result;
-        result.current_zone_name = get<1>(execute("/bin/zonename"));
```
joshcooper: I think we previously always set `current_zone_name` even if the command failed, though it seems other resolvers check `success`, which makes sense, so I did the same. Not sure if we should be failing instead?

joshcooper:
Should address failure in https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Master/view/Puppet%20Agent%20Daily/job/platform_aio-suite_pkg-build_master/BUILD_TARGET=solaris-10-i386,SLAVE_LABEL=beaker/16/console

lib/src/facts/solaris/zone_resolver.cc
kylog: This seems cleaner to me (and more consistent with the behavior on failure for the zones fact immediately below).

So I'm :+1:

joshcooper:
I ended up building puppet-agent on solaris and testing it works:

```
solaris_zones => {
  current => "global",
  zones => {
    global => {
      brand => "native",
      id => "0",
      ip_type => "shared",
      path => "/",
      status => "running"
    }
  }
}
```

PR #1177, opened 2015-09-25
(PUP-5001) Remove Ubuntu 14.04 acceptance node defs

johnduarte:
This commit removes the Ubuntu 14.04 acceptance test node
definitions.

Ubuntu 14.04 (Utopic) reached EOL on 2015-07-23 and is no longer
needed as a testing platform.

Ref: https://lists.ubuntu.com/archives/ubuntu-announce/2015-July/000197.html


whopper:
Hah, travis doesn't like that this is a PUP ticket instead of a FACT ticket. Do we care enough to have a facter ticket? @kylog?

PR #1180, opened 2015-09-27
(FACT-1243) Add Scientific/Oracle to el acceptance

johnduarte:
This commit adds Scientific and Oracle to the accepted pattern
of os.name values for the `acceptance/test/facts/el.rb` test.

Prior to this commit, this test would fail On Scientific Linux
and Oracle Linux due to the fact that the `os.name` value did
not match the available values.

### DIFF
--- a/acceptance/tests/facts/el.rb
+++ b/acceptance/tests/facts/el.rb
@@ -40,7 +40,7 @@ agents.each do |agent|
                   'os.architecture'         => os_arch,
                   'os.family'               => 'RedHat',
                   'os.hardware'             => os_hardware,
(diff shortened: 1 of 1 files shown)


kylog:
That makes sense. Do we want that in stable?

whopper:
@johnduarte doh, I must have missed your PR when I put up https://github.com/puppetlabs/facter/pull/1181. However, I did find that the test will also fail in Oracle 7 because the kernel version is different than all the other EL's. See that PR for details.

joshcooper:
Does PR #1181 supercede this one? can we close this?

PR #1181, opened 2015-09-28
(FACT-1243) Update EL fact test to support Oracle and Scientific Linux

whopper:
In addition, make the check of the os.name fact more explicit to
prevent the test from erroneously passing in the case of an EL
believing itself to be 'RedHat', which is the osfamily.


joshcooper:
Tested on redhat-7, centos-4, oracle-7, scientific-7, :+1:

PR #1182, opened 2015-09-29
WIP - DO NOT MERGE (FACT-1186) Update sparc pre-suite for solaris 11

johnduarte:
This commit adds a package removal command for solaris 11, which
is not the same command as solaris 10.


joshcooper:
Can this be handled in the jenkins job instead? I'd hate to have to do this in facter, hiera, puppet, ...

johnduarte:
@joshcooper that would be AWESOME.
@justinstoller do you have suggestions how to wire this up in Jenkins (ala AIX LPAR reset)? I can take your response offline.

justinstoller:
My understanding (per discussion in RE-5506) was that preparing non pooler platforms for testing and packaging was being abstracted away from the CI system via the pre-suite and vanagon.

I don't think anyone wants a bunch of duplication, whether it is having to duplicate the code between puppet, facter, and hiera pre-suites by the Client team or a bunch of jenkins jobs by the QE team.

This seems like a great task for a shared library of helpers that gets called. That way there is a single place to update and a small surface area that simply calls those helpers in each project/job.

I would recommend, if these steps require running Beaker, to put this into a Beaker DSL helper library (akin to what @kevpl has create at https://github.com/puppetlabs/beaker-template and has worked with @er0ck to pull out https://github.com/puppetlabs/beaker-hiera/). The pre-suite could call a generic `prepare_host()` type method and pass in the host in particular, and then the helper could contain the logic for doing this kind of thing.

This approach would allow to start with managing SPARC clean up, but could become a useful library in to reduce the shared setup code between the client projects.

An alternative would be to create a rake task akin to `rake ci:test_against_preserved_hosts`, which could be ran prior to the test rake task and could be packaged into a similar gem without the Beaker library template.

kevpl:
Since this is about uninstalling a package from a host, I think that Beaker's host abstraction would be the correct place to put general package removal details.

To @justinstoller's idea, it seems that if we put this into [beaker-facter](https://github.com/er0ck/beaker-facter) (in progress, woot @er0ck!), we would have to copy it into other Beaker libraries as well, just like @joshcooper was referring to above.

My opinion is that the work here should be done just to find the name of the package, and then `host.uninstall_package` should be used.  It looks like this does have support for [solaris 10 & 11](https://github.com/puppetlabs/beaker/blob/master/lib/beaker/host/unix/pkg.rb#L135-L138).  Is that not working in your testing, or has this avenue not come up in testing this so far?

Let me know if you have any questions about this, or reasons that this won't be a good approach, and hopefully we'll get it settled down and in soon.

justinstoller:
Just to clarify, my idea wasn't to move this to a per-component helper library (I was just using those as an example of what a DSL helper library is). I was thinking more of a helper library that would eventually be a place where we could split out the `foss_utils` DSL code in Beaker, combine it with the copy-pasta'd `acceptance/lib/puppet/acceptance/install_utils.rb` in each component, and put things like this in there, so each component can just call an entry method, and as the processes to install puppet-agent components evolve those changes only have to go to one place.

That's not a requirement for this work, but a direction I think things are going in. I was just trying to say that we could begin heading in that direction with this. It does seem like utilizing the Beaker `Host` abstraction is probably the simplest solution to this problem.

johnduarte:
On Wed, Sep 30, 2015 at 7:41 AM, Justin Stoller <notifications@github.com>
wrote:

> My understanding (per discussion in RE-5506) was that preparing non pooler
> platforms for testing and packaging was being abstracted away from the CI
> system via the pre-suite and vanagon.
>

My apologies if I misunderstood or misrepresented the item of platform
preparation.

My understanding was that Beaker, and therefore, the CI testing jobs would
not be concerned with with preparation. We are realizing this by using the
"none" hypervisor. However, I was under the impression that a Jenkins job
for preparing the non pooler platforms would still be required as part of
the process. For example, we have an AIX LPAR reset job. I presumed that an
analogous job for SPARC would be needed when OPS finished creating zones
that could be rolled back. (https://tickets.puppetlabs.com/browse/OPS-6972).

I am not assuming that the pipeline is responsible for including the
'preparation' jobs. This would be the responsibility of the tester to run
these as needed prior to executing the manual pipeline.

In other words, the workflow would look like.
1. Tester kicks off 'preparation' job for non pooler platform. (e.g.
http://jenkins-compose.delivery.puppetlabs.net/view/Build%20AIX/job/AIX_Revert_LPAR(all_HOSTS)/
)
2. After successful completion of the 'preparation' job, the tester kicks
off the AIO manual pipeline for that platform. (e.g.
https://jenkins.puppetlabs.com/view/All%20in%20One%20Agent/view/Manual/view/Puppet%20Agent%20Manual%20(stable)/
)


> I don't think anyone wants a bunch of duplication, whether it is having to
> duplicate the code between puppet, facter, and hiera pre-suites by the
> Client team or a bunch of jenkins jobs by the QE team.
>
> This seems like a great task for a shared library of helpers that gets
> called. That way there is a single place to update and a small surface area
> that simply calls those helpers in each project/job.
>
> I would recommend, if these steps require running Beaker, to put this into
> a Beaker DSL helper library (akin to what @kevpl
> <https://github.com/kevpl> has create at
> https://github.com/puppetlabs/beaker-template and has worked with @er0ck
> <https://github.com/er0ck> to pull out
> https://github.com/puppetlabs/beaker-hiera/). The pre-suite could call a
> generic prepare_host() type method and pass in the host in particular,
> and then the helper could contain the logic for doing this kind of thing.
>
> This approach would allow to start with managing SPARC clean up, but could
> become a useful library in to reduce the shared setup code between the
> client projects.
>
This clean-up step is really a stop-gap "preparation" step because we do
not yet have SPARC instances that we provision or roll back. So, I am
expecting it to be deleted within the next four weeks. However, even when
it is no longer necessary, we will still need a way to provision the SPARC
boxes. It seems like this should live outside of both the Beaker and
puppetlabs project code bases since it will be specific to our
infrastructure.

johnduarte:
@joshcooper @justinstoller I will replace the changes here with a call to a new beaker helper to address this in a more agnostic manner. I will amend this PR with the updates.

joshcooper:
@johnduarte Can you bump the minimum beaker dependency in the acceptance Gemfile (to a beaker version that supports the new method)?

PR #1184, opened 2015-10-02
(FACT-1245) Update tests for AIX

johnduarte:
This commit updates the `facts/ruby.rb` and
`ticket_1123_facter_with_invalid_locale.rb` tests to include
specific criteria for the AIX platform.


kylog:
LGTM. @johnduarte what did you sort out about the `verify_facts.rb` test? I wonder if we should tweak that test to report *which* fact was supposed to be hidden?

johnduarte:
@kylog the `verify_facts.rb` test failure looks legit. Opened FACT-1248 for the `serialnumber` failure.

I agree with your suggestion to add the name of the fact to the error. I will shoehorn that change in this PR.

johnduarte:
@kylog you were too fast for me. `verify_facts.rb` update submitted here: https://github.com/puppetlabs/facter/pull/1185

PR #1186, opened 2015-10-08
(FACT-1224) Add beaker node defs for Win10(x86,x64)

johnduarte:
[skip-ci]

### DIFF
--- a/acceptance/config/nodes/win10-rubyx64.yaml
+++ b/acceptance/config/nodes/win10-rubyx64.yaml
@@ -1,0 +1,14 @@
+---
+HOSTS:
+  agent-10-x86_64-rubyx64:
+    roles:
+    - agent
+    platform: windows-10-64
+    ruby_arch: x64
+    hypervisor: vcloud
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This only adds configs for 64-bit Windows 10 (for 32-bit and 64-bit Ruby installs). Will we also do 32-bit Windows 10?

johnduarte:
@MikaelSmith  the only 32bit Windows we currently support is 2003. So, I do not believe we intend to add 32bit Windows 10.

joshcooper:
@johnduarte @MikaelSmith I think it's okay to skip 32-bit Windows 10, but I wanted to clarify some comments.

We have supported 32-bit Windows OSes for a number of releases (Windows 7, 8, 8.1), but we don't actually test them in CI, at least not in FOSS. The only 32-bit server windows platforms are 2003 and 2008 (non-R2), and we only test 2003. 2008R2 is 64-bit only and we do test that.

/cc @Iristyle

PR #1187, opened 2015-10-08
(FACT-1224) Add beaker node defs for Win10(x86,x64)

johnduarte:
[skip-ci]


MikaelSmith:
This will be merged down from stable when https://github.com/puppetlabs/facter/pull/1186 is merged.

PR #1189, opened 2015-10-12
(FACT-1238) Respect fully-qualified hostnames if specified.

peterhuene:
Facter 2.x would respect any hostname that was already fully-qualified.  A
hostname of 'foo.bar.com' would be split into a hostname of 'foo' and a domain
of 'bar.com'. No further attempt to resolve the FQDN would take place.

Facter 3 always used fallback logic of attempting to resolve the hostname to a
FQDN and looking through `resolv.conf` for domains.  This should only occur if
the system hostname is not already fully-qualified.

This is to say, if the users have configured their machine hostnames to be
fully-qualified, Facter should respect that.

### DIFF
--- a/lib/src/facts/posix/networking_resolver.cc
+++ b/lib/src/facts/posix/networking_resolver.cc
@@ -61,21 +61,33 @@ namespace facter { namespace facts { namespace posix {
     {
         data result;
 
-        // Get the hostname
+        // Get the maximum size of the host name
         int size = sysconf(_SC_HOST_NAME_MAX);
-        vector<char> name(size == -1 ? 1024 : size);
-        if (gethostname(name.data(), name.size()) != 0) {
+        if (size <= 0) {
+            size = 1024;
+        }
+        // Get the hostname (+1 to ensure a null is returned on platforms where maximum truncation may occur)
+        vector<char> name(size + 1);
+        if (gethostname(name.data(), size) != 0) {
             LOG_WARNING("gethostname failed: %1% (%2%): hostname is unavailable.", strerror(errno), errno);
         } else {
-            // Use everything up to the first period
-            result.hostname = name.data();
-            auto pos = result.hostname.find('.');
-            if (pos != string::npos) {
-                result.hostname = result.hostname.substr(0, pos);
+            // Check for fully-qualified hostname
+            auto it = find(name.begin(), name.end(), '.');
+            if (it != name.end()) {
+                LOG_DEBUG("using the FQDN returned by gethostname: %1%.", name.data());
+                result.hostname.assign(name.begin(), it);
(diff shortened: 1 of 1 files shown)


lib/src/facts/posix/networking_resolver.cc
```diff
@@ -106,7 +110,7 @@ namespace facter { namespace facts { namespace posix {
                     result.domain.assign(parts[1].begin(), parts[1].end());
                     return false;
                 }
-                if (parts[0] == boost::as_literal("search")) {
+                if (search.empty() && parts[0] == boost::as_literal("search")) {
```
MikaelSmith: Did this fix a different problem?

lib/src/facts/posix/networking_resolver.cc
```diff
@@ -67,15 +67,19 @@ namespace facter { namespace facts { namespace posix {
         if (gethostname(name.data(), name.size()) != 0) {
             LOG_WARNING("gethostname failed: %1% (%2%): hostname is unavailable.", strerror(errno), errno);
         } else {
-            // Use everything up to the first period
-            result.hostname = name.data();
-            auto pos = result.hostname.find('.');
-            if (pos != string::npos) {
-                result.hostname = result.hostname.substr(0, pos);
+            // Check for fully-qualified hostname
```
MikaelSmith: Maps to approximately https://github.com/puppetlabs/facter/blob/2.x/lib/facter/domain.rb#L45.

lib/src/facts/posix/networking_resolver.cc
peterhuene: Noticed a discrepancy between the comment and the code: it was respecting the last `search` directive and not the first.  Can split into a separate commit if desired.

lib/src/facts/posix/networking_resolver.cc
```diff
@@ -67,15 +67,19 @@ namespace facter { namespace facts { namespace posix {
...
+            auto it = find(name.begin(), name.end(), '.');
+            if (it != name.end()) {
+                LOG_DEBUG("using the FQDN returned by gethostname: %1%.", name.data());
+                result.hostname.assign(name.begin(), it);
+                result.domain.assign(++it, name.end());
```
peterhuene: Actually, this is going to insert extra nulls into `domain` (prints ok, but the string shouldn't have nulls in it). Let me push up a fix.

lib/src/facts/posix/networking_resolver.cc
peterhuene: Since I have to push up an amended commit anyway, let me pull this out into its own commit.

peterhuene:
@MikaelSmith Pushed up changes.

PR #1191, opened 2015-10-12
(FACT-780) Prevent Puppet from altering Facter logging with -p.

peterhuene:
When the `-p` option is given to Facter, Facter loads and initializes Puppet.
However, part of the Puppet initialization is to alter Facter's logging
behavior to better integrate it with Puppet's output.

When Puppet is loaded by Facter, this is undesirable behavior as by default
this will cause no further output from Facter other than printing the facts.

To fix this, this change only defines the methods that Puppet consumes to alter
Facter's logging behavior if the `-p` option was not specified.  Puppet already
checks for these methods being present before consuming them.

### DIFF
--- a/lib/inc/internal/ruby/module.hpp
+++ b/lib/inc/internal/ruby/module.hpp
@@ -33,8 +33,9 @@ namespace facter { namespace ruby {
          * Constructs the Ruby Facter module.
          * @param facts The collection of facts to populate.
          * @param paths The search paths for loading custom facts.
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Should this be FACT-780?

MikaelSmith:
Otherwise :+1:

peterhuene:
Doesn't really matter.  Let's leave it as FACT-1227 and I'll resolve the other one as fixed as well.

peterhuene:
Actually, I'll amend the commit message.

PR #1193, opened 2015-10-14
(maint) Bump to latest Leatherman version

MikaelSmith:
Fixes builds on Mac OS X with Xcode 7.


MikaelSmith:
Should we try to do this on stable instead? I don't think any significant Leatherman changes have happened that would cause a problem. /cc @peterhuene

peterhuene:
Stable works.

PR #1195, opened 2015-10-14
(FACT-1248) Mark serialnumber as hidden on AIX

branan:
AIX does not use the DMI resolver, and instead has its own
implementation of the serial number fact. That fact is supposed to be
hidden, and since AIX has its own resolver we need to explicitly hide
the value in AIX

### DIFF
--- a/lib/src/facts/aix/serial_number_resolver.cc
+++ b/lib/src/facts/aix/serial_number_resolver.cc
@@ -34,7 +34,7 @@ namespace facter { namespace facts { namespace aix {
         const auto regex = boost::regex("^IBM,\\d\\d(.+)");
         string serial;
         if (re_search(string(result.value), regex, &serial)) {
-             facts.add(fact::serial_number, make_value<string_value>(move(serial)));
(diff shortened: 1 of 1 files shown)


peterhuene:
:+1:

MikaelSmith:
Does hiding it on AIX make sense? i.e. does this have little value outside backwards compatibility?

branan:
@MikaelSmith I'm actually not sure. We hide it on other platforms presumably because it's part of the DMI structured fact, but on AIX we don't have DMI so it *only* exists as a top-level fact. Possibly we actually want to have a proper "system information" resolver for AIX that can pull out similar stuff to DMI.

For now since serialnumber is a legacy fact in our schema, we probably should keep it hidden and think about where it belongs longer-term outside of just fixing the customer issue that led to it being added.

kylog:
Yeah :+1: on that reasoning. I'd rather be consistent about CLI hiding cross-platform, and think through whether we should add additional structure longer term.

PR #1197, opened 2015-10-14
(maint) Update boost to 1.58

mhaskel:
This updates the boost version and also updates the libraries boost is
being built with.

### DIFF
--- a/contrib/facter.ps1
+++ b/contrib/facter.ps1
@@ -52,7 +52,7 @@ if ($arch -eq 64) {
 }
 $mingwVer = "${mingwArch}_mingw-w64_${mingwVerNum}_${mingwThreads}_${mingwExceptions}"
 
-$boostVer = "boost_1_57_0"
+$boostVer = "boost_1_58_0"
 $boostPkg = "${boostVer}-${mingwVer}"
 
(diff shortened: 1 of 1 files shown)


mhaskel:
This seems to be failing travis since the update is a RE ticket not FACT or maint, so let me know if I need to update that.

peterhuene:
@mhaskel Could probably use `(maint)` for this.  Not sure if we want to loosen our restriction on having only `maint`, `FACT`, and `doc` tags.  @kylog?

kylog:
I'd rather reference public tickets in the applicable public project if it's of interest to users, or use (maint). This isn't a user-facing thing so (maint) makes sense to me.

The reason we're excluding RE is because in the past some user-facing changes were sliding in referring to RE tickets, and RE is a private repo so no one external could see the backstory.

mhaskel:
@peterhuene @kylog ok, thanks, updated!

kylog:
Seems reasonable, but I haven't tried 1.58 on Windows. Ping @MikaelSmith.

MikaelSmith:
I was for awhile until I realized we hadn't updated it. This should be fine.

MikaelSmith:
Running a puppet-agent build against this to check it.

PR #1198, opened 2015-10-15
Specify deprecation in the hidden fact note.

gguillotte:
As in [`puppet-docs` PR 551](https://github.com/puppetlabs/puppet-docs/pull/551), make sure the docs generator specifies that we're hiding facts because they're unstructured and deprecated.

Also, fix an issue causing a paragraph in `env_windows_installer` to show up as an unordered list.

### DIFF
--- a/lib/docs/template.erb
+++ b/lib/docs/template.erb
@@ -11,7 +11,7 @@
 ## `<%= name %>`
 
 <% if schema['hidden'] -%>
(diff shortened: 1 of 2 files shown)


gguillotte:
@kylog @jessereynolds Followup to the `puppet-docs` repo work.

gguillotte:
#1203 supersedes this, so I'll close this PR.

PR #1199, opened 2015-10-15
(FACT-1257) Fix Facter exception resolving dmi

MikaelSmith:
When resolving dmi facts on Windows, if WMI is unable to resolve the
Win32_ComputerSystemProduct's Name, Win32_Bios's SerialNumber or
Manufacturer it would previously throw an exception and halt the Facter
run. It will now fail to resolve specific dmi facts without halting
Facter.

### DIFF
--- a/lib/src/facts/windows/dmi_resolver.cc
+++ b/lib/src/facts/windows/dmi_resolver.cc
@@ -18,11 +18,19 @@ namespace facter { namespace facts { namespace windows {
         data result;
 
         auto vals = _wmi->query(wmi::computersystemproduct, {wmi::name});
-        result.product_name = wmi::get(vals, wmi::name);
+        if (vals.empty()) {
+            LOG_DEBUG("WMI query returned no results for %1% with value %2%.", wmi::computersystemproduct, wmi::name);
+        } else {
+            result.product_name = wmi::get(vals, wmi::name);
+        }
 
         vals = _wmi->query(wmi::bios, {wmi::manufacturer, wmi::serialnumber});
-        result.serial_number = wmi::get(vals, wmi::serialnumber);
-        result.manufacturer = wmi::get(vals, wmi::manufacturer);
+        if (vals.empty()) {
+            LOG_DEBUG("WMI query returned no results for %1% with values %2% and %3%.", wmi::bios, wmi::serialnumber, wmi::manufacturer);
+        } else {
+            result.serial_number = wmi::get(vals, wmi::serialnumber);
+            result.manufacturer = wmi::get(vals, wmi::manufacturer);
+        }
 
         return result;
     }


lib/src/facts/windows/dmi_resolver.cc
```diff
@@ -18,11 +18,19 @@ namespace facter { namespace facts { namespace windows {
...
+            LOG_DEBUG("failure resolving dmi::product::name fact, wmi could not resolve %1%::%2%", wmi::computersystemproduct, wmi::name);
+        } else {
+            result.product_name = wmi::get(vals, wmi::name);
+        }
 
         vals = _wmi->query(wmi::bios, {wmi::manufacturer, wmi::serialnumber});
-        result.serial_number = wmi::get(vals, wmi::serialnumber);
-        result.manufacturer = wmi::get(vals, wmi::manufacturer);
+        if (vals.empty()) {
+            LOG_DEBUG("failure resolving dmi::{serialnumber,manufacturer} facts, wmi could not resolve %1%", wmi::manufacturer);
```
peterhuene: This message seems a bit odd as it says "could not resolve <manufacturer>" after informing you it failed to resolve both serial number and manufacturer.  Does the latter addition provide any value?

lib/src/facts/windows/dmi_resolver.cc
```diff
@@ -18,11 +18,19 @@ namespace facter { namespace facts { namespace windows {
         data result;
 
         auto vals = _wmi->query(wmi::computersystemproduct, {wmi::name});
-        result.product_name = wmi::get(vals, wmi::name);
+        if (vals.empty()) {
+            LOG_DEBUG("failure resolving dmi::product::name fact, wmi could not resolve %1%::%2%", wmi::computersystemproduct, wmi::name);
```
peterhuene: `dmi::product::name` might be confusing as it's not a fact name.  Perhaps something like `LOG_DEBUG("WMI query returned no results for %1% with value %2%.", wmi::computersystemproduct, wmi::name);`?

peterhuene:
Could we be potentially be running into this problem in other places we call `wmi::get`?  Do we really want it to throw just because it was given an empty values array?

lib/src/facts/windows/dmi_resolver.cc
peterhuene: Perhaps `LOG_DEBUG("WMI query returned no results for %1% with values %2% and %3%.", wmi::computersystemproduct, wmi::serialnumber, wmi::manufacturer);`

MikaelSmith: It may not be worth including the details about wmi, as they'd be obvious to us. It might be useful to Windows devs to know that wmi isn't behaving correctly. Have a suggestion on improving it?

lib/src/facts/windows/dmi_resolver.cc
MikaelSmith: I feel like there's value tying it back to the dmi facts.

lib/src/facts/windows/dmi_resolver.cc
peterhuene: I guess I don't like the "failure" in the message.  The query succeeded, just gave empty results so the facts shouldn't resolve.  Not really an error condition.

MikaelSmith:
Everywhere else checks for an empty list.

PR #1200, opened 2015-10-15
(FACT-1251) Allow Oracle-specified special characters in zone schema regex

whopper:
According to Oracle: A zone name must begin with an alpha-numeric
character. The name can contain alpha-numeric characters, underbars (_),
hyphens (-), and periods (.).

Update our schema to reflect this.

### DIFF
--- a/lib/schema/facter.yaml
+++ b/lib/schema/facter.yaml
@@ -1780,7 +1780,7 @@ zfs_version:
         Solaris: the `zfs` utility must be present.
 
 zone_<name>_brand:
-    pattern: ^zone_\w+_brand$
+    pattern: ^zone_[\w\-\.]+_brand$
     type: string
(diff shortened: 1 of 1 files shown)


whopper:
Tested by creating the following zone:

```
root@at1fxrh67ztfgbl:~# zoneadm list
global
test-zone_with.specialchars
```

... and then running the test against it:

```
...
zone_test-zone_with.specialchars_brand
zone_test-zone_with.specialchars_id
zone_test-zone_with.specialchars_iptype
zone_test-zone_with.specialchars_name
zone_test-zone_with.specialchars_path
zone_test-zone_with.specialchars_status
zone_test-zone_with.specialchars_uuid
zonename
zones
./tests/verify_facts.rb passed in 1.29 seconds
```

PR #1201, opened 2015-10-16
(FACT-1231) Prevent blkid cache file generation.

peterhuene:
This prevents libblkid from creating an unused cache file on each run of
facter. Instead we use `/dev/null` for the cache file and do a probe for the
block devices.

### DIFF
--- a/lib/src/facts/linux/filesystem_resolver.cc
+++ b/lib/src/facts/linux/filesystem_resolver.cc
@@ -142,7 +142,13 @@ namespace facter { namespace facts { namespace linux {
 
 #ifdef USE_BLKID
         blkid_cache actual = nullptr;
-        if (blkid_get_cache(&actual, nullptr) == 0) {
+        if (blkid_get_cache(&actual, "/dev/null") == 0) {
+            // Do a probe since we're not using a cache file
+            if (blkid_probe_all(actual) != 0) {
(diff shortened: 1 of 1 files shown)


lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -142,7 +142,13 @@ namespace facter { namespace facts { namespace linux {
 
 #ifdef USE_BLKID
         blkid_cache actual = nullptr;
-        if (blkid_get_cache(&actual, nullptr) == 0) {
+        if (blkid_get_cache(&actual, "/dev/null") == 0) {
+            // Do a probe since we're not using a cache file
+            if (blkid_probe_all(actual) != 0) {
+                LOG_DEBUG("blkid_probe_all failed: partition attributes are not available.");
+                blkid_put_cache(actual);
+                actual = nullptr;
```
branan: Shouldn't we somehow be freeing the data pointed by by `blkid_cache` here?

peterhuene: That's done in the "aptly" named `blkid_put_cache`, of course!

branan: Everything about blkid is terrible

branan:
I'm :+1: here. Blkid is awful, but this seems pretty straightforward for what we want to do

PR #1202, opened 2015-10-17
(FACT-1238) Acceptance test, hostname as FQDN

johnduarte:
This commit adds an acceptance test to validate that facter
correctly honors the hostname value if it is set as an FQDN.

### DIFF
--- a/acceptance/tests/ticket_1238_hostname_fqdn.rb
+++ b/acceptance/tests/ticket_1238_hostname_fqdn.rb
@@ -1,0 +1,23 @@
+test_name 'ticket 1238 facter should respect hostname if fqdn'
+
+confine :except, :platform => 'windows'
+
+fqdn = 'foo.bar.example.org'
+
+agents.each do |agent|
+  orig_hostname = on(agent, 'hostname').stdout.chomp
+  teardown do
+    step 'restore original hostname' do
+      on(agent, "hostname #{orig_hostname}")
+    end
+  end
+
+  step 'set hostname as fqdn' do
+    on(agent, "hostname #{fqdn}")
+  end
+
+  step 'validate facter honors fqdn' do
+    res = on(agent, 'facter fqdn').stdout.chomp
+    assert_equal(fqdn, res, "fqdn hostname #{fqdn} did not match `facter fqdn` #{res}")
+  end
+end


branan:
Does AIX have a valid `hostname` command?

acceptance/tests/ticket_1238_hostname_fqdn.rb
```diff
@@ -0,0 +1,22 @@
+test_name 'ticket 1238 facter should respect hostname if fqdn'
+
+confine :except, :platform => 'windows'
```
er0ck: why not on windows?

acceptance/tests/ticket_1238_hostname_fqdn.rb
```diff
@@ -0,0 +1,22 @@
...
+
+fqdn = 'foo.bar.example.org'
+
+agents.each do |host|
```
er0ck: i prefer the iterator var to be agent in these cases, so it's easier for others to read the code; eg for the cases when one uses a hosts loop.  but maybe only important if you make other changes to this PR

acceptance/tests/ticket_1238_hostname_fqdn.rb
```diff
@@ -0,0 +1,22 @@
...
+      on(host, "hostname #{orig_hostname}")
+    end
+  end
+
+  step 'set hostname as fqdn' do
+    on(host, "hostname #{fqdn}")
+  end
+
+  step 'validate facter honors fqdn' do
+    assert_equal(fqdn, on(host, 'facter fqdn').stdout.chomp)
```
er0ck: this should probably have a failed assert message.

acceptance/tests/ticket_1238_hostname_fqdn.rb
johnduarte: Does Windows have a `hostname` command?

MikaelSmith: Our VMs do, but it's not used to resolve the hostname on Windows.

johnduarte:
@branan 

    [0] [AIX] root@pe-aix-53-acceptance:~ # which hostname
    /usr/bin/hostname

acceptance/tests/ticket_1238_hostname_fqdn.rb
joshcooper: Windows has a `hostname` command, but it's read-only:

```
C:\Windows\System32>hostname foo
sethostname: Use the Network Control Panel Applet to set hostname.
hostname -s is not supported.
```

johnduarte:
cc @whopper @joshcooper @kylog 

Remaining facter AIX pr

MikaelSmith:
:+1:

PR #1203, opened 2015-10-19
(docs) Split the facts reference into "modern" and "legacy" sections. 

nfagerlund:
This came out of some discussion in chat, where some people wanted to emphasize even more heavily that Puppet Labs would prefer that people use the new structured facts. 

I'm expecting we'll go back and forth a bit on wording, but wanted to get this up there to show it's really pretty easy to do.

### DIFF
--- a/lib/docs/generate.rb
+++ b/lib/docs/generate.rb
@@ -9,8 +9,17 @@ require 'ostruct'
 PATH_TO_SCHEMA = File.join(File.dirname(__FILE__), '../schema/facter.yaml')
 PATH_TO_TEMPLATE = File.join(File.dirname(__FILE__), 'template.erb')
 
-scope = OpenStruct.new({
-  :facts => YAML.load_file(PATH_TO_SCHEMA)
-})
+schema = YAML.load_file(PATH_TO_SCHEMA)
 
-puts ERB.new(File.read(PATH_TO_TEMPLATE), nil, '-').result(scope.instance_eval {binding})
+def format_facts(fact_hash)
+  scope = OpenStruct.new({
+    :facts => fact_hash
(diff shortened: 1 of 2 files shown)


gguillotte:
Looks like Facter's Travis still prefers `(doc)` over `(docs)`.

nfagerlund:
Aww, boo. Does someone with Travis access want to take a look at that? For other projects, I think we decided to allow both (doc) and (docs), with the latter being preferred due to precedent.

MikaelSmith:
It should allow both.

nfagerlund:
Ping — what do y'all think about splitting the list into two? And about the wording?

MikaelSmith:
Seems like a sensible organization. If we have them split up like that, does it make sense to mention they're hidden for every fact?

gguillotte:
We could cut it, though it might still make sense to note that in the fact
as some links point directly to the fact's heading.

MikaelSmith:
Ok, that seems like a good reason. I could see using a more subtle design to show something that applies to the whole section, and is visible from any point in that section, but I'm not sure how easy that is.

gguillotte:
Maybe add a section on the page that officially notes the facts are
deprecated, why, what the effects are, and how to work around it. Then we
just add a one-word link to that section instead of the repetitive sentence.

MikaelSmith:
Not deprecated.

And yeah, a small link back to the header of the section could cover that.

joshcooper:
Let's wait to merge until puppet-agent 1.2.7 goes out

nfagerlund:
Works for me.

MikaelSmith:
I think we could merge as-is.

PR #1205, opened 2015-10-20
(FACT-1257) Fix Facter exception resolving dmi

peterhuene:
When resolving dmi facts on Windows, if WMI is unable to resolve the
Win32_ComputerSystemProduct's Name, Win32_Bios's SerialNumber or
Manufacturer it would previously throw an exception and halt the Facter
run. It will now fail to resolve specific dmi facts without halting
Facter.


peterhuene:
Cherry-picked from master, belongs in stable.

PR #1206, opened 2015-10-20
(FACT-1254) Allow dot syntax to query ruby facts

branan:
Previously, the dot syntax could not query ruby facts, since they were
opaque to the facter collection.

This commit updates the collection code to recognize when it has
encountered a ruby value during dot lookup, and passes the value and
the remaining query segments to the facter::ruby code to finish the
query.

### DIFF
--- a/lib/inc/facter/ruby/ruby.hpp
+++ b/lib/inc/facter/ruby/ruby.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "../facts/collection.hpp"
+#include "../facts/value.hpp"
 #include "../export.h"
 #include <vector>
 #include <string>
@@ -39,6 +40,16 @@ namespace facter { namespace ruby {
      * @param paths The paths to search for custom facts.
      */
     LIBFACTER_EXPORT void load_custom_facts(facter::facts::collection& facts, std::vector<std::string> const& paths = {});
+
+    /**
+     * Traverses a ruby fact and returns a new value based on the
+     * query segments passed in the range.
+     * @param value The original value to query
+     * @param segment The beginning of the query segment range
+     * @param end The end of the query segment range
+     * @return Returns a pointer to the value queried, or nullptr if it does not exist.
+     */
+    LIBFACTER_EXPORT facts::value const* lookup(facts::value const* value, std::vector<std::string>::iterator segment, std::vector<std::string>::iterator end);
 
     /**
      * Uninitialize Ruby integration in Facter.
--- a/lib/inc/internal/ruby/ruby_value.hpp
+++ b/lib/inc/internal/ruby/ruby_value.hpp
@@ -8,6 +8,8 @@
 #include "fact.hpp"
 #include <facter/facts/value.hpp>
 
+#include <unordered_map>
+
 namespace facter { namespace ruby {
 
     /**
@@ -79,12 +81,39 @@ namespace facter { namespace ruby {
          */
         leatherman::ruby::VALUE value() const;
 
+        /**
+         * Exposes an owned ruby VALUE as a facter ruby_value
+         * @param child the owned child object to wrap
+         * @param key the query string for the child
+         * @return pointer to the ruby_value wrapper for the child object
+         */
+        ruby_value const* wrap_child(leatherman::ruby::VALUE child, std::string key) const;
+
+        /**
+         * Get a cached ruby_value wrapper for a child VALUE
+         * @param key the query string for the child
+         * @return pointer to the ruby_value wrapper, or nullptr if none exists
+         */
+        ruby_value const* child(const std::string& key) const;
+
      private:
         static void to_json(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, facts::json_allocator& allocator, facts::json_value& json);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, std::ostream& os, bool quoted, unsigned int level);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, YAML::Emitter& emitter);
 
         leatherman::ruby::VALUE _value;
+
+        // This is mutable because of constness that's passed down
+        // from the collection object during dot-syntax fact
+        // querying. That query is const over the collection, which
+        // means the collection's actions on values also need to be
+        // const. Unfortunately, we need a place to stash the owning
+        // pointer to the new value that we extract from a ruby object
+        // during lookup. The logical place for that ownership is the
+        // parent ruby object (right here). So we get this cool
+        // mutable field that owns the C++ wrappers for looked-up ruby
+        // values.
+        mutable std::unordered_map<std::string, std::unique_ptr<ruby_value>> _children;
     };
 
 }}  // namespace facter::ruby
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -4,6 +4,7 @@
 #include <facter/facts/scalar_value.hpp>
 #include <facter/facts/array_value.hpp>
 #include <facter/facts/map_value.hpp>
+#include <facter/ruby/ruby.hpp>
 #include <leatherman/file_util/directory.hpp>
 #include <leatherman/util/environment.hpp>
 #include <facter/util/string.hpp>
@@ -14,6 +15,7 @@
 #include <internal/facts/resolvers/ec2_resolver.hpp>
 #include <internal/facts/resolvers/gce_resolver.hpp>
 #include <internal/facts/resolvers/augeas_resolver.hpp>
+#include <internal/ruby/ruby_value.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
@@ -341,6 +343,7 @@ namespace facter { namespace facts {
         }
 
         bool in_quotes = false;
+        vector<string> segments;
         string segment;
         for (auto const& c : query) {
             if (c == '"') {
@@ -351,16 +354,32 @@ namespace facter { namespace facts {
                 segment += c;
                 continue;
             }
-            current = lookup(current, segment);
-            if (!current) {
-                return nullptr;
-            }
+            segments.emplace_back(move(segment));
             segment.clear();
         }
-
         if (!segment.empty()) {
-            current = lookup(current, segment);
+            segments.emplace_back(move(segment));
         }
+
+        auto segment_end = end(segments);
+        for (auto segment = begin(segments); segment != segment_end; ++segment) {
+            auto rb_val = dynamic_cast<ruby::ruby_value const *>(current);
+            if (rb_val) {
+                current = facter::ruby::lookup(current, segment, segment_end);
+                if (!current) {
+                    LOG_DEBUG("cannot lookup an element with \"%1%\" from Ruby fact", *segment);
+                }
+                // Once we hit Ruby there's no going back, so whatever we get from Ruby is the value.
+                return current;
+            } else {
+                current = lookup(current, *segment);
+            }
+            if (!current) {
+                // Break out early if there's no value for this segment
+                return nullptr;
+            }
+        }
+
         return current;
     }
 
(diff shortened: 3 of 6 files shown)


MikaelSmith:
Failed linting.

branan:
Just whitespace. Re-pushed

lib/tests/ruby/ruby.cc
```diff
@@ -271,12 +271,17 @@ SCENARIO("custom facts written in Ruby") {
...
+        THEN("the fact can be queried") {
+            REQUIRE(ruby_value_to_string(facts.query<ruby_value>("foo.5.0")) == "1");
+        }
     }
     GIVEN("a fact that resolves to a hash value") {
         REQUIRE(load_custom_fact("hash_fact.rb", facts));
         THEN("the value is a hash") {
             REQUIRE(ruby_value_to_string(facts.get<ruby_value>("foo")) == "{\n  int => 1,\n  bool_true => true,\n  bool_false => false,\n  double => 12.34,\n  string => \"foo\",\n  array => [\n    1,\n    2,\n    3\n  ]\n}");
         }
+        THEN("the fact can be subqueried") {
```
MikaelSmith: This is odd...

lib/src/ruby/ruby_value.cc
```diff
@@ -272,4 +272,10 @@ namespace facter { namespace ruby {
         emitter << Null;
     }
 
+    ruby_value const* ruby_value::wrap_child(VALUE child) const {
+        auto result = new ruby_value(child);
+        children.emplace_back(result);
```
MikaelSmith: What is this used for? Does Ruby GC it otherwise?

lib/inc/internal/ruby/ruby_value.hpp
```diff
@@ -79,12 +79,20 @@ namespace facter { namespace ruby {
...
+         */
+        ruby_value const* wrap_child(leatherman::ruby::VALUE child) const;
+
      private:
         static void to_json(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, facts::json_allocator& allocator, facts::json_value& json);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, std::ostream& os, bool quoted, unsigned int level);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, YAML::Emitter& emitter);
 
         leatherman::ruby::VALUE _value;
+        mutable std::vector<std::unique_ptr<ruby_value>> children;
```
MikaelSmith: I'm not sure why this is mutable.

lib/src/ruby/ruby_value.cc
peterhuene: Should be done in the emplace_back call; if someone mistakenly adds an early return before it is emplaced, it will leak.

```
children.emplace_back(new ruby_value(child));
return children.back();
```

lib/src/facts/collection.cc
```diff
@@ -351,16 +354,32 @@ namespace facter { namespace facts {
                 segment += c;
                 continue;
             }
-            current = lookup(current, segment);
+            segments.push_back(segment);
```
peterhuene: Micro-optimization: how about a `segments.empalce_back(std::move(segment))` instead?

lib/src/facts/collection.cc
```diff
@@ -351,16 +354,32 @@ namespace facter { namespace facts {
...
+            segment.clear();
+        }
+        if (!segment.empty()) {
+            segments.push_back(segment);
```
peterhuene: Likewise.

lib/src/ruby/ruby_value.cc
MikaelSmith: Seconded, that's what I had in my head.

lib/inc/facter/ruby/ruby.hpp
```diff
@@ -41,6 +42,16 @@ namespace facter { namespace ruby {
...
 
     /**
+     * Traverses a ruby fact and returns a new value based on the
+     * query segments passed in the range.
+     * @param value The original value to query
+     * @param segment The beginning of the query segment range
+     * @param end The end of the query segment range
+     * @return Returns a pointer to the value queried, or nullptr if it does not exist.
+     */
+    LIBFACTER_EXPORT facts::value const* lookup(facts::value const* value, std::vector<std::string>::iterator segment, std::vector<std::string>::iterator end);
```
peterhuene: Rather than `dynamic_cast`ing twice, can this just take a `ruby_value const*`?

lib/inc/internal/ruby/ruby_value.hpp
peterhuene: Yeah, unfortunately the `lookup` contract assumes the value already exists and doesn't lend itself to creating one on demand.  Not sure how to solve this that doesn't require a larger refactoring.

peterhuene: Also, naming convention here differs from one uses for member variables.

peterhuene:
Other than the above, I'm generally :+1:.  Would love to see us not have to use a mutable member, though.

lib/inc/internal/ruby/ruby_value.hpp
branan: @MikaelSmith it should probably be cached for repeated lookups, yes. I hadn't thought of that :/

This exists because somebody needs to own the facter-side value object. It is mutable because the lookup contract passes values by const pointer, and also because this does not affect the logical constness of the object.

lib/inc/facter/ruby/ruby.hpp
branan: ruby_value isn't public API, but this header is

branan:
@peterhuene I think avoiding mutable here means either using a const_cast (which is worse, IMO) or making the entire lookup/query call stack non-const, which also feels weird/wrong to me. I'd be happy to toss it but of the options a `mutable` cache member isn't the worst

lib/inc/facter/ruby/ruby.hpp
peterhuene: (sadpanda)

branan: I'm happy to expose ruby_value (the rest of the value types are public API, if you think that's a reasonable answer... but it probably would pull in more of the leatherman ruby stuff as public API, which I think we didn't want.

peterhuene:
I can live with it as mutable.  I don't think caching is currently necessary as we don't provide a way to query using this syntax from the Ruby API currently.

lib/inc/facter/ruby/ruby.hpp
peterhuene: I think we can live with this for now rather than having to shuffle around headers and what not.

branan:
It's not that much harder to make it check for an existing object, and it makes the API behave more cleanly. I'd rather just do it

lib/src/ruby/ruby_value.cc
```diff
@@ -272,4 +272,17 @@ namespace facter { namespace ruby {
         emitter << Null;
     }
 
+    ruby_value const* ruby_value::wrap_child(VALUE child, string key) const {
+        return _children.emplace(key, std::unique_ptr<ruby_value>(new ruby_value(child))).first->second.get();
```
peterhuene: Taking `key` by value, should be moved into the map.

lib/src/ruby/ruby_value.cc
```diff
@@ -272,4 +272,17 @@ namespace facter { namespace ruby {
...
+    }
+
+    ruby_value const* ruby_value::child(string key) const {
```
peterhuene: `key` should be const&.

lib/src/ruby/ruby.cc
```diff
@@ -55,6 +58,67 @@ namespace facter { namespace ruby {
...
+                }
+                val = result;
+            } else {
+                LOG_DEBUG("cannot lookup element \"%1%\": container is not an array or hash", *segment);
+            }
+            if (ruby.is_nil(val)) {
+                return nullptr;
+            }
+        }
+        return rb_value->wrap_child(val, key);
```
peterhuene: `key` can be moved here.

peterhuene:
Other than those last comments, :+1:

branan:
annnd I actually somehow broke this when I added the caching. Amazing

lib/tests/ruby/ruby.cc
```diff
@@ -271,12 +271,18 @@ SCENARIO("custom facts written in Ruby") {
...
+            REQUIRE(ruby_value_to_string(facts.query<ruby_value>("foo.5.0")) == "1");
+        }
     }
     GIVEN("a fact that resolves to a hash value") {
         REQUIRE(load_custom_fact("hash_fact.rb", facts));
         THEN("the value is a hash") {
             REQUIRE(ruby_value_to_string(facts.get<ruby_value>("foo")) == "{\n  int => 1,\n  bool_true => true,\n  bool_false => false,\n  double => 12.34,\n  string => \"foo\",\n  array => [\n    1,\n    2,\n    3\n  ]\n}");
         }
+        THEN("the members of the fact can be queried") {
+            REQUIRE(ruby_value_to_string(facts.query<ruby_value>("array.1")) == "2");
```
peterhuene: Seems like the test is not correct to me.  Should be `foo.array.1`?

branan: Yup. There was also a code bug - I was comparing a VALUE against 0 instead of calling is_nil.

lib/src/ruby/ruby.cc
```diff
@@ -55,6 +58,69 @@ namespace facter { namespace ruby {
...
     }
 
+    value const* lookup(value const* value, vector<string>::iterator segment, vector<string>::iterator end) {
+        auto rb_value = dynamic_cast<ruby_value const*>(value);
+        if (!rb_value) {
+             return nullptr;
+        }
+
+        // Check for a cached lookup
+        auto key = accumulate(segment, end, string{}, [](const string& a, const string& b) -> string {
```
branan: Oh, there's a bug here. If a hash has a key called "foo.bar" as well as a key called "foo" that has a hash with a key "bar", this won't work. I need to wrap in quotes when there are dots in a string (just like the actual query string would have)

branan: Fixed below

MikaelSmith:
Compilation failed. b.find doesn't return b.end, it returns string::npos.

MikaelSmith:
Doxygen this time, `child` is missing docs.

branan:
@MikaelSmith added doxygen, all green now

lib/inc/internal/ruby/ruby_value.hpp
```diff
@@ -79,12 +81,28 @@ namespace facter { namespace ruby {
...
+         */
+        ruby_value const* child(const std::string& key) const;
+
      private:
         static void to_json(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, facts::json_allocator& allocator, facts::json_value& json);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, std::ostream& os, bool quoted, unsigned int level);
         static void write(leatherman::ruby::api const& ruby, leatherman::ruby::VALUE value, YAML::Emitter& emitter);
 
         leatherman::ruby::VALUE _value;
+        mutable std::unordered_map<std::string, std::unique_ptr<ruby_value>> _children;
```
MikaelSmith: I'd like a comment here about why mutable, and what changes might require that we rethink this approach.

MikaelSmith:
Other than adding a comment about mutable, I'm :+1:

branan:
@MikaelSmith comment added

lib/inc/internal/ruby/ruby_value.hpp
```diff
@@ -79,12 +81,39 @@ namespace facter { namespace ruby {
...
         leatherman::ruby::VALUE _value;
+
+        // This is mutable because of constness that's passed down
+        // from the collection object during dot-syntax fact
+        // querying. That query is const over the collection, which
+        // means the collection's actions on values also need to be
+        // const. Unfortunately, we need a place to stash the owning
+        // pointer to the new value that we extract from a ruby object
+        // during lookup. The logical place for that ownership is the
+        // partent ruby object (right here). So we get this cool
```
MikaelSmith: Typo: 'partent'

peterhuene: Partent is perfectly cromulent.

branan:
@MikaelSmith typo fixed

PR #1207, opened 2015-10-21
(maint) Allow 'docs' in commit summaries

kylog:


kylog:
For #1203 (and friends).

PR #1208, opened 2015-10-21
(maint) Run Valgrind on Facter (w/o Ruby)

MikaelSmith:
Runs Valgrind against Facter in Travis, with errors causing a failure.
Doesn't use Ruby, as Ruby messes with the stack in ways that don't work
with Valgrind.


MikaelSmith:
Doesn't work yet.

PR #1211, opened 2015-10-22
(FACT-1258) Guard all Ruby callbacks for exceptions

MikaelSmith:
Ruby doesn't know how to handle C++ exceptions (it's a C program) so we
must catch any C++ exceptions in the ruby_* calls to avoid crashing the
interpreter.

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -444,357 +444,428 @@ namespace facter { namespace ruby {
         return from_self(ruby.lookup({"Facter"}));
     }
 
+    static VALUE safe_eval(char const* scope, function<VALUE()> body)
+    {
+        try {
+            return body();
+        } catch (exception const& e) {
+            LOG_ERROR("%1% uncaught exception: %2%", scope, e.what());
+        }
+        return api::instance().nil_value();
+    }
+
     VALUE module::ruby_version(VALUE self)
     {
-        auto const& ruby = api::instance();
-        return ruby.lookup({ "Facter", "FACTERVERSION" });
+        return safe_eval("Facter.version", [&]() {
+            auto const& ruby = api::instance();
+            return ruby.lookup({ "Facter", "FACTERVERSION" });
+        });
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/ruby/module.cc
```diff
@@ -444,357 +444,428 @@ namespace facter { namespace ruby {
...
+            }
+            return array;
+        });
     }
 
     VALUE module::ruby_which(VALUE self, VALUE binary)
     {
-        // Note: self is Facter::Core::Execution
-        auto const& ruby = api::instance();
+        return safe_eval("Facter.which", [&]() {
```
peterhuene: Technically this method is `Facter::Core::Execution::which`.

lib/src/ruby/module.cc
```diff
@@ -444,357 +444,428 @@ namespace facter { namespace ruby {
...
+            return ruby.utf8_value(path);
+        });
     }
 
     VALUE module::ruby_exec(VALUE self, VALUE command)
     {
-        // Note: self is Facter::Core::Execution
-        auto const& ruby = api::instance();
-        return execute_command(ruby.to_string(command), ruby.nil_value(), false);
+        return safe_eval("Facter.exec", [&]() {
```
peterhuene: Technically this method is `Facter::Core::Execution::exec`.

lib/src/ruby/module.cc
```diff
@@ -444,357 +444,428 @@ namespace facter { namespace ruby {
...
+            auto const& ruby = api::instance();
+            return execute_command(ruby.to_string(command), ruby.nil_value(), false);
+        });
     }
 
     VALUE module::ruby_execute(int argc, VALUE* argv, VALUE self)
     {
-        // Note: self is Facter::Core::Execution
-        auto const& ruby = api::instance();
+        return safe_eval("Facter.exec", [&]() {
```
peterhuene: Technically this method is `Facter::Core::Execution::execute`.

MikaelSmith:
Hah, should've told you to ignore whitespace. I can fix those up though.

MikaelSmith:
Updated.

PR #1212, opened 2015-10-22
(FACT-1241) Use dmidecode from puppet-agent when available

branan:


branan:
I want to kick an acceptance run of this from my fork before this is emrged, just to be on the safe side. It's kind of fidgety to test manually

lib/inc/internal/util/agent.hpp
```diff
@@ -0,0 +1,22 @@
+/**
+ * @file
+ * Declares methods for Facter's integration with the puppet-agent package.
+ */
+#pragma once
+
+#include <leatherman/execution/execution.hpp>
+#include <string>
+
+namespace facter { namespace util { namespace agent {
```
branan: I'm not married to this namespace

lib/inc/internal/util/agent.hpp
```diff
@@ -0,0 +1,22 @@
...
+    inline std::string which(std::string exe) {
```
peterhuene: Should be `const&`.

lib/inc/internal/util/agent.hpp
```diff
@@ -0,0 +1,22 @@
...
+#ifdef FACTER_PATH
+        std::string fixed = leatherman::execution::which(exe, {FACTER_PATH});
+        if (fixed.empty()) {
```
peterhuene: Can this be:

```c++
if (!fixed.empty()) {
    return fixed;
}
LOG_WARNING(...)
#endif

That makes it a little more clear that we're falling through the ifdef to return exe, imo.

lib/inc/internal/util/agent.hpp
peterhuene: Perhaps just `facter::util` but call the function `agent_which`?  I don't really have strong opinion on this though; completely fine as is.

lib/inc/internal/util/agent.hpp
peterhuene: Missing doc comments.

peterhuene:
Other than the above, :+1:

lib/inc/internal/util/agent.hpp
branan: Yeah, that's much better. This was just copied from what was already being copypasta'd around the codebase

lib/inc/internal/util/agent.hpp
branan: Assuming that this will almost always be called with a string literal, taking the string by value and moving on the return involves only a single construction (as opposed to two when taking the const&)... but that's a micro-optimization, and const& is probably cleaner looking

branan:
@peterhuene give it another look over?

branan:
CMake install failed on appveyor :/

branan:
This is going to require one more change to be compatible with PA-2. I'm going to push up the additional commit momentarily

peterhuene:
:+1:

peterhuene:
I assume we can merge this in without waiting on the PA ticket?

branan:
@peterhuene if we merge without the PA ticket, we'll get warnings because our dmidecode can't be found. We shouldn't ship it that way, but I wouldn't be totally heartbroken about it sitting merged for a couple days

branan:
actually, nevermind... that would break the facter tests (we ensure there's no output on stderr)

joshcooper:
Since we're close to release, let's hold off on merging this to stable. Or retarget for facter#master

joshcooper:
Since the sprint ends before stable will reopen, shall we retarget for facter#master so we can do final review and merge?

branan:
I don't know if we want to retarget this one - I think we were planning on doing a stable z-release of Facter for puppet-agent 1.3.0, which would mean we want this to go into stable as soon as 3.1.1 is tagged

branan:
I guess if there's nothing in master we *don't* want in puppet-agent 1.3.0 I'm fine retargeting

joshcooper:
Yeah, good questions. I was assuming after facter 3.1.1 was released we were going to fast forward stable to master, and then release either 3.1.2 or 3.2.0 out of stable in preparation for puppet-agent 1.3.0.

branan:
This requires puppet-agent#387

branan:
I'm fine retargeting this to master in that case

PR #1213, opened 2015-10-23
(FACT-1254) Fix compilation with Clang

MikaelSmith:
`std::accumulate` lives in the `numeric` header, but GCC includes that
from the `algorithm` header. Clang is more strict, so switch to
including `numeric`.


branan:
yes

PR #1217, opened 2015-11-02
(maint) Update cmake URL but ignore certificate

MikaelSmith:
cmake.org updated to redirect to HTTPS and use a shorter URL. However
the certificate they use requires validation from an SSL chain that's
not installed on Ubuntu 12.04. Since we were previously using HTTP, just
ignore the certificate for now while this is resolved.


branan:
Everything is terrible

PR #1218, opened 2015-11-04
(FACT-1266) Remove leading dot from .gemspec.in

MikaelSmith:
The task that produces a tar file for a release ignores dot-files. This
is likely a property of the glob used when specifying files to tar. The
template file used to produce .gemspec is expected for CMake
configuration, so we shouldn't hide it anyway.


MikaelSmith:
Requires https://github.com/puppetlabs/facter/pull/1217 to pass CI.

MikaelSmith:
Rebased to fix CI.

mmoll:
this is looking good, thanks! :+1:

melissa:
Does this change actually work, and get the file into the tarball?

MikaelSmith:
Looks like. Testing notes in https://tickets.puppetlabs.com/browse/FACT-1266.

PR #1220, opened 2015-11-04
(FACT-1267) Fix os.macosx.version.full for .0 releases and update test

MikaelSmith:


MikaelSmith:
I've tested both changes independently and believe this will work.

MikaelSmith:
Passed CI.

PR #1221, opened 2015-11-16
(FACT-1277) Add puppetversion to `facter -p`

MikaelSmith:
When `facter -p` was restored to Facter 3, the `puppetversion` fact was
overlooked. It was moved to Puppet, but the code path we call to
initialize `facter -p` doesn't set it up. Explicitly set it by calling
`Puppet.version`, with support for a Puppet helper to be setup in
PUP-5508 to future-proof `facter -p` for new core Puppet facts.

### DIFF
--- a/lib/src/ruby/ruby.cc
+++ b/lib/src/ruby/ruby.cc
@@ -15,9 +15,16 @@ static const char load_puppet[] =
 "require 'puppet'\n"
 "Puppet.initialize_settings\n"
 "unless $LOAD_PATH.include?(Puppet[:libdir])\n"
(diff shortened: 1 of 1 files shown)


kylog:
I updated facter's protected branches configs to point to the new appveyor status endpoint.

kylog:
Should we point this one at stable?

MikaelSmith:
Oops, I'll fix tomorrow.

MikaelSmith:
Retargeted stable as https://github.com/puppetlabs/facter/pull/1222.

PR #1222, opened 2015-11-17
(FACT-1277) Add puppetversion to `facter -p`

MikaelSmith:
When `facter -p` was restored to Facter 3, the `puppetversion` fact was
overlooked. It was moved to Puppet, but the code path we call to
initialize `facter -p` doesn't set it up. Explicitly set it by calling
`Puppet.version`, with support for a Puppet helper to be setup in
PUP-5508 to future-proof `facter -p` for new core Puppet facts.


peterhuene:
:+1:

PR #1224, opened 2015-11-18
(maint) Fix Facter API to export scalar_value<T>.

peterhuene:
Missing an export declaration on scalar_value<T>.  This prevents Facter
from exporting needed type info on Linux.


peterhuene:
This seems to work with GCC 5.2 on Windows where it was previously failing.  May need to update AppVeyor to get this to pass.

peterhuene:
I'll add an `#ifdef` to this so we can at least get this in and working for Linux, with the expectation that it won't work on Windows until we move our tooling to GCC 5+.

branan:
:+1:

PR #1228, opened 2015-11-20
Support multilib installation on RPM distributions

hguemar:
Major RPM distributions define the CMake variable LIB_SUFFIX to determine
if libraries should be shipped in /usr/lib64 or /usr/lib

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -241,8 +241,7 @@ endif()
 if (WIN32)
     set(LIBFACTER_INSTALL_DESTINATION bin)
 else()
-    # TODO: lib64 for certain operating systems?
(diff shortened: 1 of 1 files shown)


kylog:
@hguemar interesting, thanks, I wasn't aware of `LIB_SUFFIX`.

Do you mind filing a ticket at https://tickets.puppetlabs.com/browse/FACT describing the issue and then referencing the ticket in the commit summary like "(FACT-1234) Support multilib installation on RPM distributions"? That will get it past our commit checker in travis-ci, and also appeases our release  script.

Thanks!

MikaelSmith:
I'll take care of this in a helper I'm creating in Leatherman.

MikaelSmith:
Created LTH-64, this will be dealt with downstream from that change.

PR #1229, opened 2015-11-22
(FACT-959) Incorrect processor counts on Windows

jasminen:
Facter is showing the CPU socket count on processorcount fact which should be the the logical count and not the physical count (which is already represented by the physicalprocessorcount fact).
Apparently it is a bug in the proc.respond_to?(:NumberOfLogicalProcessors) which always returns false since the NumberOfLogicalProcessors is not a regular method but an OLE method. Therefore it should use the ole_respond_to? method (WIN32OLE class).

### DIFF
--- a/spec/unit/processors/os_spec.rb
+++ b/spec/unit/processors/os_spec.rb
@@ -221,6 +221,7 @@ describe Facter::Processors::Windows do
       before :each do
         proc = stubs 'proc'
         proc.stubs(:Name).returns("Intel(R)    Celeron(R)   processor")
+        proc.stubs(:ole_respond_to?).with("NumberOfLogicalProcessors").returns(false)
         load(Array.new(2, proc))
       end
 
@@ -235,6 +236,7 @@ describe Facter::Processors::Windows do
       before :each do
         proc = stubs 'proc'
         proc.stubs(:NumberOfLogicalProcessors).returns(2)
+        proc.stubs(:ole_respond_to?).with("NumberOfLogicalProcessors").returns(true)
         proc.stubs(:Name).returns("Intel(R)    Celeron(R)   processor")
         load(Array.new(2, proc))
       end
--- a/spec/unit/processors/os_spec.rb
+++ b/spec/unit/processors/os_spec.rb
@@ -221,7 +221,7 @@ describe Facter::Processors::Windows do
       before :each do
         proc = stubs 'proc'
         proc.stubs(:Name).returns("Intel(R)    Celeron(R)   processor")
-        proc.stubs(:ole_respond_to?).with("NumberOfLogicalProcessors").returns(false)
+        proc.stubs(:ole_respond_to?).with(:NumberOfLogicalProcessors).returns(false)
         load(Array.new(2, proc))
       end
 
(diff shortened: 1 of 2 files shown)


ferventcoder:
@peterhuene @MikaelSmith @kylog status of 2.x?

MikaelSmith:
The 2.x branch is still open to fixes, and this looks correct. However it appears some spec tests need to be updated with this change.

kylog:
@MikaelSmith do we know if facter 3 does The Right Thing in this regard?

MikaelSmith:
Yes, the WMI call implementation is completely different and correctly queries for NumberOfLogicalProcessors.

ferventcoder:
@jasminen can you update specs for this? Thanks!

joshcooper:
Thanks @jasminen! It looks like this has been broken since https://github.com/puppetlabs/facter/commit/0da8bd75c482466f1212f2cbd3d24d35532416dd when I updated the code to support ruby 1.9.

jasminen:
@ferventcoder, I've updated the specs accordingly. Sorry for that :)
Thanks.

spec/unit/processors/os_spec.rb
```diff
@@ -221,6 +221,7 @@ def load(procs)
       before :each do
         proc = stubs 'proc'
         proc.stubs(:Name).returns("Intel(R)    Celeron(R)   processor")
+        proc.stubs(:ole_respond_to?).with(:NumberOfLogicalProcessors).returns(false)
```
ferventcoder: Additionally you should ensure that `proc.NumberOfLogicalProcessors` is not called.

ferventcoder:
@jasminen Thanks for updating the specs. I've added at least one comment that I think will help ensure the code is working properly.

Also when you update this time, would you mind squashing the commits together into one commit? If you need help on this, let us know.

jasminen:
@ferventcoder Thanks for the comment. I am not sure this addition is necessary since if the  proc.NumberOfLogicalProcessors will be called the test will fail on unknown method because i am not stubbing it in 2003. This also was the logic before i changed the specs, i only added the response of the ole_respond_to? method. WDYT?
Also, can you please help me understand how to squash my commits into one commit? Thanks :)

ferventcoder:
@jasminen Here are some articles to help you squash/rebase:

* https://help.github.com/articles/about-git-rebase/
* http://stackoverflow.com/q/14534397/18475
* http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

ferventcoder:
@jasminen For specs, I'm stating you should ensure that the function is never called. That is a `proc.expects(:function_name).never` (I believe)

spec/unit/processors/os_spec.rb
joshcooper: I don't think we need an explicit `never` expectation since it's a stub object and it will raise if any other methods (besides `Name`) are called on it.

joshcooper:
I thought closing and reopening would rekick appveyor, but I just remembered we don't have it enabled on old branches. We'll need to manually very this on windows before merging.

MikaelSmith:
Closed in favor of squash and rebase in https://github.com/puppetlabs/facter/pull/1242.

PR #1230, opened 2015-11-24
(FACT-1246) Return the first value found in ID_LIKE

kylog:
The ID_LIKE field may contain a space-separated list of related OS
families. Some Cisco platforms exercise this feature, and when they
do so, we only want to return the first value found.


peterhuene:
:+1:

MikaelSmith:
Cherry-picking to stable?

MikaelSmith:
AppVeyor failure is due to using 64-bit Git in AppVeyor. I should have fixes for that in a sec.

kylog:
Rebased and green now. This should be good to go.

PR #1234, opened 2015-12-14
(FACT-1286) Only return files, never directories

p8952:
Without this patch 'Facter::Util::Resolution.which' will return a
directory if a directory with a matching name is found earlier in the
PATH than a executable file with a matching name.

This patch ensures that 'Facter::Util::Resolution.which' will only
return items which are files.


MikaelSmith:
:+1:

PR #1237, opened 2015-12-16
(FACT-1288) Bump leatherman and use `leatherman_install`

MikaelSmith:
Update to the latest Leatherman, and use `leatherman_install` to install
targets. Linking `libfacter.bundle` now explicitly points to `lib`.

Also picks up static libnowide.


MikaelSmith:
ping @branan

branan:
:+1: this looks sane to me

PR #1239, opened 2015-12-19
(maint) remove git & package methods from acceptance

er0ck:
This change removes unused install methods from acceptance

[skip ci]

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -189,7 +189,6 @@ def sha
 end
 
 namespace :ci do
-
   namespace :test do
 
     USAGE = <<-EOS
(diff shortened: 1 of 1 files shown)


MikaelSmith:
:+1:

acceptance/Rakefile
```diff
@@ -197,82 +196,58 @@ May specify project SHA environment variable, which can be a branch, release ver
...
-    preserved = list_preserved_configurations
-    print_preserved(preserved)
-    config_number = (ENV['CONFIG_NUMBER'] || 0).to_i
-    config_path = preserved[config_number][0]
-    puts "Using ##{config_number}: #{config_path}"
-    beaker_test(beaker_run_type,
-      :hosts_file => "log/latest/#{PRESERVED_HOSTS_FILENAME}"
-    )
+    EOS
+    task :test_against_preserved_hosts do
```
MikaelSmith: This is now part of the ci:test namespace?

er0ck: woops.  removed an extra end and didn't catch the namespacing...  will fix

er0ck:
@MikaelSmith fixed...

acceptance/Rakefile
MikaelSmith: I think know they're no longer in any namespace. They were originally in the `ci` namespace. I'd expect to see a lot less line changes for this PR, indentation of tasks shouldn't really change.

er0ck:
this should probably just wait for the changes coming via the POC against puppet to unify the puppet agent component acceptance workflows.  these should also be against stable.

PR #1240, opened 2015-12-21
(FACT-1289) fix detection of a virtualbox guest

duritong:
A virtualbox guest on top of a KVM hypervisor gets reported as kvm,
as we only look at the last line of virt-what's output. We need
to be more specific here.

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -256,6 +256,8 @@ Facter.add("virtual") do
         'xenu'
       when /ibm_systemz/i
         'zlinux'
+      when /virtualbox/i
+        'virtualbox'
       else
         output.to_s.split("\n").last
       end
--- a/spec/unit/virtual_spec.rb
+++ b/spec/unit/virtual_spec.rb
@@ -410,6 +410,7 @@ describe "Virtual fact" do
         'xen-dom0' => 'xen0',
         'xen-domU' => 'xenu',
         'ibm_systemz' => 'zlinux',
+        "virtualbox\nkvm" => 'virtualbox',
       }
 
       virt_what_map.each do |input,output|


MikaelSmith:
This seems reasonable. :+1:

kylog:
Not directly related, but a different case of facter getting tripped up by treating virt-what's output as a contract: https://tickets.puppetlabs.com/browse/FACT-1244.

MikaelSmith:
The output appears to suggest virt-what will report multiple levels of virtualization. Parsing should be updated to match the first virtualization layer reported.

This change will need to be applied to Facter 3 as well.

branan:
I've requested physical hardware from our IT team so I can play with different hypervisors and hypervisor nestings and get Facter's virtualization facts into better shape generally.

I'm going to close this for now, as it only fixes a special case and wouldn't have a chance to be released before it gets replaced with my rewrite of these facts

branan:
If it turns out priorities change and I can't spend the time I want on virtualization stuff, I'm happy to re-visit this as something we can ship in the meantime (since we already regex match a bunch of other hypervisors)

5nafu:
> The output appears to suggest virt-what will report multiple levels of virtualization. 
> Parsing should be updated to match the first virtualization layer reported.
As you can see in the source code from [virt-what](http://git.annexia.org/?p=virt-what.git;a=blob;f=virt-what.in;hb=HEAD), the output is _not_ ordered by level. 
virt-what will test (in a static but arbitrary order) for visualization technologies and print an output if the technology matches. Getting the last level of virtualization with virt-what might not be anything the software can deliver.

MikaelSmith:
There may actually have been some development on this. Is that the intention of https://github.com/puppetlabs/libwhereami? @branan @caseywilliams

caseywilliams:
That is the intention, yes - we've worked around this problem in libwhereami, but it will take a little longer before it starts to inform existing facts (although replacing virt-what is the goal). We've just merged some changes that will add a new `hypervisors` fact that includes libwhereami's results (this is a list of detected hypervisors plus metadata as opposed to a single name).

PR #1242, opened 2015-12-22
(FACT-959) Incorrect processor counts on Windows

MikaelSmith:
Facter is showing the CPU socket count on processorcount fact which
should be the the logical count and not the physical count (which is
already represented by the physicalprocessorcount fact).

Apparently it is a bug in the
proc.respond_to?(:NumberOfLogicalProcessors) which always returns
false since the NumberOfLogicalProcessors is not a regular method but
an OLE method. Therefore it should use the ole_respond_to? method
(WIN32OLE class).


MikaelSmith:
I've manually verified this on Windows, processorcount looks good and the specs pass.

PR #1243, opened 2015-12-23
(FACT-1281) Allow host configs to be generated at runtime

joshcooper:
This commit adds a dependency on beaker-hostgenerator in the acceptance
Gemfile, which may be overridden using the
`BEAKER_HOSTGENERATOR_VERSION` environment variable.

The acceptance Rakefile will continue to use the `CONFIG` environment
variable, which points to an existing host config file.

If `PLATFORMS` is specified, it will override `CONFIG`, and the Rakefile
will use beaker-hostgenerator to dynamically generate a host config. The
host config is generated in the `facter/acceptance/tmp` directory, as
opposed to `/tmp` so that it doesn't conflict with concurrent acceptance
tests running on the same CI coordinator. It also appends SecureRandom
to the hostfile, because it seems like a good idea. The resulting host
config is copied to the `log/latest/config.yml`.

For example, to run against a tagged build:

    $ bundle exec rake ci:test:aio PLATFORMS=redhat7-64a SHA=1.3.2

To run against a non-tagged commit:

    $ bundle exec rake ci:test:aio PLATFORMS=redhat7-64a \
      SHA=<full sha> SUITE_VERSION=<git describe>

This commit doesn't remove the static host configs as CI jobs need to be
updated first.

PR #1244, opened 2015-12-28
(FACT-1281) Allow host configs to be generated at runtime

kylog:


kylog:
This adds a one-line, temporary change to #1243. This change can be backed out when we merge https://github.com/puppetlabs/ci-job-configs/pull/804, hopefully later today (or Real Soon Now).

PR #1245, opened 2015-12-28
(maint) Use 'hosts' not 'config' in acceptance/Rakefile

kylog:
On the stable->master mergeup I just did, I missed that master
has dropped the 'config' method. So this moves the new changes
to use 'hosts'.


er0ck:
:+1:

er0ck:
we should backport all this stuff to stable as well...

PR #1250, opened 2016-01-07
(FACT-380) Use default gateway device for `ipaddress` fact on linux

branan:
Previously, Facter would treat whichever device was first in `ifconfig` as the default device. This kinda-sorta-works in a world where "first ethernet device" is your default, and there are no devices alphabetically before "e".

The new version of the code will check `/proc/net/route` and use the IP address of whichever interface provides the default route.

### DIFF
--- a/lib/facter/ipaddress.rb
+++ b/lib/facter/ipaddress.rb
@@ -3,7 +3,9 @@
 # Purpose: Return the main IP address for a host.
 #
 # Resolution:
-#   On the Unixes does an ifconfig, and returns the first non 127.0.0.0/8
+#   On Linux and AIX, it examines the routing table and uses the IP address
+#   of the default interface.
+#   On other Unixes does an ifconfig, and returns the first non 127.0.0.0/8
 #   subnetted IP it finds.
 #   On Windows, it attempts to use the socket library and resolve the machine's
 #   hostname via DNS.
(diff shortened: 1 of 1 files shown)


lib/facter/ipaddress.rb
```diff
@@ -28,15 +30,10 @@
...
-      output.split("\n").each do |line|
-        match = regexp.match(line)
-        if match and not /^127\./.match(match[1])
-          ip = match[1]
-          break
-        end
+    File.read("/proc/net/route").split("\n").each do |line|
+      parts = line.split()
+      if parts[1] == "00000000"
+        ip ||= Facter.value("ipaddress_#{parts[0]}".intern)
```
MikaelSmith: Should this break after finding a valid ip?

MikaelSmith:
Looks like a spec needs updating.

lib/facter/ipaddress.rb
branan: meh. The route table shouldn't be too big, and `|||=` ensures it'll take the first one. I'd kind rather not complicate the code with extra conditionals

branan:
Ah,yup. I also need to update netmask and ipaddress6 to follow the same convention

branan:
I'm an idiot and opened a new PR for this. checkout out #1258

PR #1257, opened 2016-01-13
(maint) acceptance-fix previous merge-up

er0ck:
The previous merge-up from stable had a few conflicts improperly
resolved.  There were at least two refactors that conflicted.  This
change takes the pure stable changes, as intended.

[skip ci]

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -5,18 +5,30 @@ require 'securerandom'
 require 'fileutils'
 require 'beaker-hostgenerator'
 
+$LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), 'lib'))
+require 'puppet/acceptance/git_utils'
+extend Puppet::Acceptance::GitUtils
+
 ONE_DAY_IN_SECS = 24 * 60 * 60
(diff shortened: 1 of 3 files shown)


acceptance/Rakefile
```diff
@@ -28,7 +40,7 @@ module HarnessOptions
     end
 
     def get_options(file_path)
-      puts "Attempting to merge config file: #{file_path}"
+      puts file_path
```
MikaelSmith: Looks like this change might be worth preserving.

acceptance/config/git/options.rbgit
```diff
@@ -0,0 +1,7 @@
+{
```
MikaelSmith: The file name here looks like a mistake.

MikaelSmith:
Some of these improvements will return as part of PA-139 (https://github.com/puppetlabs/puppet/pull/4539).

acceptance/Rakefile
er0ck: there are a few things here that are worth preserving that we can do later. they should all be captured here: https://github.com/puppetlabs/puppet/pull/4539

PR #1258, opened 2016-01-14
(FACT-380) Use the interface of the default route for network facts

branan:
Previously, Facter would use whichever non-loopback device came first
in the ifconfig output for the networking facts. This is not always
what most people would consider the "ipaddress of a node", since
things like virtual network devices can show up in this list and might
sort before physical interfaces.

This changes the code to check the routing table for the default
interface, and uses the values from that interface.


MikaelSmith:
That's some serious test refactoring.

MikaelSmith:
Looks like OS isn't mocked out for a few tests, and can't read /etc/lsb-release.

MikaelSmith:
Seems reasonable except for the remaining test failures.

branan:
@MikaelSmith trying to fight my File.read stub was getting super weird with whatever ancient rspec/mocha combo we're using here. I went ahead and just split out the read of `/proc/net/route` to its own method that I can stub without fighting the rest of the program that wants to read files

MikaelSmith:
Going to check this out on Windows.

spec/unit/ipaddress6_spec.rb
```diff
@@ -44,8 +50,11 @@ def ifconfig_fixture(filename)
...
     Facter::Util::IP.stubs(:get_ifconfig).returns("/sbin/ifconfig")
-    Facter::Util::IP.stubs(:exec_ifconfig).with(["2>/dev/null"]).
+    Facter::Util::IP.stubs(:exec_ifconfig).
       returns(ifconfig_fixture('linux_ifconfig_all_with_multiple_interfaces_and_fe80'))
+    routes = ifconfig_fixture('net_route_all_with_multiple_interfaces')
+    Facter::Util::IP.stubs(:read_proc_net_route).returns(routes)
+    Facter.collection.internal_loader.load(:interfaces)
 
     Facter.value(:ipaddress6).should == "2610:10:20:209:212:3fff:fe80:2201"
   end
```
MikaelSmith: It looks like the next test needs to be updated as well, it's still stubbing ifconfig with "2>/dev/null" and produces an error message on Windows: "Could not retrieve fact='ipaddress6', resolution='<anonymous>': No such file or directory @ rb_sysopen - /proc/net/route"

branan:
@MikaelSmith Ah, yeah. that passed on my machine because my interface names don't match the fixture data, so it returned 'nil' when comparing my route to ifconfig. Updated to load the proper fixture data there. Should pass on Windows now

spec/unit/ipaddress6_spec.rb
MikaelSmith: Changing it to the following seems to work
```
  it "should not return ipaddress6 link-local address for Linux" do
    Facter::Core::Execution.stubs(:exec).with('uname -s').returns('Linux')
    Facter::Util::IP.stubs(:get_ifconfig).returns("/sbin/ifconfig")
    Facter::Util::IP.stubs(:exec_ifconfig).
      returns(ifconfig_fixture('linux_ifconfig_all_with_multiple_interfaces_and_no_public_ipv6'))
    routes = ifconfig_fixture('net_route_all_with_multiple_interfaces')
    Facter::Util::IP.stubs(:read_proc_net_route).returns(routes)
    Facter.collection.internal_loader.load(:interfaces)

    Facter.value(:ipaddress6).should be_false
  end
```

MikaelSmith:
:+1: once it passes CI.

PR #1259, opened 2016-01-14
(FACT-1313) Acceptance - sleep after hostname

johnduarte:
On Solaris 11, the hostname command returns before the system
has set the new hostname. This commit adds a sleep statement
to allow the hostname value to be updated on the system before
validating that facter can retrieve the expected fqdn.

[skip ci]

### DIFF
--- a/acceptance/tests/ticket_1238_hostname_fqdn.rb
+++ b/acceptance/tests/ticket_1238_hostname_fqdn.rb
@@ -14,6 +14,7 @@ agents.each do |agent|
 
   step 'set hostname as fqdn' do
     on(agent, "hostname #{fqdn}")
+    sleep 1 # on Solaris 11 hostname returns before the hostname is updated
   end
 
   step 'validate facter honors fqdn' do


acceptance/tests/ticket_1238_hostname_fqdn.rb
```diff
@@ -14,6 +14,7 @@
 
   step 'set hostname as fqdn' do
     on(agent, "hostname #{fqdn}")
+    sleep 1 # on Solaris 11 hostname returns before the hostname is updated
```
joshcooper: Is 1 sec long enough?

johnduarte: It has been when tested locally on the Solaris 11 SPARC box.
This is a recent failure, so I don't know what has changed from a performance standpoint on the box. I am trying not to introduce any additional unneeded latency. Having said that, I am open to bumping it up.

MikaelSmith:
:frowning: but I don't have any strong objections.

PR #1261, opened 2016-01-22
(FACT-1317) Fix YAML output of non-numeric types

MikaelSmith:
Several formats intended to be non-numeric are interpreted as numbers in
(Ruby) YAML.

Mac and IPv6 addresses as sexagesimal types:
  10:10:10:10:10:10:10:10
  01:23:45:54:32:21

Comma-delimited lists as comma-delimited integers:
  1,2,3,4,5
  123,456

Modify YAML quoting to ensure these patterns are quoted.

### DIFF
--- a/acceptance/tests/verify_facts.rb
+++ b/acceptance/tests/verify_facts.rb
@@ -29,9 +29,7 @@ def validate_fact(name, node, fact_value, hidden)
               # YAML-CPP seems to drop the decimal whenever it feels like it, so just match Int too.
               fact_value.is_a? Float or fact_value.is_a? Integer or fact_value.to_s =~ /^(\.inf|\.nan|-\.inf)$/
             when 'string'
-              # Any string can be interpreted as a more specific type by ruby
-              # YAML.
-              true
+              fact_value.is_a? String or fact_value.is_a? TrueValue or fact_value.is_a? FalseValue
             when 'ip'
               fact_value.is_a? String and fact_value =~ @ip_pattern
             when 'ip6'
--- a/lib/src/util/string.cc
+++ b/lib/src/util/string.cc
@@ -114,26 +114,33 @@ namespace facter { namespace util {
             return true;
         }
 
-        // Check that it doesn't start with the ':' special character. This interferes with parsing.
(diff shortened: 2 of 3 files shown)


MikaelSmith:
Merge after release

lib/src/util/string.cc
```diff
@@ -114,26 +114,33 @@ namespace facter { namespace util {
...
+        // Poor man's check for a numerical string or list of numbers
         // May start with - or +
-        // May contain at most one . or ,
+        // May contain one . and one or more ,
         // All other characters should be digits
-        bool has_separator = false;
+        // Note: Ruby YAML interprets 1,2,3 as a number, but not 1.2.3.
+        // It doesn't appear to honor locales when parsing YAML.
+        bool has_dot = false;
         for (size_t i = 0; i < str.size(); ++i) {
```
branan: any reason why this isn't `for(char c: str)` ?

MikaelSmith: I don't think there is.

acceptance/tests/verify_facts.rb
```diff
@@ -29,9 +29,7 @@ def validate_fact(name, node, fact_value, hidden)
               # YAML-CPP seems to drop the decimal whenever it feels like it, so just match Int too.
               fact_value.is_a? Float or fact_value.is_a? Integer or fact_value.to_s =~ /^(\.inf|\.nan|-\.inf)$/
             when 'string'
-              # Any string can be interpreted as a more specific type by ruby
-              # YAML.
-              true
+              fact_value.nil? or fact_value.is_a? String or fact_value.is_a? TrueValue or fact_value.is_a? FalseValue
```
joshcooper: nit, it looks like `validate_fact` will never be called with a nil `fact_value`? https://github.com/MikaelSmith/facter/blob/FACT-1317/acceptance/tests/verify_facts.rb#L68

MikaelSmith: Hmm, I don't remember why that case came up.

MikaelSmith:
Updated.

PR #1262, opened 2016-01-25
Revert "(FACT-380) Use the interface of the default route for network facts"

branan:
This update was more breaking than we anticipated (see comments on FACT-380, but also FACT-1316 which was filed almost immediately)

For users that really need the new behavior, Facter 3 provides the new behavior.

This reverts commit ec73549be0c26aab40a8968330726b369ab080f7.


MikaelSmith:
:+1:

MikaelSmith:
Is there a new ticket for this change going into 2.4.6?

branan:
@MikaelSmith not currently. This is the first time I've had to deal with a revert first-hand, not really sure what our general policy/plan is here

MikaelSmith:
Probably associate it with FACT-1316 with a fix target of 2.4.6. /cc @kylog

kylog:
We don't have a general policy/precedent for this type of scenario, but, yeah, that makes sense to me here.

PR #1263, opened 2016-01-26
(FACT-1321) Install facter batch files on windows

melissa:
As we switch to building puppet-agent for windows with vanagon, we need
access to resources that currently live in puppet_for_the_win. Though
these resources depend on resources that are made available in
puppet-agent, it makes sense to store these in the project specific
repo.

This takes advantage of `environment.bat` in puppet-agent installed with https://github.com/puppetlabs/puppet-agent/pull/512

### DIFF
--- a/ext/windows/facter.bat
+++ b/ext/windows/facter.bat
@@ -1,0 +1,8 @@
+@echo off
+SETLOCAL
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET "PATH=%~dp0;%PATH%"
+)
+facter.exe %*
--- a/ext/windows/run_facter_interactive.bat
+++ b/ext/windows/run_facter_interactive.bat
@@ -1,0 +1,9 @@
+@echo Running facter on demand ...
+@echo off
+SETLOCAL
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET "PATH=%~dp0;%PATH%"
+)
+elevate.exe "%~dp0facter_interactive.bat"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,3 +150,8 @@ else()
 endif()
 
 install(FILES ${PROJECT_SOURCE_DIR}/man/man8/facter.8 DESTINATION ${MANDIR})
+if (WIN32)
+  install(FILES ${PROJECT_SOURCE_DIR}/ext/windows/facter.bat DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+  install(FILES ${PROJECT_SOURCE_DIR}/ext/windows/facter_interactive.bat DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+  install(FILES ${PROJECT_SOURCE_DIR}/ext/windows/run_facter_interactive.bat DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+endif()
--- a/ext/windows/facter_interactive.bat
+++ b/ext/windows/facter_interactive.bat
@@ -1,0 +1,6 @@
+@echo off
+SETLOCAL
+echo Running Facter on demand ...
+cd "%~dp0"
+call .\facter.bat %*
+PAUSE


ext/windows/run_facter_interactive.bat
```diff
@@ -0,0 +1,7 @@
+@echo Running facter on demand ...
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET PATH=%~dp0;%PATH%
+)
+elevate.exe "%~dp0facter_interactive.bat"
```
melissa: `facter.bat` and `facter_interactive.bat` work fine, but I'm having trouble with `run_facter_interactive.bat`. It seems like the call to `elevate.exe` isn't right. Any thoughts?

cc/ @joshcooper @MikaelSmith

melissa: Also, https://github.com/puppetlabs/puppet-agent/pull/515 will install elevate.exe and elevate.exe.config

MikaelSmith: It's just Cygwin.

melissa: This wasn't working in cygwin. Running from not-cygwin worked as expected

shrug:
:+1:

melissa:
cc/ @MikaelSmith @kylog @joshcooper could one of you review this?

ext/windows/run_facter_interactive.bat
joshcooper: Do we know the path to `elevate.exe` relative to the directory where `run_facter_interactive.bat` is installed?

ext/windows/facter.bat
```diff
@@ -0,0 +1,8 @@
+@echo off
+SETLOCAL
+if exist "%~dp0environment.bat" (
```
joshcooper: When will `environment.bat` not exist?

melissa: See comments on https://github.com/puppetlabs/hiera/pull/340

The interesting part about facter is that these files aren't installed by the system, they have to be purposefully copied over. I figured that since they're in the public facter repo, theoretically they should be able to be used on their own. So, if someone just installs this on their system without environment.bat, it should still work. The other batch scripts assume ruby is in the working directory, so I kept that assumption here if environment.bat doesn't exist.

ext/windows/run_facter_interactive.bat
melissa: if it's installed as a part of puppet-agent, it should be in the same directory

MikaelSmith:
These look like they need to be installed. We could make them part of the CMake install task on Windows, something like what we do for man pages at https://github.com/puppetlabs/facter/blob/master/CMakeLists.txt#L152.

melissa:
@MikaelSmith yeah, that seems totally legit. I wasn't sure how to do that. Thanks for pointing me to that example!

melissa:
It installs the files as expected on windows
```
C:\ProgramData\chocolatey\lib\cmake\content\cmake-3.2.2-win32-x86\bin\cmake.exe -P cmake_install.cmake
-- Install configuration: "Release"
-- Installing: C:/Program Files/Puppet Labs/puppet/share/man/man8/facter.8
-- Installing: C:/Program Files/Puppet Labs/puppet/bin/facter.bat
-- Installing: C:/Program Files/Puppet Labs/puppet/bin/facter_interactive.bat
-- Installing: C:/Program Files/Puppet Labs/puppet/bin/run_facter_interactive.bat
-- Installing: C:/Program Files/Puppet Labs/puppet/lib/libfacter.so.a
-- Installing: C:/Program Files/Puppet Labs/puppet/bin/libfacter.so
-- Installing: C:/Program Files/Puppet Labs/puppet/include/facter
```

melissa:
A change to puppet-agent that installed these new files before the cmake change was merged, and now needs to be reverted. https://github.com/puppetlabs/puppet-agent/pull/526

MikaelSmith:
Fine with this; I think it'd be cleaner to only have the batch files in "C:/Program Files/Puppet Labs/bin" and point to puppet/bin, but we can leave that for later. I'm not sure how you want to deal with that.

ext/windows/facter.bat
```diff
@@ -0,0 +1,8 @@
...
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET PATH=%~dp0;%PATH%
```
MikaelSmith: I think this needs to be quoted. When I try it out and `%~dp0` is `C:\Program Files\FACTER\bin` I get the error
```
\FACTER\bin\ was unexpected at this time.
```

ext/windows/run_facter_interactive.bat
```diff
@@ -0,0 +1,8 @@
+@echo Running facter on demand ...
```
MikaelSmith: Should this be followed by `@echo off`? Otherwise we get all the following commands printed out.

MikaelSmith:
:+1:

ext/windows/facter.bat
MikaelSmith: I was almost right, using `SET PATH="%~dp0;%PATH%"` creates an invalid PATH. It needs to be `SET "PATH=%~dp0;%PATH%"`.

MikaelSmith:
Looks good in stand-alone install and a puppet-agent build. :+1:

PR #1264, opened 2016-01-30
OpenBSD swap warnings

faust64:
When no swap device is configured, prevents from showing the following message on OpenBSD:
```
swapctl: no swap devices configured
```


kylog:
Hi @faust64! A few bits of hygiene: 
* can you file a ticket at https://tickets.puppetlabs.com/browse/FACT?
* and then reference that ticket in the commit message?
* also we like our commit summaries in the imperative so that someone skimming summaries has an idea what the change was, can you do so?
* last, it's failing spec tests - you may need a matching spec test change

Thanks!

faust64:
closing that one.
issue opened on Jira, will re-do my messages, ...

PR #1265, opened 2016-02-01
(FACT-1338) Resolve weird pthread dependencies

MikaelSmith:
AIX shows a dependency on `.pthread_cancel` from `libboost_date_time`,
which is now included in linking the `facter` binary from
`leatherman_util`. Other links succeed because they include
`libboost_thread`, bringing in a dependency on pthread.

Explicitly depend on pthread everywhere until LTH-81 is resolved.
Resolving LTH-81 exposes an implicit dependency of `libboost_date_time`
and `libboost_regex` on pthread on some platforms that's not fulfilled
by `find_package(Boost)`, so the libfacter and libfacter_test
dependencies on pthread will be permanent.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,9 @@ include(options)
 
 # We use program_options, system, filesystem, date_time, and regex directly.
 find_package(Boost 1.54 REQUIRED COMPONENTS program_options system filesystem date_time regex)
+# date_time and regex need threads on some platforms, and find_package Boost only includes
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Resolves AIX builds @johnduarte. I believe this started happening when we switched to doing shared library builds of Leatherman. /cc @branan

johnduarte:
:+1:
@MikaelSmith I will take your word that these changes allow for building on AIX.

@branan can you provide a technical review here?

PR #1266, opened 2016-02-02
(FACT-1339) Prevents OpenBSD swap warnings

faust64:
Addressing https://tickets.puppetlabs.com/browse/FACT-1339
Fixes swap fact on swap-less OpenBSD.

### DIFF
--- a/lib/facter/util/memory.rb
+++ b/lib/facter/util/memory.rb
@@ -150,7 +150,7 @@ module Facter::Memory
     when /AIX/i
       (Facter.value(:id) == "root") ? Facter::Core::Execution.exec('swap -l 2>/dev/null') : nil
     when /OpenBSD/i
-      Facter::Core::Execution.exec('swapctl -s')
+      Facter::Core::Execution.exec('swapctl -s 2>/dev/null')
     when /FreeBSD/i
       Facter::Core::Execution.exec('swapinfo -k')
     when /Darwin/i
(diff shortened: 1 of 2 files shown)


faust64:
Now regarding unit tests, I might need help.
I can see where it fails ('memory facts on OpenBSD ...'), which is definitely my doing.
Something around line 176, in spec/unit/memory_spec.rb
Not sure to understand how redirecting stderr could be related to these errors.

branan:
Line 182 of that test file has an expectation that `Facter::Execution.exec('swapctl -s')` is called. This matches parameters exactly, so you need to update that to include the stderr redirection that you added. The other failing expectations are almost certainly because it bails early as soon as it gets a call with unexpected parameters.

Could you also please update your commit message to include the ticket number? See the example commit message in [contributing.md](https://github.com/puppetlabs/facter/blob/master/CONTRIBUTING.md)

faust64:
Indeed, thanks.

branan:
:+1: on the test change, looks fine to me.

Would you mind squashing these commits down to one? Having a commit with broken tests makes git bisect super hard to use.

Once it's squashed I'll hit the big green button.

HAIL9000:
Hey @faust64, it looks like you have a merge commit and a commit that doesn't match the commit message format. It looks like this PR needs to be rebased and then the commits need to be squashed so we just have one or two (all with correctly formatted commit messages).

Do you need some help or guidance on how to do this?

peterhuene:
@faust64 Thanks again for the contribution.  I am closing this PR in favor of #1287, which contains your fix, just squashed and rebased.

faust64:
Sorry I missed the last reply.
Indeed, I'm not familiar with the squashing process. I though I had it fixed... Thanks for your follow-up!

PR #1268, opened 2016-02-08
DO NOT MERGE (FACT-1344) Acceptance - Replace TrueValue/FalseValue

johnduarte:
This commit replaces occurrences of the constants `TrueValue` and
`FalseValue` with `TrueClass` and `FalseClass` respectively.

The previous constants were not recognized as ruby or beaker
constants and resulted in a `NameError` exception when triggered
by the acceptance test.

[skip ci]

### DIFF
--- a/acceptance/tests/verify_facts.rb
+++ b/acceptance/tests/verify_facts.rb
@@ -29,7 +29,7 @@ def validate_fact(name, node, fact_value, hidden)
               # YAML-CPP seems to drop the decimal whenever it feels like it, so just match Int too.
               fact_value.is_a? Float or fact_value.is_a? Integer or fact_value.to_s =~ /^(\.inf|\.nan|-\.inf)$/
             when 'string'
-              fact_value.is_a? String or fact_value.is_a? TrueValue or fact_value.is_a? FalseValue
+              fact_value.is_a? String or fact_value.is_a? TrueClass or fact_value.is_a? FalseClass
             when 'ip'
               fact_value.is_a? String and fact_value =~ @ip_pattern
             when 'ip6'


kylog:
This makes sense, but I'm not understanding why it ever worked .... What changed?

johnduarte:
@kylog this is unclear. We have been using beaker version 2.23.0 since Jan 12, 2016. Running this test against `puppet-agent 1.3.4` results in the same error, where previously it had passed in CI.

MikaelSmith:
I don't remember why I put those in my last PR...

joshcooper:
The failure only shows up if you have a fact that is supposed to be a `string` according to the schema, but the fact value that we parse is not an instance of the `String` class. I presume there is something unique about solaris causing us to try to evaluate `TrueValue` and failing? This raises two questions.

1. Should we revert the [change that added `{True,False}Value`](https://github.com/puppetlabs/facter/commit/812fac739c8ceb36815655fe42f059e56aa61b7e#diff-73a561352755f031ea4efc306183ccb7L34
)? or should we follow through and change it to `{True,False}Class`? Either way we probably want to tag that change with FACT-1317.

2. Which Solaris fact does the schema say should be a `string` but we're not parsing as such? That sounds like a bug that should be fixed for FACT-1344. Or is this a case of running latest acceptance tests, against an older puppet-agent build, where the test requires a quoted value, but the older code doesn't know to quote base-60 integer values like mac address, so we trigger the `TrueValue` behavior above?

MikaelSmith:
My preference would be to quote whatever output is causing this failure on Solaris.

johnduarte:
If the true/false check is removed, the following error is emitted.
```
RuntimeError: type of zfs_featurenumbers did not match schema, 12345 of type Fixnum is not a string
```

MikaelSmith:
That's a case I thought I'd handled. It should be quoting `1,2,3,4,5`.

joshcooper:
Are we running the latest acceptance tests against an older puppet-agent build that doesn't have the fix?

MikaelSmith:
It's not failing on stable or master nightly. On stable, the characters are quoted
```
zfs_featurenumbers: "1,2,3,4,5"
```

On master it's not, because we hadn't yet merged stable to master. I've just done so.

So I think this can change to just remove checking TrueValue and FalseValue, because they're obviously never called. I'm checking what happens if we create a fact that reports true or false (or string versions of those).

joshcooper:
@johnduarte based on @MikaelSmith's comments can you update the commit summary to reference  FACT-1317 as that introduced `TrueValue`? And then I think we can just close FACT-1344.

MikaelSmith:
Emitting a "true" string with Facter 3 as YAML doesn't quote it, which still seems like a problem.

johnduarte:
I am still scratching my head on this. The first manifestation of this issue was over the weekend in the `puppet-agent` stable branch run against SHA `eb30f32564d15376c20292c608d78481656578da` using `facter` stable tests at SHA `fab5f4107ac7cf056fb8f5dd4503f62870009ef3`.

This only impacted Solaris on SPARC while Solaris on x86 passed this test.

MikaelSmith:
Ah. It'd be helpful to see the output for the test.

johnduarte:
Flagging as DO NOT MERGE until we sort this out.

MikaelSmith:
Fixed in https://github.com/puppetlabs/facter/pull/1269.

PR #1269, opened 2016-02-10
(FACT-1317) Fix string values that look like booleans

MikaelSmith:
YAML will read boolean strings as boolean types. When the schema states
the Fact type is a string, we should ensure it's read as a string by
YAML parsers. Quote any boolean literal representation in a
string_value.

Also fix a prior mistake in the verify_facts acceptance test, which no
longer needs to check for boolean classes for a string.

### DIFF
--- a/acceptance/tests/verify_facts.rb
+++ b/acceptance/tests/verify_facts.rb
@@ -29,7 +29,7 @@ def validate_fact(name, node, fact_value, hidden)
               # YAML-CPP seems to drop the decimal whenever it feels like it, so just match Int too.
               fact_value.is_a? Float or fact_value.is_a? Integer or fact_value.to_s =~ /^(\.inf|\.nan|-\.inf)$/
             when 'string'
-              fact_value.is_a? String or fact_value.is_a? TrueValue or fact_value.is_a? FalseValue
+              fact_value.is_a? String
             when 'ip'
               fact_value.is_a? String and fact_value =~ @ip_pattern
             when 'ip6'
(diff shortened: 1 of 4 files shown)


joshcooper:
Looks like travis is complaining about trailing white space:

```
/home/travis/build/puppetlabs/facter/lib/src/util/string.cc:124:  Line ends in whitespace.  Consider deleting these extra spaces.  [whitespace/end_of_line] [4]
```

joshcooper:
For ruby 2.1.5p273, `YAML.load` doesn't parse single character `y|n` as boolean. I don't know if that we care or if it depends on a particular yaml version.

```
irb(main):012:0> inputs = "y|Y|yes|Yes|YES|n|N|no|No|NO|true|True|TRUE|false|False|FALSE|on|On|ON|off|Off|OFF".split('|')
=> ["y", "Y", "yes", "Yes", "YES", "n", "N", "no", "No", "NO", "true", "True", "TRUE", "false", "False", "FALSE", "on", "On", "ON", "off", "Off", "OFF"]
irb(main):013:0> outputs = inputs.select {|value| YAML.load(value).is_a?(TrueClass) }
=> ["yes", "Yes", "YES", "true", "True", "TRUE", "on", "On", "ON"]
irb(main):014:0> inputs - outputs
=> ["y", "Y", "n", "N", "no", "No", "NO", "false", "False", "FALSE", "off", "Off", "OFF"]
```

MikaelSmith:
Overquoting shouldn't be an issue, I just tried to avoid doing it in some legacy cases (like ipv4 addresses). Pushed a fix for the whitespace.

johnduarte:
@MikaelSmith does this need to get merged into stable as well?

MikaelSmith:
oh oops, yes. I'll cherry-pick up.

PR #1270, opened 2016-02-10
(FACT-1317) Fix string values that look like booleans

MikaelSmith:
YAML will read boolean strings as boolean types. When the schema states
the Fact type is a string, we should ensure it's read as a string by
YAML parsers. Quote any boolean literal representation in a
string_value.

Also fix a prior mistake in the verify_facts acceptance test, which no
longer needs to check for boolean classes for a string.


johnduarte:
:+1:  will merge after travis run

PR #1271, opened 2016-02-18
(FACT-1348) Remove usage of the Win32-Security gem

glennsarti:
The Win32-Security gem uses API calls which taget the ANSI implementation
instead of Wide String.  Puppet will remove its requirement for the gem
as part of PUP-5735 however Facter still had a dependency on the gem.
This commit
- Ports the required code from Puppet into Facter for the windows root
  fact.  This required the following classes and files to be ported into
  Facter;
  - Puppet::Util::Windows
  - Puppet::Util::Windows::User
  - Puppet::Util::Windows::Process
  - Puppet::Util::Windows::Error
  - Puppet::Util::Windows::ApiTypes
  - Various FFI extensions
- Ported the applicable integration tests from Puppet to Facter.  No
  additional unit or acceptance tests were applicable for porting

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ mingw << :x64_mingw if Bundler::Dsl::VALID_PLATFORMS.include?(:x64_mingw)
 platform(*mingw) do
   gem 'ffi', '~> 1.9.5', :require => false
   gem 'win32-dir', '~> 0.4.8', :require => false
+  # Use of the win32-security gem is deprecated
   gem 'win32-security', '~> 0.2.5', :require => false
 end
 
--- a/lib/facter/util/windows/api_types.rb
+++ b/lib/facter/util/windows/api_types.rb
@@ -1,0 +1,106 @@
+require 'ffi'
+
+module Facter::Util::Windows::ApiTypes
+  module ::FFI
+    WIN32_FALSE = 0
+
+    # standard Win32 error codes
+    ERROR_SUCCESS = 0
+  end
+
+  module ::FFI::Library
+    # Wrapper method for attach_function + private
+    def attach_function_private(*args)
+      attach_function(*args)
+      private args[0]
+    end
+  end
+
+  class ::FFI::Pointer
+    NULL_HANDLE = 0
+
+    def read_win32_bool
+      # BOOL is always a 32-bit integer in Win32
+      # some Win32 APIs return 1 for true, while others are non-0
+      read_int32 != FFI::WIN32_FALSE
+    end
+    #
+    alias_method :read_dword, :read_uint32
+
+    def read_handle
+      type_size == 4 ? read_uint32 : read_uint64
+    end
+
+    def read_wide_string(char_length, dst_encoding = Encoding::UTF_8)
+      # char_length is number of wide chars (typically excluding NULLs), *not* bytes
+      str = get_bytes(0, char_length * 2).force_encoding('UTF-16LE')
+      str.encode(dst_encoding)
+    end
+
+    def read_win32_local_pointer(&block)
+      ptr = nil
+      begin
+        ptr = read_pointer
+        yield ptr
+      ensure
+        if ptr && ! ptr.null?
+          if FFI::WIN32::LocalFree(ptr.address) != FFI::Pointer::NULL_HANDLE
+            Puppet.debug "LocalFree memory leak"
+          end
+        end
+      end
+
+      # ptr has already had LocalFree called, so nothing to return
+      nil
+    end
+  end
+  # FFI Types
+  # https://github.com/ffi/ffi/wiki/Types
+
+  # Windows - Common Data Types
+  # https://msdn.microsoft.com/en-us/library/cc230309.aspx
+
+  # Windows Data Types
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx
+
+  FFI.typedef :uint32, :dword
+  FFI.typedef :uintptr_t, :handle
+
+  # pointer in FFI is platform specific
+  # NOTE: for API calls with reserved lpvoid parameters, pass a FFI::Pointer::NULL
+  FFI.typedef :pointer, :lpcvoid
+  FFI.typedef :pointer, :lpvoid
+  FFI.typedef :pointer, :lpdword
+  FFI.typedef :pointer, :pdword
+  FFI.typedef :pointer, :phandle
+  FFI.typedef :pointer, :pbool
+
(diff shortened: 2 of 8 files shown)


lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+    NULL_TERMINATOR_WCHAR = 0
+
+    def read_win32_bool
+      # BOOL is always a 32-bit integer in Win32
+      # some Win32 APIs return 1 for true, while others are non-0
+      read_int32 != FFI::WIN32_FALSE
+    end
+
+    alias_method :read_dword, :read_uint32
+    alias_method :read_win32_ulong, :read_uint32
```
Iristyle: These two aliases are unused

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+            Puppet.debug "LocalFree memory leak"
+          end
+        end
+      end
+
+      # ptr has already had LocalFree called, so nothing to return
+      nil
+    end
+
+    def read_com_memory_pointer(&block)
```
Iristyle: `read_com_memory_pointer` is unused

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+      ensure
+        FFI::WIN32::CoTaskMemFree(ptr) if ptr && ! ptr.null?
+      end
+
+      # ptr has already had CoTaskMemFree called, so nothing to return
+      nil
+    end
+
+
+    alias_method :write_dword, :write_uint32
```
Iristyle: `write_word` and `write_dword` are unused

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+  # FFI Types
+  # https://github.com/ffi/ffi/wiki/Types
+
+  # Windows - Common Data Types
+  # https://msdn.microsoft.com/en-us/library/cc230309.aspx
+
+  # Windows Data Types
+  # https://msdn.microsoft.com/en-us/library/windows/desktop/aa383751(v=vs.85).aspx
+
+  FFI.typedef :uint16, :word
```
Iristyle: Unused typedefs:

* `:hwnd`
* `:lpwstr`
* `:lpcwstr`
* `:lpcolestr`
* `:lpcstr`
* `:lpword`
* `:lpbyte`
* `:ulong_ptr`
* `:lpunknown`
* `:win32_ulong`
* `:hresult`

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+    extend ::FFI::Library
+
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/aa373931(v=vs.85).aspx
+    # typedef struct _GUID {
+    #   DWORD Data1;
+    #   WORD  Data2;
+    #   WORD  Data3;
+    #   BYTE  Data4[8];
+    # } GUID;
+    class GUID < FFI::Struct
```
Iristyle: `GUID` not used anywhere (is really here to help define COM methods, which weren't not using at all)

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,256 @@
...
+    end
+
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/ms724284(v=vs.85).aspx
+    # Contains a 64-bit value representing the number of 100-nanosecond
+    # intervals since January 1, 1601 (UTC).
+    # typedef struct _FILETIME {
+    #   DWORD dwLowDateTime;
+    #   DWORD dwHighDateTime;
+    # } FILETIME, *PFILETIME;
+    class FILETIME < FFI::Struct
```
Iristyle: `SYSTEMTIME` and `FILETIME` can be axed.

Iristyle:
I think we can trim a bunch of code out of `api_types`, and then take another look at the PR.  One bit that might not be obvious as well, is that `FFI.typedef` is a global call which adds these aliases for anyone calling FFI in the current Ruby process - so it's best to use as sparingly as possible.

lib/facter/util/windows/api_types.rb
```diff
@@ -0,0 +1,113 @@
...
+    # );
+    ffi_lib :kernel32
+    attach_function_private :CloseHandle, [:handle], :win32_bool
+
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/ms680722(v=vs.85).aspx
+    # void CoTaskMemFree(
+    #   _In_opt_  LPVOID pv
+    # );
+    ffi_lib :ole32
+    attach_function :CoTaskMemFree, [:lpvoid], :void
```
Iristyle: `CoTaskMemFree` is another COM memory function that we can remove...

Gemfile
```diff
@@ -43,7 +43,6 @@ mingw << :x64_mingw if Bundler::Dsl::VALID_PLATFORMS.include?(:x64_mingw)
 platform(*mingw) do
   gem 'ffi', '~> 1.9.5', :require => false
   gem 'win32-dir', '~> 0.4.8', :require => false
-  gem 'win32-security', '~> 0.2.5', :require => false
```
Iristyle: Removing this would affect a dev workflow using gems (and not the MSI) where a user has written custom facts, that depend on code in `win32-security`.

I feel like that scenario is pretty limited in scope, so this is probably OK.

Thoughts @ferventcoder / @joshcooper ?

joshcooper: I'm thinking we should leave the dependency so that we don't break PE 3.8

joshcooper:
Can you create a FACT ticket and update this PR to refer to that?

Gemfile
glennsarti: Ok, I'll add the dependency back in and a comment saying that use of the gem is deprecated, as I did in the Puppet repo.

glennsarti:
I'll create a mirrored FACT ticket and update this PR

joshcooper:
nit, can you remove "Removes the requirement of the gem from Gemfile" from the commit message? Otherwise :+1: from me.

Also I'm wondering if we should move this code to a puppetlabs specific gem and have facter#2.x and puppet depend on it? I think we could do that in the future by assigning the `Facter::Util::Windows` constant to be the new constant, whatever that happens to be. Doesn't need to be addressed in this PR, but would be great for future consideration.

glennsarti:
@joshcooper "Removes requirement..." message removed from the commit and PR text

PR #1272, opened 2016-02-23
(FACT-1024) Fix AWS EC2 Gzip userdata

Pryz:
This patch is for the 2.x branch of facter. It fixes when the AWS EC2 User Data are Gzip and can't be read directly.

Error message :
`Fact resolution fact='ec2_userdata', resolution='rest' resolved to an invalid value: String`

Jira ticket : https://tickets.puppetlabs.com/browse/FACT-1024

### DIFF
--- a/lib/facter/ec2.rb
+++ b/lib/facter/ec2.rb
@@ -46,7 +46,14 @@ Facter.define_fact(:ec2_userdata) do
     end
 
     setcode do
-      @querier.fetch
+      userdata = @querier.fetch
+      begin
+        Facter::Util::Normalization.normalize_string(userdata)
+      rescue Facter::Util::Normalization::NormalizationError => e
+        Facter.debug("Failed to read ec2 userdata: #{e.message}")
+        userdata = nil
+      end
+      userdata
     end
   end
 end


MikaelSmith:
CI is complaining that the call to `open` has changed. However I believe we've handled this problem before, and slightly differently. In https://github.com/puppetlabs/puppet/blob/master/lib/puppet/forge/repository.rb#L67 we end up setting `Accept-Encoding` to `gzip;q=1.0,deflate;q=0.6,identity;q=0.3`.

Pryz:
Hi @MikaelSmith, thanks for taking care of my PR.

Issue is even if you change the request header you will have to "decrypt" the gzip.

I see that you are doing the same here : https://github.com/puppetlabs/puppet/blob/master/lib/puppet/forge/repository.rb#L102

If you validate the approach I will take care of fixing the specs :)

MikaelSmith:
Good point, I overlooked that.

The OpenURI API provides access to `content-encoding`, I'd rather query that to decide whether to use gzip, rather than trying and failing.

Also, I can't imagine how zlib would be unavailable, but we have the feature check in Puppet. @joshcooper or @Sharpie, any recollection of why that exists?

Sharpie:
Wild guess: Running Puppet under a Ruby interpreter which has been built `--with-out-ext=zlib`? Obscure platforms?

MikaelSmith:
Probably need a fallback behavior if `require 'zlib'` fails.

Pryz:
Thanks for the review. I will add a fallback and fix the specs :)

Pryz:
Fallback added for zlib and spec fixed.
I'm not really sure about the tests since I didn't write much specs yet. Let me know :)

MikaelSmith:
What Ruby version are you using to reproduce this problem? It seems like it doesn't occur in all cases.

Pryz:
I have this problem on all my Ubuntu Trusty instances :

```
# ruby --version
ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]
# facter --version
2.4.4
```

lib/facter/ec2/rest.rb
```diff
@@ -125,7 +125,14 @@ def initialize(uri = DEFAULT_URI)
       end
 
       def fetch
-        open(@baseurl).read
+        open(@baseurl, 'Accept-Encoding' => '') do |udata|
```
MikaelSmith: Not clear why this is needed, it doesn't seem to change the response.

MikaelSmith:
Unfortunately, the ec2_userdata can be any binary data. Trying gzip is a bit of a hack.

An alternative approach would be to fix it to emit the binary data, so you can process it as gzip externally.

Pryz:
You're right. Setting ec2_userdata to nil and dealing with it in a custom fact seems to be the right approach.

In my mind people were only using gzip for the userdata since it's automatically handled by cloud-init.

lib/facter/ec2.rb
```diff
@@ -46,8 +46,15 @@
     end
 
     setcode do
-      @querier.fetch
+      userdata = @querier.fetch
+      begin
+        userdata.encode(Encoding::UTF_8)
+      rescue EncodingError
```
MikaelSmith: Apparently encoding handling is a problem in Ruby 1.8.7; I think it'd be ok to only do this check for Ruby 1.9+ /cc @joshcooper

joshcooper: Yeah... you'd have to use iconv for 1.8.7, or just skip that?

Also watch out for ruby 2.1, as that was the version that added http auto-decompression, which is the issue PMT ran into.

Pryz: yes, I just saw how it's currently working here : https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/normalization.rb#L47

Do you want me to do the same here ?

MikaelSmith: It looks like you could call normalize_string directly and catch the NormalizationError.

Pryz:
I've pushed a new version using Facter::Util::Normalization. Let me know what you think

MikaelSmith:
A debug message when it fails would be nice - mentioning that ec2_userdata was not text, so ignoring it - but I'm ok merging this as-is.

Pryz:
Sorry for the delay. I've added a simple message. Feel free to change it ;)

MikaelSmith:
`log_exception` will still generate a warning, which seemed to be what you wanted to avoid. `Facter.debug` could be used instead to log the failure but not show it normally.

Pryz:
Well the main target was to avoid having the full blob of gzipped data displaying when we are running 'facter -p' :D

But I agree using .debug is more elegant. I've pushed the change ;)

MikaelSmith:
Normally we request signing a CLA via http://cla.puppetlabs.com/. This change is trivial enough I don't think it's necessary.

Pryz:
Thanks !

PR #1277, opened 2016-03-03
(FACT-1341) Ensure facter handles unicode in search

HAIL9000:
Prior to this commit, there was no test to make sure that facter
will not throw and unhandled exception when a path contains
unicode characters in Windows. The fix for this bug was actually
merged into leatherman, but we want to ensure that the issue does
not occur in facter by adding a test.


HAIL9000:
Closing this to reopen against stable

PR #1279, opened 2016-03-04
(FACT-1282) Populate interface bindings from routing table

branan:
Previously, we only populated interface bindings directly from the IPs
that were set on that interface. It's possible, though, that an
interface doesn't have a direct IP binding, but instead only has a
'source' parameter on a route that sends outgoing traffic through that
interface. This happens when incoming and outgoing traffic is split
between two interfaces.

This patch queries the source parameter of routes, and treats them as
additional bindings on the interfaces referenced in the routing table.


branan:
Just a linter error on travis. I'll fix up when I push up my next set of tweaks

lib/inc/internal/facts/linux/networking_resolver.hpp
```diff
@@ -47,6 +54,22 @@ namespace facter { namespace facts { namespace linux {
          * @return Returns the primary interface or empty string if one could not be determined.
          */
         virtual std::string get_primary_interface() const override;
+
+     private:
+        void read_routing_table();
+        void populate_from_routing_table(data&) const;
+
+         struct route {
```
MikaelSmith: Indentation off here.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -12,9 +14,18 @@ using namespace std;
 using namespace facter::util::posix;
 
 namespace lth_file = leatherman::file_util;
+namespace lth_exe  = leatherman::execution;
 
 namespace facter { namespace facts { namespace linux {
 
+    networking_resolver::data networking_resolver::collect_data(collection& facts)
+    {
+        read_routing_table();
```
MikaelSmith: Using side-effects?

branan: Yeah. In the `get_primary_interface` method below I check to see if we read the routing table using `ip` here so that we don't re-read it from proc and duplicate work/logic. I could have this method return the vector and just set a bool on the class instance, but just storing the vector of routes as a class members seemed easier seemed easier.

lib/inc/internal/facts/linux/networking_resolver.hpp
branan: autoindent--

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+        if (ip_command.empty()) {
+            LOG_DEBUG("Could not find the 'ip' command. Network bindings will not be populated from routing table");
+            return;
+        }
+
+        auto parse_route_line = [](string& line, std::vector<route>& routes) {
+            vector<boost::iterator_range<string::iterator>> parts;
+            boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+            if (parts.size() % 2 != 1) {
+                LOG_WARNING("Could not process routing table entry: Expected a destination followed by key/value pairs, got '{}'", line);
```
MikaelSmith: `{}` works for substitution?

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+                return true;
+            }
+
+            route r;
+            r.destination.assign(parts[0].begin(), parts[0].end());
```
MikaelSmith: How about a move operation?

branan: parts isn't a vector of strings, it's a vector of iterator_ranges which point directly into the original string. If I wanted to move here I'd have to use the version of `split` that creates copies of every field, which would include the values for the key/value pairs I'm not using. It would maybe simplify the code, but since I know there's at least a couple standard route fields I'm ignoring (via, proto, and meetric), this seemed "more right".

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+            // Iterate over key/value pairs and add the ones we care
+            // about to our routes entries
+            for (size_t i = 1; i < parts.size(); i += 2) {
+                std::string key(parts[i].begin(), parts[i].end());
```
MikaelSmith: Could also do `move` here.

lib/src/facts/linux/networking_resolver.cc
branan: I have no idea where I copypasta'd this from, but it's obviously wrong.

branan: This is mostly a guard against a future version of the `ip` command changing its output format, so obviously never actually came up in my testing

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+            if (parts.size() % 2 != 1) {
+                LOG_WARNING("Could not process routing table entry: Expected a destination followed by key/value pairs, got '{}'", line);
+                return true;
+            }
+
+            route r;
+            r.destination.assign(parts[0].begin(), parts[0].end());
+            // Iterate over key/value pairs and add the ones we care
+            // about to our routes entries
+            for (size_t i = 1; i < parts.size(); i += 2) {
```
MikaelSmith: This could be an algorithm (group by n and iterator over tuples). Candidate for a later date maybe.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+    }
+
+    void networking_resolver::populate_from_routing_table(networking_resolver::data& result) const
+    {
+        for (const auto& r : routes4) {
+            if (r.destination == "default" && result.primary_interface.empty()) {
+                 result.primary_interface = r.interface;
+            }
+            if (!r.source.empty()) {
+                auto iface = find_if(result.interfaces.begin(), result.interfaces.end(), [&](const interface& iface) {
```
MikaelSmith: Can we pull the updating logic in both of these as a helper function? The only difference is `iface->ipv4_bindings` vs `iface->ipv6_bindings`.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -71,4 +89,85 @@ namespace facter { namespace facts { namespace linux {
...
+                    return iface.name == r.interface;
+                });
+                if (iface != result.interfaces.end()) {
+                    auto existing_binding = find_if(iface->ipv4_bindings.begin(), iface->ipv4_bindings.end(), [&](const binding& b) {
```
MikaelSmith: This is `O(n^2)`, but currently we don't see the number of bindings being large enough to require additional code complexity and the overhead of setting up a `set`. /cc @Sharpie on whether that's a bad assumption.

branan:
@MikaelSmith I pushed up a version that deduplicates. I'm honestly not sure it's better than the code duplication :/

I'm building for cisco to test it out now. I'll let you know how that goes

MikaelSmith:
:+1:

branan:
@MikaelSmith this commit still looks good on Cisco

```
{
  domain => "delivery.puppetlabs.net",
  fqdn => "wczlp85z7wqlhpj.delivery.puppetlabs.net",
  hostname => "wczlp85z7wqlhpj",
  interfaces => {
    Gi0_0_0_0 => {
      bindings => [
        {
          address => "192.168.122.100",
          netmask => "255.255.255.0",
          network => "192.168.122.0"
        }
      ],
      bindings6 => [
        {
          address => "fe80::5046:beff:fe1f:4125",
          network => "fe80::"
        }
      ],
      ip => "192.168.122.100",
      ip6 => "fe80::5046:beff:fe1f:4125",
      mac => "52:46:be:1f:41:25",
      mtu => 1514,
      netmask => "255.255.255.0",
      network => "192.168.122.0",
      network6 => "fe80::"
    },
    fwd_ew => {
      bindings6 => [
        {
          address => "fe80::200:ff:fe00:b",
          network => "fe80::"
        }
      ],
      ip6 => "fe80::200:ff:fe00:b",
      mac => "00:00:00:00:00:0b",
      mtu => 1500,
      network6 => "fe80::"
    },
    fwdintf => {
      bindings => [
        {
          address => "192.168.122.100"
        }
      ],
      bindings6 => [
        {
          address => "fe80::200:ff:fe00:a",
          network => "fe80::"
        }
      ],
      ip => "192.168.122.100",
      ip6 => "fe80::200:ff:fe00:a",
      mac => "00:00:00:00:00:0a",
      mtu => 1496,
      network6 => "fe80::"
    },
    lo => {
      bindings => [
        {
          address => "127.0.0.1",
          netmask => "255.0.0.0",
          network => "127.0.0.0"
        }
      ],
      bindings6 => [
        {
          address => "::1",
          netmask => "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
          network => "::1"
        }
      ],
      ip => "127.0.0.1",
      ip6 => "::1",
      mtu => 65536,
      netmask => "255.0.0.0",
      netmask6 => "ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff",
      network => "127.0.0.0",
      network6 => "::1"
    }
  },
  ip => "192.168.122.100",
  ip6 => "fe80::200:ff:fe00:a",
  mac => "00:00:00:00:00:0a",
  mtu => 1496,
  network6 => "fe80::",
  primary => "fwdintf"
}
```

branan:
@MikaelSmith if you have no more comments I'm going to close and re-target against master

branan:
#1282 replaces this targeting stable

PR #1280, opened 2016-03-04
(maint) Skip check for locale warning on OS X

MikaelSmith:
i18n work in Leatherman 0.4.0 introduced using Boost.Locale to generate
locales on platforms besides Windows. For OS X, this has a side-effect
that an exception is no longer thrown if a bad locale is set. Add OS X
to the list of platforms we don't check for a warning when a bad locale
is set.

### DIFF
--- a/acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
+++ b/acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
@@ -3,7 +3,7 @@ confine :except, :platform => 'windows'
 agents.each do |host|
   step 'set an invalid value for locale and run facter'
   result = on host, 'LANG=ABCD facter facterversion'
-  if host['platform'] !~ /solaris|aix|cumulus|osx/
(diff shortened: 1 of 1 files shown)


acceptance/tests/ticket_1123_facter_with_invalid_locale.rb
```diff
@@ -3,7 +3,7 @@
 agents.each do |host|
   step 'set an invalid value for locale and run facter'
   result = on host, 'LANG=ABCD facter facterversion'
-  if host['platform'] !~ /solaris|aix|cumulus/
+  if host['platform'] !~ /solaris|aix|cumulus|osx/
```
joshcooper: do we need to add cisco too?

MikaelSmith: Seems likely.

MikaelSmith:
Added cisco exclusion.

PR #1281, opened 2016-03-04
(maint) Rely on Leatherman CMake logic to decide if we should try to imbue a locale

branan:
This also bumps the leatherman we test against to 0.4.2 to pull in that behavior


branan:
Don't merge until we tag and release Leatherman 0.4.2

branan:
CI is expected to fail until we have LTH 0.4.2 packages to test against here

PR #1283, opened 2016-03-07
(FACT-1282) parse routing table rules with a starting 'type' field

branan:
The `ip` command is the worst, and thinks it's helpful to show the
route type as an optional field before the route destination. This
patch detects that case and allows us to keep parsing key/value pairs
that appear in these rules.

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -6,6 +6,7 @@
 #include <boost/algorithm/string.hpp>
 #include <algorithm>
 #include <cstring>
+#include <unordered_set>
 #include <netpacket/packet.h>
 #include <net/if.h>
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This should still be part of FACT-1282.

branan:
@MikaelSmith commit message updated. Not fixing my branch name since that means closing and re-opening the PR

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -97,19 +98,36 @@ namespace facter { namespace facts { namespace linux {
...
+                }
             }
 
             route r;
-            r.destination.assign(parts[0].begin(), parts[0].end());
+            r.destination.assign(parts[dst_idx].begin(), parts[dst_idx].end());
             // Iterate over key/value pairs and add the ones we care
             // about to our routes entries
-            for (size_t i = 1; i < parts.size(); i += 2) {
+            for (size_t i = dst_idx+1; i < parts.size(); i += 2) {
```
MikaelSmith: This looks wrong; why start at 1 or 2?

branan:
```
[root@bqpdzd6yyu6gzck ~]# ip -6 route
unreachable ::/96 dev lo  metric 1024  error -101
unreachable ::ffff:0.0.0.0/96 dev lo  metric 1024  error -101
unreachable 2002:a00::/24 dev lo  metric 1024  error -101
unreachable 2002:7f00::/24 dev lo  metric 1024  error -101
unreachable 2002:a9fe::/32 dev lo  metric 1024  error -101
unreachable 2002:ac10::/28 dev lo  metric 1024  error -101
unreachable 2002:c0a8::/32 dev lo  metric 1024  error -101
unreachable 2002:e000::/19 dev lo  metric 1024  error -101
unreachable 3ffe:ffff::/32 dev lo  metric 1024  error -101
fe80::/64 dev ens192  proto kernel  metric 256 
```

MikaelSmith:
:+1:

lib/src/facts/linux/networking_resolver.cc
branan: It's entirely the way the `ip` command prints stuff. It's great for humans, but not awesome for parsing, unfortunately.

PR #1284, opened 2016-03-07
(FACT-1362) Add platform identification for HuaweiOS

ScottGarman:
HuaweiOS runs as an LXC container based on Debian 8 with some
modifications (for example, it uses Debian sysv instead of sytemd).

The firmware version is a long string which they store in
/etc/huawei-release which does not follow semver, but looks
like "V100R006C00SPC300B188", so we just pass along this
version information without further parsing.

I've tested this on a Huawei CloudEngine switch and it returns the following for 'facter os':
```
{
  architecture => "ppc",
  family => "Debian",
  hardware => "ppc",
  name => "HuaweiOS",
  release => {
    full => "V100R006C00SPC300B188",
    major => "V100R006C00SPC300B188"
  },
  selinux => {
    enabled => false
  }
}
```
Also, someone on the client team should especially double-check that I'm not likely breaking the Debian os.name section by injecting the Huawei check before it.

### DIFF
--- a/lib/src/facts/linux/os_linux.cc
+++ b/lib/src/facts/linux/os_linux.cc
@@ -46,6 +46,10 @@ namespace facter { namespace facts { namespace linux {
     {
         // Check for Debian variants
         bs::error_code ec;
+        if (is_regular_file(release_file::huawei, ec)) {
+          return os::huawei;
+        }
+
         if (is_regular_file(release_file::debian, ec)) {
             if (distro_id == os::ubuntu || distro_id == os::linux_mint) {
                 return distro_id;
@@ -206,6 +210,7 @@ namespace facter { namespace facts { namespace linux {
             { string(os::oracle_enterprise_linux),  string(os_family::redhat) },
             { string(os::amazon),                   string(os_family::redhat) },
             { string(os::xen_server),               string(os_family::redhat) },
+            { string(os::huawei),                   string(os_family::debian) },
             { string(os::linux_mint),               string(os_family::debian) },
             { string(os::ubuntu),                   string(os_family::debian) },
             { string(os::debian),                   string(os_family::debian) },
@@ -273,6 +278,12 @@ namespace facter { namespace facts { namespace linux {
         // Alpine uses the entire contents of the release file as the version
         if (value.empty() && name == os::alpine) {
             value = lth_file::read(release_file::alpine);
+            boost::trim_right(value);
+        }
+
+        // HuaweiOS uses the entire contents of the release file as the version
+        if (value.empty() && name == os::huawei) {
+            value = lth_file::read(release_file::huawei);
             boost::trim_right(value);
         }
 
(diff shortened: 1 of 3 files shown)


mckern:
@ScottGarman I rubbed a little code fence formatting on that PR message for you.

MikaelSmith:
Can you file a ticket for adding acceptance tests of the platform facts?

ScottGarman:
@MikaelSmith sure, it's FACT-1363.

kylog:
That's weird - I forgot I could edit other people's comments, but anyway, I fixed up the ticket number in the comment above.

kylog:
Does family of Linux seem right? As opposed to Debian?

ScottGarman:
@kylog I think a family of Debian would be better if the fact that the customizations include changing the init system won't be an issue. I assume Debian will help providers select the apt and dpkg tools?

lib/src/facts/linux/os_linux.cc
```diff
@@ -46,6 +46,10 @@ namespace facter { namespace facts { namespace linux {
     {
         // Check for Debian variants
         bs::error_code ec;
+        if (is_regular_file(release_file::huawei, ec)) {
```
MikaelSmith: Is this a debian variant?

kylog:
Yes, the decision on what `os.family` reports should consider the implications for the service provider and the package provider.

You mentioned that it is sysv-based instead of systemd-based, and some of the debian family do use sysv - there is an `init` provider and a `debian` provider (for debian-flavored sysv). Do you know which of those providers is a better suit to HuaweiOS? That might partially guide the decision.

Anyway, if it's easiest to predicate any such decisions on `os.name`, not surface the debian relationship at all, and sprinkle `HuaweiOS` comparisons where needed, that certainly could make sense to me.

lib/src/facts/linux/os_linux.cc
ScottGarman: @MikaelSmith it's Debian 8 PowerPC but they are using sysv instead of systemd. There is an /etc/os-release file which has the details of Debian 8 in it.

ScottGarman:
The OS has update-rc.d and it looks like the Debian sysv provider is the right one. I'll work on getting os.family set to Debian shortly.

kylog:
@ScottGarman does HuaweiOS include a package manager? And if so, is it apt or something custom?

ScottGarman:
@kylog HuaweiOS has apt and dpkg. When we eventually release it, we'll be creating a repo for this platform on apt.puppetlabs.com.

MikaelSmith:
This seems ok. Sounds like a change in Puppet may be necessary?

MikaelSmith:
If you want it in the next release, it needs to target stable.

ScottGarman:
Ok, I will re-target this for stable and create an new PR. What Puppet changes do you think will be necessary? We can simply document the need to specify 'debian' for service management for the time being. Huawei's main use case for the module is to issue netconf commands to the switch via their module.

MikaelSmith:
@ScottGarman I meant it should have an acceptance test similar to https://github.com/puppetlabs/facter/blob/master/acceptance/tests/facts/debian.rb.

ScottGarman:
Closing this PR in favor of PR-1285 which is targeted at stable

PR #1285, opened 2016-03-08
(FACT-1362) Add platform identification for HuaweiOS

ScottGarman:
HuaweiOS runs as an LXC container based on Debian 8 with some
modifications (for example, it uses Debian sysv instead of
systemd). It still uses apt and dpkg for package management.

The firmware version is a long string stored in
/etc/huawei-release which does not follow semver, but looks
like "V100R006C00SPC300B188", so we just pass along this
version information without further parsing.

So the os.name we're using is "HuaweiOS", and os.family is
"Debian"

Confirmed on the target that the output from 'facter os' is:
```
{
  architecture => "ppc",
  family => "Debian",
  hardware => "ppc",
  name => "HuaweiOS",
  release => {
    full => "V100R006C00SPC300B188",
    major => "V100R006C00SPC300B188"
  },
  selinux => {
    enabled => false
  }
}
```

### DIFF
--- a/lib/inc/facter/facts/os.hpp
+++ b/lib/inc/facter/facts/os.hpp
@@ -185,6 +185,10 @@ namespace facter { namespace facts {
          * The AristaEOS operating system.
          */
         constexpr static char const* arista_eos = "AristaEOS";
+       /**
+        * The HuaweiOS operating system.
+        */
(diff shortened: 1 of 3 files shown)


stahnma:
:+1: LGTM

PR #1286, opened 2016-03-10
(FACT-1276) Make external facts dir absolute instead of canonicalizing

branan:
It's not necessary to get the full canonical path for external facts
dir - we only need it to be absolute. Attempting to canonicalize paths
is more expensive, and can fail on Windows when boost::filesystem is
built to support Windows 2003, as we do.

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -125,9 +125,9 @@ namespace facter { namespace facts {
         // If dir is relative, make it an absolute path before passing to can_resolve.
         bool found = false;
         boost::system::error_code ec;
-        path search_dir = canonical(dir, ec);
+        path search_dir = absolute(dir);
 
-        if (ec || !is_directory(search_dir, ec)) {
+        if (!is_directory(search_dir, ec)) {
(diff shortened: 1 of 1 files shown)


kylog:
> when boost::filesystem is built to support Windows 2003, as we do

I thought we had dropped the 2003-supporting builds? I may have confused intention and reality (or branches). /cc @MikaelSmith

branan:
In that case, another fix for this is to build boost::filesystem with `-D_WIN32_WINNT=0x0600`

MikaelSmith:
Ah, possible but a little more involved. Is there value to canonicalizing the path?

branan:
Regardless of the windows situation, though, canonicalization can be a very expensive operation (lots of stat calls) vs just creating an absolute path and letting the kernel do it's job to follow symlinks

branan:
@MikaelSmith I don't think there's any value to canonicalizing in this case

MikaelSmith:
I'll add a ticket for using `_WIN32_WINNT=0x0600` for Windows builds; that's unrelated to this change then.

PR #1287, opened 2016-03-15
(FACT-1339) Prevent swap warnings on OpenBSD.

peterhuene:
When no swap device is configured, prevents from showing the following message on OpenBSD:
```
swapctl: no swap devices configured
```


peterhuene:
@branan Here's that fixed up PR.

PR #1289, opened 2016-03-16
(FACT-1345) Exclude automatic mounts on solaris

branan:
Solaris has a concept of "automounts" that are used to transparently
map chunks of the filesystem to other filesystems, or across the
network. The number of automounted paths can become very large when a
number of zones are in use.

This patch excludes automounted paths from the `mountpoints` fact, so
that the fact can stay focused on mounts that an admin might actually
care about.


branan:
I want to hold off on merging this until we get confirmation that the exclusions look correct to somebody who knows what they're doing, but otherwise this is ready for review

MikaelSmith:
This looks good as an implementation. :+1:

PR #1290, opened 2016-03-18
(FACT-1372) IPv6 are incorrectly parsed.

sathlan:
The regexps matching IPv6 is incorrect.  This fix them all.

### DIFF
--- a/spec/unit/interfaces_spec.rb
+++ b/spec/unit/interfaces_spec.rb
@@ -41,6 +41,16 @@ describe "Per Interface IP facts" do
     fact = Facter.fact("macaddress_net0".intern)
     fact.value.should eq("00:12:34:56:78:90")
   end
+
+  it "should support ipv6 address starting by [180ef]" do
+    Facter.fact(:kernel).stubs(:value).returns("Linux")
+    Facter::Util::IP.stubs(:exec_ifconfig) \
+      .returns(my_fixture_read('ifconfig_net_tools_1.60_v6.txt'))
+    Facter.collection.internal_loader.load(:interfaces)
+    Facter.fact('ipaddress6_em1'.intern).value \
+      .should eq('1610:10:20:209:212:3fff:febe:2201')
+  end
+
 end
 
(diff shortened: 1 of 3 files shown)


MikaelSmith:
Spec tests demonstrating the broken behavior is fixed would be very appropriate for this change.

HAIL9000:
Hey @sathlan, just wanted to check in to see about signing the CLA and potentially updating the spec tests to go with this change. Let us know if you have questions about either of those.

sathlan:
Hi @HAIL9000,

I've signed the CLA, and added the spec.  Without the patch it returns the new test fails with nil returned instead of the ipv6's value.

MikaelSmith:
👍 Thanks for the update, that looks great.

spec/unit/interfaces_spec.rb
```diff
@@ -41,6 +41,16 @@
     fact = Facter.fact("macaddress_net0".intern)
     fact.value.should eq("00:12:34:56:78:90")
   end
+
+  it "should support ipv6 address starting by [180ef]" do
+    Facter.fact(:kernel).stubs(:value).returns("Linux")
+    Facter::Util::IP.stubs(:exec_ifconfig)
+      .returns(my_fixture_read('ifconfig_net_tools_1.60_v6.txt'))
```
kylog: Apparently ruby 1.8.7 needs a line continuation character (`\`) at the end of line 47 (or lines 47 and 48 to be collapsed to one line).

spec/unit/interfaces_spec.rb
```diff
@@ -41,6 +41,16 @@
...
+    Facter.collection.internal_loader.load(:interfaces)
+    Facter.fact("ipaddress6_em1".intern).value
+      .should eq('1610:10:20:209:212:3fff:febe:2201')
```
kylog: Same for lines 50 and 51.

sathlan:
Thanks @kylog for the debug.  ruby 1.8.7, tell me about legacy :)

PR #1291, opened 2016-03-18
(FACT-1294) Use relative path in facter.rb

MikaelSmith:
Change `facter.rb` to always use a relative path to find libfacter, based
on the relative path of the install location of facter.rb and
CMAKE_INSTALL_PREFIX. This fixes libfacter lookup when doing `make
install` on a Windows machine, and removes hard-coding the MSI layout.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -307,13 +307,8 @@ symbol_exports(libfacter "${CMAKE_CURRENT_LIST_DIR}/inc/facter/export.h")
 leatherman_install(libfacter)
 install(DIRECTORY inc/facter DESTINATION include)
 
-# Always generate facter.rb and spec_helper.rb, as they might be used in packaging
+# Always generate spec_helper.rb, as they might be used in packaging
 # and testing on another machine.
-configure_file (
-    "facter.rb.in"
-    "${CMAKE_BINARY_DIR}/lib/facter.rb"
-)
-
 # Generate a file for ruby testing
 configure_file (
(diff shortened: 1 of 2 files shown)


MikaelSmith:
This requires a change to puppet-agent builds on Windows to set the CMAKE_INSTALL_PREFIX appropriately.

lib/CMakeLists.txt
```diff
@@ -343,6 +338,12 @@ else()
 endif()
 
 if(RUBY_VENDORDIR)
+    file(RELATIVE_PATH LIBFACTER_INSTALL_RELATIVE ${RUBY_VENDORDIR} ${CMAKE_INSTALL_PREFIX})
```
branan: Do you need to quote `RUBY_VENDORDIR` and `CMAKE_INSTALL_PREFIX` here?

MikaelSmith:
Did Windows and el-7 builds, both look good. `puppet facts` and `irb 'facter'` work correctly.

lib/CMakeLists.txt
MikaelSmith: We didn't other places, so I was hoping not. I'll try a run installing to "C:/Program Files" just to be sure.

branan:
:+1:

lib/CMakeLists.txt
MikaelSmith: Testing "C:/Program Files" worked fine.

MikaelSmith:
PA-252 was merged, this should be ready to go.

PR #1292, opened 2016-03-24
(FACT-1375) Only recognize default route with netmask of 0.0.0.0

jearls:
Without this patch, Facter will recognize any route with a destination
of 0.0.0.0 as a default route, even if that route has a non-0 netmask.
The default route always has a netmask of 0.0.0.0, and should be the
one chosen to determine the primary network interface.

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -81,7 +81,7 @@ namespace facter { namespace facts { namespace linux {
         lth_file::each_line("/proc/net/route", [&interface](string& line) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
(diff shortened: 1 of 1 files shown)


lib/src/facts/linux/networking_resolver.cc
```diff
@@ -81,7 +81,8 @@ namespace facter { namespace facts { namespace linux {
         lth_file::each_line("/proc/net/route", [&interface](string& line) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
-            if (parts.size() > 1 && parts[1] == boost::as_literal("00000000")) {
+            if (parts.size() > 1 && parts[1] == boost::as_literal("00000000")
```
peterhuene: The size check needs to be updated in case the split didn't have 8 parts.

jearls: Whoops, fixed.  Sorry about that; it's been a very long time since I've done non-scripting-language programming.

PR #1295, opened 2016-04-01
(FACT-1385) Fix FixNum interpretation in Ruby 2.2

MikaelSmith:
64-bit Ruby 2.2 on Windows returns a 32-bit integer coerced to a intptr_t
from rb_num2long. Switch to using rb_num2ll instead to get a true 64-bit
integer back.

Ruby's behavior was probably changed to implement the actual `long` type,
which is 32-bit on Windows and 64-bit on other platforms. Our Ruby API
layer still expects it to return a 64-bit type, and doesn't know any
better.

### DIFF
--- a/lib/src/ruby/ruby_value.cc
+++ b/lib/src/ruby/ruby_value.cc
@@ -87,7 +87,7 @@ namespace facter { namespace ruby {
             return;
         }
         if (ruby.is_fixednum(value)) {
-            json.SetInt64(ruby.rb_num2long(value));
+            json.SetInt64(ruby.rb_num2ll(value));
             return;
         }
         if (ruby.is_float(value)) {
@@ -156,7 +156,7 @@ namespace facter { namespace ruby {
             return;
         }
         if (ruby.is_fixednum(value)) {
-            os << ruby.rb_num2long(value);
+            os << ruby.rb_num2ll(value);
             return;
         }
         if (ruby.is_float(value)) {
@@ -242,7 +242,7 @@ namespace facter { namespace ruby {
             return;
         }
         if (ruby.is_fixednum(value)) {
-            emitter << ruby.rb_num2long(value);
+            emitter << ruby.rb_num2ll(value);
             return;
         }
         if (ruby.is_float(value)) {


MikaelSmith:
(╯°□°）╯︵ ┻━┻ /cc @peterhuene

lib/src/ruby/ruby_value.cc
```diff
@@ -87,7 +87,10 @@ namespace facter { namespace ruby {
             return;
         }
         if (ruby.is_fixednum(value)) {
-            json.SetInt64(ruby.rb_num2long(value));
+            // On Windows 64-bit with Ruby 2.2, rb_num2long returns the wrong value for negative numbers.
+            // FixedNums can only represent a word (32-bit) so explicitly cast to an int.
+            // TODO: Handle BigNum.
+            json.SetInt64(static_cast<int>(ruby.rb_num2long(value)));
```
peterhuene: Instead of casting to int, perhaps using `rb_num2ll` instead of `rb_num2long` would help?

MikaelSmith: I'll give it a try. The behavior in general seems weird.

peterhuene: Looking at the impl, this function can handle big nums too.

MikaelSmith: Yes, we just skip calling it for BigNums. That'll be the fix for another FACT ticket.

MikaelSmith: rb_num2ll seems to work. I'll update Leatherman, and this will have to be pending a Leatherman release.

MikaelSmith:
rb_num2long changed to returning a long at some point. https://github.com/ruby/ruby/blob/trunk/include/ruby/ruby.h#L650

MikaelSmith:
Depends on https://github.com/puppetlabs/leatherman/pull/154.

MikaelSmith:
This might need to target stable if we need other Leatherman updates on stable.

MikaelSmith:
Covered by https://github.com/puppetlabs/facter/pull/1310.

PR #1297, opened 2016-04-07
(FACT-1296) Avoid transient unit test failure

MikaelSmith:
We've seen transient unit test failures due to the test fixture that
creates and deletes a directory. Add a loop check to ensure the
directory is deleted (with a 5-second timeout) to avoid continuing until
previous tests are cleaned up.

### DIFF
--- a/lib/tests/fixtures.cc
+++ b/lib/tests/fixtures.cc
@@ -4,6 +4,12 @@
 #include <iostream>
 #include <sstream>
 
+#pragma GCC diagnostic push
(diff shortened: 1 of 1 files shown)


kylog:
It looks like mingw wants the same treatment you provided in https://github.com/puppetlabs/cpp-pcp-client/commit/f478d7dcc4e41deae63d1de1bf777a2db0c7617a.

MikaelSmith:
Updated.

MikaelSmith:
Woo, @puppetcla party!

PR #1300, opened 2016-04-09
(FACT-1387) Make Google Windows VMs work as 'is_virtual'

leepa:
As noted in the subject - the manufacturer check was missed for Windows
Google VMs and the result is that you cannot use any more the meta-data
as is_virtual is set to False.


leepa:
So the travis CI fails because a 'bug' has to exist and be referenced!?

MikaelSmith:
Thanks for the contribution!

PR #1301, opened 2016-04-12
(FACT-1043) Fix virtual fact on Openstack Windows instances

jasminen:
(FACT-1043) Falsely reporting virtual as physical under Windows for OpenStack instance

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -215,6 +215,8 @@ Facter.add("virtual") do
           result = "kvm"
         when /Bochs/
           result = "bochs"
+        when /OpenStack/
+          result="openstack"
         end
 
         if result.nil? and computersystem.manufacturer =~ /Xen/


ferventcoder:
@MikaelSmith Thoughts on this guy?

ferventcoder:
@jasminen I'm not sure we are still releasing a 2.x line of Facter. Does this fix still need to be applied in the master/stable branches?

ferventcoder:
@puppetlabs/windows

MikaelSmith:
We still release 2.x. It should also be applied to master or stable. I'll follow up on it. This seems inconsistent with other platforms, so I want to see what they report on OpenStack infrastructure.

Iristyle:
IMHO, this seems like the wrong spot for this fact to be addressed, given OpenStack runs atop an underlying hypervisor like `xen`, `kvm`, `hyperv`, or `esxi`.  OpenStack should be a higher level abstraction like Azure, Google, AWS, etc.

glennsarti:
It seems that OpenStack (via Nova) is showing the virtual BIOS as 'Openstack Nova'.  A quick search found similar issues with DMIDecode on Linux based systems as well.  Which may mean this fix should also be applied at;

https://github.com/jasminen/facter/blob/2.x/lib/facter/virtual.rb#L167

jasminen:
@glennsarti We run a very big Openstack environment (using Nova and KVM) while most instances are running Linux, and we do not have this issue on Linux based systems. 
The value we get is 'kvm' following this line: https://github.com/jasminen/facter/blob/2.x/lib/facter/virtual.rb#L163
and even if it won't find it there, the DMIDecode returns Bochs which is also included:
https://github.com/jasminen/facter/blob/2.x/lib/facter/virtual.rb#L177

@Iristyle I agree but like stated in the bug report: "This is down to OpenStack setting the System Manufacturer and System Model to its own values. This also results in the ec2 metadata not being fetched on Windows instances."
If you know of any other way to get the hypervisor from these Windows instances i will change the fix, but i couldn't find any other way.

jasminen:
Output from powershell:
 
    PS > Get-WmiObject -Class Win32_ComputerSystem
    
    
    Domain              : ######
    Manufacturer        : OpenStack Foundation
    Model               : OpenStack Nova
    Name                : ######
    PrimaryOwnerName    : ######
    TotalPhysicalMemory : 6442033152

MikaelSmith:
That's strange configuration, but I don't have a problem with this otherwise.

lib/facter/virtual.rb
```diff
@@ -215,6 +215,8 @@
           result = "kvm"
         when /Bochs/
           result = "bochs"
+        when /OpenStack/
+          result="openstack"
```
kylog: Can we make the whitespace consistent with the surrounding code?

kylog:
Yeah no problem with this functionally. Thanks for the contribution!

MikaelSmith:
Cleaned up in https://github.com/puppetlabs/facter/pull/1315.

PR #1303, opened 2016-04-12
(FACT-1379) Query original hardware address for bonded interfaces

branan:
The list of interfaces on Linux contains the bonded mac, but previous
version of Facter reported the original hardware mac in this
case. This restores the behavior of querying the system for bonded
devices and retrieving the original mac.

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -24,6 +24,35 @@ namespace facter { namespace facts { namespace linux {
         read_routing_table();
         data result = bsd::networking_resolver::collect_data(facts);
         populate_from_routing_table(result);
+
+        // On linux, the macaddress of bonded interfaces is reported
+        // as the address of the bonding master. We want to report the
+        // original HW address, so we dig it out of /proc
+        for (auto& interface : result.interfaces) {
+            // For each interface we check if we're part of a bond,
+            // and update the `macaddress` fact if we are
+            auto bond_master = get_bond_master(interface.name);
+            if (!bond_master.empty()) {
+                bool in_our_block = false;
+                lth_file::each_line("/proc/net/bonding/"+bond_master, [&](string& line) {
+                    // /proc/net/bonding files are organized into chunks for each slave
+                    // interface. We want to grab the mac address for the block we're in.
+                    if (line == "Slave Interface: " + interface.name) {
+                        in_our_block = true;
+                    } else if (line.find("Slave Interface") != string::npos) {
+                        in_our_block = false;
+                    }
+
+                    // If we're in the block for our iface, we can grab the HW address
+                    if (in_our_block && line.find("Permanent HW addr: ") != string::npos) {
+                        auto split = line.find(':') + 2;
+                        interface.macaddress = line.substr(split, string::npos);
+                        return false;
+                    }
+                    return true;
+                });
+            }
+        }
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


lib/src/facts/linux/networking_resolver.cc
```diff
@@ -184,4 +213,43 @@ namespace facter { namespace facts { namespace linux {
             }
         }
     }
+
+    string networking_resolver::get_bond_master(const std::string& name) const {
+        static bool have_logged_about_bonding = false;
+        auto ip_command = lth_exe::which("ip");
+        if (ip_command.empty()) {
+            if (!have_logged_about_bonding) {
+                 LOG_DEBUG("Could not find the 'ip' command. Physical macaddressed for bonded interfaces will be incorrect.");
```
MikaelSmith: Was `macaddressed` intentional?

branan: typo. I'll fix

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -184,4 +213,43 @@ namespace facter { namespace facts { namespace linux {
...
+            }
+            return {};
+        }
+
+        string bonding_master;
+
+        lth_exe::each_line(ip_command, {"link", "show", name}, [&bonding_master](string& line) {
+            if (line.find("SLAVE") != string::npos) {
+                // Look for the word "master", and strip everything before that
+                auto master_tag_location = line.find("master");
```
MikaelSmith: Could also be a regex_search, but this seems fine.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -24,6 +24,35 @@ namespace facter { namespace facts { namespace linux {
...
+                    if (line == "Slave Interface: " + interface.name) {
+                        in_our_block = true;
+                    } else if (line.find("Slave Interface") != string::npos) {
+                        in_our_block = false;
+                    }
+
+                    // If we're in the block for our iface, we can grab the HW address
+                    if (in_our_block && line.find("Permanent HW addr: ") != string::npos) {
+                        auto split = line.find(':') + 2;
+                        interface.macaddress = line.substr(split, string::npos);
```
MikaelSmith: This is messy. But seems to do what we want.

MikaelSmith:
I guess 👍

lib/src/facts/linux/networking_resolver.cc
branan: Yeah. It's this or a regex, and I've *tried* to move away from using regex parsers all over facter in Facter 3

branan: I guess I could use the boost tokenizer, rather than searching the string myself. We're already using that throughout facter

branan:
Typo fix pushed up. I think I want to try swapping in the tokenizer to see if that makes the code more legible before we merge this, though

branan:
Updated to use the boost tokenizer for parsing `ip` output, but not for parsing the /proc/net/bonding file. The gymnastics to use the tokenizer weren't worth it for the simple "string after a colon" case, but definitely simplified the IP out code

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -184,4 +213,45 @@ namespace facter { namespace facts { namespace linux {
...
+        string bonding_master;
+
+        lth_exe::each_line(ip_command, {"link", "show", name}, [&bonding_master](string& line) {
+            if (line.find("SLAVE") != string::npos) {
+                vector<boost::iterator_range<string::iterator>> parts;
+                boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+                // We have to use find_if here since a boost::iterator_range doesn't compare properly to a string.
+                auto master = find_if(parts.begin(), parts.end(), [](boost::iterator_range<string::iterator>& part){
+                    string p {part.begin(), part.end()};
```
MikaelSmith: I suspect `boost::algorithm::equals(part, "master")` could handle this.

branan: I tried a few variations on that and couldn't get it to work - I think a `const char*` is *just* different enough from a real iterator range to break `equals`, despite pointers being nominally "iterators"

PR #1304, opened 2016-04-13
(FACT-1390) Add static node config for huawei

LuvCurves:


acceptance/config/nodes/huaweios-6-powerpc.yaml
```diff
@@ -0,0 +1,20 @@
+---
+HOSTS:
+  master:
```
johnduarte: @LuvCurves there should not be a master node in the configs for facter.

acceptance/config/nodes/huaweios-6-powerpc.yaml
```diff
@@ -0,0 +1,14 @@
+---
+HOSTS:
+  huaweios-6-powerpc:
+    roles:
+      - agent
+    platform: huaweios-6-powerpc
+    hypervisor: none
+    ip: 10.0.7.26
+    vmhostname: huawei-ce6850-2-debian-vm-eth0.ops.puppetlabs.net
+CONFIG:
```
johnduarte: @LuvCurves since hypervisor is 'none', I presume that the pooler CONFIG options are not required. If so, this section can be removed.

LuvCurves:
@johnduarte Fixed

LuvCurves:
@johnduarte Fixed again

acceptance/config/nodes/huaweios-6-powerpc.yaml
MikaelSmith: Agreed, looks like CONFIG doesn't need to be there. Look at the other configs for comparable examples.

acceptance/config/nodes/huaweios-6-powerpc.yaml
```diff
@@ -0,0 +1,13 @@
+---
+HOSTS:
+  huaweios-6-powerpc:
+    roles:
+      - agent
+    platform: huaweios-6-powerpc
+    hypervisor: none
+    ip: huawei-ce6850-2-debian-vm-eth0.ops.puppetlabs.net
```
johnduarte: The above is incorrect. The ip is set to the hostname. I believe it should look like the following:
```
ip: 10.0.7.26
vmhostname: huawei-ce6850-2-debian-vm-eth0.ops.puppetlabs.net
```

acceptance/config/nodes/huaweios-6-powerpc.yaml
```diff
@@ -0,0 +1,13 @@
...
+CONFIG:
```
johnduarte: This line and all below it can be removed.

johnduarte:
:+1:

PR #1305, opened 2016-04-13
(FACT-1389) Fix acceptance test on systems with Ruby installed

MikaelSmith:
Previously the path to find Ruby was setup with system paths first, so
we would find any system Ruby installed. That caused the test to fail
when that Ruby is 1.8.7. Fix to find the puppet-agent Ruby instead.

[skip ci]


kylog:
Makes sense. Don't need travis/appveyor since this is in acceptance.

PR #1307, opened 2016-04-14
(FACT-1280) Enable setting aio_agent_version during compilation

MikaelSmith:
Adds a new CMake option to create an aio_agent_version fact. This allows
AIO builds to set the package version during compilation.


MikaelSmith:
Added to schema.

PR #1308, opened 2016-04-14
(FACT-1392) Avoid locale failures in corrupted environments

MikaelSmith:
On puppet-dev, someone observed issues running Facter when all locale
variables were set to UTF-8 but their environment had no locale files
for UTF-8. Facter's fallback to LANG=C appeared to be insufficient.

Add a stronger fallback of LC_ALL=C, as done in
https://github.com/puppetlabs/puppet/blob/4.4.1/lib/puppet/util/execution.rb#L282-L286,
to ensure only the C locale is used.

### DIFF
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -54,10 +54,13 @@ namespace facter { namespace logging {
         try {
             setup_logging_internal(os);
         } catch (exception const&) {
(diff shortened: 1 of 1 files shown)


kylog:
Did you reproduce this config and verify that adding `LC_ALL=C` fixes it?

Regardless, this makes sense to me.

MikaelSmith:
Nope, couldn't figure out how to reproduce the failure. And not sure how to contact the original commenter.

PR #1309, opened 2016-04-15
(FACT-1392) Restore logging after logging is setup

MikaelSmith:
Prior change for FACT-1392 mistakenly moved logging to before logging
can occur, because it hasn't been initialized. Restore it to occur
after.

### DIFF
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -54,8 +54,6 @@ namespace facter { namespace logging {
         try {
             setup_logging_internal(os);
         } catch (exception const&) {
-            log(level::warning, "locale environment variables were bad; continuing with LANG=C LC_ALL=C");
-
             for (auto var : lc_vars) {
                 environment::clear(var);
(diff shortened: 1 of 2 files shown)


lib/src/logging/logging.cc
```diff
@@ -71,8 +69,12 @@ namespace facter { namespace logging {
                 // Since logging is busted, we raise an exception that
                 // signals to the consumer that a special action must
                 // be taken to alert the user.
-                throw locale_error(e.what());
+                throw locale_error(string("could not initialize logging, even with locale variables reset to LANG=C LC_ALL=C: ")
```
branan: This loses any potential information that e.what() may have. Are we sure there's nothing potentially useful there?

branan: wait, next line. I'm dumb

kylog:
Oh right. I think I may have suggested moving the line about logging to before logging was setup Doh!

PR #1310, opened 2016-04-18
(FACT-1385) (FACT-1364) BigNum and FixedNum fix ups on Windows 

HAIL9000:
This pull request addresses two Facter tickets. The first is FACT-1385 which
involved updating Facter so that our tests which relied on the rb_num2long
returning a long did not fail on Windows with Ruby 2.2.

The second is FACT-1364, which involved updating the printing methods
in Facter to recognize integers larger than 64-bits and display them
properly.

Both these changes rely on the new release of Leatherman.

### DIFF
--- a/lib/src/ruby/chunk.cc
+++ b/lib/src/ruby/chunk.cc
@@ -60,7 +60,7 @@ namespace facter { namespace ruby {
                     ruby.rb_gc_register_address(&values[0]);
                 } else if (ruby.is_array(_dependencies)) {
                     // Resize the vector now to ensure it is fully allocated before registering with GC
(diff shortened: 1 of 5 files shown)


HAIL9000:
Suspecting tests will fail for now because the Leatherman version in Facter needs to be bumped first.

MikaelSmith:
The comment in the 1st commit isn't quite accurate where it says `This fix also relies on an update to leatherman which adds the ability to recognize bignums to the ruby API.` It relies on an update to Leatherman to get num2size_t and rb_num2ll.

HAIL9000:
Closing this to retarget at stable

PR #1311, opened 2016-04-18
(FACT-1385) (FACT-1364) BigNum and FixedNum fix ups on Windows

HAIL9000:
This pull request addresses two Facter tickets. The first is FACT-1385 which
involved updating Facter so that our tests which relied on the rb_num2long
returning a long did not fail on Windows with Ruby 2.2.

The second is FACT-1364, which involved updating the printing methods
in Facter to recognize integers larger than 64-bits and display them
properly.

Both these changes rely on the new release of Leatherman.


MikaelSmith:
We messed up tagging Leatherman 0.5.0, so you'll need to use 0.5.1.

branan:
:+1:

MikaelSmith:
I'm ok to merge this now so Branan can continue some other work using new Leatherman, and add tests in a follow-on PR.

PR #1312, opened 2016-04-18
(FACT-1395) Use Windows API for Windows version

Iristyle:
- GetVersionEx was historically the way to do this, except that with
   Windows 8.1 it was deprecated.  Depending on how an application is
   built, GetVersionEx may return an unexpected version - for instance
   it returns '6.2.9200' on a Windows 10 machine, when the previous
   WMI query returns '10.0.10586'

   Instead use the kernel call RtlGetVersion from ntoskrnl.exe, which
   returns exactly the same structure as GetVersionEx does, but will
   return the correct versions of the kernel.

- Reimplement the kernelrelease and OS facts so that they no longer use
  WMI. To support Windows 2003 R2, use the function GetSystemMetrics
  which can return a special value when on that OS.

- Add the current Nano build number (though a more appropriate way of
  detecting Nano is probably more appropriate).

- This implementation shaves off about 400ms+ of runtime in some micro
  benchmarks.

### DIFF
--- a/lib/facter/kernelrelease.rb
+++ b/lib/facter/kernelrelease.rb
@@ -11,6 +11,7 @@
 # Caveats:
 #
 require 'facter/util/posix'
+require 'facter/util/windows'
 
 Facter.add(:kernelrelease) do
   setcode 'uname -r'
@@ -39,12 +40,6 @@ end
 Facter.add(:kernelrelease) do
   confine :kernel => "windows"
   setcode do
-    require 'facter/util/wmi'
-    version = ""
-    Facter::Util::WMI.execquery("SELECT Version from Win32_OperatingSystem").each do |ole|
-      version = "#{ole.Version}"
-      break
-    end
-    version
+    Facter::Util::Windows::Process.os_version_string
   end
 end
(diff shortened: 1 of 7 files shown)


MikaelSmith:
This looks reasonable. @branan @kylog we should consider pulling this into Facter 3 as well.

branan:
It's maybe a moot question now that the work is done, but the ticket says that this is for Nano Server support - are we actually planning to target Nano Server with Facter 2 / PE 3.8?

ferventcoder:
LGTM

ferventcoder:
Should we have this run against appveyor?

Iristyle:
@branan I specifically PR'd against Facter 2 to support dev workflows - i.e. git cloning the source, and running the Puppet spec suite.  In production, we of course would run Facter 3.

Iristyle:
I've also filed https://tickets.puppetlabs.com/browse/FACT-1415 in case we want to do the same in Facter 3.

kylog:
Merged. Note that there's no scheduled release in the 2.x series at this time. It will probably happen at some point, but nothing scheduled. If that becomes a blocker at some point, we can schedule a release. And of course, if you want an MSI as opposed to a gem, that probably means a puppet 3.x release as well.

Iristyle:
@kylog that's fine - can always modify Gemfile / Facter env var to point to a commit-ish if needed.

We're still in the early days here of evaluating what is needed for Nano, and FYI - MSIs don't work / exist on Nano.

PR #1314, opened 2016-04-19
(FACT-1284) Set ruby $? variable from Facter execution bindings

branan:
This is based on #1311 because my leatherman changes in leatherman#160 are built on top of lth 0.5. Once 1311 is merged I'll rebase.

This needs my Leatherman PR to be merged and released before it will build, because it relies on changes over there.


lib/tests/fixtures/ruby/uses_exit_code.rb
```diff
@@ -0,0 +1,6 @@
+Facter.add(:foo) do
+    setcode do
+      Facter::Core::Execution.execute("sh -c 'exit 99'");
```
MikaelSmith: Is this a safe test on Windows?

lib/tests/fixtures/ruby/uses_exit_code.rb
```diff
@@ -0,0 +1,7 @@
+Facter.add(:foo) do
+  setcode do
+    cmd = Gem.win_platform? ? 'cmd /k "exit 99"' : 'sh -c "exit 99"'
```
branan: This seemed like the most expedient way to find out if we were windows-y, but if there's another quick way I'm happy to tweak this line. the 'Gem' module exists in all rubies 1.9+

MikaelSmith: I mean, this is Facter, but I think I've used that test before and feel comfortable relying on it.

PR #1317, opened 2016-04-20
(FACT-1364) Add test for int value larger than 32-bits

HAIL9000:
To ensure that facts with bignum integer values continue to work on
windows, add a test.


HAIL9000:
This was meant to be targeted at stable, closing to reopen.

PR #1318, opened 2016-04-20
(FACT-1364) Add test for int value larger than 32-bits

HAIL9000:
To ensure that facts with bignum integer values continue to work on
windows, add a test.


MikaelSmith:
Seems like it should work.

HAIL9000:
I made sure this failed on Windows with my fix commented out, so hopefully it should work!

PR #1319, opened 2016-04-20
(FACT-1399) Fix build error on AIX due to Leatherman changes

MikaelSmith:
Leatherman dropped pulling in lots of dependencies when building with
shared libraries. On AIX, this results in failing to build tests because
they depend on Boost.Thread, and we didn't include that dependency
explicitly in Facter. Add it now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,8 +65,8 @@ find_package(Leatherman REQUIRED COMPONENTS ${LEATHERMAN_COMPONENTS})
 # we need for finding all our other libraries.
 include(options)
 
-# We use program_options, system, filesystem, date_time, and regex directly.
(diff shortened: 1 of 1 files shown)


underscorgan:
👍

MikaelSmith:
This may also need a Leatherman change.

branan:
I thought we weren't building the tests on AIX? Or do we build them and just not run them?

MikaelSmith:
I think the latter.

MikaelSmith:
I'm not happy about it, but I think https://github.com/puppetlabs/leatherman/pull/164 is necessary as well. Going to kick off an AIX test build with two PRs.

branan:
I'm :+1:, since we're building the tests

PR #1320, opened 2016-04-22
(maint) Fix building Facter with Leatherman DLLs

MikaelSmith:
On Windows, building Facter with Leatherman shared libraries leaves an
unresolved symbol from chrono. Add the chrono library dependency for
testing.


MikaelSmith:
@kylog @branan this is needed to start using Leatherman shared libraries on Windows for PCP-372.

PR #1321, opened 2016-04-26
(FACT-1402) check also upper case letter in virtual kvm lspci check

chiemseesurfer:
If virtual = kvm should be detected, the virtual facter just triggers for lower case and is not able to catch up for example "00:03.0 Ethernet controller: Red Hat, Inc Virtio network device"

### DIFF
--- a/lib/src/facts/linux/virtualization_resolver.cc
+++ b/lib/src/facts/linux/virtualization_resolver.cc
@@ -237,7 +237,7 @@ namespace facter { namespace facts { namespace linux {
             make_tuple(boost::regex("XenSource"),                     string(vm::xen_hardware)),
             make_tuple(boost::regex("Microsoft Corporation Hyper-V"), string(vm::hyperv)),
             make_tuple(boost::regex("Class 8007: Google, Inc"),       string(vm::gce)),
-            make_tuple(boost::regex("virtio", boost::regex::icase),   string(vm::kvm)),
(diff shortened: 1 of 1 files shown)


lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -237,7 +237,7 @@ namespace facter { namespace facts { namespace linux {
             make_tuple(boost::regex("XenSource"),                     string(vm::xen_hardware)),
             make_tuple(boost::regex("Microsoft Corporation Hyper-V"), string(vm::hyperv)),
             make_tuple(boost::regex("Class 8007: Google, Inc"),       string(vm::gce)),
-            make_tuple(boost::regex("virtio", boost::regex::icase),   string(vm::kvm)),
+            make_tuple(boost::regex("[Vv]irtio", boost::regex::icase),   string(vm::kvm)),
```
kylog: As with #1322, the `boost::regex::icase` on the regex means case-insensitive, so this shouldn't be needed. Can you provide an input string that doesn't match with the current code?

kylog:
@chiemseesurfer did you get a chance to look at my regex comment above?

kylog:
@chiemseesurfer I'm going to close this for now, but feel free to re-open this (if it's still an issue, see my comments above). Thanks for the contribution!

chiemseesurfer:
Hey,

like the part on the 2.x branch. I was not able to reproduce the problem, but I'm pretty sure, there was one :)

If I'm able to reproduce it again, I'll come back and reopen this ticket.

PR #1322, opened 2016-04-26
(FACT-1402) check also upper case letter in virtual kvm lspci check

chiemseesurfer:
If virtual = kvm should be detected, the virtual facter just triggers for lower case and is not able to catch up for example "00:03.0 Ethernet controller: Red Hat, Inc Virtio network device"

### DIFF
--- a/lib/facter/virtual.rb
+++ b/lib/facter/virtual.rb
@@ -160,7 +160,7 @@ Facter.add("virtual") do
       next "xenhvm"     if lines.any? {|l| l =~ /XenSource/ }
       next "hyperv"     if lines.any? {|l| l =~ /Microsoft Corporation Hyper-V/ }
       next "gce"        if lines.any? {|l| l =~ /Class 8007: Google, Inc/ }
-      next "kvm"        if lines.any? {|l| l =~ /virtio/i }
+      next "kvm"        if lines.any? {|l| l =~ /[Vv]irtio/i }
     end
 
     # Parse dmidecode


lib/facter/virtual.rb
```diff
@@ -160,7 +160,7 @@
       next "xenhvm"     if lines.any? {|l| l =~ /XenSource/ }
       next "hyperv"     if lines.any? {|l| l =~ /Microsoft Corporation Hyper-V/ }
       next "gce"        if lines.any? {|l| l =~ /Class 8007: Google, Inc/ }
-      next "kvm"        if lines.any? {|l| l =~ /virtio/i }
+      next "kvm"        if lines.any? {|l| l =~ /[Vv]irtio/i }
```
kylog: Hmm, the `/i` on the regex means case-insensitive, so this shouldn't be needed. Can you provide an input string that doesn't match with the current code?

kylog:
@chiemseesurfer did you get a chance to look at my regex comment above?

kylog:
@chiemseesurfer I'm going to close this for now, but feel free to re-open this (if it's still an issue, see my comments above). Thanks for the contribution!

chiemseesurfer:
Hey, sure...
sorry, was not able to reproduce it. I thought I was using version 2.2, but there the kvm part is not there...
And with some other 2.x versions I was not able to reproduce it the last weeks...
So, It's fine that you closed the ticket. Sure the /i means case-insensitive.

PR #1326, opened 2016-05-06
(FACT-1411) Regex issue with util/netmask.rb in 2.x

mattjbarker:
Probably a better way to fix this, but figured I'd bring attention to it. Thanks!

### DIFF
--- a/lib/facter/util/netmask.rb
+++ b/lib/facter/util/netmask.rb
@@ -8,7 +8,7 @@ module Facter::NetMask
     when 'Linux'
       ops = {
         :ifconfig_opts => ['2>/dev/null'],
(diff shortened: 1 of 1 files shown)


branan:
This makes sense to me based on your example in the ticket. Changing fact values always scares me, so I'm gonna chat with some other folks at Puppet before I hit the button, but this will probably make it in.

MikaelSmith:
ping @branan, did you get a chance to follow-up with others?

MosesMendoza:
Given the 2.x series of facter is EOL (or maintenance mode at best) I don't believe we can accept this change, as folks may be relying on the existing behavior. I'm going to close this PR. If you come across this issue in the facter 3 series, can you reopen the JIRA issue? Thanks

PR #1328, opened 2016-05-09
(FACT-1373) Prevent fact calls from external facts

HAIL9000:
Prior to this commit, it was possible to make a facter call from
within an external fact. This would result in a fork bomb where
facter would be constantly calling into itself. In order to prevent
this, set an environment variable that allows us to ensure facter
is not already running when we make the call to evaluate external
facts.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -2,6 +2,7 @@
 #include <facter/logging/logging.hpp>
 #include <facter/facts/collection.hpp>
 #include <facter/ruby/ruby.hpp>
+#include <leatherman/util/environment.hpp>
 #include <leatherman/util/scope_exit.hpp>
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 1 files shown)


exe/facter.cc
```diff
@@ -251,7 +255,15 @@ int main(int argc, char **argv)
         facts.add_default_facts(ruby);
 
         if (!vm.count("no-external-facts")) {
+          string inside_facter;
+          environment::get("INSIDE_FACTER", inside_facter);
+
+          if (inside_facter == "true") {
+            LOG_WARNING("Facter was called recursively, skipping external facts. Add '--no-external-facts' to silence this warning");
```
MikaelSmith: It might be worth mentioning that we detected this because INSIDE_FACTER was set to true, in case (somehow) someone accidentally has that in their environment.

PR #1330, opened 2016-05-12
DO NOT MERGE (PE-14501) add superuser flag to the identity fact

mruzicka:
This commit adds the `superuser` flag as another attribute of the `identity` structured fact.

It is a boolean flag which is set to `true` if the facter process runs with UID of 0 on *NIX systems or with the privileges of the local Administrators group on Windows. Under any other circumstances the flag is set to `false`.

### DIFF
--- a/lib/inc/internal/facts/resolvers/identity_resolver.hpp
+++ b/lib/inc/internal/facts/resolvers/identity_resolver.hpp
@@ -51,6 +51,14 @@ namespace facter { namespace facts { namespace resolvers {
              * Stores the name of the user's primary group.
              */
             std::string group_name;
+
+            /**
+             * Stores whether the user is a superuser (a user
+             * with the UID of 0 on *NIX systems or a member
+             * of the local Administrators group on Windows)
+             * or not.
+             */
+            boost::optional<bool> superuser;
         };
 
         /**
--- a/lib/src/facts/windows/identity_resolver.cc
+++ b/lib/src/facts/windows/identity_resolver.cc
@@ -1,5 +1,6 @@
 #include <internal/facts/windows/identity_resolver.hpp>
 #include <leatherman/windows/system_error.hpp>
+#include <leatherman/windows/user.hpp>
 #include <leatherman/windows/windows.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/nowide/convert.hpp>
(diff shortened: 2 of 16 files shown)


lib/src/facts/windows/identity_resolver.cc
```diff
@@ -20,20 +20,37 @@ namespace facter { namespace facts { namespace windows {
...
+            LOG_DEBUG("failure resolving identity facts: %1% (%2%)", leatherman::windows::system_error(), GetLastError());
             return result;
         }
 
         // Resize the buffer to the returned string size.
         buffer.resize(size);
         result.user_name = boost::nowide::narrow(buffer);
+
+        // Build the well known local Administrators group SID
+        PSID local_administrators_sid;
```
MikaelSmith: There's an is_admin helper in Leatherman. I'm pretty sure it does the same thing. https://github.com/puppetlabs/leatherman/blob/master/windows/src/user.cc#L14-L45

mruzicka: oh, cool

lib/inc/internal/facts/resolvers/identity_resolver.hpp
```diff
@@ -51,6 +51,14 @@ namespace facter { namespace facts { namespace resolvers {
...
              */
             std::string group_name;
+
+            /**
+             * Stores whether the user is a superuser (a user
+             * with the UID of 0 on *NIX systems or a member
+             * of the local Administrators group on Windows)
+             * or not
+             */
+            bool superuser;
```
MikaelSmith: You'll also need code in https://github.com/puppetlabs/facter/blob/master/lib/src/facts/resolvers/identity_resolver.cc#L22-L45 to add it to the fact collection.

lib/src/facts/windows/identity_resolver.cc
```diff
@@ -20,20 +20,37 @@ namespace facter { namespace facts { namespace windows {
         auto nameformat = NameSamCompatible;
         GetUserNameExW(nameformat, nullptr, &size);
         if (GetLastError() != ERROR_MORE_DATA) {
-            LOG_DEBUG("failure resolving identity facts: %1% (%2%)", leatherman::windows::system_error());
+            LOG_DEBUG("failure resolving identity facts: %1% (%2%)", leatherman::windows::system_error(), GetLastError());
```
mruzicka: OK

lib/inc/internal/facts/resolvers/identity_resolver.hpp
mruzicka: right

MikaelSmith:
This seems pretty good. We'll want to change the PE ticket to a FACT ticket however (or open a new one).

kylog:
Another nit: in addition to referencing a FACT ticket, can you put the commit messages in the imperative? (Or squash them to the first one, which is in the imperative.) Thanks @mruzicka !

mruzicka:
I'm going to create the new FACT ticket and re-submit the changes in this PR in a new PR.

mruzicka:
Closing in favor of https://github.com/puppetlabs/facter/pull/1332.

PR #1332, opened 2016-05-12
(FACT-1419) add privileged element to identity fact

mruzicka:
This commit adds the `privileged` flag as another element of the `identity` structured fact.

It is a boolean flag which is set to `true` if the facter process runs with UID of 0 on *NIX systems or with the privileges of the local Administrators group on Windows. Under any other circumstances the flag is set to `false`.

### DIFF
--- a/acceptance/tests/facts/aix.rb
+++ b/acceptance/tests/facts/aix.rb
@@ -82,10 +82,11 @@ agents.each do |agent|
 
   step "Ensure the identity fact resolves as expected"
   expected_identity = {
-                        'identity.gid'   => '0',
-                        'identity.group' => 'system',
-                        'identity.uid'   => '0',
-                        'identity.user'  => 'root'
+                        'identity.gid'        => '0',
+                        'identity.group'      => 'system',
+                        'identity.uid'        => '0',
+                        'identity.user'       => 'root',
+                        'identity.privileged' => true
                       }
 
   expected_identity.each do |fact, value|
--- a/acceptance/tests/facts/debian.rb
+++ b/acceptance/tests/facts/debian.rb
@@ -102,10 +102,11 @@ agents.each do |agent|
 
   step "Ensure the identity fact resolves as expected"
   expected_identity = {
-                        'identity.gid'   => '0',
-                        'identity.group' => 'root',
-                        'identity.uid'   => '0',
-                        'identity.user'  => 'root'
+                        'identity.gid'        => '0',
+                        'identity.group'      => 'root',
+                        'identity.uid'        => '0',
+                        'identity.user'       => 'root',
+                        'identity.privileged' => true
                       }
 
   expected_identity.each do |fact, value|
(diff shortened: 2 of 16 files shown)


MikaelSmith:
👍  lgtm if it passes CI.

MikaelSmith:
@branan thoughts on this as a new fact?

branan:
I'm fine with adding this. Superuser is easy on *nix, but on Windows you really have to query the system, so we should provide that information.

I'd like docs or ux feedback on "superuser" vs "administrator" vs. whatever else we might be able to come up with

MikaelSmith:
@jtappa we're looking at adding/naming a new core fact here, you want to weigh in (or point us to someone you think should)?

jtappa:
I'm sure @nfagerlund would be interested. I tend to lean more towards administrator, myself, but if we document it for users I think either term is fine.

nfagerlund:
I think I agree with Jorie on both counts: Administrator sounds better (slightly more general), but I'm not really against superuser.

mruzicka:
@jtappa @nfagerlund I deliberately tried to avoid the administrator, as it feels (at least to me) quite Windows specific, whereas the implemented fact extension is cross platform. If that doesn't feel like that to you I'm happy to rename to the administrator.

nfagerlund:
Hmm. I don't really come from Windows land; would it make sense to talk about a "superuser" there? I feel like in either case, one of the platforms will have to just read the description or think for a moment.

joshcooper:
On Windows Vista and up, a user may be a member of the Administrators group, but not running in an elevated context. So they effectively have the same privileges as a normal user, and need to elevate privileges via UAC (kind of like sudo).

Note 2003 behaves differently, but we no longer support that platform, so its behavior can be ignored.

I assume the new fact is reporting whether the process is running elevated (or not)? As opposed the user is a member of the Administrators group?

Perhaps call it "privileged"?

mruzicka:
@joshcooper it uses https://github.com/puppetlabs/leatherman/blob/master/windows/src/user.cc#L14-L45 to determine the value of flag on Windows, i.e. it takes the privileges elevation status into account

+1 on the ```privileged```

mruzicka:
@joshcooper what about service processes on Windows? Does the privileges elevation apply to them too?

MikaelSmith:
I like `privileged` as well. That privilege can come from being granted it via elevation, or running as a user that implicitly has privilege (root or Administrator/SYSTEM).

joshcooper:
> what about service processes on Windows? Does the privileges elevation apply to them too?

I believe that code works correctly for service processes too. In the process explorer example below, `ruby.exe` is running as a service as localsystem, and it has the "Mandatory Label\System Mandatory level" integrity level showing that it's running at the highest level:

screen shot 2016-05-16 at 3 40 44 pm

And the code should detect that it's running elevated.

mruzicka:
ok, thanks @joshcooper
@jtappa @nfagerlund what are your thoughts on naming the flag ```privileged```?

jtappa:
`privileged` is fine with me.

nfagerlund:
I'm fine with "privileged" too!

mruzicka:
Thanks for the feedback, I'm going to update the PR.

mruzicka:
I've updated the PR to use the name `privileged`.

MikaelSmith:
👍

PR #1333, opened 2016-05-16
(FACT-1409) fallback to no locales if initialization fails

branan:
Previously, we would assume that a failed locale initialization meant
that the sky was falling. Unfortunately, it looks like some vendors
have removed locale support from containers as way to make them
lighter-weight. We need to handle that case and proceed with locale
support.

### DIFF
--- a/lib/src/logging/logging.cc
+++ b/lib/src/logging/logging.cc
@@ -8,16 +8,22 @@ using namespace std;
 using namespace leatherman::util;
 namespace lm = leatherman::logging;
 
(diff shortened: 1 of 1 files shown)


branan:
This requires puppetlabs/leatherman#178 to be merged and released before it will build

MikaelSmith:
Oh right, this'll need a Leatherman release.

branan:
This still needs lth 0.7.0 in puppet-agent before it can be merged, but is passing travis now and ready for review

PR #1334, opened 2016-05-17
(FACT-1421) Add NO_BATCH_FILES option to cmake

McdonaldSeanp:
puppet-agent carries the batch files for all executables with the new move to
vanagon for windows projects, we will need an option to omit batch file
creation during installation of facter so there are not more than one set of
batch files. So this commit adds a NO_BATCH_FILES option for cmake that omits
those files.

The change has a double negative in order to preserve expected behavior if the
option is not specified


kylog:
Can you move this ticket to FACT? Facter is an open-source project so we try to reference public jira projects. Thanks!!

McdonaldSeanp:
👍 @kylog

McdonaldSeanp:
I created https://tickets.puppetlabs.com/browse/FACT-1421 as the ticket for this

MikaelSmith:
The double-negative isn't necessary, you can use `option` to specify a default. Like
```
option(INSTALL_BATCH_FILES "Install batch files on Windows" ON)
```

McdonaldSeanp:
Yeah that makes more sense, updated

PR #1335, opened 2016-05-18
(FACT-1419) Use string comparison for `privileged` in tests

MikaelSmith:
The `identity` fact output is grabbed by running `facter` and grabbing
stdout, so we'll always be comparing against a string. Change to
expecting a string rather than a boolean.


whopper:
👍 verified that this gets Facter passing again on redhat 7. @MikaelSmith I see you `[skip ci]`'d, so cool to ignore the yellow dots and merge?

PR #1336, opened 2016-05-18
(FACT-1423) Update Ubuntu fact test with new platform names

whopper:
Ubuntu beaker platform names no longer use codenames, and
instead just use the version.

In addition, this commit adds test support for Ubuntu
15.10 and 16.04.

### DIFF
--- a/acceptance/tests/facts/ubuntu.rb
+++ b/acceptance/tests/facts/ubuntu.rb
@@ -1,35 +1,45 @@
-test_name "Facts should resolve as expected in Ubuntu 12.04, 14.04, and 14.10"
+test_name "Facts should resolve as expected in Ubuntu"
 
 #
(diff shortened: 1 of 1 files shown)


acceptance/tests/facts/ubuntu.rb
```diff
@@ -1,35 +1,43 @@
...
     os_kernel  = '3.19'
+  elsif agent['platform'] =~ /ubuntu-15.10/
+    os_name    = 'wily'
+    os_version = '15.10'
+    os_kernel  = '4.2'
+  elsif agent['platform'] =~ /ubuntu-16.04/
+    os_name    = 'xenial'
+    os_version = '16.04'
+    os_kernel  = '4.4'
   end
```
MikaelSmith: Probably add `else fail`.

MikaelSmith:
👍

PR #1337, opened 2016-05-18
(maint) bump travis/appveyor lth version to 0.7

branan:


lib/CMakeLists.txt
```diff
@@ -302,6 +302,7 @@ target_link_libraries(libfacter PRIVATE
     ${CMAKE_THREAD_LIBS_INIT}
 )
 
+symbol_exports(libfactersrc "${CMAKE_CURRENT_LIST_DIR}/inc/facter/export.h")
```
MikaelSmith: Ah, does Leatherman need to change for this to work?

branan: Gonna find out as soon as appveyor hits it

branan: I'm also not *entirely* certain that's the cause of our problem, but it seems to be something that approximates "GCC is assuming declarations are dllimport, but definitions aren't". I actually can't find `dllimport` or `dllexport` anywhere in facter or leatherman, so I have no idea how our export stuff works on Windows.

branan:
@MikaelSmith I've had no luck trying to add the export helper macro to other targets to get this running on Windows. Adding it to just the `libfactersrc` target resulted in the same error as before, and it doesn't work to add it to `libfacter_test` because it only works on library targets. Do you have any thoughts?

lib/tests/CMakeLists.txt
```diff
@@ -126,7 +126,9 @@ target_link_libraries(libfacter_test
     ${LEATHERMAN_LIBRARIES}
     ${CURL_LIBRARIES}
     ${CMAKE_THREAD_LIBS_INIT}
-)
+    )
```
MikaelSmith: No longer matches other formatting...

branan: I'm blame my editor. Fixed.

MikaelSmith:
👍

PR #1340, opened 2016-05-20
(FACT-1428) Fixes FreeBSD fact for processor count

petems:
* FreeBSD is missing a lot of core facts
* I'm no C++ guy, but I can sure steal existing knowledge 💃 
* This is the beginning of adding back a lot of core facts
* Added instructions for FreeBSD dev

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -190,11 +190,12 @@ elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
     )
 elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "FreeBSD")
     set(LIBFACTER_PLATFORM_SOURCES
-        "src/facts/bsd/collection.cc"
+        "src/facts/freebsd/collection.cc"
         "src/facts/glib/load_average_resolver.cc"
         "src/facts/bsd/filesystem_resolver.cc"
         "src/facts/bsd/networking_resolver.cc"
         "src/facts/bsd/uptime_resolver.cc"
+        "src/facts/freebsd/processor_resolver.cc"
         "src/util/bsd/scoped_ifaddrs.cc"
     )
 elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "OpenBSD")
--- a/lib/src/facts/freebsd/processor_resolver.cc
+++ b/lib/src/facts/freebsd/processor_resolver.cc
@@ -1,0 +1,63 @@
+#include <internal/facts/freebsd/processor_resolver.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <sys/types.h>
+#include <sys/sysctl.h>
+
+using namespace std;
+
+namespace facter { namespace facts { namespace freebsd {
+
+    processor_resolver::data processor_resolver::collect_data(collection& facts)
+    {
+        auto result = posix::processor_resolver::collect_data(facts);
+        size_t len;
+        int mib[2];
+        mib[0] = CTL_HW;
+
+        // Get the logical count of processors
+        len = sizeof(result.logical_count);
+        mib[1] = HW_NCPU;
+
+        if (sysctl(mib, 2, &result.logical_count, &len, nullptr, 0) == -1) {
+            LOG_DEBUG("sysctl hw.ncpu failed: %1% (%2%): logical processor count is unavailable.", strerror(errno), errno);
+        }
+
+        // For each logical processor, collect the model name
+        if (result.logical_count > 0) {
+            // Note: we're using the same description string for all logical processors:
+            // using different CPUs is not even likely to work.
+            vector<char> buffer(256);
+
+            while (true) {
+                size_t size = buffer.size();
+                mib[1] = HW_MODEL;
+                if (sysctl(mib, 2, buffer.data(), &size, nullptr, 0) == 0) {
+                    buffer.resize(size + 1);
+                    result.models.resize(result.logical_count, buffer.data());
+                    break;
+                }
+                if (errno != ENOMEM) {
+                    LOG_DEBUG("sysctl hw.model failed: %1% (%2%): processor models are unavailable.", strerror(errno), errno);
+                    break;
+                }
+                buffer.resize(buffer.size() * 2);
+            }
+        }
+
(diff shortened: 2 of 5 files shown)


lib/src/facts/freebsd/processor_resolver.cc
```diff
@@ -0,0 +1,50 @@
...
+                    result.models.resize(result.logical_count, buffer.data());
+                    break;
+                }
+                if (errno != ENOMEM) {
+                    LOG_DEBUG("sysctl hw.model failed: %1% (%2%): processor models are unavailable.", strerror(errno), errno);
+                    break;
+                }
+                buffer.resize(buffer.size() * 2);
+            }
+        }
```
petems: This code is the same as OpenBSD, so it could probably be moved into a bsd helper and included.

However, processor speed is different. Here's what I found in a FreeBSD portfile:

```c++
inline double getMHZ_x86()
{
   int mhz = -1;
   size_t mhz_sz = sizeof(mhz);
#if defined(__FreeBSD__)
   int cmd[2];
   size_t two = 2;
   sysctlnametomib("hw.clockrate", cmd, &two);
#elif defined(__OpenBSD__)
   int cmd[] = {CTL_HW, HW_CPUSPEED};
#endif
   sysctl(cmd, sizeof(cmd)/sizeof(int), &mhz, &mhz_sz, NULL, 0);
   return (double)mhz;
}
#endif // FreeBSD || OpenBSD
```

If someone can help me adapt this I would be happy 👍

lib/CMakeLists.txt
```diff
@@ -190,11 +190,12 @@ elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "Linux")
     )
 elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "FreeBSD")
     set(LIBFACTER_PLATFORM_SOURCES
-        "src/facts/bsd/collection.cc"
```
MikaelSmith: This file appears to no longer be used. Seems like we could remove it (i.e. we're renaming it to `freebsd/collection`).

petems: I think it would still theoritically be used for DragonflyBSD, NetBSD or other BSDs if people were to build Facter 3.X? Should I still remove?

lib/src/facts/freebsd/processor_resolver.cc
petems: Ok, thanks to some help from Johnny B, who showed me that `sysctlnametomib` just gives you the ref to give to sysctl:

```c++
// Set the speed
        len = sizeof(result.speed);

        int cmd[2];
        size_t two = 2;
        sysctlnametomib("hw.clockrate", cmd, &two);

        if (sysctl(cmd, 2, &result.speed, &len, nullptr, 0) == -1) {
            LOG_DEBUG("sysctl hw.cpuspeed failed: %1% (%2%): processor speed is unavailable.", strerror(errno), errno);
        }
        // Scale the speed to something resolve() can correctly map
        result.speed *= 1000 * 1000;

        return result;
```

jasperla: Also note that OpenBSD doesn't have `sysctlnametomib`. So if this code were to be moved to a more generic class, that's something to keep in mind as well.

petems: Yeah, I hit my head against the wall with this for a while. It looks like FreeBSD's [kenv](https://www.freebsd.org/cgi/man.cgi?kenv(1)) method seems like the way to go:

```c
#ifdef ENABLE_LLDPMED
	/* Fill in inventory stuff:
	    - hardware version: smbios.system.version
	    - firmware version: smbios.bios.version
	    - software version: `uname -r`
	    - serial number: smbios.system.serial
	    - manufacturer: smbios.system.maker
	    - model: smbios.system.product
	    - asset: smbios.chassis.tag
	*/

static char*
dmi_get(char *file)
{
	char buffer[100] = {};

	log_debug("localchassis", "DMI request for %s", file);
	if (kenv(KENV_GET, file, buffer, sizeof(buffer) - 1) == -1) {
		log_debug("localchassis", "cannot get %s", file);
		return NULL;
	}
	if (strlen(buffer))
		return strdup(buffer);
	return NULL;
}

char*
dmi_sn()
{
	return dmi_get("smbios.system.serial");
}
```

jasperla: Well OpenBSD obviously has `sysctl`, that's how basically all the OpenBSD facts are implemented. Can't you just use the plain MIB in a fashion similar to https://github.com/puppetlabs/facter/blob/master/lib/src/facts/openbsd/dmi_resolver.cc#L10 ?

petems: The FreeBSD doesn't seem to have 1:1 mapping with OpenBSD, the only constants(Are the all caps references to sysctl stuff called constants) I can find in the docs are:

```
           HW_MACHINE		      string	    no
	   HW_MODEL		      string	    no
	   HW_NCPU		      integer	    no
	   HW_BYTEORDER		      integer	    no
	   HW_PHYSMEM		      integer	    no
	   HW_USERMEM		      integer	    no
	   HW_PAGESIZE		      integer	    no
	   HW_FLOATINGPT	      integer	    no
	   HW_MACHINE_ARCH	      string	    no
	   HW_REALMEM		      integer	    no
```

For other settings, all the references and code I can find use the kenv lookup:

```
bios_release_date = kenv -q smbios.bios.reldate
bios_vendor = kenv -q smbios.bios.vendor
bios_version = kenv -q smbios.bios.version
boardmanufacturer = kenv -q smbios.planar.maker
boardproductname = kenv -q smbios.planar.product
boardserialnumber = kenv -q smbios.planar.serial
serialnumber = kenv -q smbios.system.serial
productname = kenv -q smbios.system.product
manufacturer = kenv -q smbios.system.maker
uuid = kenv -q smbios.system.uuid
```

```
root@freebsd-10-facter3:~ # sysctl smbios.bios.reldate
sysctl: unknown oid 'smbios.bios.reldate': No such file or directory
root@freebsd-10-facter3:~ # kenv smbios.bios.reldate
12/01/2006
```

MikaelSmith:
👍  This looks reasonable.

PR #1341, opened 2016-05-20
(FACT-1408) Use prtdiag for productname on Solaris

HAIL9000:
Prior to this commit, facter was using prtconf to resolve the
productname and manufacturer facts on Solaris SPARC machines. We
made this change because it was believed that prtdiag was not present
on all SPARC boxes, but as far as we can tell that doesn't seem to be
the case.

Using prtconf gives a less specific fact value and doesn't work on
some Solaris machines, so we are falling back to using prtdiag since
that is what was used in Facter 2 and seems to produce better results.

### DIFF
--- a/lib/src/facts/solaris/dmi_resolver.cc
+++ b/lib/src/facts/solaris/dmi_resolver.cc
@@ -67,16 +67,9 @@ namespace facter { namespace facts { namespace solaris {
                 return result.chassis_type.empty() || result.chassis_asset_tag.empty();
             });
         } else if (isa && isa->value() == "sparc") {
-            static boost::regex manufacturer_re("^System Configuration: (.+) sun\\d.$");
-            static boost::regex product_name_re("^SUNW,(.*)$");
-            // prtdiag is not implemented in all sparc machines, so we cant get product name this way.
-            each_line("/usr/sbin/prtconf", [&](string& line) {
-                if (result.manufacturer.empty()) {
-                    re_search(line, manufacturer_re, &result.manufacturer);
-                }
-                if (result.product_name.empty()) {
-                    re_search(line, product_name_re, &result.product_name);
-                }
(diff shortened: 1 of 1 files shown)


lib/src/facts/solaris/dmi_resolver.cc
```diff
@@ -67,16 +67,10 @@ namespace facter { namespace facts { namespace solaris {
                 return result.chassis_type.empty() || result.chassis_asset_tag.empty();
             });
         } else if (isa && isa->value() == "sparc") {
-            static boost::regex manufacturer_re("^System Configuration: (.+) sun\\d.$");
-            static boost::regex product_name_re("^SUNW,(.*)$");
+            static boost::regex product_name_manufacturer_re("^System Configuration:\\s+(.+?)\\s+sun\\d+\\S+\\s+(.+)");
             // prtdiag is not implemented in all sparc machines, so we cant get product name this way.
```
kylog: Toss this comment now? Or modify?

HAIL9000: Good point... seems like we should say something but I'm not sure what's most helpful

kylog: My $.02 is to just remove the comment as it's no longer applicable (and thus might confuse the reader).

branan: :+1: to that

HAIL9000:
@branan @kylog Updated but not sure why appveyor would be failing now

branan:
@HAIL9000 weird. You can ammend the commit with no new changes to get an updated timestamp/sha, and force-push that. It might kick it?

HAIL9000:
@branan Cool thanks! Giving that a shot

HAIL9000:
All great now! 👍

PR #1342, opened 2016-05-30
(FACT-1431) Enable curl on OpenBSD

jasperla:
No need to disable curl on OpenBSD when it builds fine.


MikaelSmith:
Curl is enabled only for Amazon EC2 and GCE. Are there OpenBSD images for those?

jasperla:
[Yep!](http://blog.d2-si.fr/2016/02/15/openbsd-on-aws/)

PR #1346, opened 2016-06-02
(MAINT) mention setup instructions on OpenBSD

jasperla:


README.md
```diff
@@ -112,6 +112,15 @@ In Powershell:
 
 Note that OpenSSL isn't needed on Windows.
 
+### Setup on OpenBSD
```
kylog: Should this specify anything about which version of OpenBSD?

jasperla: There isn't a really strict requirement from the base system, rather the ports/packages have to be up to date. And base/packages out of sync is no go. So `OpenBSD 6.0` would be more accurate.

HAIL9000: Hey @jasperla, could you update these docs with the version so they're more accurate?

jasperla:
Pardon the delay, it's properly annotated with a version now.

whopper:
Thanks @jasperla!

PR #1352, opened 2016-06-13
(FACT-1438) Fix tests on s390x

johnduarte:
Fix `acceptance/tests/facts/el.rb` to accommodate s390x instances.

[skip-ci]

### DIFF
--- a/acceptance/tests/facts/el.rb
+++ b/acceptance/tests/facts/el.rb
@@ -10,6 +10,8 @@ test_name "Facts should resolve as expected in EL 4, 5, 6 and 7"
 # Don't test on RedHat 4, as is does not include yum (See BKR-512)
 confine :to, :platform => /el-[5-7]|centos-[4-7]/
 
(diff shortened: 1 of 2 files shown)


johnduarte:
This change has been successfully run against the following platforms:

- redhat-6-s390x
- centos-6-i386
- scientific-6-i386
- oracle-6-i386

acceptance/tests/facts/el.rb
```diff
@@ -10,6 +10,8 @@
 # Don't test on RedHat 4, as is does not include yum (See BKR-512)
 confine :to, :platform => /el-[5-7]|centos-[4-7]/
 
+@ip_regex = /^([1-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])){3}$/
```
MikaelSmith: Why not go with something like https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9780596802837/ch07s16.html? Or are you specifically disallowing `0.0.0.0` (as an example).

johnduarte: Thanks for the suggestion @MikaelSmith . I have updated the regex in the el.rb file and the previously committed sles.rb file.

MikaelSmith:
Works for me.

PR #1353, opened 2016-06-14
(FACT-1244) Correctly report Xen Dom0 vs DomU

branan:
Previously, our virt-what reader would return the first line of output
it found. For cases where the machine is Xen, however, this first line
is simple "xen", with Dom0 or DomU information being provided
after. In that case we want to skip the first line and move on to the
detailed hypervisor info.

### DIFF
--- a/lib/src/facts/linux/virtualization_resolver.cc
+++ b/lib/src/facts/linux/virtualization_resolver.cc
@@ -117,6 +117,11 @@ namespace facter { namespace facts { namespace linux {
                 return true;
             }
             // Take the first line that isn't an error/warning
+            // unless it's "xen", in which case we expect a second
+            // line with more useful information
+            if (line == "xen") {
+                return true;
+            }
             value = move(line);
             return false;
         });


MikaelSmith:
So this didn't lead to a generalized approach to virt-what parsing?

lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -117,6 +117,11 @@ namespace facter { namespace facts { namespace linux {
                 return true;
             }
             // Take the first line that isn't an error/warning
+            // unless it's "xen", in which case we expect a second
+            // line with more useful information
```
kylog: Is that always the case? Or is the output of virt-what ever just one line with `xen`?

branan: Not as far as I can tell. Running as a nonpriviliged user spits out "xen0" but not "xen" on the hypervisor, so I think we're safe here.

I could have it latch 'xen' into the value field to be on the safe side. Then if somehow we don't get the detailed line we'll still have *an* answer

kylog: After re-reading the ticket and the virt-what man page, I agree that we're safe here, so let's not add YAGNI complexity to the code.

PR #1354, opened 2016-06-15
FACT-1439 Facter detecting only enabled network card

adeleglise:
Changing the way facter list NIC on Windows. Get rid of the NetEnabled statement so that every NIC are listed, even the disabled.
modified :         lib/facter/util/ip/windows.rb

### DIFF
--- a/lib/facter/util/ip/windows.rb
+++ b/lib/facter/util/ip/windows.rb
@@ -43,7 +43,7 @@ class Facter::Util::IP::Windows
     interface_names = []
 
     network_adapter_configurations.map do |nic|
-      Facter::Util::WMI.execquery("SELECT * FROM Win32_NetworkAdapter WHERE Index = #{nic.Index} AND NetEnabled = TRUE").each do |sub_nic|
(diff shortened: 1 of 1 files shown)


branan:
Hi, thanks for 

The change that added the NetEnabled check was intended to [https://projects.puppetlabs.com/issues/22619](fix a bug), and merged in #537  I think in order to effectively list disabled interfaces, additional logic would have to be added elsewhere to facter to handle the cases that ticket was trying to correct for. @ferventcoder do you have any thoughts here?

Beyond that, facter 2 is essentially in "extended support" at this point. We try not to make changes to it that could negatively impact the users who rely on older versions of software for stability. Can you explain a bit more about what your use case is for having facter include the extra information, so we can determine if it's something that's worth bringing in to the old code?

adeleglise:
Sorry, for the delay the summer was very busy.

I'm just relaying this contribution. I think that this is a very specific need, and that this could be helpful for someone else.

However, all the travis builds failed because the spec were not modified accordingly. So I think that we can just put this contribution aside. 

Thanks for coming back.

branan:
OK. I'm going to go ahead and close this, then.

Thanks for reaching out to us, and please don't hesitate with other contributions in the future.

PR #1356, opened 2016-06-21
(FACT-1421) Add NO_BATCH_FILES option to cmake

McdonaldSeanp:
puppet-agent carries the batch files for all executables with the new move to
vanagon for windows projects, we will need an option to omit batch file
creation during installation of facter so there are not more than one set of
batch files. So this commit adds a NO_BATCH_FILES option for cmake that omits
those files.

This is a cherry-pick of https://github.com/puppetlabs/facter/pull/1334 to stable


geoffnichols:
👍

MikaelSmith:
👍

PR #1358, opened 2016-06-23
(FACT-1442) Fix regex used to validate windows username in acceptance tests

glennsarti:
Previously the windows acceptance test was expecting that the user is always
cyg_server, however this is not always true and can cause test failures even
when facter is operating correctly.  This commit changes the regular
expression used to validate the username to conform to Windows username
standards.

### DIFF
--- a/acceptance/tests/facts/windows.rb
+++ b/acceptance/tests/facts/windows.rb
@@ -88,8 +88,13 @@ agents.each do |agent|
   end
 
   step "Ensure the identity fact resolves as expected"
+  # Regular expression to validate the username from facter in the form of '<domain>\<username>'
+  # Reference - https://msdn.microsoft.com/en-us/library/bb726984.aspx
+  # - The domain name can be any character or empty
+  # - Must contain a backslash between the domain and username
+  # - Username must be at least one character and not contain the following charaters; " / \ [ ] : ; | = , + * ? < >
   expected_identity = {
-                        'identity.user'       => /.*\\cyg_server/,
+                        'identity.user'       => /.*\\[^\\\/\"\[\]:|<>+=;,?*@]+$/,
                         'identity.privileged' => 'true'
                       }
 


glennsarti:
I have verified this regular expression works by running beaker locally

acceptance/tests/facts/windows.rb
```diff
@@ -88,8 +88,13 @@
...
 
   step "Ensure the identity fact resolves as expected"
+  # Regular expression to validate the username from facter in the form of '<domain>\<username>'
+  # Reference - https://msdn.microsoft.com/en-us/library/bb726984.aspx
+  # - The domain name can be any character or empty
+  # - Must contain a backslash between the domain and username
+  # - Username must be at least one character and not contain the following charaters; " / \ [ ] : ; | = , + * ? < >
   expected_identity = {
-                        'identity.user'       => /.*\\cyg_server/,
+                        'identity.user'       => /.*\\[^\\\/\"\[\]:|<>+=;,?*@]+$/,
```
Iristyle: One thing you could do is call `net user` on the remote machine to get the list of accounts available.

Then you could ensure that the identity returns is one of those accounts.

We're almost to the point of this test not being super valuable... any comments from anyone else?

kylog:
Do we want this on stable?

glennsarti:
I did consider this, but we don't test Windows 10 in stable so it wasn't an issue.

acceptance/tests/facts/windows.rb
MikaelSmith: Noting it returns a value for the fact is still valuable; it'd be even better if we could verify it's the value we expect.

glennsarti: Given that the pipelines are red, I'd like to get this PR merged and then raise a ticket for changing the test to actually verify the content of the value coming from facter.

Iristyle: My bigger point was that `.*` is too lenient - sorry, I probably should have been more clear about that.  

You also match strings like:

* `foo\bar\baz` 
* `\\foo\\bar`
* `123!<>!@#$%^\\foo`

At least before we at least had `cyg_server` in there, so the `.*`, while it could accept junk, wasn't as bad.  Now the regex accepts a lot more stuff.

MikaelSmith: I'm ok with leaving it for a later ticket to match the actual value. It's pretty unlikely we'll get a value, but miss something because that regex was too broad.

PR #1360, opened 2016-06-23
(FACT-1395) Fix FFI load of ntoskrnl.exe

Iristyle:
- The ffi_lib call is failing under certain circumstances in testing
   - in particular, a failure has been observed on Windows 2008 x64
   running 32-bit Ruby 1.9.3 with Facter 1.9.5.

   Attempts to reproduce the problem have been unsuccessful, but in
   such cases, the error reported is:

   Could not open library 'ntoskrnl.exe':
   The specified module could not be found. (LoadError)

 - FFI supports a special loading mechanism that will access a
   currently loaded module.  Since Ruby always loads `ntoskrnl.exe`,
   this should remove any issues in FFI around finding modules based
   on searching load path, and should always use the correct
   bitness automatically.

### DIFF
--- a/lib/facter/util/windows/process.rb
+++ b/lib/facter/util/windows/process.rb
@@ -273,7 +273,7 @@ module Facter::Util::Windows::Process
   # NTSTATUS RtlGetVersion(
   #   _Out_ PRTL_OSVERSIONINFOW lpVersionInformation
   # );
-  ffi_lib 'ntoskrnl.exe'
+  ffi_lib [FFI::CURRENT_PROCESS, 'ntoskrnl.exe']
   attach_function :RtlGetVersion, [:pointer], :int32
 
   # C++ int is a signed 32-bit integer


Iristyle:
I still don't entirely understand why it can't find `ntoskrnl.exe` in some cases, but not sure it's worth chasing down given the fix here is short / sensible.

Iristyle:
@whopper / @MikaelSmith I believe that https://jenkins.puppetlabs.com/view/legacy_Facter/view/2.x/ verifies this as valid, particularly run #19 at https://jenkins.puppetlabs.com/view/legacy_Facter/view/2.x/job/legacy_facter_unit-windows_2.x/

kylog:
Funky. But thanks for running this down @Iristyle !!

Iristyle:
Note that the commit msg was slightly off - it reads `Facter 1.9.5` when it should read `FFI 1.9.5`

kylog:
Ah I missed that :(

PR #1368, opened 2016-06-30
(FACT-1449) Fix YAML external facts to respect string types.

peterhuene:
Facter was converting quoted scalars in YAML external fact files that could be
converted to integers, doubles, and booleans.  For example, 'true', '123', and
'123.456' would convert to boolean, integer, and double respectively, despite
being quoted in the YAML.

This fix makes use of a node's tag to determine if the scalar value in the YAML
was quoted (has a tag value of "!").  If the scalar was quoted, Facter does not
attempt to do a conversion and respects the fact that it is a string.


MikaelSmith:
👍

PR #1369, opened 2016-07-01
(FACT-1451) Use regex to match kernel versions in ubuntu fact test

whopper:
Previously, the Ubuntu kernel version test expected specific,
hardcoded kernel versions for each Ubuntu release. This caused
failures when our Ubuntu VM templates were updated, which caused
kernel version bumps. This commit updates the test to use a more
relaxed regex which only expects a specific major kernel version.

[skip ci]


whopper:
Successfully ran this against 12.04, 14.04, 15.10, and 16.04

Magisus:
Seems legit to me 👍

PR #1372, opened 2016-07-07
(FACT-1405) Fix network route parsing when "onlink" flag is present

ahenroid:


lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,12 @@ namespace facter { namespace facts { namespace linux {
         auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+            // remove terminating "onlink" flag
+            std::string last_token(parts.back().begin(), parts.back().end());
+            if (last_token == "onlink")
```
branan: We should probably also look for "pervasive" (according to the manpage http://www.dsm.fordham.edu/cgi-bin/man-cgi.pl?topic=ip-route&ampsect=8, see the NHFLAGS line)

Other than that, I'm not sure if there are any other peculiarities of parsing the route format. It's probably not worth writing a full-blown parser unless we get more reports of issues with this simple one, though

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,17 @@ namespace facter { namespace facts { namespace linux {
         auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+            // remove trailing "onlink" or "pervasive" flags
+            for (;;) {
+                std::string last_token(parts.back().begin(),
+                                       parts.back().end());
+                if (last_token == "onlink" || last_token == "pervasive")
+                    parts = decltype(parts)(parts.begin(), parts.end() - 1);
```
branan: Can we just use `parts.erase(parts.end()-1)` here to avoid copying the entire list?

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,17 @@ namespace facter { namespace facts { namespace linux {
         auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+            // remove trailing "onlink" or "pervasive" flags
+            for (;;) {
+                std::string last_token(parts.back().begin(),
```
branan: I don't think we need to make a copy here, since we don't reference it after we do the erasure. `auto& last_token = parts.back()` should be fine, and will be cleaner code to boot

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,17 @@ namespace facter { namespace facts { namespace linux {
         auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+            // remove trailing "onlink" or "pervasive" flags
+            while (parts.size() > 0) {
```
MikaelSmith: I would've gone for something more like `boost::algorithm::trim_right_if(parts, [](string &s) { return s == "onlink" || s == "pervasive"; });`.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,17 @@ namespace facter { namespace facts { namespace linux {
...
+                std::string last_token(parts.back().begin(),
+                                       parts.back().end());
+                if (last_token == "onlink" || last_token == "pervasive")
+                    parts.erase(parts.end() - 1);
```
MikaelSmith: This could be `pop_back`.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -142,6 +142,17 @@ namespace facter { namespace facts { namespace linux {
         auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
+
+            // remove trailing "onlink" or "pervasive" flags
+            while (parts.size() > 0) {
+                std::string last_token(parts.back().begin(),
```
MikaelSmith: Could still use `auto last_token = parts.back()` here. It's automatically a reference because that's what `back()` returns.

PR #1373, opened 2016-07-07
(FACT-1448) Add acceptance test to ensure json handles bools

ccaviness:
On macOS (at least), json output converts all boolean values to true.

### DIFF
--- a/acceptance/tests/options/json.rb
+++ b/acceptance/tests/options/json.rb
@@ -14,7 +14,7 @@ extend Facter::Acceptance::UserFactUtils
 content = <<EOM
 Facter.add('structured_fact') do
   setcode do
-    { "foo" => {"nested" => "value1"}, "bar" => "value2", "baz" => "value3" }
+    { "foo" => {"nested" => "value1"}, "bar" => "value2", "baz" => "value3", "true" => true, "false" => false }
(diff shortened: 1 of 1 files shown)


ccaviness:
I'm not sure how to actually _fix_ this, nor am I sure how to run the acceptance tests without all of puppet inc's infrastructure, but I _think_ this is a test that will prove the bug in any case.

whopper:
https://github.com/puppetlabs/facter/pull/1374 looks to fix the underlying issue - yay! This is a reasonable addition to this test. Once we get the fix merged I can give this a run and merge. Thanks for the contribution!

ccaviness:
Someone more clever than I am should also probably add `lib/test/ruby/ruby_value.cc`

ccaviness:
Also, it would be worth checking that this test actually _fails_ with the previous code and passes with the fix before relying on it.

branan:
:+1: this looks right eyeballing it. We're gonna go ahead and merge so it runs through our CI system, as that's the easiest way to verify test changes. We can fix it up if anything breaks. Thanks for taking the time to write this test! We really appreciate test/doc fixes when somebody finds a bug. It makes our lives so much easier :D

PR #1374, opened 2016-07-08
(FACT-1448) Fix json output to correctly print booleans from custom facts

Magisus:
Fixes a typo in the ruby_value::to_json method that was converting all booleans to true.


branan:
OH, it's RUBY booleans. When I did my testing I looked at built-in facts rather than their ruby test. *sigh*. Good catch :)

whopper:
@Magisus nit pick: can you rewrite your commit message in the imperative? :D  Perhaps something like "fix typo in... blah blah"

whopper:
Building this now for fun, then will run the acceptance update against it and merge 👍

PR #1375, opened 2016-07-08
(FACT-1429) Add `--strict` flag to command line options

Magisus:
Enabling this flag currently only changes one error code: when a non-existing fact is queried, Facter will return 1 instead of 0 as before.

In the future this flag may enable more detailed error codes for different errors. The discussion thread on what this functionality should look like is here: https://groups.google.com/forum/#!topic/puppet-dev/qtiPcJAtKZw

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -268,10 +268,10 @@ namespace facter { namespace facts {
 
     ostream& collection::write(ostream& stream, format fmt, set<string> const& queries)
     {
-        return write(stream, fmt, queries, false);
+        return write(stream, fmt, queries, false, false);
     }
 
-    ostream& collection::write(ostream& stream, format fmt, set<string> const& queries, bool show_legacy)
+    ostream& collection::write(ostream& stream, format fmt, set<string> const& queries, bool show_legacy, bool strict_errors)
     {
         if (queries.empty()) {
             // Resolve all facts
@@ -279,11 +279,11 @@ namespace facter { namespace facts {
         }
 
         if (fmt == format::hash) {
-            write_hash(stream, queries, show_legacy);
+            write_hash(stream, queries, show_legacy, strict_errors);
         } else if (fmt == format::json) {
-            write_json(stream, queries, show_legacy);
+            write_json(stream, queries, show_legacy, strict_errors);
         } else if (fmt == format::yaml) {
-            write_yaml(stream, queries, show_legacy);
+            write_yaml(stream, queries, show_legacy, strict_errors);
         }
         return stream;
     }
@@ -334,7 +334,7 @@ namespace facter { namespace facts {
         return it == _facts.end() ? nullptr : it->second.get();
     }
 
-    value const* collection::query_value(string const& query)
+    value const* collection::query_value(string const& query, bool strict_errors)
     {
         // First attempt to lookup a fact with the exact name of the query
         value const* current = get_value(query);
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Not putting this behind a `--detailed-exitcodes` flag?

Magisus:
Do we have any examples of propagating actually detailed exit codes in any of our C++ projects yet? Because if so I'd like the approach here to be consistent. And if no such example exists, does anyone have a recommended approach? I've never worked much with error codes.

lib/src/facts/collection.cc
```diff
@@ -383,12 +383,17 @@ namespace facter { namespace facts {
         return current;
     }
 
-    value const* collection::lookup(value const* value, string const& name)
+    value const* collection::lookup(value const* value, string const& name, bool detailed_exit_codes)
```
MikaelSmith: This could obviously be simplified by making the flag part of the `collection` object. If these were all public functions, I'd lean towards adding it on construction; since they're all implementation and only affect one public function, I like keeping it restricted to that function rather than a property of the `collection`.

MikaelSmith:
I'm not aware of prior art. Have you defined what the detailed exit code should be? It seems like this should return a new error code (not necessarily EXIT_FAILURE), but saving that probably becomes part of the logging library (`error_logged()`).

Magisus:
Making it part of the logging library makes the most sense to me, I think. It would keep this a lot cleaner. We are collecting opinions on what situations should generate which error codes, but this is a place to start. From what I saw we are already using 2 to represent problems setting up the logger, but I think that's the only one. Is there a convention for this we'd like to follow, or is any number fine (e.g. just keep counting up)?

MikaelSmith:
I'm not too familiar with error code conventions, but Puppet certainly has one to look at. @whopper has some background with them.

whopper:
My puppet-dev thread hasn't generated much on specific codes, so looks like we'll just want to follow puppet's lead as best we can. In puppet, `--detailed-exitcodes` provides the following:

```
--detailed-exitcodes
Provide extra information about the run via exit codes; only works if '--test' or '--onetime' is also specified. If enabled, 'puppet agent' will use the following exit codes:

0: The run succeeded with no changes or failures; the system was already in the desired state.
1: The run failed, or wasn't attempted due to another run already in progress.
2: The run succeeded, and some resources were changed.
4: The run succeeded, and some resources failed.
6: The run succeeded, and included both changes and failures.
```

Looks like for our purposes for this ticket (being unable to resolve a specific queried fact), we'd want to exit 1. I don't know that the second error case (any fact failed to resolve on a general facter run) is in the scope of this ticket, but it'd probably make sense to return 1 there as well.

Magisus:
For the couple cases suggested in the thread, do those count as a run "success" with some failures though? As in, does the Facter run actually complete, or does it crash if a fact can't be resolved?

I personally like the idea of adding support for error codes to the logging library and integrating that into Facter, because it seems like there are "degrees of failure" that we want to be able to report, especially if we currently aren't crashing when these more modest failures occur.

But if we do still want to just use 1 everywhere, then I feel like the flag "detailed-exitcodes" doesn't make much sense, because they aren't any more detailed.

Magisus:
If we wanted to add error code support to the Leatherman logging library: https://github.com/Magisus/leatherman/commit/b11bddca7888049573d025dc2191353d2a1e4627

Magisus:
After talking to @whopper and @branan it sounds like we're going to just stick with the success/failure dichotomy for the time being. In which case, barring any other comments, this should be good to go, unless we want tests for the new flag.

branan:
Now that I've had a while to mull it over, I wonder if it isn't better naming this `--strict` instead. That's a more commonly used term for what we're doing here.

Thoughts @MikaelSmith and @Magisus ?

MikaelSmith:
That seems better than `detailed_exit_codes`. 👍

Magisus:
I agree too. I'll update this!

PR #1377, opened 2016-07-13
(FACT-1457) Spike an access layer around cpp-hocon

Magisus:
This is a proof of concept for configuring facter via a HOCON config
file. It currently assumes that file is located at
/etc/puppetlabs/facter/facter.conf.

It adds a class for wrapping the calls to cpp-hocon and presenting
a simple API for accessing settings.

Since I don't know exactly what our requirements are, this is a generic initial draft. Looking for feedback!

### DIFF
--- a/scripts/travis_target.sh
+++ b/scripts/travis_target.sh
@@ -19,6 +19,7 @@ elif [ ${TRAVIS_TARGET} == DEBUG ]; then
   pip install --user cpp-coveralls
 fi
 
+
 # Generate build files
 if [ ${TRAVIS_TARGET} == COMMITS ]; then
   shopt -s nocasematch
@@ -55,6 +56,15 @@ elif [ ${TRAVIS_TARGET} == DOXYGEN ]; then
 elif [ ${TRAVIS_TARGET} == CPPCHECK ]; then
   make cppcheck
 else
+  # clone and install cpp-hocon
+  git clone https://github.com/puppetlabs/cpp-hocon.git
+  cd cpp-hocon
+  mkdir build
+  cd build
+  cmake ..
+  make install
+  cd ..
+
   make -j2
 
   # Install the bundle before testing
--- a/scripts/travis_target.sh
+++ b/scripts/travis_target.sh
@@ -62,8 +62,9 @@ else
   mkdir build
   cd build
   cmake ..
-  make install
-  cd ..
+  make DESTDIR=$USERDIR install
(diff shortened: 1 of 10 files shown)


Magisus:
Note: currently this only allows us to fetch lists of `hocon::unwrapped_value` and maps of `(string, hocon::unwrapped_value)`. In order to return homogeneous lists and maps of plain C++ types, the best approach would probably be implementing [HC-75](https://tickets.puppetlabs.com/browse/HC-75) and writing a few template specializations to call those methods.

Magisus:
Update: can now retrieve homogeneous lists

adrienthebo:
It looks like the builds are failing because HOCON isn't available at build time; we'll need to fetch that library for Appveyor and Travis.

That aside, the design seems reasonable to me.

Magisus:
Yeah I never got an answer on how we plan to add cpp-hocon to Appveyor and Travis. How does that process work?

adrienthebo:
It looks like we have a somewhat ad-hoc process for downloading prebuilt binaries; for Leatherman we have proper releases but for cmake and yaml-cpp we're fetching files from a random S3 bucket. Since cpp-hocon is still in flux (doesn't seem to have any tagged releases and we're actively waiting for PRs to get merged), the quick and dirty solution that I see is to just clone the hocon-cpp repo and build the project in the build step of the project.

Ultimately I would expect we'll have prebuilt binaries available somewhere, but that task is worth ticketing out and putting into a sprint to deal with separately.

kylog:
The commit checker is going to squawk at the commit summary of `Getting Travis to work` because it wants a ticket or `(maint)`. You could probably amend that commit.

Magisus:
Yeah I'll fix that. Right now I'm focusing on getting the build to fail in the expected way; it relies on code that hasn't been merged into master, so it's going to break regardless.

scripts/travis_target.sh
```diff
@@ -55,6 +56,16 @@ elif [ ${TRAVIS_TARGET} == DOXYGEN ]; then
 elif [ ${TRAVIS_TARGET} == CPPCHECK ]; then
   make cppcheck
 else
+  # clone and install cpp-hocon
```
Magisus: Any ideas as to what is wrong here that Travis still can't find the headers?

Magisus:
Closing this for the time being, since we're going to favor using boost::program_options instead of this interface.

PR #1378, opened 2016-07-18
(maint) Fix networking.dhcp fact in ubuntu.rb for ubuntu-10.04.

seangriff:
The acceptance test facts/ubuntu.rb assumes the ip address in the
networking.dhcp fact begins with '10.'  On ubuntu-10.04, this is not always
the case.  This fix allows ip addresses beginning with '192.' as well as '10.'.

### DIFF
--- a/acceptance/tests/facts/ubuntu.rb
+++ b/acceptance/tests/facts/ubuntu.rb
@@ -83,7 +83,7 @@ agents.each do |agent|
 
   step "Ensure the Networking fact resolves with reasonable values for at least one interface"
   expected_networking = {
-                          "networking.dhcp"     => /10\.\d+\.\d+\.\d+/,
+                          "networking.dhcp"     => /(?:10|192)\.\d+\.\d+\.\d+/,
                           "networking.ip"       => /10\.\d+\.\d+\.\d+/,
                           "networking.ip6"      => /[a-f0-9]+:+/,
                           "networking.mac"      => /[a-f0-9]{2}:/,


kylog:
Is there a reason for any of the networking facts on any of the platforms to assume the first octet is `10` (vs just `\d`)?

kylog:
Better than `\d`, should they all use something like https://github.com/puppetlabs/facter/blob/master/acceptance/tests/facts/sles.rb#L12?

seangriff:
That is a good regular expression (in sles.rb), allowing only and all well-formed ip addresses.  I don't know why the code was written to allow only 10-dot addresses.  It's that way in debian.rb, fedora.rb, macosx.rb, solaris.rb, and windows.rb.  Apparently so far only ubuntu1004 gets a dhcp server address of 192.168.x.x.  I am now inclined to use the sles.rb expression, but only fix  ubuntu.rb at this point. If another arch fails then replace the regex for the rest of the other architectures too.  Does that sound reasonable?

kylog:
I'm tempted to just fix them all to be a regex for good IP address. There's no reason to hardwire `10.` or `192.` anywhere AFAICT. But /cc @whopper or @johnduarte who might have more insight than I.

(That said, it is nerd-snipingly interesting that this one platform would see this failure, so I don't want to gloss over that. Seems unlikely to be a facter change though.)

seangriff:
I'm good with fixing all of them.  It's not difficult.  I was tending toward least disruption to fix the problem at hand.  Uniformity across platforms is a plus.

johnduarte:
:+1:  for updating the `10.` assumptions to general IP addresses. I presume the `10.` was a simple paradigm to use for internal testing. However, when running the test against infrastructure with other network configurations, this becomes brittle. We should generalize them.

MikaelSmith:
Yes we should fix the rest. However, builds been blocked for 6 days so merging.

PR #1381, opened 2016-07-19
(FACT-1465) Use @ip_regex for validating network address facts

seangriff:
Prior to this fix, many architecture specific fact tests were verifying
ip addresses and netmasks were in the 10.x.x.x range.  This cause ubuntu.rb
to fail when the dhcp server address was 192.x.x.x instead.  After this fix,
all network addresses are validated using the same regex pattern as used in
sles.rb, which matches any and only valid IPv4 addresses.  See also 
https://github.com/puppetlabs/facter/pull/1378

### DIFF
--- a/acceptance/tests/facts/debian.rb
+++ b/acceptance/tests/facts/debian.rb
@@ -1,5 +1,7 @@
 test_name "Facts should resolve as expected in Debian 6 and 7"
 
+@ip_regex = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
+
 #
 # This test is intended to ensure that facts specific to an OS configuration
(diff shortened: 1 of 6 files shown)


acceptance/tests/facts/debian.rb
```diff
@@ -1,5 +1,7 @@
 test_name "Facts should resolve as expected in Debian 6 and 7"
 
+@ip_regex = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
```
MikaelSmith: Why an instance variable?

seangriff: Was copying the model of sles.rb

seangriff: downside?

MikaelSmith: None I know of.

PR #1383, opened 2016-07-26
(FACT-1458 & 1459) Basic Facter configuration with hocon and boost::program_options

Magisus:
This PR contains two commits which together allow settings many of Facter's command line options in a config file, the location of which defaults to /etc/puppetlabs/facter/facter.conf, but which is also settable from the command line (for ease of testing).


Magisus:
Currently blocked on [HC-85](https://tickets.puppetlabs.com/browse/HC-85).

Magisus:
This passes all existing Facter acceptance tests, so the exisitng CLI has not changed.

lib/CMakeLists.txt
```diff
@@ -296,6 +298,7 @@ endif()
 
 # Link in additional libraries
 target_link_libraries(libfacter PRIVATE
+    ${CPPHOCON_LIBRARIES}
```
MikaelSmith: Will libfacter eventually be reading from config? i.e. when running as Ruby? I don't see a current use for linking here, but that seems like something we'd need.

Magisus: Yes it will, though I think you're right, it's not actually needed for this PR (not sure how it slipped in). Happy to remove it if you'd like.

exe/CMakeLists.txt
```diff
@@ -12,13 +12,29 @@ include_directories(
...
+if (${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    target_link_libraries(facter rt)
+endif()
+
 target_link_libraries(facter libfacter
+    ${CPPHOCON_LIBRARIES}
     ${Boost_PROGRAM_OPTIONS_LIBRARY}
+    ${Boost_THREAD_LIBRARY}
+    ${LEATHERMAN_RUBY_LIB}
+    ${LEATHERMAN_FILE_UTIL_LIB}
```
MikaelSmith: Windows build is failing because leatherman_file_util has a dependency on boost_filesystem. You'll want to put the boost_filesystem dependency after leatherman_file_util. I'd also suggest moving the other boost libraries after leatherman libraries, as boost won't depend on leatherman.

There are certain scenarios where that suggestion can cause trouble - dynamic leatherman but static boost, and leatherman only importing some of the functions from boost, can result in some symbols binding to leatherman dlls and others pulling them in from the static boost libraries; in that case you may have problems with global static members, where there are two copies of the same global and different functions refer to different versions. Hence don't use dynamic leatherman and static boost.

lib/CMakeLists.txt
MikaelSmith: Didn't seem like it caused problems, so you can leave it.

HAIL9000:
@Magisus @MikaelSmith With HC-85 merged, do you think this is ready to go?

Magisus:
No, we're still waiting on a fix for the OSX boxes, which we wanted to get sorted before we retag cpp-hocon. Then we have to bump puppet-agent, which will include a fix to get cpp-hocon building properly on Windows.

Magisus:
If CI goes green with https://github.com/puppetlabs/puppet-agent/pull/779 in it, this should be ready to go.

.gitignore
```diff
@@ -16,6 +16,7 @@ compile_commands.json
 /lib/spec/spec_helper.rb
 /debug*/
 /*release*/
+/*build*/
```
branan: If you feel like a maint patch to the cpp-project-template to make sure it has debug, release, and build flagged to ignore, that'd be cool. I'm not sure we've been consistent here

cmake/FindCPPHOCON.cmake
```diff
@@ -0,0 +1,6 @@
+include(FindDependency)
```
branan: We should set up cpp-hocon itself to do the normal CMake library export soon-ish, but since Facter is the only consumer I think we can fix that when we switch cpp-hocon to be shared.

.gitignore
Magisus: Yeah I can do that!

Magisus: Actually the project template already has it, but I'm going to change the way they match, for ease of use on VMs using shared folders, where you might need multiple build directories in one project.

Magisus:
@er0ck This is the PR that needs a little QA review before we merge it. Thanks!

acceptance/tests/options/config_file/command_line_override.rb
```diff
@@ -0,0 +1,39 @@
+test_name "flags set on the command line override config file settings"
```
er0ck: i'm sure the other tests didn't do this but `test_name` and `step` should have do/end blocks.  this allows beaker to indent the logs properly. this mode is deprecated, if not officially.

Magisus: Okay sounds good. Do you want me to update the command line flag tests as well?

acceptance/tests/options/config_file/command_line_override.rb
```diff
@@ -0,0 +1,39 @@
...
+}
+cli : {
+    debug : false
+}
+EOM
+
+agents.each do |agent|
+  step "Agent #{agent}: create config file"
+  config_dir = agent.tmpdir("config_dir")
+  config_file = File.join(config_dir, "facter.conf")
```
er0ck: this might not work on windows agents.  but we might get lucky because of cygwin (don't think i've ever said that before)

seangriff:
I think this looks good.  The acceptance tests are thorough and seem correct.  Let's get this building.

acceptance/tests/options/config_file/command_line_override.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  step "--external_dir flag should override external_dir in config file"
+  on(agent, facter("--config '#{config_file}' --external-dir 'cmd/line/dir'")) do
+    assert_match(/cmd\/line\/dir/, stderr, "Facter should attempt to find external fact dir 'cmd/line/dir'")
+    assert_no_match(/config\/file\/dir/, stderr, "Facter should not attempt to find external fact dir 'config/file/dir'")
+  end
+
+  step "conflict logic applies across settings sources"
+  on(agent, facter("--config '#{config_file}' --no-external-facts"), :acceptable_exit_codes => [1]) do
+    assert_match(/no-external-facts and external-dir options conflict/, stderr, "Facter should have warned about conflicting settings")
```
er0ck: you really can't set `--external-dir` and `--no-external-facts` at the same time?

i'd think it'd just ignore the dir.  but i guess that immaterial to this test...

acceptance/tests/options/config_file/verbose.rb
```diff
@@ -0,0 +1,23 @@
...
+
+agents.each do |agent|
+  step "Agent #{agent}: create config file"
+  config_dir = agent.tmpdir("config_dir")
+  config_file = File.join(config_dir, "facter.conf")
+  create_remote_file(agent, config_file, config)
+
+  step "debug output should print when config file is loaded"
+  on(agent, facter("--config '#{config_file}'")) do
+    assert_match(/INFO/, stderr, "Expected stderr to contain verbose (INFO) statements")
```
er0ck: is there a default place to find a facter config file?  we should probably have one of these test that it finds that file.

Magisus: Yes, though come to think of it, it only has the unix location by default right now. I can add the test and the default windows location when I do the update fixing the rest of these comments.

acceptance/tests/options/config_file/command_line_override.rb
Magisus: I was actually wondering about that. We may want to change that behavior now because of this new interaction.

PR #1384, opened 2016-07-27
(FACT-1413) Fix fact precedence and resolution order

ahenroid:
This change formalizes fact precedence and resolution order when there are
multiple top-level facts with the same name, and it fixes application of the
"has_weight" field from custom facts. With this change, fact precedence
with default weights is as follows:

  external fact (highest precendence)
  custom fact
  environment fact
  core fact (lowest precedence)

To align with Facter 2.x behavior, external facts are assigned an inherent
weight of 10000. Therefore, a custom fact with "has_weight" >= 10001 will
override an external fact with the same name. Also, preserving existing
Facter behavior, a custom fact with a "has_weight" of 0 will be overriden
by an environment or core fact with the same name.


ahenroid:
Tests coming next...

PR #1385, opened 2016-07-27
(FACT-1476) acceptance test for facter --strict flag

phongdly:
The flag --strict has been recently added to facter cli, so this PR is for its acceptance test

### DIFF
--- a/acceptance/tests/ticket_1429_facter_with_strict_flag.rb
+++ b/acceptance/tests/ticket_1429_facter_with_strict_flag.rb
@@ -1,0 +1,13 @@
+test_name 'ticket FACT-1476 - C98098 facter cli with --strict flag'
+
+agents.each do |host|
+  step 'facter should return exit code 0 for querying non-existing-fact without --strict flag'
+  on(host, facter('-p non-existing-fact')) do |result|
+    assert_match(/WARN\s+puppetlabs\.facter\s+-\s+skipping external facts for/, result.stderr, 'Unexpected error was detected!')
+  end
+
+  step 'facter should return exit code 1 for querying non-existing-fact with --strict flag'
+  on(host, facter('-p non-existing-fact --strict'), :acceptable_exit_codes => 1) do |result|
+    assert_match(/ERROR\s+puppetlabs\.facter - fact "non-existing-fact" does not exist/, result.stderr, 'Unexpected error was detected!')
+  end
+end


acceptance/tests/ticket_1429_facter_with_strict_flag.rb
```diff
@@ -0,0 +1,11 @@
+test_name 'ticket FACT-1429  facter should return exit code 1 for querying non-existing fact when --strict is on'
+
+agents.each do |host|
+  step 'facter should return exit code 0 for querying non-existing-fact without --strict flag'
+  on(host, facter('-p non-existing-fact'))
```
er0ck: should we assert on a warning here?

acceptance/tests/ticket_1429_facter_with_strict_flag.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  step 'facter should return exit code 1 for querying non-existing-fact with --strict flag'
+  on(host, facter('-p non-existing-fact --strict'), :acceptable_exit_codes => 1) do
+    assert_match(/ERROR puppetlabs.facter - fact "non-existing-fact" does not exist/, stderr.chomp, 'Unexpected error was detected!')
```
er0ck: does this work?  i'm not sure what stderr here is.  it looks like it's the last output of `#on`.  
i think we prefer the pattern where we explicitly place the result in a variable and compare to its stderr like:
```
on(blah, 'blah') do |result|
  assert(result.exit_code)
end
```

acceptance/tests/ticket_1429_facter_with_strict_flag.rb
```diff
@@ -0,0 +1,11 @@
+test_name 'ticket FACT-1429  facter should return exit code 1 for querying non-existing fact when --strict is on'
```
er0ck: we should create a testrail case for this and place its number (including the 'C') where `ticket FACT-1429` is above.

when the case is creaeted, please reference that ticket, and create a new ticket in the same sprint with a link to it from the test case in the automation ticket field.  also place a reference to the automation ticket in the reference field (with commas separating the references).

then you can set FACT-1429 `qa-status` to `reviewed`

acceptance/tests/ticket_1429_facter_with_strict_flag.rb
phongdly: Yes we should, will have it in the next push

acceptance/tests/ticket_1429_facter_with_strict_flag.rb
phongdly: yes, it works with the assert_match, but it's better if we check the result.exit_code that way, will update in the next push

phongdly: @er0ck, I decided to go back to assert_match instead because result.exit_code is a number (either 0 or 1 in this case), so it is very likely to pass the test most of the time regardless of result.exit_code is 0 or 1,  because the output has date/time which usually has both 0 and 1 in it.

acceptance/tests/ticket_1429_facter_with_strict_flag.rb
phongdly: Added the Fact Jira ticket and the TestRail test case in the 'test_name'

er0ck:
👍

PR #1387, opened 2016-08-01
(FACT-1405) Properly compare ip route flags

branan:
Previously, we were attempting to compare a string::iterator range
with a const char* C-string via an iterator comparison, treating the
C-string as a "range". This does not work, because when treated as a
range the c-string includes the null byte, while the string iterator
pair does not. This means the ranges are unequal even if they contain
the same character data.

This converts the iterator range to a string before comparison, so
that we use std::string::operator== for the comparison.


MikaelSmith:
Oh, joy. I wonder if string_view (C++17) would do the correct thing. Oh well, this is at least pretty memory-efficient.

PR #1389, opened 2016-08-03
(maint) Fix typo in path to custom fact

Magisus:


branan:
This just changes an acceptance test, so doesn't need Travis. I'm merging early.

PR #1390, opened 2016-08-03
(FACT-1361) Create networking fact acceptance test

phongdly:
Prior to this PR, networking facts were verified in each individual test cases in acceptance/tests/facts/<platform>.rb

This PR pulls out the networking portion of all the tests in acceptance/tests/facts into new acceptance test, acceptance/tests/facts/networking_facts.rb

This will also help to test networking facts in many more platforms in puppet-agent CI, instead of being confined to only 9 platforms in acceptance/tests/facts

### DIFF
--- a/acceptance/tests/facts/networking_facts.rb
+++ b/acceptance/tests/facts/networking_facts.rb
@@ -1,0 +1,80 @@
+test_name 'FACT-1361 - C59029 networking facts should be fully populated'
+
+#
+# This test is intended to ensure that networking facts resolves
+# as expected in AIO across supported platforms.
+#
+
+skip_test "Networking fact test is confined to AIO" if @options[:type] != 'aio'
+
+@ip_regex = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
+
+expected_networking = {
+    "networking.dhcp"     => @ip_regex,
+    "networking.ip"       => @ip_regex,
+    "networking.ip6"      => /[a-f0-9]+:+/,
+    "networking.mac"      => /[a-f0-9]{2}:/,
+    "networking.mtu"      => /\d+/,
+    "networking.netmask"  => /\d+\.\d+\.\d+\.\d+/,
+    "networking.netmask6" => /[a-f0-9]+:/,
+    "networking.network"  => @ip_regex,
+    "networking.network6" => /([a-f0-9]+)?:([a-f0-9]+)?/
+}
+
+agents.each do |agent|
+
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  refute_empty(primary_interface)
+
+  expected_bindings = {
+      "\"networking.interfaces.#{primary_interface}.bindings.0.address\"" => /\d+\.\d+\.\d+\.\d+/,
+      "\"networking.interfaces.#{primary_interface}.bindings.0.netmask\"" => /\d+\.\d+\.\d+\.\d+/,
+      "\"networking.interfaces.#{primary_interface}.bindings.0.network\"" => /\d+\.\d+\.\d+\.\d+/,
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.address\"" => /[a-f0-9:]+/,
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.netmask\"" => /[a-f0-9:]+/,
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.network\"" => /[a-f0-9:]+/
+  }
+
+  case agent['platform']
+    when /solaris/, /eos/
+      #remove the invalid networking facts on aix and solaris
+      expected_networking.delete("networking.ip6")
+      expected_networking.delete("networking.netmask6")
+      expected_networking.delete("networking.network6")
+
+      #remove invalid bindings for the primary networking interface on AIX and Solaris
+      expected_bindings.delete("\"networking.interfaces.#{primary_interface}.bindings6.0.address\"")
+      expected_bindings.delete("\"networking.interfaces.#{primary_interface}.bindings6.0.netmask\"")
+      expected_bindings.delete("\"networking.interfaces.#{primary_interface}.bindings6.0.network\"")
+
+    when /sparc/, /aix/, /cisco/
+      #remove the invalid networking facts on SPARC or AIX
+      #Our SPARC testing platforms don't use DHCP
+      expected_networking.delete("networking.dhcp")
+      expected_networking.delete("networking.ip6")
+      expected_networking.delete("networking.netmask6")
+      expected_networking.delete("networking.network6")
+
+      #remove invalid bindings for the primary networking interface on SPARC or AIX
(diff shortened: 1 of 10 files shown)


acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
+test_name 'FACT-1361 - C59029 networking facts should be fully populated'
```
er0ck: test_name and step should have `do/end` to create blocks, and proper indentation in logs

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
...
+
+#
+# This test is intended to ensure that networking facts resolves
+# as expected in AIO across supported platforms.
+#
+
+skip_test "Networking fact test is confined to AIO" if @options[:type] != 'aio'
```
er0ck: this is targeted at master, so i don't think this is required.  even if merged down to stable, it'll still always be aio.  unless i'm missing something about network devices, where they aren't run with aio type?

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
...
+
+@ip_regex = /^(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/
+
+expected_networking = {
+    "networking.dhcp"     => @ip_regex,
+    "networking.ip"       => @ip_regex,
+    "networking.ip6"      => /[a-f0-9]+:+/,
+    "networking.mac"      => /[a-f0-9]{2}:/,
+    "networking.mtu"      => /\d+/,
+    "networking.netmask"  => /\d+\.\d+\.\d+\.\d+/,
```
er0ck: i think each of these octets can be restricted to... octets?  i.e.: this will always be 1-3 digits per octet, AFAIK.
@branan are these never in the `192.168.0.0/24` type format?

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
...
+    "networking.netmask6" => /[a-f0-9]+:/,
```
er0ck: i think these are always `2001:0db8:0123:4567:89ab:cdef:1234:5678`  
as in 8 sets of 4 hex encoded chars separated by colons

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
...
+agents.each do |agent|
+
+  step "Ensure a primary networking interface was determined."
+  primary_interface = fact_on(agent, 'networking.primary')
+  refute_empty(primary_interface)
+
+  expected_bindings = {
+      "\"networking.interfaces.#{primary_interface}.bindings.0.address\"" => /\d+\.\d+\.\d+\.\d+/,
+      "\"networking.interfaces.#{primary_interface}.bindings.0.netmask\"" => /\d+\.\d+\.\d+\.\d+/,
+      "\"networking.interfaces.#{primary_interface}.bindings.0.network\"" => /\d+\.\d+\.\d+\.\d+/,
```
er0ck: can we re-use the netmask and ip regexes here?

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,80 @@
...
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.address\"" => /[a-f0-9:]+/,
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.netmask\"" => /[a-f0-9:]+/,
+      "\"networking.interfaces.#{primary_interface}.bindings6.0.network\"" => /[a-f0-9:]+/
+  }
+
+  case agent['platform']
+    when /solaris/, /eos/
+      #remove the invalid networking facts on aix and solaris
```
er0ck: comment appears incorrect

acceptance/tests/facts/networking_facts.rb
seangriff: The hex characters do come in groups for four, and there are 8 groups, but there is a short hand that allows skipping one run of contiguous 0s.  An address may look like 2001:0db8::1234:5678, to mean 2001:0db8:0000:0000:0000:0000:1234:5678 (zeros between the two contiguous colons).  This makes a recognizing regexp very nasty (see http://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses)

acceptance/tests/facts/networking_facts.rb
seangriff: Nice work Phong.  This is much better than before.

acceptance/tests/facts/networking_facts.rb
er0ck: hmmm, there is also:
http://ruby-doc.org/stdlib-2.0.0/libdoc/ipaddr/rdoc/IPAddr.html#method-i-ip6_int

TL;DR: use ruby's networking code to check them

acceptance/tests/facts/networking_facts.rb
phongdly: will add do/end for test_name in the next push

acceptance/tests/facts/networking_facts.rb
phongdly: will remove the lind in the next push

acceptance/tests/facts/networking_facts.rb
phongdly: That makes sense, netmask should be restricted to octets. 
Netmask type A format: 255.0.0.0
Netmask type B format: 255.255.0.0
Netmask type C format: 255.255.255.0

But in some customized netmask, any 255 in any octet can be replaced by either 128, 193, 224, 240, 248, 252, or 254.

So I will add the @netmask_regex = /^(((128|192|224|240|248|252|254)\.0\.0\.0)|(255\.(0|128|192|224|240|248|252|254)\.0\.0)|(255\.255\.(0|128|192|224|240|248|252|254)\.0)|(255\.255\.255\.(0|128|192|224|240|248|252|254)))$/ in the next push

er0ck: that seems a bit hard to read.
i'd probably just do:
`/\d{1,3}?\.\d{1,3}?\.\d{1,3}?\.\d{1,3}?\./`

acceptance/tests/facts/networking_facts.rb
```diff
@@ -0,0 +1,79 @@
...
+        expected_bindings.delete("\"networking.interfaces.#{primary_interface}.bindings6.0.netmask\"")
+        expected_bindings.delete("\"networking.interfaces.#{primary_interface}.bindings6.0.network\"")
+
+      when /sles/
+        #some sles VMs do not have networking.dhcp
+        expected_networking.delete("networking.dhcp")
+
+    end
+
+    step "Ensure the Networking fact resolves with reasonable values for at least one interface"
```
er0ck: `step` needs do/end blocks also

er0ck:
looks good.  please add do/end blocks to all `step`'s and i can merge.

acceptance/tests/facts/networking_facts.rb
er0ck: @phongdly please add do/end blocks around your steps

PR #1391, opened 2016-08-04
(FACT-1478) EC2 facts intermittently unavailable due to overly optimistic timeouts

saj:


HAIL9000:
👍 From me and @Magisus

PR #1392, opened 2016-08-05
(FACT-1271) Generate msg catalogs, review and mark translatable strings

ahenroid:
Updates include:
* Apply leatherman::locale::format() to mark and translate log messages and error strings.
* Update format strings from boost::format style %N% to boost::locale::format style {N}.
* Update build for string externalization with xgettext (generate FACTER.pot).
* Add log/format helpers to facter::logging, intended for the Facter executable
  (exe/facter.cc) which cannot use leatherman/locale/locale.hpp or link to boost::regex.


lib/src/facts/aix/networking_resolver.cc
```diff
@@ -12,6 +13,9 @@
 #include <netinet/in.h>
 #include <sys/kinfo.h>
 
+// Mark string for translation (a helper macro for simple cases)
+#define _(...) leatherman::locale::format(__VA_ARGS__)
```
ahenroid: When Facter catches up to Leatherman 0.8.2, this unfortunate macro definition (here and in the other changed files) will be replaced with the sturdier: "using leatherman::locale::_"

exe/facter.cc
```diff
@@ -200,41 +202,41 @@ int main(int argc, char **argv)
...
+            boost::nowide::cerr << _("error: {1}", ex.what()) << endl;
             colorize(boost::nowide::cerr);
             help(visible_options);
             return EXIT_FAILURE;
         }
 
         // Check for printing the version
         if (vm.count("version")) {
-            boost::nowide::cout << LIBFACTER_VERSION_WITH_COMMIT << endl;
+            boost::nowide::cout << _(LIBFACTER_VERSION_WITH_COMMIT) << endl;
```
MikaelSmith: This definitely isn't necessary.

exe/facter.cc
```diff
@@ -122,25 +124,25 @@ int main(int argc, char **argv)
...
-            ("no-custom-facts", po::bool_switch()->default_value(false), "Disables custom facts.")
-            ("no-external-facts", po::bool_switch()->default_value(false), "Disables external facts.")
-            ("no-ruby", po::bool_switch()->default_value(false), "Disables loading Ruby, facts requiring Ruby, and custom facts.")
-            ("puppet,p", "(Deprecated: use `puppet facts` instead) Load the Puppet libraries, thus allowing Facter to load Puppet-specific facts.")
-            ("trace", po::bool_switch()->default_value(false), "Enable backtraces for custom facts.")
-            ("verbose", po::bool_switch()->default_value(false), "Enable verbose (info) output.")
-            ("version,v", "Print the version and exit.")
-            ("yaml,y", "Output in YAML format.")
-            ("strict", "Enables more aggressive error reporting.");
+            ("color", _("Enables color output.").c_str())
```
MikaelSmith: I'm not sure what's going to happen if locale throws an exception. We have some special handling for exceptions in setup_logging.

exe/facter.cc
ahenroid: You're right. I'll remove.

exe/facter.cc
ahenroid: I'll take a look at setup_logging. We may want to catch and report errors in leatherman.

ahenroid: I looked back at leatherman::locale::translate*, which is where these calls to
facter::logging::format eventually end up. The leatherman translate functions
catch any and all exceptions and return the original untranslated message,
obviating the need for exception handling here. So I believe this is a non-issue,
unless you have some specific concern.

MikaelSmith: `get_locale` can throw an exception, and is called for all the `format*` functions. I believe that's what happens on AIX, Solaris, and some network device platforms.

ahenroid: Thanks for pointing this out. I had forgotten about the call to get_locale inside format.  I believe it makes sense to fix this in Leatherman (adding a try-catch block to leatherman::locale::format_common and then some sensible fallback). If this sounds reasonable, I'll open a ticket against Leatherman.

MikaelSmith: I think so as well. Pinging @branan, he was the last to work on exception handling for locale.

branan: My thought/plan here was that we should communicate any initialization failure in the logging/locale setup step, and transparently catch any exceptions when doing actual logging/translation, with the assumption that the consumer knows what's up at that point

CMakeLists.txt
```diff
@@ -139,8 +139,10 @@ enable_cppcheck()
 # Pull in helper macros for working with leatherman libraries
 include(leatherman)
 
+file(GLOB_RECURSE ALL_SOURCES lib/src/*.cc lib/inc/*.hpp lib/inc/version.h exe/*.cc exe/*.hpp exe/*.h)
```
branan: you can probably do something like ${CMAKE_PROJECT_DIR}/etc/etc for each of these so that it can be moved into the locale/CMakeLists.txt file

MikaelSmith: We also already do this thing about 10 lines up for cpplint.

ahenroid: The duplication in initializing ALL_SOURCES was to work around an earlier build issue. I thought ALL_SOURCES was being modified by one of the included cmake files. However, I checked and I was wrong. There is no need for the second initialization, here or in locales/CMakeLists.txt. I will now push the corrected commit.

ahenroid: However, I'll need to rebase/merge this after the configurable facter changes are merged.

ahenroid:
Rebased to master.

Magisus:
This looks good to me, merging.

PR #1393, opened 2016-08-05
(FACT-1460) Load from config file when calling Facter from Ruby

Magisus:
This inludes a refactor which builds config file access out into
a utility file. It also enables config file loading when Facter is
called from Ruby. For this mode, only two of the currently implemented
settings are used: external-dir and custom-dir, which allow the user to
specify additional directories for custom and external facts. These are
added to any paths specified by Puppet when running Facter from within
Puppet.


exe/facter.cc
```diff
@@ -299,6 +281,9 @@ int main(int argc, char **argv)
             log(level::warning, "Facter was called recursively, skipping external facts. Add '--no-external-facts' to silence this warning");
           } else {
             environment::set("INSIDE_FACTER", "true");
+            if (vm.count("external-dir")) {
+                external_directories.push_back(vm["external-dir"].as<string>());
```
MikaelSmith: vm["external-dir"] is listed as supporting a vector of strings. How does this work?

lib/src/util/config/windows/config.cc
```diff
@@ -0,0 +1,8 @@
+#include <facter/util/config.hpp>
+
+namespace facter { namespace util { namespace config {
+
+    hocon::shared_config load_default_config_file() {
+        return load_config_from("C:\\ProgramData\\PuppetLabs\\facter\\etc\\facter.conf");
```
MikaelSmith: Should probably use https://github.com/puppetlabs/leatherman/blob/master/windows/inc/leatherman/windows/file_util.hpp#L21 here.

lib/src/util/config/config.cc
```diff
@@ -0,0 +1,53 @@
...
+        if (hocon_config && hocon_config->has_path("cli")) {
+            auto cli_settings = hocon_config->get_object("cli")->to_config();
+            po::store(hocon::program_options::parse_hocon<char>(cli_settings, cli_config_options(), true), vm);
+        }
+    }
+
+    po::options_description global_config_options() {
+        po::options_description global_options("");
+        global_options.add_options()
+            ("custom-dir", po::value<string>(), "A directory to use for custom facts.")
```
MikaelSmith: These don't match the `po::value` declarations in exe/facter.cc.

exe/facter.cc
Magisus: I have been unable to get it to properly parse any kind of list, no matter what formatting of cli arg I used. Since the description says the user is only supposed to be able to specify one directory, was there any reason to have it grab a vector other than to allow it to automatically save into `external_directories`? I had to refactor this functionality when I pulled out the config file code (to avoid requiring array references to be passed into the utility functions), but I think it would work just as well to use `vector<string>` as `string`.  Would doing the assignment be better here?

MikaelSmith: I don't know, honestly that predates me. /cc @peterhuene 

You can use multiple arguments on the command-line via `multitoken`, but I don't think we actually want to. So this should maybe just be a string everywhere (that we `move` into our internal vectors).

Magisus: I'm going to go with `string` for now and add the `move`. It makes more sense to me given the description of the option (which matches online docs I found for Facter, which say you can only specify one directory with this option).

Magisus: Edit: figured out how multiple directories could be specified on the command line. You have to use the flag multiples times, and each time it will concatenate into the vector. I assume we do not want to break this behavior. Luckily, this also seems to just work with hocon list syntax (yay @branan  being thorough with the program_options extension!), so I will go back to using `vector<string>`.

lib/tests/ruby/ruby.cc
```diff
@@ -528,7 +528,7 @@ SCENARIO("custom facts written in Ruby") {
         }
     }
     GIVEN("a custom on_message block") {
-        log_capture capture(level::debug);
+        log_capture capture(level::warning);
```
Magisus: This test was failing because Leatherman emits a debug message when file_util::file_readable emits a debug message when the file does not exist, as happens when it attempt to load the default config file, but there isn't one. Does this change still test what it was intended to test? It looks like we are doing the same thing to both debug and warn in the ruby fixture: https://github.com/puppetlabs/facter/blob/master/lib/tests/fixtures/ruby/on_message.rb.

MikaelSmith: Hmm, file_readable emitting a debug message isn't great :( I'd probably change the capture.result() check to verify that the specific warning/debug messages in the test aren't present.

Magisus: I can do that. Should I use another way of checking if the file is there though, to avoid that in general? I just wanted to make use of our existing utility functions, but could be it's not a good use case for that method.

MikaelSmith: file_readable is fine; the idea is to make it clear whether the file is missing, or unreadable because of permissions.

Magisus:
Green acceptance tests are here: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-facter-ad-hoc/69/

adrienthebo:
@Magisus we have merged conflicts on this (probably from #1394), could you rebase?

Magisus:
If we decide to do a z release for puppet-agent before Davis, the stuff in this PR would be a candidate to go in it, since it fleshes out the stuff we shipped with Facter 3.4.0 (including supporting the documented default config location on Windows).

branan:
@Magisus if you want this in the Z it should be retargeted at stable.

I'm +1 on the code - I haven't taken the time to read through all the acceptance changes, though.

Magisus:
Nearly all the acceptance changes are indentation that Eric asked me to fix before my previous PR got merged a little quickly. There is only one new test, acceptance/tests/options/config_file/default_file_location.rb, which just verifies that the file can be loaded from the default locations specified in the RFC, on both Unix and Windows.

Magisus:
@branan Is the only way to retarget to close the PR and open a new one?

branan:
unfortunately, yes

PR #1394, opened 2016-08-06
(FACT-1413) Fix fact precedence and resolution order

ahenroid:
This change formalizes fact precedence and resolution order when there are
multiple top-level facts with the same name, and it fixes application of the
"has_weight" field from custom facts. With this change, fact precedence
with default weights is as follows:

* external fact (highest precendence)
* custom fact
* environment fact
* core fact (lowest precedence)

To align with Facter 2.x behavior, external facts are assigned an inherent
weight of 10000. Therefore, a custom fact with "has_weight" >= 10001 will
override an external fact with the same name. Also, preserving existing
Facter behavior, a custom fact with a "has_weight" of 0 will be overriden
by an environment or core fact with the same name.

Also, this commit adds a check and error message for negative "has_weight"
values. Previously, these were silently cast to large unsigned numbers.

### DIFF
--- a/lib/src/ruby/resolution.cc
+++ b/lib/src/ruby/resolution.cc
@@ -183,9 +183,14 @@ namespace facter { namespace ruby {
     {
         auto const& ruby = api::instance();
 
+        int64_t val = ruby.rb_num2ll(value);
+        if (val < 0) {
+            ruby.rb_raise(*ruby.rb_eTypeError, "expected a non-negative value for has_weight (not %lld)", val);
+        }
+
         auto instance = ruby.to_native<resolution>(self);
         instance->_has_weight = true;
-        instance->_weight = (ruby.num2size_t(value));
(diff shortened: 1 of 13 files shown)


lib/src/ruby/resolution.cc
```diff
@@ -183,9 +183,14 @@ namespace facter { namespace ruby {
     {
         auto const& ruby = api::instance();
 
+        int64_t val = ruby.rb_num2ll(value);
+        if (val < 0) {
```
adrienthebo: Before this change would we silently cast negative numbers to huge unsigned numbers?

adrienthebo:
Aside from the new sign check, :+1:, will merge after that question is resolved.

lib/src/ruby/resolution.cc
ahenroid: Yes, that's exactly what would happen. I added a note to the PR comment.

PR #1396, opened 2016-08-11
mergeup

Magisus:


MikaelSmith:
Doesn't like your commit labeled "Updated to version 3.4.1". That should be have (maint) or (packaging) at the beginning.

PR #1399, opened 2016-08-12
(FACT-1485) Fix Amazon Linux operating system release parsing

downsj2:
Facter 3 reverted behavior that was already fixed in facter 2.  See FACT-1485 for details.

### DIFF
--- a/lib/src/facts/linux/os_linux.cc
+++ b/lib/src/facts/linux/os_linux.cc
@@ -251,6 +251,7 @@ namespace facter { namespace facts { namespace linux {
     {
         // Map of release files that contain a "release X.X.X" on the first line
         static map<string, string> const release_files = {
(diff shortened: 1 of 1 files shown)


MikaelSmith:
👍

MikaelSmith:
Seems eligible for merging up to stable for a bug-fix release.

downsj2:
@puppetcla I do not believe this patch is subject to copyright and that it should be exempt from the CLA.

a) It's one line of code, which is simply a copy of other members of the same map with slight changes.
b) Due to the addition of that map entry, some redundant code is then removed further down the function.

There's nothing creative or novel about this patch, it's just an expression of code which was already present.

MikaelSmith:
Thanks, closing in favor of https://github.com/puppetlabs/facter/pull/1403.

PR #1401, opened 2016-08-15
(QENG-4186) Restrict mime-types gem for Ruby 1.x

ferventcoder:
The mime-types v3+ gem flipped to requiring Ruby >= 2 with
the dependency on mime-types-data. This means when running
beaker against versions of Ruby less than 2, we need to
restrict the mime-types gem to 2.99.2 or less.


johnduarte:
:+1: LGTM

PR #1402, opened 2016-08-15
(FACT-1460) Load global settings from config file when loading Facter from Ruby

Magisus:
New acceptance tests: loading default file, setting multiple external/custom directories on the command line and in the config file. Green here: https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_intn-van-sys_suite-manual-facter-ad-hoc/73/


lib/src/ruby/module.cc
```diff
@@ -44,6 +46,7 @@ namespace facter { namespace ruby {
         {
             // Create a collection and a facter module
             _facts.reset(new collection());
+
```
branan: random whitespace change

lib/src/ruby/module.cc
```diff
@@ -170,8 +173,16 @@ namespace facter { namespace ruby {
...
-        // Initialize the search paths
-        initialize_search_paths(paths);
+        // Load global settigs from config file
+        load_global_settings(load_default_config_file(), _config_file_settings);
+
+        // Initialize custom fact paths
+        vector<string> custom_fact_path = paths;
+        if (_config_file_settings.count("custom-dir")) {
+            auto config_paths = _config_file_settings["custom-dir"].as<vector<string>>();
+            custom_fact_path.insert(custom_fact_path.end(), config_paths.begin(), config_paths.end());
```
branan: Can you verify if `vector::insert` does one allocation or N when passed a random access iterator pair?

lib/src/ruby/module.cc
```diff
@@ -780,8 +791,15 @@ namespace facter { namespace ruby {
                     return true;
                 }
                 instance->_external_search_paths.emplace_back(ruby.to_string(element));
+
```
branan: another random whitespace change

Magisus:
@branan De-duplicating the options descriptions won't work for most of the options that overlap, because the command line needs to use bool switches, while the config file uses explicit boolean key-value pairs. So the only ones that are identical are log-level, external-dir, and custom-dir, which are split across two groupings. So it feels like it'd only be worth it if we did some other refactoring around the way settings are parsed and handled.

PR #1403, opened 2016-08-15
(FACT-1485) Fix Amazon Linux version facts

MikaelSmith:
Facter 2 previously reported the Amazon Linux release version, rather
than an underlying RedHat version. Fix Facter 3 to do the same.

### DIFF
--- a/lib/src/facts/linux/os_linux.cc
+++ b/lib/src/facts/linux/os_linux.cc
@@ -251,6 +251,7 @@ namespace facter { namespace facts { namespace linux {
     {
         // Map of release files that contain a "release X.X.X" on the first line
         static map<string, string> const release_files = {
+                { string(os::amazon),                   string(release_file::amazon) },
                 { string(os::centos),                   string(release_file::redhat) },
                 { string(os::redhat),                   string(release_file::redhat) },
                 { string(os::scientific),               string(release_file::redhat) },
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@branan eligible for a bug fix?

@peterhuene do you remember a reason for the original change in behavior (using the distro release)?

branan:
@MikaelSmith I'm fine with this being a bug fix, unless we have a good historical reason why we changed it in the first place. I'm going to guess it was an oversight, though

MikaelSmith:
Merge it?

whopper:
👍 I remember when we made the original change in 2.2.0. It was one of the two breaking changes we accidentally released and made everyone mad :) (though it's the correct way to report this fact)

MikaelSmith:
Why's it the correct way to report it? The "release" is Amazon's, even though it's a fork of an EL platform.

whopper:
@MikaelSmith I meant that this PR is the correct thing to do. We used to report the kernel version as the "os version" but that was wrong. See Amazon's release notes for 2016.09, for example. Releases are versioned with year and month, like Ubuntu: https://aws.amazon.com/amazon-linux-ami/2016.09-release-notes/

MikaelSmith:
Ah, gotcha.

whopper:
Output with this change:

```
root@jvolfu3lxol7sft ~]# /opt/puppetlabs/bin/facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "Amazon",
  release => {
    full => "2014.03",
    major => "2014",
    minor => "03"
  },
  selinux => {
    enabled => false
  }
}
```

This matches the desired output: 

```
"release"=>{"major"=>"2016", "minor"=>"03", "full"=>"2016.03"}}
```

whopper:
Appveyor seems to be infinitely hanging on PRs today. Noticed the same thing in a puppet PR earlier. I'll check again in the morning.

MikaelSmith:
Probably just backed up, the job seems to be just finishing.

PR #1406, opened 2016-08-17
(FACT-1488) Add git pack config on Windows

johnduarte:
This commit adds git configuration values for pack memory usage.
Prior to this commit cloning the `puppet-win32-ruby` repo failed on
Window 2003.

This change is a belt and suspenders attempt to mitigate the failure
mentioned above. The packing issue should be addressed on the git
remote server side.

### DIFF
--- a/acceptance/setup/common/00_EnvSetup.rb
+++ b/acceptance/setup/common/00_EnvSetup.rb
@@ -46,6 +46,9 @@ hosts.each do |host|
                  '1.9.3-x86'
                end
 
+    on host, 'mkdir .git'
+    on host, 'cmd /c git config pack.windowMemory 10m'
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Does this need to be coordinated with any other changes?

johnduarte:
@MikaelSmith this change is atomic and should not interact with other changes. Thanks.

ferventcoder:
Well that was fast!

kylog:
Perhaps OT but: can we just drop testing on 2003?

ferventcoder:
@kylog I'm under the assumption this is against a version that supported Windows 2003.

ferventcoder:
Other than that 

sure

PR #1408, opened 2016-08-19
(FACT-1489 & 1490) Framework for fact blocking within resolvers

Magisus:
This adds an argument to the `resolve` method of each resolver, allowing a blocklist to be passed into the method. Since we have decided to support blocking of individual top-level facts, and not all resolvers map cleanly to these chunks, it was necessary to have each resolve method handle the list according to its implementation.

As an example of what this might look like, I have implemented the blocking of facts associated with the `filesystem_resolver`, which is an example of a resolver that deals with multiple top-level facts.

I would like to be able to test this blocking functionality, but I can't seem to find any tests that verify the behavior of this resolver. Is it unfeasible to test? If not, what would be the best approach to add these tests?

### DIFF
--- a/lib/src/facts/solaris/filesystem_resolver.cc
+++ b/lib/src/facts/solaris/filesystem_resolver.cc
@@ -1,3 +1,4 @@
+#include <facter/facts/fact.hpp>
 #include <internal/facts/solaris/filesystem_resolver.hpp>
 #include <internal/util/solaris/k_stat.hpp>
 #include <internal/util/scoped_file.hpp>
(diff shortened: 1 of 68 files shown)


lib/src/facts/solaris/filesystem_resolver.cc
```diff
@@ -23,11 +24,21 @@ using namespace boost::filesystem;
...
-        collect_filesystem_data(result);
+
+        if (blocklist.count(fact::mountpoints)) {
+            log_fact_blockage(fact::mountpoints);
+        } else {
+            collect_mountpoint_data(result);
+        }
+
+        if (blocklist.count(fact::filesystems)) {
+            log_fact_blockage(fact::mountpoints);
```
MikaelSmith: Typo here.

MikaelSmith:
@Magisus is this getting reworked based on our discussions?

Magisus:
Yes it is. I'll close this.

PR #1409, opened 2016-08-25
(FACT-1494) facter.conf with external-dir and custom-dir

phongdly:
Acceptance test for default facter.conf file having external-dir and custom-dir defined. 

This PR to ensure that facter will search for external facts and custom facts
in the directory paths defined with external-dir as an array of paths and custome-dir as a single path in the facter.conf file

Default path for facter.conf:
- on *nix: /etc/puppetlabs/facter/facter.conf
- on Windows version older than 6.0: C:\Documents and Settings\All Users\Application Data\PuppetLabs\facter\facter.conf
- on Windows version is 6.0 or newer: C:\ProgramData\PuppetLabs\facter\facter.conf

[skip ci]

### DIFF
--- a/acceptance/lib/facter/acceptance/user_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/user_fact_utils.rb
@@ -42,6 +42,20 @@ module Facter
           '.sh'
         end
       end
+
+      # Retrieve the path to default location of facter.conf file.
+      #
+      def get_default_fact_dir(platform, version)
+        if platform =~ /windows/
+          if version < 6.0
+            File.join('C:', 'Documents and Settings', 'All Users', 'Application Data', 'PuppetLabs', 'facter')
+          else
+            File.join('C:', 'ProgramData', 'PuppetLabs', 'facter')
+          end
+        else
+          File.join('/', 'etc', 'puppetlabs', 'facter')
+        end
+      end
     end
   end
 end
--- a/acceptance/tests/options/config_file/facter_config_file.rb
+++ b/acceptance/tests/options/config_file/facter_config_file.rb
@@ -1,0 +1,73 @@
+# This test is intended to demonstrate that Facter will load a config file
+# saved at the default location without any special command line flags.
+# On *nix, this location is /etc/puppetlabs/facter/facter.conf.
+# On Windows 6.0 or newer, it is C:\ProgramData\PuppetLabs\facter\facter.conf
+# On Windows older than 6.0, it is
+# C:\Documents and Settings\All Users\Application Data\PuppetLabs\facter\facter.conf
+#
+# The test also verifies that facter will search for external facts and custom facts
+# in the directory paths defined with external-dir and custome-dir in the facter.conf file
+test_name "FACT-1494 - C98142 Load facter.conf having external-dir" do
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  agents.each do |agent|
+    # creat paths for default facter.conf, external-dir, and custom-dir
+    #
+
+    # defaul facter.conf
+    facter_conf_default_dir = get_default_fact_dir(agent['platform'], on(agent, facter('kernelmajversion')).stdout.chomp.to_f)
+    facter_conf_default_path = "#{facter_conf_default_dir}/facter.conf"
+
+    # external-dir
+    ext_fact_dir1 = agent.tmpdir('ext_fact_dir1')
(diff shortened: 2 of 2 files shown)


acceptance/lib/facter/acceptance/user_fact_utils.rb
```diff
@@ -42,6 +42,20 @@ def get_external_fact_script_extension(platform)
           '.sh'
         end
       end
+
+      # Retrieve the path to default location of facter.conf file.
+      #
+      def get_default_fact_dir(platform, version)
+        if platform =~ /windows/
+          if version < 6.0
+            File.join('C:', 'Documents and Settings', 'All Users', 'Application Data', 'PuppetLabs', 'facter')
```
kylog: We have dropped Windows 2003 support, which I *think* means we don't need this. @Iristyle can you confirm or correct me?

kylog:
Two more nits:

To spare travis/appveyor bandwidth, can you include `[skip ci]` in the commit message body for changes to acceptance (which are a no-op for what travis/appveyor actually test on this repo)
* https://docs.travis-ci.com/user/customizing-the-build#Skipping-a-build
* https://www.appveyor.com/docs/how-to/filtering-commits/#skip-directive-in-commit-message

Can you put the commit message in the imperative, e.g. as discussed here https://github.com/puppetlabs/facter/blob/master/CONTRIBUTING.md

Thanks!

phongdly:
Thanks @kylog , I added [skip ci] in the commit message body

Magisus:
All of this functionality already has tests written for it, though they are more broken up (see [default file location](https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/default_file_location.rb), [custom dir](https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/custom_facts.rb), and [external dir](https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/external_facts.rb)). The duplication seems unnecessary to me. Since Phong's tests are more concise, maybe we should use this file and remove the other three.

Magisus:
Though come to think of it, that will remove the tests for the disable options (`no-external-facts` and `no-custom-facts`). @phongdly do you feel like those tests have a place in this one you've written, or should they be covered elsewhere?

acceptance/lib/facter/acceptance/user_fact_utils.rb
Magisus: I did not have this case in the tests I wrote for the default location, and it currently passes all of our platforms.

phongdly:
@Magisus thanks for the review. I think we should keep those existing tests because as you mentioned about the tests for disable options, and I might be wrong but they are a little bit different from the test in this PR.
  [default file location] (https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/default_file_location.rb) verify facter.conf is loaded as default but does not test it with external-dir and custom-dir paths.

 [custom dir](https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/custom_facts.rb), and [external dir] (https://github.com/puppetlabs/facter/blob/master/acceptance/tests/options/config_file/external_facts.rb) test the custom and external facts when facter is used with --config flag.

This PR is to test facter will search external and custom facts in the paths provided in facter.conf  and when facter is used without --config flag

Magisus:
Arguably we should refactor all the existing config file tests to use the default location. The only reason they currently don't use it is because that functionality didn't work on all platforms when those tests were written. That refactor, in combination with a test in the command line section that verifies the functionality of the `--config` flag, would make more sense than what we currently have and be a lot cleaner than all these kind-of-overlapping tests.

As a side note: was this ticket (FACT-1494) intended to be a follow up to [FACT-1460](https://tickets.puppetlabs.com/browse/FACT-1460) which causes the config file to be loaded when Facter is called from Ruby? I only ask because this test deals with pieces of that functionality without actually testing the "loading from Ruby" bit. We currently don't have any tests for that behavior, which is kind of scary because that's the part that will impact Puppet.

So if possible, I would be in favor of repurposing THIS test verify that the config file is respected when calling Facter from Ruby, refactoring the existing tests to use the default location, and adding a separate test for the `--config` flag. Since that's kind of a lot of things, I'd also volunteer to do it, if QA has better things to do :P 

Short of that, I think the test itself looks fine :)

phongdly:
@Magisus sounds great if we can also cover the  'loading from Ruby' testing in this test.

MikaelSmith:
This appears to be have been pulled into another PR that I merged. Can I close this one?

Magisus:
Yes I incorporated this into my refactor. I'll close this.

PR #1411, opened 2016-08-26
(FACT-1272) Add tooling for facts POT file

HAIL9000:
In order to support the translation of information about core facts,
add a ruby script which takes the fact.yaml file and generates a POT
file with the important strings so they can be translated.

### DIFF
--- a/lib/schema/core_facts.pot
+++ b/lib/schema/core_facts.pot
@@ -1,0 +1,2017 @@
+# CORE FACTS SCHEMA
+# Copyright (C) 2016 Puppet, LLC
+# This file is distributed under the same license as the FACTER package.
+# FIRST AUTHOR <docs@puppet.com>, 2016.
+#
+#, fuzzy
+msgid ""
+msgstr ""
+"Project-Id-Version: FACTER \n"
+"Report-Msgid-Bugs-To: docs@puppet.com\n"
+"POT-Creation-Date: \n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\n"
+
+
+#. aio_agent_version description
+msgid "Return the version of the puppet-agent package that installed facter."
+msgstr ""
+
+#. aio_agent_version resolution
+msgid "All platforms: use the compile-time enabled version definition."
+msgstr ""
+
+#. architecture description
+msgid "Return the operating system's hardware architecture."
+msgstr ""
+
+#. architecture resolution
+msgid "POSIX platforms: use the `uname` function to retrieve the OS hardware architecture."
+msgstr ""
+
+#. architecture resolution
+msgid "Windows: use the `GetNativeSystemInfo` function to retrieve the OS hardware architecture."
+msgstr ""
+
+#. architecture caveats
+msgid "Linux: Debian, Gentoo, kFreeBSD, and Ubuntu use "amd64" for "x86_64" and Gentoo uses "x86" for "i386"."
+msgstr ""
+
+#. augeas description
+msgid "Return information about augeas."
+msgstr ""
+
+#. augeas resolution
+msgid "All platforms: query augparse for augeas metadata."
+msgstr ""
+
+#. augeasversion description
+msgid "Return the version of augeas."
+msgstr ""
+
+#. augeasversion resolution
+msgid "All platforms: query augparse for the augeas version."
+msgstr ""
+
+#. blockdevices description
+msgid "Return a comma-separated list of block devices."
+msgstr ""
+
+#. blockdevices resolution
+msgid "Linux: parse the contents of `/sys/block/<device>/`."
+msgstr ""
+
+#. blockdevices resolution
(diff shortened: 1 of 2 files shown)


HAIL9000:
@MikaelSmith and @ahenroid thoughts?

HAIL9000:
I was thinking maybe of making it into a command line utility where you specify whether you want to generate a yaml schema from a translate PO file or a POT file from the yaml schema. You could pass in files as arguments, etc

lib/schema/translation_tooling.rb
```diff
@@ -0,0 +1,56 @@
...
+  file.puts <<-HEADER
+# CORE FACTS SCHEMA
+# Copyright (C) 2016 Puppet Labs, LLC
+# This file is distributed under the same license as the Sinatra i18n demo package.
+# FIRST AUTHOR <info@puppet.com>, 2016.
+#
+#, fuzzy
+# msgid ""
+# msgstr ""
+"Project-Id-Version: Sinatra i18n demo init-11-ga552a06\\n"
```
ahenroid: Tiny change, but change the Project-Id-Version to something sensible.

lib/schema/translation_tooling.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  fact_data = {:name => fact_name,
+               :description => fact_details['description'],
+               :resolution => fact_details['resolution'],
+               :caveats => fact_details['caveats']}
+
+  facts << fact_data
+end
+
+File.open(File.join(File.dirname(__FILE__), "core_facts.pot"), 'w') do |file|
```
ahenroid: Do we want the output file to live down in lib/schema/core_facts.pot or possibly up in the top-level locales/ directory? Perhaps @LarissaLane or @MikaelSmith have an opinion?

lib/schema/translation_tooling.rb
```diff
@@ -0,0 +1,56 @@
...
+"MIME-Version: 1.0\\n"
+"Content-Type: text/plain; charset=UTF-8\\n"
+"Content-Transfer-Encoding: 8bit\\n"
+"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\\n"
+  \n
+  HEADER
+
+  facts.each do |fact|
+    if fact[:description]
+      file.puts "# #{fact[:name]} description\nmsgid \"#{fact[:description].tr("\n","")}\"\nmsgstr \"\"\n\n"
```
ahenroid: Also very minor consideration, but the comments in the .pot file do have some special formatting. There is a single character after the '#' that indicates the type of information. I believe "#: [PATH]" indicates the pathname, "#. [NAME]" perhaps indicates the method. There are probably other specifiers. Pick the one that best matches your usage.

MikaelSmith:
I'm not sure what you mean by "generate a yaml schema from a translate PO file". Where would we use that?

lib/schema/translation_tooling.rb
MikaelSmith: `lib/schema`; it's not something we ship or use in the facter executable. If that ever changes we can move it.

lib/schema/translation_tooling.rb
```diff
@@ -0,0 +1,56 @@
...
+# Copyright (C) 2016 Puppet Labs, LLC
+# This file is distributed under the same license as the Sinatra i18n demo package.
+# FIRST AUTHOR <info@puppet.com>, 2016.
+#
+#, fuzzy
+# msgid ""
+# msgstr ""
+"Project-Id-Version: Sinatra i18n demo init-11-ga552a06\\n"
+"\\n"
+"Report-Msgid-Bugs-To: docs@puppetlabs.com\\n"
```
MikaelSmith: Should be `docs@puppet.com`.

lib/schema/translation_tooling.rb
```diff
@@ -0,0 +1,56 @@
...
+"POT-Creation-Date: 2016-08-26 09:00-0000\\n"
+"PO-Revision-Date: #{Time.now}\\n"
```
MikaelSmith: In the Leatherman `.pot` file this is a pattern: `"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"`. I don't think we want the actual date. I'd use the leatherman.pot file as a guide for what the header should look like.

lib/schema/core_facts.pot
```diff
@@ -0,0 +1,1354 @@
...
+# augeasversion resolution
+msgid "All platforms: query augparse for the augeas version."
+msgstr ""
+
+# blockdevices description
+msgid "Return a comma-separated list of block devices."
+msgstr ""
+
+# blockdevices resolution
+msgid "Linux: parse the contents of `/sys/block/<device>/`.Solaris: use the `kstat` function to query disk information."
```
MikaelSmith: I kind of think we should split the resolutions into multiple segments. This might depend on how they're pulled in by the docs script.

HAIL9000: @MikaelSmith Could you say a little bit more about what you mean by that? Or maybe we should chat about it when I'm back in the office

MikaelSmith: I mean right now, all the entries under the `resolution` key are concatenated.

If the script to generate a webpage from the schema does lookup on each entry, i.e. something like
```
schema.each do |fact|
  fact['resolution'].each do |resolution|
    puts resolution
  end
end
```
then lookup for the whole string will fail. So this is an area where your script and the docs script to generate webpages need to align.

HAIL9000: Ah, so are you suggesting inserting like a space or a new line?

MikaelSmith: Or dealing with them as separate translation strings. You can iterate over each item under `resolution`.

HAIL9000:
@MikaelSmith @ahenroid Updated and addressed comments!

MikaelSmith:
👍

PR #1412, opened 2016-08-29
(FACT-1482) Add mountpoints fact entries for tmpfs-mounted filesystems

ahenroid:
This change updates code that previously filtered-out all non-device
entries, now allowing tmpfs.


MikaelSmith:
👍

MikaelSmith:
Restricted to Linux, so this shouldn't regress FACT-1345.

ahenroid:
Yes, this should not regress FACT-1345 or significantly enlarge the mountpoints fact.
There was some discussion about adding all mountpoints for virtual filesystems to the
fact, but in the end, we settled on just adding tmpfs.

PR #1415, opened 2016-08-30
(maint) Fix string conversion bug

ahenroid:


Magisus:
This went through a full AIX build successfully, and Appveyor seems happy now, so merging.

locales/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-gettext_templates(${CMAKE_CURRENT_SOURCE_DIR} ${ALL_SOURCES})
-gettext_compile(${CMAKE_CURRENT_SOURCE_DIR} share/locale)
-SET_DIRECTORY_PROPERTIES(PROPERTIES CLEAN_NO_CUSTOM TRUE)
-export_var(GETTEXT_ENABLED)
+if (NOT WIN32)
+    gettext_templates(${CMAKE_CURRENT_SOURCE_DIR} ${ALL_SOURCES})
```
MikaelSmith: What happened here?

lib/inc/internal/util/aix/odm.hpp
```diff
@@ -66,13 +63,13 @@ namespace facter { namespace util { namespace aix {
           * Get the error string for an ODM error state.
           * @return an error string owned by the ODM subsystem
           */
-         static const char* error_string() {
+         static std::string error_string() {
              char* msg;
              int result = odm_err_msg(odmerrno, &msg);
              if (result < 0) {
-                 return _("failed to retrieve ODM error message");
+                 return leatherman::locale::format("failed to retrieve ODM error message");
```
MikaelSmith: Why couldn't you continue to use the `_` shortcut?

locales/CMakeLists.txt
ahenroid: Oh, I'll take a look at this in more detail. I didn't realize there was a prevention mechanism already
in place. Appveyor suddenly started trying to build the .pot file, breaking all Windows builds.

MikaelSmith: Ah, for that we should probably update the version of gettext used in AppVeyor.

ahenroid: Note that the build breakage is not actually caused by gettext itself. I believe the issue is a Windows command line length limit. With a huge number of files on the command line to xgettext, some number of arguments are truncated. We may need to change how xgettext is called for Windows. For projects with smaller number of source files (e.g. cpp-hocon and even leatherman), xgettext works just fine on Windows.

MikaelSmith: Ah, sounds like it. Can you file a ticket to make sure that gets fixed?

ahenroid: Yes, I just filed FACT-1500 for this. To me, it made sense to file this under Facter, even though much
of the fix will happen in Leatherman. The notes should make this clear.

PR #1416, opened 2016-08-31
(maint) Sign packages with new signing key

melissa:
This commit updates the key used to signed packages with the automation
in the packaging repo. We are shipping a new gpg key for all publicly
available packages, so this is a part of that effort.


melissa:
Currently blocked on getting some releases out the door, and then we'll go and bulk merge everything. I'm not sure which branches are currently in development, so I did everything that could potentially have a release in the future. If for some reason a release is cut from a branch that does not have these changes, we will see failures.

ScottGarman:
:+1:

PR #1419, opened 2016-09-01
(FACT-718) Allow blocking resolvers via a list in the config file

Magisus:
New approach: resolvers are blockable by name, and this mapping is documented in the schema with a `blockgroup` tag. Blocking a fact with this tag will also result in all other facts in that group (associated with that resolver) being blocked. Resolvers implement a method describing whether they are blockable or not. If the schema does not list a block group for a given fact, that fact (or rather its associated resolver) cannot be blocked. Usually this is due to some dependency, i.e. other facts rely on the output of that resolver for their own resolution. 

"Blockability" is currently opt-in: to make a resolver blockable, override the is_blockable method to return true, and add the appropriate blockgroup tag to its associated facts in the schema.

If this approach looks sane, I will more thoroughly investigate which resolvers it makes sense to block, and add the appropriate code.

### DIFF
--- a/lib/tests/facts/schema.cc
+++ b/lib/tests/facts/schema.cc
@@ -96,6 +96,15 @@ struct dmi_resolver : resolvers::dmi_resolver
     }
 };
 
+struct ec2_resolver : resolvers::ec2_resolver
+{
+    void resolve(collection& facts) override
+    {
+        facts.add(fact::ec2_metadata, make_value<map_value>());
+        facts.add(fact::ec2_userdata, make_value<string_value>("user data"));
+    }
+};
+
 struct filesystem_resolver : resolvers::filesystem_resolver
 {
  protected:
@@ -434,8 +443,7 @@ void add_all_facts(collection& facts)
     facts.add(make_shared<dmi_resolver>());
     facts.add(make_shared<filesystem_resolver>());
     // TODO: refactor the EC2 resolver to use the "collect_data" pattern
-    facts.add(fact::ec2_metadata, make_value<map_value>());
-    facts.add(fact::ec2_userdata, make_value<string_value>("user data"));
+    facts.add(make_shared<ec2_resolver>());
     // TODO: refactor the GCE resolver to use the "collect_data" pattern
     facts.add(fact::gce, make_value<map_value>());
     facts.add(make_shared<identity_resolver>());
@@ -475,7 +483,8 @@ void validate_attributes(YAML::Node const& node)
             Catch::Equals("resolution")).add(
             Catch::Equals("caveats")).add(
             Catch::Equals("elements")).add(
-            Catch::Equals("validate"))
+            Catch::Equals("validate")).add(
+            Catch::Equals("blockgroup"))
         );
     }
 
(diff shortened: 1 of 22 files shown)


lib/tests/facts/schema.cc
```diff
@@ -560,6 +561,13 @@ void validate_attributes(YAML::Node const& node)
         REQUIRE_FALSE(caveats.empty());
     }
 
+    auto blockgroup_attribute = node["blockgroup"];
```
Magisus: Ideally I would also like to be able to test that each blockgroup attribute corresponds to the name of a resolver that has is_blockable() == true. But like elsewhere in this code base, these tests do not currently assume or validate any mapping between facts and specific resolvers. Since this attribute depends on that mapping, is that something we would want to start validating?

MikaelSmith: We could create new collections with the blockgroup specified, and confirm that the fact doesn't resolve. It doesn't seem unreasonable to test mapping between facts and resolvers, though.

Magisus:
This makes blockable the two fact groups (mountpoints and EC2) most requested by users.

glennsarti:
Should there be documentation examples on how to use this added to this PR or is that a later ticket?

MikaelSmith:
There's a Facter man page we could add this to. It probably needs an update for the config file in general.

Magisus:
Are there any more fact groups that come to mind as something we definitely want to be able to block? So far I've added code for file system facts (per complaints about mountpoints being slow) and EC2 facts (per complaints about hitting an invalid IP when trying to pointlessly detect EC2).

glennsarti:
:+1:  on doco updates!

glennsarti:
@Magisus Would you be able to add some simple spec tests for the behavior e.g. Should return some fact info if not blocked, should return a warning and no fact info if blocked.

Magisus:
Do you want some tests above and beyond those I added as part of schema validation? Those construct a blocklist of all blockable resolvers, then attempt to resolve all resolvers, and verify that no facts from blocked resolvers appear.

MikaelSmith:
The spec tests should be runnable everywhere. I'd rather see an acceptance test.

Magisus:
I would like to have an acceptance test as well. We don't currently have any acceptance tests that deal with either of the resolvers made blockable by this PR (though there is one ready to merge that exercises the `mountpoints` fact), but it's probably feasible to add one. That would give us end-to-end validation.

The collection unit tests mock out a resolver, and it would be fairly easy to make that resolver blockable and test the collection with and without a blocklist specified (passed directly into the collection, not via the config file). Should still be runnable everywhere.

glennsarti:
@Magisus I meant unit/integration tests.  IMO unit tests for that behaviour would be preferrable over an acceptance test.  Quicker feedback and _should_ be simple to do although I don't know the C++ testing framework at all.

Magisus:
Yeah sounds good. I do already have some of those (see linked section in lib/tests/facts/schema.cc) that rely on mocked resolvers. Those tests verify that if a fact has a blockgroup in the schema, it corresponds to a blockable resolver, and if the blockgroup appears in the blocklist passed to a collection, the facts will not be resolved.

glennsarti:
FACT-1507 has been raised to address acceptance tests.

PR #1421, opened 2016-09-06
(FACT-1502) acceptance: mount a tmpfs directory and check the fact

phongdly:
New acceptance test for verifying that 'facter mountpoints' show a tmpfs mount.

[skip ci]

### DIFF
--- a/acceptance/tests/verify_mountpoints.rb
+++ b/acceptance/tests/verify_mountpoints.rb
@@ -1,7 +1,7 @@
 # This test is intended to demonstrate that mount resource can mount tmpfs directory
-# and the mount facter mountpoints shouls show the mount on tmpfs
+# and the mount facter mountpoints should show the mount on tmpfs
 
-test_name "FACT-1052 - C98163 mountpoints fact should show mounts on tmpfs" do
+test_name "FACT-1502 - C98163 mountpoints fact should show mounts on tmpfs" do
   require 'facter/acceptance/user_fact_utils'
   extend Facter::Acceptance::UserFactUtils
 
--- a/acceptance/tests/verify_mountpoints.rb
+++ b/acceptance/tests/verify_mountpoints.rb
@@ -1,0 +1,41 @@
+# This test is intended to demonstrate that mount resource can mount tmpfs directory
+# and the mount facter mountpoints shouls show the mount on tmpfs
+
+test_name "FACT-1052 - C98163 mountpoints fact should show mounts on tmpfs" do
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  confine :except, :platform => 'windows'
+
+  agents.each do |agent|
+    dir = '/tmp/tempdir'
+    on(agent, "mkdir -p #{dir}")
+    manifest  = 'mount_manifest.pp'
+
+    teardown do
+      on(agent, "umount #{dir}")
+      on(agent, "rm -rf #{dir} #{manifest}")
+    end
+
+    create_remote_file(agent, manifest, <<-FILE)
+      mount {"#{dir}":
+        ensure  => mounted,
+        options => 'noexec',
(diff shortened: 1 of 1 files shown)


acceptance/tests/verify_mountpoints.rb
```diff
@@ -0,0 +1,41 @@
+# This test is intended to demonstrate that mount resource can mount tmpfs directory
+# and the mount facter mountpoints shouls show the mount on tmpfs
+
+test_name "FACT-1052 - C98163 mountpoints fact should show mounts on tmpfs" do
```
Magisus: Ticket number should be 1502.

And out of curiosity, what does the C-number mean?

acceptance/tests/verify_mountpoints.rb
```diff
@@ -0,0 +1,41 @@
+# This test is intended to demonstrate that mount resource can mount tmpfs directory
+# and the mount facter mountpoints shouls show the mount on tmpfs
```
Magisus: typo here: "shouls" -> "should"

Magisus:
Verified that the test passes, 👍 from me except for those nitpicks

acceptance/tests/verify_mountpoints.rb
glennsarti: That's a reference to a TestRail case.

phongdly:
Thanks @Magisus and @glennsarti for the review, I just pushed a new update based on @Magisus feedback

acceptance/tests/verify_mountpoints.rb
```diff
@@ -0,0 +1,41 @@
...
+        options => 'noexec',
+        fstype  => 'tmpfs',
+        device  => 'tmpfs',
+        atboot  => true,
+      }
+    FILE
+
+    step "Apply the manifest to mount" do
+      on(agent, puppet("apply #{manifest}")) do
+        assert_no_match(/Error/, stdout, "Unexpected error was detected!")
```
MosesMendoza: Is there a way to use exit codes to tell if this was successful instead of matching against stdout? This will help future-proof us against tests against different languages.

Magisus: I think you should be able to add `:acceptable_exit_codes => [0]` as a third argument to the on(agent, ...) call. It will fail if it gives any non-zero exit code.

glennsarti: I was under the impression that it was implicitly `:acceptable_exit_codes => [0]` unless you specify otherwise

MosesMendoza:
Perhaps we want to check that puppet successfully applied the catalog and resources were changed? If so I think we want detailed exit codes and to check for the right one.

acceptance/tests/verify_mountpoints.rb
glennsarti: https://github.com/puppetlabs/beaker/blob/master/docs/concepts/shared_options_for_executing_beaker_commands.md#acceptable_exit_codes

Magisus:
+1 to @MosesMendoza 's comment

phongdly:
@Magisus @MosesMendoza @glennsarti : updated acceptable exit codes.

acceptance/tests/verify_mountpoints.rb
```diff
@@ -27,8 +27,8 @@
       }
     FILE
 
-    step "Apply the manifest to mount" do
-      on(agent, puppet("apply #{manifest}")) do
+    step "Apply the manifest to mount directory '#{dir}'" do
+      on(agent, puppet("apply #{manifest}"), :acceptable_exit_codes => [0,2]) do
```
Magisus: Did we also want to add the `--detailed-exitcodes` flag?

Magisus: Oh, good to know!

phongdly: image

PR #1424, opened 2016-09-14
	(FACT-1356) Allow executable facts to output Yaml/Json 

branan:
Previously, executable facts could only return simple key/value pairs. This code now attempts to parse executable output as Yaml, and only falls back to the key/value parsing if we do not get a useful result. Because Yaml is a superset of Json, we do not have to separately try to parse Json output.

This also moves the helper code for converting a Yaml object into facts into a new shared location where both the flat-file yaml resolver and the executable resolver can share it.

### DIFF
--- a/lib/src/facts/external/execution_resolver.cc
+++ b/lib/src/facts/external/execution_resolver.cc
@@ -37,7 +37,8 @@ namespace facter { namespace facts { namespace external {
                 {
                     execution_options::trim_output,
                     execution_options::merge_environment,
-                    execution_options::throw_on_failure
+                    execution_options::throw_on_failure,
+                    execution_options::convert_newlines
                 });
 
             try {
--- a/lib/src/facts/external/execution_resolver.cc
+++ b/lib/src/facts/external/execution_resolver.cc
@@ -3,15 +3,18 @@
 #include <facter/facts/array_value.hpp>
 #include <facter/facts/map_value.hpp>
 #include <facter/facts/scalar_value.hpp>
+#include <internal/util/yaml.hpp>
 
 #include <leatherman/execution/execution.hpp>
 #include <leatherman/logging/logging.hpp>
 #include <boost/algorithm/string.hpp>
+#include <yaml-cpp/yaml.h>
 
 using namespace std;
 using namespace leatherman::execution;
 using namespace facter::facts;
 using namespace facter::facts::external;
+using namespace facter::util::yaml;
 
 namespace facter { namespace facts { namespace external {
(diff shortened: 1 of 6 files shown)


glennsarti:
Do we need to be concerned about UTF8 being parsed correctly into a structured fact?

branan:
@glennsarti As long as the system locale/encoding is set correctly we will be well-behaved. If a program gives us output that doesn't match that we can't really be trusted... but that was also the case for key/value facts

glennsarti:
Ahh ok. So no worse off.

branan:
Quick force-push for the linter. I'm not sure what happened on appveyor :/

branan:
The windows failure is newline conversion. I'm going to go ahead and open a leatherman ticket to add a "this program outputs text, please convert its newlines" flag to the execution helpers.

branan:
I've just pushed up a commit to use the newline conversion code I'm working on for leatherman. This will cause travis to fail until we get leatherman updated and released with that feature

branan:
https://github.com/puppetlabs/leatherman/pull/208 contains the leatherman PR. We'll need to cut a leatherman release once that's landed in order to get this in.

Iristyle:
As I commented on the other PR, I don't understand why the parsing doesn't use the appropriate line endings for a given platform?  Where does the failure occur - in libYaml?

lib/src/facts/external/execution_resolver.cc
```diff
@@ -29,38 +32,51 @@ namespace facter { namespace facts { namespace external {
...
+                    // interpretation in those cases.
+                    resolved = true;
+                }
+            }  catch (YAML::Exception& ex) {
+                LOG_DEBUG("Could not parse executable fact output as YAML ({1})", ex.msg);
+            }
+
+            if (!resolved) {
+                std::vector<string> lines;
+                boost::split(lines, result.output, boost::is_any_of("\n"));
```
branan: @Iristyle The windows issue is this split. I could definitely just deal with the `\r` here, but it feels cleaner to me to let the I/O layer deal with that sort of platform-specific stuff. Maybe that just comes from me being used to `stdio` doing it, though.

glennsarti:
You'd probably want a consistent line ending for all plats.  What would happen on Windows if you gave it only LFs instead of CRLFs

Magisus:
I've opened https://github.com/puppetlabs/facter/pull/1449 with everything in this PR, plus additional tests and the leatherman bump. Closing this in favor of that one.

PR #1425, opened 2016-09-14
(FACT-348) Allow fact caching

Magisus:
This allows a `ttls` element to be specified in the config file with the following structure:
`facts : { ttls : [ { "block group" : 30 days } ] }`. Facts in the specified block groups will be cached as json files and loaded from those files until the files expire, i.e. the file is older than the specified duration.

This can be used with any fact that has a blockgroup specified in the schema.

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -128,6 +128,7 @@ int main(int argc, char **argv)
 
         vector<string> external_directories;
         vector<string> custom_directories;
+        unordered_map<string, int64_t> ttls;
 
         // Build a list of options visible on the command line
         // Keep this list sorted alphabetically
@@ -190,7 +191,9 @@ int main(int argc, char **argv)
                 load_global_settings(hocon_conf, vm);
                 load_cli_settings(hocon_conf, vm);
                 load_fact_settings(hocon_conf, vm);
+                ttls = load_ttls(hocon_conf);
             }
+
 
             // Check for a help option first before notifying
             if (vm.count("help")) {
@@ -294,7 +297,7 @@ int main(int argc, char **argv)
             auto facts_to_block = vm["blocklist"].as<vector<string>>();
             blocklist.insert(facts_to_block.begin(), facts_to_block.end());
         }
-        collection facts(blocklist);
+        collection facts(blocklist, ttls);
         facts.add_default_facts(ruby);
 
         // Add the environment facts
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -148,6 +148,7 @@ int main(int argc, char **argv)
             ("help,h", _("Print this help message.").c_str())
             ("json,j", _("Output in JSON format.").c_str())
             ("show-legacy", _("Show legacy facts when querying all facts.").c_str())
+            ("list-cache-groups", _("List the name of each cacheable group of facts").c_str())
             ("log-level,l", po::value<level>()->default_value(level::warning, "warn"), _("Set logging level.\nSupported levels are: none, trace, debug, info, warn, error, and fatal.").c_str())
             ("no-cache", _("Disable loading and refreshing facts from the cache").c_str())
             ("no-color", _("Disables color output.").c_str())
@@ -237,6 +238,17 @@ int main(int argc, char **argv)
             colorize(boost::nowide::cerr);
             help(visible_options);
             return EXIT_FAILURE;
+        }
+
+        // Check for listing fact groups
+        if (vm.count("list-cache-groups")) {
+            collection facts;
+            facts.add_default_facts(!vm.count("no-ruby"));
+            vector<string> fact_groups = facts.get_fact_groups();
+            for (auto group : fact_groups) {
+                boost::nowide::cout << group << endl;
+            }
+            return EXIT_SUCCESS;
         }
 
         // Check for printing the version
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -149,6 +149,7 @@ int main(int argc, char **argv)
             ("json,j", _("Output in JSON format.").c_str())
             ("show-legacy", _("Show legacy facts when querying all facts.").c_str())
             ("log-level,l", po::value<level>()->default_value(level::warning, "warn"), _("Set logging level.\nSupported levels are: none, trace, debug, info, warn, error, and fatal.").c_str())
+            ("no-cache", _("Disable loading and refreshing facts from the cache").c_str())
             ("no-color", _("Disables color output.").c_str())
             ("no-custom-facts", po::bool_switch()->default_value(false), _("Disables custom facts.").c_str())
             ("no-external-facts", po::bool_switch()->default_value(false), _("Disables external facts.").c_str())
@@ -191,9 +192,10 @@ int main(int argc, char **argv)
                 load_global_settings(hocon_conf, vm);
                 load_cli_settings(hocon_conf, vm);
                 load_fact_settings(hocon_conf, vm);
-                ttls = load_ttls(hocon_conf);
+                if (!vm.count("no-cache")) {
+                    ttls = load_ttls(hocon_conf);
+                }
             }
-
 
             // Check for a help option first before notifying
             if (vm.count("help")) {
--- a/lib/src/facts/cache.cc
+++ b/lib/src/facts/cache.cc
@@ -1,0 +1,109 @@
+#include <internal/facts/cache.hpp>
+#include <internal/facts/external/json_resolver.hpp>
+
+#include <rapidjson/document.h>
+#include <rapidjson/prettywriter.h>
+#include <leatherman/logging/logging.hpp>
+#include <leatherman/file_util/directory.hpp>
+#include <leatherman/file_util/file.hpp>
+#include <boost/nowide/fstream.hpp>
+#include <time.h>
+
+using namespace std;
+using namespace rapidjson;
+using namespace facter::facts::external;
+namespace boost_file = boost::filesystem;
+
+namespace facter { namespace facts { namespace cache {
+
+    bool cache_is_valid(boost_file::path const& cache_dir, int64_t ttl) {
+        time_t last_mod = boost_file::last_write_time(cache_dir);
+        time_t now;
+        double lifetime_seconds = difftime(time(&now), last_mod);
+        return static_cast<int64_t>(lifetime_seconds) < ttl;
+    }
+
+    void load_facts_from_cache(boost_file::path const& cache_dir, vector<string> const& cached_facts, collection& facts) {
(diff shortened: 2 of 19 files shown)


Magisus:
I have manually verified this both on my laptop and on a Windows VM, but I would like some ideas about what kind of automated tests we would like to see. Does it make sense to try to mock out some stuff and add unit tests? Or should this only be tested at the integration level?

lib/src/facts/cache.cc
```diff
@@ -0,0 +1,108 @@
...
+        if (boost_file::is_directory(cache_dir) && cache_is_valid(cache_dir, ttl)) {
+            LOG_DEBUG("loading cached values for {1} facts", res->name());
+            load_facts_from_cache(cache_dir, res->names(), facts);
+        } else {
+            LOG_DEBUG("caching values for {1} facts", res->name());
+            refresh_cache(res, cache_dir, facts);
+        }
+    }
+
+    struct stream_adapter
```
Magisus: This is copied from the collection, what is a good spot to put this so they both can share?

.travis.yml
```diff
@@ -43,7 +43,7 @@ env:
     - PATH=$USERDIR/bin:$PATH
     - LD_LIBRARY_PATH=$USERDIR/lib:$LD_LIBRARY_PATH
     - LEATHERMAN_VERSION=0.9.0
-    - CPPHOCON_VERSION=0.1.2
+    - CPPHOCON_VERSION=0.1.3
```
glennsarti: Should this be split into a seprate maint commit?

appveyor.yml
```diff
@@ -2,7 +2,7 @@ version: 3.1.0.{build}
 clone_depth: 10
 environment:
   LEATHERMAN_VERSION: 0.9.0
-  CPPHOCON_VERSION: 0.1.2
+  CPPHOCON_VERSION: 0.1.3
```
glennsarti: Should this be split into a seprate maint commit?

lib/src/facts/cache.cc
```diff
@@ -0,0 +1,108 @@
...
+using namespace rapidjson;
+using namespace facter::facts::external;
+namespace boost_file = boost::filesystem;
+
+namespace facter { namespace facts { namespace cache {
+
+    bool cache_is_valid(boost_file::path const& cache_dir, int64_t ttl) {
+        time_t last_mod = boost_file::last_write_time(cache_dir);
+        time_t now;
+        double lifetime_seconds = difftime(time(&now), last_mod);
```
Magisus: Both of these times are reported as unix timestamps (i.e. the number of seconds since 00:00 hours, Jan 1, 1970 UTC), and should not be affected by changing time zones.

glennsarti:
I've manually tested modifying the timezone on Windows and it does not affect the TTL calculation.  Changing the time itself does affect the TTL calculation but that is to be expected.

lib/src/facts/cache.cc
```diff
@@ -0,0 +1,116 @@
...
+
+using namespace std;
+using namespace rapidjson;
+using namespace facter::facts::external;
+namespace boost_file = boost::filesystem;
+
+namespace facter { namespace facts { namespace cache {
+
+    bool cache_is_valid(boost_file::path const& cache_dir, int64_t ttl) {
+        time_t last_mod = boost_file::last_write_time(cache_dir);
```
glennsarti: The last modified datetime stamps on directories under Windows are notoriously erroneous. Given that only puppet _should_ be modifying these it may be ok, but I'm not comfortable with the directory based timestamps.  The file based stamps are far more reliable

.travis.yml
Magisus: I'm not sure, given that the rest of the functionality in this PR doesn't work at all without it. What do you think? Do we usually always split it out?

glennsarti: Ahh ok..In that case perhaps a mention in the commit message as to why hocon is being bumped for a fact caching PR.

Magisus: Will do!

lib/src/facts/cache.cc
Magisus: Okay good to know. That is easier to do now that we have changed it to always produce a file for every fact, collected or not. Originally I checked the directory because I couldn't predict which facts would actually be in it.

Magisus:
This can't be merged until the new version of HOCON is tagged and some Facter CMake files are updated to handle new i18n features, see https://github.com/puppetlabs/cpp-hocon/pull/79 and https://github.com/puppetlabs/facter/pull/1432

Magisus:
Everything has been updated in puppet-agent to handle the cpp-hocon version bump, this is good to go through when review is complete.

lib/src/facts/cache.cc
Magisus: @glennsarti I did some investigation into writing all the facts for a given resolver into one file, implementation is here: https://github.com/Magisus/facter/commit/bc9087e84805a3378d086cd714346ef51c57a4d8

Magisus:
I would love to get a second set of eyes on this, @tylerpace or maybe @MikaelSmith , would one of you mind taking a look when you get a chance?

acceptance/tests/options/ignore_cache.rb
```diff
@@ -0,0 +1,53 @@
...
+  config = <<-FILE
+  cli : { debug : true }
+  facts : { ttls : [ "kernel" : 30 minutes ] }
+  FILE
+
+  agents.each do |agent|
+    config_dir = get_default_fact_dir(agent['platform'], on(agent, facter('kernelmajversion')).stdout.chomp.to_f)
+    config_file = File.join(config_dir, "facter.conf")
+
+    cached_facts_dir = get_cached_facts_dir(agent['platform'], on(agent, facter('kernelmajversion')).stdout.chomp.to_f)
```
MikaelSmith: Might be nice to re-use the kernelmajversion so we don't query Facter twice.

acceptance/tests/options/ignore_cache.rb
```diff
@@ -0,0 +1,53 @@
+test_case "--ignore-cache command-line option causes the fact cache to be ignored" do
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  # the kernel fact should be resolvable on ALL systems
+  cached_fact_name = "kernel"
+
+  config = <<-FILE
+  cli : { debug : true }
+  facts : { ttls : [ "kernel" : 30 minutes ] }
```
MikaelSmith: Can we also test expiration by modifying mtime on the cache file (make it appear older than 30 minutes) and trying again?

exe/facter.cc
```diff
@@ -147,7 +148,9 @@ int main(int argc, char **argv)
             ("help,h", _("Print this help message.").c_str())
             ("json,j", _("Output in JSON format.").c_str())
             ("show-legacy", _("Show legacy facts when querying all facts.").c_str())
+            ("list-cache-groups", _("List the name of each cacheable group of facts").c_str())
```
MikaelSmith: Can we acceptance test `list-cache-groups`?

lib/src/facts/cache.cc
MikaelSmith: I don't have a great recommendation. File stat can be a little slow, so avoiding stat-ing every file is useful, but that trades off against potentially reading and updating a large file if a lot of facts are cached (even for a single update). With that trade-off, I'd lean towards following the original RFC.

MikaelSmith: Actually, I'm not sure why we're treating the whole directory as valid/invalid. Isn't the ttl per resolver? So we'd want to check validity per resolver?

lib/src/facts/cache.cc
MikaelSmith: Anonymous namespace in collection.hpp makes sense to me.

lib/src/facts/collection.cc
```diff
@@ -319,20 +329,33 @@ namespace facter { namespace facts {
...
+
+        // Check if the resolver has been blocked
+        if (try_block(res)) {
+            return;
+        }
+
+        // Check if the resolver should be cached
+        auto resolver_ttl = _ttls.find(res->name());
+        if (resolver_ttl != _ttls.end()) {
+           cache::use_cache(*this, res, (*resolver_ttl).second);
```
MikaelSmith: This happens before resolving facts. So I don't understand how this will result in caching the correct values if the cache is deemed invalid.

MikaelSmith: Ah right, the resolution happens in refresh_cache.

lib/src/facts/cache.cc
Magisus: There is a subdirectory for the facts from each resolver. I'm not checking the whole cache directory, but each subdirectory.

exe/facter.cc
Magisus: added

lib/src/facts/cache.cc
MikaelSmith: Ah, then a single file per resolver seems fine.

Magisus: Okay I'll get that update merged in then.

ahpook: @Magisus do the contents of a given file map directly back to the config file? that is, i would expect that if i configure 

`facts : { ttls : [ { "os" : 30 days } ] }`

all of the `os` facts would be collected in a batch and kept in one file. I *think* that's what you're saying - if so i am enthusiastic 👍

acceptance/tests/options/ignore_cache.rb
Magisus: Added this test case

acceptance/tests/options/ignore_cache.rb
Magisus: udpated

lib/src/facts/cache.cc
Magisus: Yes, this is what would happen. Thanks for the input, I'll make this change.

Magisus:
@glennsarti has some acceptance tests [here](https://github.com/glennsarti/facter/tree/fact-348-tests) for the feature itself. We can either try to get those into this PR, or open a separate ticket/PR for them.

MikaelSmith:
I don't have much time to review. Overall seems good.

Magisus:
No problem @MikaelSmith, thanks so much for your input!

tylerpace:
@Magisus +1 from me for `list-cache-groups` and `no-cache`.

Magisus:
The new tests look good. I edited my `--no-cache` tests to not use the kernel fact. Running these through a small pipeline, if green, I'm 👍  to merge this!

PR #1426, opened 2016-09-14
(maint) Use explicit set constructor

Magisus:
Some compilers do not allow the use of `{}` to initialize an empty std::set. Use the explicit constructor instead.


MosesMendoza:
@Magisus it would be good to include the information and related ticket in the commit message too? thoughts?

Magisus:
Sorry about that @MosesMendoza , yes that probably would have been useful.

MosesMendoza:
👍

PR #1429, opened 2016-09-20
(FACT-1507) acceptance: verify that blocklist prevents fact resolution

Magisus:


acceptance/tests/options/config_file/blocklist.rb
```diff
@@ -0,0 +1,57 @@
...
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  agents.each do |agent|
+    step "facts should be blocked when Facter is run from the command line" do
+      step "Agent #{agent}: create config file" do
+        custom_conf_dir = agent.tmpdir("config_dir")
+        config_file = File.join(custom_conf_dir, "facter.conf")
+        create_remote_file(agent, config_file, <<-FILE)
+        facts : { blocklist : [ "file system" ] }
```
glennsarti: Would like to add the EC2 blocking here too.  That way we can test the `--debug` output on all platforms, as the `file systems` block group does not attempt resolution on all platforms

acceptance/tests/options/config_file/blocklist.rb
```diff
@@ -0,0 +1,57 @@
...
+        FILE
+
+        teardown do
+          on(agent, "rm -rf '#{custom_conf_dir}'", :acceptable_exit_codes => [0, 1])
+        end
+
+        step "blocked facts should not be resolved" do
```
glennsarti: Would like to add an additional test for debug output shows when a fact is being blocked e.g.

Add `cli: { debug : true }` to config file

```
assert_match(/blocking collection of .+ facts/, stderr, "Expected stderr to contain statement about blocking fact collection")
```

Magisus:
Originally I was going to examine the debug output, but I got worried after hearing about the I18N issues that might cause. Do we foresee that ever being a problem here? Or will we only ever run these tests on an English system?

If it's not a problem, I think adding those things is a great idea. I wasn't sure if there are systems where file system facts ALL don't work.

glennsarti:
Hrmmm good point about the localisation.  The test still seems valid though.  `As a user I would like to know when facts are blocked as opposed to not resolved`.  As these are acceptance tests we do have control over what language we test against, and I don't see us testing non-english error messages in the near term.  So perhaps we accept the technical debt for this test?

Magisus:
Sounds good, I'll add those cases.

acceptance/tests/options/config_file/blocklist.rb
Magisus: Updated

acceptance/tests/options/config_file/blocklist.rb
Magisus: Updated

acceptance/tests/options/config_file/blocklist.rb
```diff
@@ -0,0 +1,57 @@
...
+          on(agent, facter("--config '#{config_file}'")) do
+            assert_no_match(/filesystems/, stdout, "filesystems fact should have been blocked")
```
Magisus: Updated. I think it's useful to still test both, because filesystem will actually produce some output on some systems at least.

PR #1433, opened 2016-09-23
(FACT-1484) Change failure of getaddrinfo from error to warning

Magisus:
When this is logged as an error, it frequently causes the Facter smoke
tests to fail on Cisco WR Linux when running though CI pipelines.
However, Facter is still able to return good output. Changing this to a
warning allows the test to pass, while still reporting the failure of
the system call. This is more consistent with the way we handle these
failures elsewhere in Facter.


ScottGarman:
:+1:

branan:
I agree this is more consistent with the way we generally handle resolution failures.

PR #1437, opened 2016-09-30
(FACT-1492) Correctly report version numbers for Windows 10 and Server 2016

Magisus:
This adds consistent reporting of the Windows version fact for Windows
10 and Windows Server 2016. Before this, we were returning the raw
kernel version string.

### DIFF
--- a/acceptance/tests/facts/windows.rb
+++ b/acceptance/tests/facts/windows.rb
@@ -22,7 +22,10 @@ agents.each do |agent|
     os_version  = '2012 R2'
     kernel_version = '6.3'
   elsif agent['platform'] =~ /-10/
-    os_version  = /^10\./
+    os_version  = '10'
+    kernel_version = /^10\./
+  elsif agent['platform'] =~ /2016/
+    os_version = '2016'
     kernel_version = /^10\./
   else
     raise "Unknown agent platform of #{agent['platform']}"
--- a/lib/src/facts/windows/operating_system_resolver.cc
+++ b/lib/src/facts/windows/operating_system_resolver.cc
@@ -100,8 +100,8 @@ namespace facter { namespace facts { namespace windows {
         // Override default release with Windows release names
         auto version = result.release.substr(0, lastDot);
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
-        if (version == "6.4") {
-            result.release = consumerrel ? "10" : result.release;
+        if (version == "10.0") {
+            result.release = consumerrel ? "10" : "2016";
         } else if (version == "6.3") {
             result.release = consumerrel ? "8.1" : "2012 R2";
         } else if (version == "6.2") {


lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -100,8 +100,8 @@ namespace facter { namespace facts { namespace windows {
         // Override default release with Windows release names
         auto version = result.release.substr(0, lastDot);
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
-        if (version == "6.4") {
```
Magisus: Does anyone know of any reason we should keep the 6.4 in here? According to everything I found, this version doesn't exist; it was replaced in favor of 10.0. I tried this on both a Windows 10 and a 2016 box, and both returned 10.0.x as the kernel version.

MosesMendoza:
Reviewing

lib/src/facts/windows/operating_system_resolver.cc
Iristyle: The 6.4 bit was added in the original 3.0.0 release at https://github.com/puppetlabs/facter/commit/c88eaa356081cde780613d11199c5e829880152a

I removed `6.4` in the Facter 2 branch at https://github.com/puppetlabs/facter/commit/3fedaa4ad106a2d445261f1b8a6d4133b16880a8

And it looks like we can trace the original `6.4`back to @whopper at https://github.com/puppetlabs/facter/commit/1faa112b173ee305bbffcfeb9b1b764f68e9a9ac

My assessment here is that we don't want 6.4 anywhere in our code ;0

whopper: 👍 oh yeah, I remember that! That was just meant to try to futureproof that code.

Iristyle: So one question that comes to mind here.... What about Windows 10 anniversary update?  That build has different behavior than the existing Windows 10 builds (and appears to start around build 14295).

Microsoft's intent is that that it should be no different, but the reality is that it does have some different behavior (at least with respect to PowerShell).

I'm not sure that we have a good solution ... maybe we just punt on worrying about that?

Any thoughts @puppetlabs/windows ?

Iristyle:
Also - since this ticket clearly needs a moniker

lib/src/facts/windows/operating_system_resolver.cc
MosesMendoza: Interesting... so the kernel major.minor won't change in Win 10 Anniversary, just the last string (build number?) My inclination is to punt and not worry about additional differentiation... perhaps with a solution that would allow us to also address if/how we want to handle the incongruity of server versions strings ie "2008 R2" currently returned by this fact vs kernel version strings on consumer versions. Open to other thoughts too.

acceptance/tests/facts/windows.rb
```diff
@@ -22,7 +22,10 @@
     os_version  = '2012 R2'
```
MosesMendoza: Side note unrelated to this PR, it seems like if we add / start testing against server 2012 this test will start failing

glennsarti: Same issue with 2008 as well.

acceptance/tests/facts/windows.rb
```diff
@@ -22,7 +22,10 @@
...
     kernel_version = '6.3'
   elsif agent['platform'] =~ /-10/
-    os_version  = /^10\./
+    os_version  = '10'
```
MosesMendoza: @Magisus I think there are a couple bugs in the code prior to this PR and I want to confirm if you were looking to fix both of them?

The first is that the `os.release.major` and `os.release.full` are reporting the same version string on consumer windows 10, i.e. this output on Windows 10 running facter 3.x:
```
os => {
  architecture => "x64",
  family => "windows",
  hardware => "x86_64",
  name => "windows",
  release => {
    full => "10.0.10586",
    major => "10.0.10586"
  },
...
```
whereas the output should actually be something like:
```
os => {
  architecture => "x64",
  family => "windows",
  hardware => "x86_64",
  name => "windows",
  release => {
    full => "10.0.10586",
    major => "10"
  },
...
```

The second bug is that on Windows Server 2016, the `os.release.full` and `os.release.major` facts return the kernel version string, ie `10.0.10586`, rather than the string `2016` as it would have done under Facter 2.x.

Is this PR intended to address both of these, or just one or the other? 

I noticed this because down below on lines 42, 43 this facter test it appears to assert that `os.release.full` and `os.release.major` are the same string, which shouldn't be the case for Windows 10 consumer edition.

Magisus: The way the resolver is written now, the lines below are correct on all currently tested platforms, and this test would pass with or without my change on all unsupported platforms, because the code in the resolver _always_ sets the major version equal to the full version. In the case of an "unrecognized" kernel version, it sets them both equal to the full kernel version string. I assumed that we would want to treat Windows 10 the same as we are treating Windows 7 and 8, i.e. return "10" as the full and major version, which is also identical to the way we treat server versions. The full kernel string should _only_ appear in the "kernel" fact.

Magisus: If we were intentionally returning the full kernel string on consumer Windows 10, then we need to change the resolver code to continue to do that. But I doubt that was the case.

acceptance/tests/facts/windows.rb
Magisus: Yes, it would hit the exception below. Do we intend to start testing against those platforms?

lib/src/facts/windows/operating_system_resolver.cc
Magisus: To me this code block looks like we never actually want to be returning the kernel version string for _any_ platform. It will do that as a default, but I think we might actually want to be catching as many as we can and returning a meaningful OS version.

ferventcoder: The full version should show up differently. The trick is that Windows 10 will just continually be updated so behavior could be changed over time and it still be Windows 10 (with slight differences between versions).

I think as long as we have full release number as a way of tracking that, we will be fine.

acceptance/tests/facts/windows.rb
MosesMendoza: Ah right. I think what I'm saying is that having the major version fact value equal to the full version fact value is a regression on consumer Windows OS versions. In Facter 2.x, these two facts didn't report the same value. Ie, here's the relevant output from facter on Windows 10 Enterprise:
```
C:\Users\moses\development\puppet [master ≡]> bundle exec facter os
{"name"=>"windows", "family"=>"windows", "release"=>{"major"=>"10", "minor"=>"0", "full"=>"10.0.10586"}}
C:\Users\moses\development\puppet [master ≡]> bundle exec facter kernelmajversion
10.0
C:\Users\moses\development\puppet [master ≡]> bundle exec facter kernelversion
10.0.10586
C:\Users\moses\development\puppet [master ≡]> bundle exec facter kernelrelease
10.0.10586
C:\Users\moses\development\puppet [master ≡]> bundle exec facter --version
2.4.6
C:\Users\moses\development\puppet [master ≡]>
```

Magisus: Does anyone have context on whether this was intentional? Because someone very deliberately added the code to override that behavior.

Magisus: What do older versions of Windows look like under Facter 2? Could the above output be an artifact of not updating special logic for new versions of windows, same as it was in Facter 3?

MosesMendoza: @Magisus I think the code in question from facter 2.x is here https://github.com/puppetlabs/facter/blob/2.x/lib/facter/operatingsystem/base.rb#L15-L54. we essentially would take the `kernelrelease` fact value and split on dots to get major/minor, and just return the `kernelrelease` value for `full`

Magisus: Actually I just found it [here](https://github.com/puppetlabs/facter/blob/2.x/lib/facter/operatingsystem/windows.rb). If I'm reading this right, it looks like we never updated it for Windows Server 2016, same as Facter 3. So, given the output you pasted above, @MosesMendoza , does Windows 10 Enterprise not count as a consumer release? And if not, what do we want to return for it?

MosesMendoza: Yes! Great find! That's it and you're totally right! Yes, Windows 10 Ent does count at a consumer operating system ("workstation", at least that's what it's [ProductType](https://msdn.microsoft.com/en-us/library/aa394239(v=vs.85).aspx) returns.

So it looks like if that case were to be updated to catch the right version value, it would actually be just as you say where the full and major versions are identical!

MosesMendoza:
I'm 👍  on this PR

Iristyle:
@MosesMendoza to update this thread with info on the ticket that results from analyzing the different Facter output across different Windows hosts.

PR #1438, opened 2016-10-04
(FACT-1518) Add quotes to ttl keys so hocon lookup works correctly

Magisus:
When the key string for each ttl object is passed directly to
config::get_duration, cpp-hocon will parse it as an unquoted string and
choke on special symbols. This adds triple quotes, to treat the string as a
single element with all special symbols ignored.

Note: this disables escape sequences and path splitting on '.'

Note 2: This is in response to [HC-88](https://tickets.puppetlabs.com/browse/HC-88), but I can't use that in the commit message within Facter or Travis will get angry.


branan:
@Magisus if HC-88 is actually "facter is using hocon wrong", then should that be closed and a new Facter ticket opened for this fix?

Magisus:
@branan new Facter ticket is [FACT-1518](https://tickets.puppetlabs.com/browse/FACT-1518).

PR #1440, opened 2016-10-10
(FACT-1516) Update beaker and tests for MacOS Sierra

McdonaldSeanp:
Update the beaker version and one facter test to prepare for MacOS Sierra support in the pipelines.


branan:
:+1: Nice and simple. Just waiting for tests

Magisus:
Verified that this is the correct kernel version for 10.12 👍

PR #1444, opened 2016-10-13
(packaging) Add beaker-abs dependency for CI

joshcooper:
This adds a dependency on our beaker-abs hypervisor when running beaker
acceptance tests. This will cause the gem to be installed in production
CI pipelines (the gem has already been mirrored internally), but the gem
will not be used until we switch to CI.next.


joshcooper:
current gem is mirrored here http://rubygems.delivery.puppetlabs.net/gems/beaker-abs-0.1.3.gem and released to rubygems.org too

branan:
:+1:

Magisus:
Sounds good, I'm excited for Facter to be one of the guinea pigs for this! 👍

PR #1448, opened 2016-10-14
(FACT-1384) Fix failure to return external fact path when admin

MikaelSmith:
The previous commits for FACT-1384 failed to return the path resolved for
external facts when running with elevated privileges on Windows. Return
the value to fix resolving external facts. This was caught by acceptance
tests on Windows.


MosesMendoza:
👍

PR #1449, opened 2016-10-19
(FACT-1356) Allow executable facts to return structured data

Magisus:
Previously, executable facts could only return simple key/value
pairs. This code now attempts to parse executable output as Yaml, and
only falls back to the key/value parsing if we do not get a useful
result. Because Yaml is a superset of Json, we do not have to
seperately try to parse Json output.

This also bumps the Leatherman version in Travis and Appveyor.

This supersedes https://github.com/puppetlabs/facter/pull/1424.


Magisus:
Successful test run on Windows: https://gist.github.com/Magisus/e17f1354f2294478754bff58de31967e

Successful test run on Centos 7:
https://gist.github.com/Magisus/0d19ccae4c9f7ce6f9e3afa823af5e86

McdonaldSeanp:
Seems good to me 👍

MosesMendoza:
reviewing

lib/src/facts/external/execution_resolver.cc
```diff
@@ -29,38 +32,51 @@ namespace facter { namespace facts { namespace external {
...
+                    // If YAML doesn't correctly parse, it will
+                    // sometimes just return an empty node instead of
+                    // erroring. Only claiming we've resolved if we
+                    // add at least one child value from the YAML
+                    // allows us to still pass on to the keyval
+                    // interpretation in those cases.
+                    resolved = true;
+                }
+            }  catch (YAML::Exception& ex) {
+                LOG_DEBUG("Could not parse executable fact output as YAML ({1})", ex.msg);
```
Magisus: Good call, updated

MosesMendoza:
I think it would also be good to have test coverage that executable fact resolution errors/logs as expected in the case of unresolvable/malformed facts. This comment can be ignored if this is already covered elsewhere..

MosesMendoza:
Aside from error case test coverage, 👍

Magisus:
I don't believe we are testing the error behavior anywhere. I'll add it.

Magisus:
@moses error case tests added.

MosesMendoza:
👍  for merge

PR #1451, opened 2016-10-19
(FACT-1520) Various fixes for Freebsd facts

jmslagle:
This adds networking facts, and fixes a case where the wrong system serial was used.

### DIFF
--- a/lib/src/facts/freebsd/dmi_resolver.cc
+++ b/lib/src/facts/freebsd/dmi_resolver.cc
@@ -14,7 +14,7 @@ namespace facter { namespace facts { namespace freebsd {
         result.bios_version = kenv_lookup("smbios.bios.version");
         result.bios_release_date = kenv_lookup("smbios.bios.reldate");
         result.uuid = kenv_lookup("smbios.system.uuid");
-        result.serial_number = kenv_lookup("smbios.planar.serial");
+        result.serial_number = kenv_lookup("smbios.system.serial");
         result.product_name = kenv_lookup("smbios.system.product");
         result.manufacturer = kenv_lookup("smbios.system.maker");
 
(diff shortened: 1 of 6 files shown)


lib/src/facts/freebsd/dmi_resolver.cc
```diff
@@ -14,7 +14,7 @@ namespace facter { namespace facts { namespace freebsd {
         result.bios_version = kenv_lookup("smbios.bios.version");
         result.bios_release_date = kenv_lookup("smbios.bios.reldate");
         result.uuid = kenv_lookup("smbios.system.uuid");
-        result.serial_number = kenv_lookup("smbios.planar.serial");
```
MikaelSmith: Was this ever accurate?

branan:
How much of the freebsd networking resolver is copied from openbsd? Do we need to do some refactoring here to clean that up? I'm happy to help guide you through that

lib/src/facts/freebsd/dmi_resolver.cc
jmslagle: No.

Well kinda?

It's the Planar serial - likely the MB serial.  Almost certainly NOT what you want.

jmslagle:
Almost or even all of it is from OpenBSD - unlike the other PR I need to submit to do memory and others which are largely different.

I can attempt to refactor this to move these two functions upstream to the bsd resolver, but you had indicated it's in use by other things.

Having FreeBSD include the openbsd file seemed just as bad.  But DRY.

jmslagle:
I can lurk on Slack if we want to discuss.

MikaelSmith:
@branan is refactoring something you planned to help with? If not, this is small enough I'm not too worried about merging it.

PR #1460, opened 2016-10-26
(PA-657) Update network fact tests to correctly exclude SPARC

Magisus:
This refactors which networking facts are excluded from eccentric platforms to
correctly handle SPARC and Cisco. Before this commit, some tests were
failing because the case statement didn't correctly constrain certain
platforms.


LuvCurves:
:+1: :shipit:

PR #1461, opened 2016-10-28
(packaging) Use beaker-abs that supports beaker 3

joshcooper:
Facter now requires beaker 3, so we need to require at least beaker-abs
0.2.0.

This is a cherry-pick of master back to stable. We're requiring at least 0.2 because that version supports both beaker 2 and 3.


Magisus:
👍

PR #1462, opened 2016-10-28
(FACT-1527) Fix Gentoo not reading os release fact correctly

ace13:


lib/src/facts/linux/os_linux.cc
```diff
@@ -265,6 +265,7 @@ namespace facter { namespace facts { namespace linux {
                 { string(os::oracle_enterprise_linux),  string(release_file::oracle_enterprise_linux) },
                 { string(os::oracle_vm_linux),          string(release_file::oracle_vm_linux) },
                 { string(os::arista_eos),               string(release_file::arista_eos) },
+                { string(os::gentoo),                   string(release_file::gentoo) },
```
MikaelSmith: Looks like this works because the 1st thing we do after is look for the pattern `release (\d[\d.]*)` in the first line. That matches `Gentoo Base System release 2.2`, and the rest falls together.

MikaelSmith:
Thanks! Looks like that was easy; you've verified it works on your system?

ace13:
Certainly seems that way;
```
$ facter os
...
  family => "Gentoo",
  hardware => "x86_64",
  name => "Gentoo",
  release => {
    full => "4.8.0",
    major => "4",
    minor => "8"
  },
...
$ ./facter os
...
  family => "Gentoo",
  hardware => "x86_64",
  name => "Gentoo",
  release => {
    full => "2.2",
    major => "2",
    minor => "2"
  },
...
$ cat /etc/gentoo-release
Gentoo Base System release 2.2
```

MikaelSmith:
Manually confirmed CLA has been signed. Thanks for the contribution!

ace13:
Happy to help.

PR #1463, opened 2016-10-28
(maint) Update MAINTAINERS file

MikaelSmith:
Kylo is no longer part of the puppetlabs GitHub org, remove him as a
maintainer.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -5,11 +5,6 @@
   "internal_list": "https://groups.google.com/a/puppet.com/forum/?hl=en#!forum/discuss-facter-maintainers",
   "people": [
     {
-      "github": "kylog",
-      "email": "kylo@puppet.com",
-      "name": "Kylo Ginsberg"
-    },
-    {
       "github": "magisus",
       "email": "maggie@puppet.com",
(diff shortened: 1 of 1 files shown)


Magisus:
👍

Magisus:
As a future reference note, going forward we would like to be more deliberate about how we land commits that are intended for multiple branches. A commit that is supposed to land on both master and stable should target STABLE first. Then after that PR has been merged, the committer should merge stable into master according to the process outlined [here](https://github.com/puppetlabs/puppet/blob/master/COMMITTERS.md#how-to-commit-a-change-set-to-multiple-base-branches).

(note: that guideline file is not present in the Facter repo. Perhaps it should be there?)

MikaelSmith:
This file isn't read from stable, so I had no intention of merging it there.

Magisus:
Ah okay. Since it still appears in the stable branch, I assumed we wanted them to be in sync.

Side note, out of curiosity: when you say "read", what reads this file? Something to do with Confluence, or Google groups?

MikaelSmith:
https://github.com/puppetlabs/maintainers reads the MAINTAINERS file from each project's default branch (using GitHub APIs).

PR #1464, opened 2016-10-31
(FACT-1394) Fix warning for linkdown routes

ace13:
Currently, Facter fails to process routes that are marked as linkdown, instead generating warnings for them.

This patch will therefore skip such entries entirely, giving the same outcome but without the warnings that the earlier method generates.

```
$ facter -p networking.interfaces.docker0
2016-10-31 08:54:40.260618 WARN  puppetlabs.facter - Could not process routing table entry: Expected a destination followed by key/value pairs, got '172.17.0.0/16 dev docker0  proto kernel  scope link  src 172.17.0.1 linkdown'
2016-10-31 08:54:40.261368 WARN  puppetlabs.facter - Could not process routing table entry: Expected a destination followed by key/value pairs, got 'fe80::/64 dev docker0  proto kernel  metric 256 linkdown  pref medium'
{
  bindings => [
    {
      address => "172.17.0.1",
      netmask => "255.255.0.0",
      network => "172.17.0.0"
    }
  ],
  bindings6 => [
    {
      address => "fe80::42:d5ff:fe36:1a50",
      network => "fe80::"
    }
  ],
  ip => "172.17.0.1",
  ip6 => "fe80::42:d5ff:fe36:1a50",
  mac => "02:42:d5:36:1a:50",
  mtu => 1500,
  netmask => "255.255.0.0",
  network => "172.17.0.0",
  network6 => "fe80::"
}
$ ./facter -p networking.interfaces.docker0
{
  bindings => [
    {
      address => "172.17.0.1",
      netmask => "255.255.0.0",
      network => "172.17.0.0"
    }
  ],
  bindings6 => [
    {
      address => "fe80::42:d5ff:fe36:1a50",
      network => "fe80::"
    }
  ],
  ip => "172.17.0.1",
  ip6 => "fe80::42:d5ff:fe36:1a50",
  mac => "02:42:d5:36:1a:50",
  mtu => 1500,
  netmask => "255.255.0.0",
  network => "172.17.0.0",
  network6 => "fe80::"
}
```

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -143,6 +143,13 @@ namespace facter { namespace facts { namespace linux {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
 
+            // skip links that are linkdown
+            if (std::find_if(parts.cbegin(), parts.cend(), [](const boost::iterator_range<string::iterator>& range) {
+                return std::string(range.begin(), range.end()) == "linkdown";
+            }) != parts.cend()) {
+                return true;
+            }
+
             // remove trailing "onlink" or "pervasive" flags
             while (parts.size() > 0) {
                 std::string last_token(parts.back().begin(), parts.back().end());


branan:
It looks like `linkdown` is actually being printed as a sort of fake key-value pair with an empty value - the first line in your warning has two spaces after it. We ignore redundant whitespace right now in two different places. Paying attention to that whitespace will avoid the extra iteration you're doing right now.

lib/src/facts/linux/networking_resolver.cc
```diff
@@ -143,6 +143,13 @@ namespace facter { namespace facts { namespace linux {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
```
branan: This is where we're ignoring the extra space in the middle of the line. You can change this to `boost::token_compress_off`

ace13:
If linkdown is attempted as a key-value, then they do it really badly. 

Here's the route output when wrapped in a string
```
$ ip r s | xargs -rn1 -d '\n' printf "'%s'\n" | grep docker
'172.17.0.0/16 dev docker0  proto kernel  scope link  src 172.17.0.1 linkdown '
$ ip -6 r s | xargs -rn1 -d '\n' printf "'%s'\n" | grep docker
'fe80::/64 dev docker0  proto kernel  metric 256 linkdown  pref medium'
```

The `linkdown` token seems more just plopped down somewhere in the string rather than being a key with empty value, all the true key-value pairs have double whitespace surrounding them after all.

`ip route show` seems to throw in an extra piece of whitespace at the end for all IPv4 routes as well, which probably doesn't help.

branan:
@ace13 ugh, I think you're right. In that case I think this is probably fine. It might be possible to beat it into slightly more idiomatic shape, but I'm not too concerned about that.

Apologies for how long it took to get back to you on this

@puppetlabs/facter-maintainers anybody else have opinions on this?

PR #1465, opened 2016-11-01
(FACT-1441) Add basic "cloud" fact 

smcclellan:
TODOs for future tickets:
- OpenBSD implementation
- OSX implementation
- Windows - convert or include [existing script](https://gallery.technet.microsoft.com/scriptcenter/Detect-Windows-Azure-aed06d51) in implementation
- Add more detail to the `cloud` fact for non-Azure stuff (may be separate PR)

Since there currently exists no fact that can identify if a machine is running
inside Azure, we need to add a more robust virtualization fact. This fact
will be differentiated from the existing `is_virtual` and `virtual` facts
in that it will be structured to include cloud and metadata info as well.

This step adds the `cloud` fact and checks for azure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,3 +279,16 @@ To generate API documentation, install doxygen 1.8.7 or later.
     $ doxygen
 
 To view the documentation, open `lib/html/index.html` in a web browser.
+
+Debugging
+---------
+
+If when running the tests you encounter this error message:
+
+"could not locate a ruby library"
+
+You may need to use a different static ruby library in Leatherman. To do
+this, run this command, where the location below is the default for a
+puppet agent installation:
+
+    $ export LEATHERMAN_RUBY=/opt/puppetlabs/puppet/lib/libruby.dylib
--- a/README.md
+++ b/README.md
@@ -287,7 +287,7 @@ If when running the tests you encounter this error message:
 
 "could not locate a ruby library"
 
-You may need to use a different static ruby library in Leatherman. To do
+You may need to use a different shared ruby library in Leatherman. To do
 this, run this command, where the location below is the default for a
 puppet agent installation:
 
--- a/lib/inc/facter/facts/fact.hpp
+++ b/lib/inc/facter/facts/fact.hpp
@@ -439,6 +439,10 @@ namespace facter { namespace facts {
          * The fact for whether or not the machine is virtual or physical.
          */
         constexpr static char const* is_virtual = "is_virtual";
+        /**
+         * The fact for the cloud info, including provider, for a node.
+         */
+        constexpr static char const* cloud = "cloud";
 
         /**
          * The structured fact for identity information.
--- a/lib/inc/internal/facts/linux/virtualization_resolver.hpp
+++ b/lib/inc/internal/facts/linux/virtualization_resolver.hpp
@@ -20,7 +20,21 @@ namespace facter { namespace facts { namespace linux {
          * @param facts The fact collection that is resolving facts.
          * @return Returns the name of the hypervisor or empty string if no hypervisor.
          */
-        virtual std::string get_hypervisor(collection& facts) override;
+        std::string get_hypervisor(collection& facts) override;
+        /**
+         * Gets the name of the cloud provider.
+         * @param facts The fact collection that is resolving facts.
+         * @return Returns the name of the cloud provider or empty string if no hypervisor.
+         */
+        std::string get_cloud_provider(collection& facts) override;
+
+        /**
+         * Gets whether the machine is running in Azure.
+         * @param facts The fact collection that is resolving facts.
+         * @param leases_file The location of where the leases file exists.
+         * @return Returns "azure" if running on azure, otherwise an empty string.
+         */
+        static std::string get_azure(collection& facts, std::string const& leases_file = "/var/lib/dhcp/dhclient.eth0.leases");
 
      private:
         static std::string get_cgroup_vm();
--- a/lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
+++ b/lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
@@ -10,6 +10,36 @@
 namespace facter { namespace facts { namespace resolvers {
 
     /**
+     *  Represents cloud data.
+     */
+    struct cloud_
+    {
+        /**
+         * Stores the cloud provider.
+         */
+        std::string provider;
+    };
+
+    /**
+     *  Represents virtualization data.
+     */
+    struct data
+    {
+        /**
+         * Stores the cloud data.
+         */
+        cloud_ cloud;
+        /**
+         * Stores the hypervisor data.
+         */
+        std::string hypervisor;
+        /**
+         * Stores the is_virtual data.
+         */
+        bool is_virtual;
+    };
+
+    /**
      * Responsible for resolving virtualization facts.
      */
     struct virtualization_resolver : resolver
@@ -34,6 +64,13 @@ namespace facter { namespace facts { namespace resolvers {
         virtual std::string get_hypervisor(collection& facts) = 0;
 
         /**
+         * Gets the name of the cloud provider.
+         * @param facts The fact collection that is resolving facts.
+         * @return Returns the name of the cloud provider or empty string if no cloud provider.
+         */
+        virtual std::string get_cloud_provider(collection& facts);
+
+        /**
          * Determines if the given hypervisor is considered to be virtual.
          * @param hypervisor The hypervisor to check.
          * @return Returns true if the hypervisor is virtual or false if it is physical.
@@ -47,6 +84,13 @@ namespace facter { namespace facts { namespace resolvers {
          * @return Returns the hypervisor product name if matched.
          */
         static std::string get_product_name_vm(std::string const& product_name);
+
+        /**
+         * Collects the virtualization data.
+         * @param facts The fact collection that is resolving facts.
+         * @return Returns the virtualization data.
+         */
+        virtual data collect_data(collection& facts);
     };
 
 }}}  // namespace facter::facts::resolvers
--- a/lib/schema/facter.yaml
+++ b/lib/schema/facter.yaml
@@ -184,6 +184,14 @@ chassistype:
     caveats: |
         Linux: kernel 2.6+ is required due to the reliance on sysfs.
 
+cloud:
+    type: map
+    description: Information about the cloud instance of the node. This is currently only populated on Linux nodes running in Microsoft Azure.
+    elements:
+        provider:
+            type: string
+            description: The cloud provider for the node.
+
 dhcp_servers:
     type: map
     hidden: true
--- a/lib/src/facts/linux/virtualization_resolver.cc
+++ b/lib/src/facts/linux/virtualization_resolver.cc
@@ -4,14 +4,10 @@
 #include <facter/facts/collection.hpp>
 #include <facter/facts/fact.hpp>
 #include <facter/facts/vm.hpp>
-#include <leatherman/execution/execution.hpp>
 #include <leatherman/file_util/file.hpp>
-#include <leatherman/logging/logging.hpp>
 #include <leatherman/util/regex.hpp>
 #include <boost/filesystem.hpp>
 #include <boost/algorithm/string.hpp>
-#include <vector>
-#include <tuple>
 
 using namespace std;
 using namespace facter::facts;
@@ -24,6 +20,34 @@ namespace bs = boost::system;
 namespace lth_file = leatherman::file_util;
 
 namespace facter { namespace facts { namespace linux {
+
+    string virtualization_resolver::get_cloud_provider(collection& facts)
+    {
+        // Check for Azure
+        std::string provider = get_azure(facts);
+
+        return provider;
+    }
+
+    string virtualization_resolver::get_azure(collection& facts, string const& leases_file)
+    {
+        std::string provider;
+        if (boost::filesystem::exists(leases_file))
+        {
+            lth_file::each_line(leases_file, [&](string& line) {
+                // Search for DHCP option 245. This is an accepted method of determining
+                // whether a machine is running inside Azure. Source:
+                // https://social.msdn.microsoft.com/Forums/azure/en-US/f7fbbee6-370a-41c2-a384-d14ab2a0ac12/what-is-the-correct-method-in-linux-on-azure-to-test-if-you-are-an-azure-vm-?forum=WAVirtualMachinesforWindows
+                if (line.find("option 245") != std::string::npos ||
+                        line.find("option unknown-245") != std::string::npos) {
+                    provider = "azure";
+                    return false;
+                }
+                return true;
+            });
+        }
+        return provider;
+    }
 
     string virtualization_resolver::get_hypervisor(collection& facts)
     {
--- a/lib/src/facts/resolvers/virtualization_resolver.cc
+++ b/lib/src/facts/resolvers/virtualization_resolver.cc
@@ -1,6 +1,7 @@
 #include <internal/facts/resolvers/virtualization_resolver.hpp>
 #include <facter/facts/collection.hpp>
 #include <facter/facts/fact.hpp>
+#include <facter/facts/map_value.hpp>
 #include <facter/facts/scalar_value.hpp>
 #include <facter/facts/vm.hpp>
 #include <set>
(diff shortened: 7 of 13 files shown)


README.md
```diff
@@ -274,3 +274,16 @@ To generate API documentation, install doxygen 1.8.7 or later.
     $ doxygen
 
 To view the documentation, open `lib/html/index.html` in a web browser.
+
+Debugging
```
briancain: Could also be worth mentioning running make in verbose mode to see what tests are failing like:

```
$ make test ARGS=-V
```

lib/inc/facter/facts/fact.hpp
```diff
@@ -434,11 +434,15 @@ namespace facter { namespace facts {
         /**
          * The fact for virtualization hypervisor.
          */
-        constexpr static char const* virtualization = "virtual";
+        constexpr static char const* virt = "virtual";
```
briancain: Maybe this should be `virtual`?

peterhuene: Unfortunately that's a keyword in C++.

briancain: Maybe the fact should be `hypervisor` then?

briancain: Or maybe we leave this one back to how it is, and come up with a better name for the fact below. It might be weird to rename this fact and bring in a new fact with the same name but different information.

lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -25,6 +25,29 @@ namespace lth_file = leatherman::file_util;
...
+
+    }
+
+    string virtualization_resolver::get_azure(collection& facts)
+    {
+        std::string provider;
+        if ( boost::filesystem::exists( "/var/lib/dhcp/dhclient.eth0.leases" ) )
+        {
+            string virt_what = agent::which("cat /var/lib/dhcp/dhclient.eth0.leases | grep option 245");
+            each_line(virt_what, [&](string& line) {
```
briancain: What happens when the grep fails and virt_what is nil? Will the entire block be skipped? And then is it okay to return a null string provider?

README.md
smcclellan: I considered that, but I think `ctest -V` works too.

lib/inc/facter/facts/fact.hpp
smcclellan: Reserved word in C++.

smcclellan: Considering this is just a constant defined for internal usage, I think either name is fine, doesn't matter IMO.

briancain: Ah ok. Didn't realize this was all internal.

briancain:
Maybe it'd be good to list out the TODO's as a checklist in the pull request description so it's clear what work is left.

smcclellan:
Good call, I have updated the description to include the TODOs.

README.md
```diff
@@ -274,3 +274,16 @@ To generate API documentation, install doxygen 1.8.7 or later.
...
+---------
+
+If when running the tests you encounter this error message:
+
+"could not locate a ruby library"
+
+You may need to use a different static ruby library in Leatherman. To do
```
MikaelSmith: It's technically a shared ruby library.

briancain:
@smcclellan @MikaelSmith @petems @garethr 

Since this ticket seems to be way bigger than originally estimated, do you all think it makes sense to repurpose this PR to be for adding a new `cloud` (or whatever we settle on) fact like what was described in [FACT-1441](https://tickets.puppetlabs.com/browse/FACT-1441), and then from there we can split up identifying various azure instances into discrete tickets to be worked on after this gets merged?

petems:
@briancain I'm just a rubber-necker, because I've had a few customers asked about facts like this so I started watching the PR 😄 

My 2 cents: changing the original fact is a can of worms, I'd rather a new namespaced fact

briancain:
@petems oh my I think I tagged the wrong Peter. Apologies! Thanks though

smcclellan:
@petems This isn't changing an original fact, but rather creating a new one (`cloud` or the like). What do you mean by a namespaced fact?

petems:
@smcclellan Sorry, meant to say named. I think overloading the original $::virtual is going to lead to some upset peoples.

petems:
Ok, basically ignore me, this is what I get for drive-by reviewing! 👍

petems:
Ok, I saw the change above about changing virtualisation to virtual or something, got the wrong impression. Basically ignore me, this is what I get for drive-by reviewing! 😄

Magisus:
I just realized that this does not follow the `collect_data` pattern used by most of our resolvers. There is a [note](https://github.com/puppetlabs/facter/blob/master/lib/tests/facts/schema.cc#L445-L448) that we would like a couple other non-conforming resolvers to be refactored to use that pattern, so I assume that applies to this work as well? For an example of the pattern, see the [file system resolver](https://github.com/puppetlabs/facter/blob/master/lib/inc/internal/facts/resolvers/filesystem_resolver.hpp)

lib/inc/internal/facts/linux/virtualization_resolver.hpp
```diff
@@ -7,20 +7,36 @@
...
+
 namespace facter { namespace facts { namespace linux {
 
     /**
      * Responsible for resolving virtualization facts.
      */
     struct virtualization_resolver : resolvers::virtualization_resolver
     {
-     protected:
+     public:
```
briancain: @smcclellan - what's the reasoning for switching this from protected to public?

lib/src/facts/openbsd/virtualization_resolver.cc
```diff
@@ -22,4 +22,10 @@ namespace facter { namespace facts { namespace openbsd {
         return {};
     }
 
+    string virtualization_resolver::get_cloud_provider(collection& facts)
+    {
+        // TODO Implement this
+        return "";
```
briancain: Is there something more explicit we can return here other than empty string so that it's clear this resolver hasn't been implemented yet? Or is this pattern pretty standard at the moment for facter?

Magisus: In theory the pattern is, only fill what elements you can of the data struct, then check each to make sure it exists before adding a corresponding value to the collection. So really, what this returns could be arbitrary, depending on how it's handled in the collect_data method.

lib/tests/fixtures/facts/linux/cloud/azure-unknown
```diff
@@ -0,0 +1 @@
+this has option unknown-245 defined
```
briancain: I'm not sure what lease files typically look like, but maybe it's worth adding some extra garbage lines here and on the `azure` file just to make sure it can find this option string? Since it looks like the implementation is iterating over each line in the given file.

lib/inc/internal/facts/windows/virtualization_resolver.hpp
```diff
@@ -29,6 +29,13 @@ namespace facter { namespace facts { namespace windows {
          * @return Returns the name of the hypervisor or empty string if no hypervisor.
          */
         virtual std::string get_hypervisor(collection& facts) override;
+        /**
+         * Gets the name of the cloud provider.
+         * @param facts The fact collection that is resolving facts.
+         * @return Returns the name of the cloud provider or empty string if no hypervisor.
+         */
+        virtual std::string get_cloud_provider(collection& facts) override;
```
Magisus: If you are going to declare that this subclass overrides the virtual function, you have to write an actual implementation. This is why appveyor is failing right now. Same goes for the Solaris one above. It might make more sense for now to leave out this declaration and let the default implementation be used on these non-linux platforms.

lib/src/facts/solaris/virtualization_resolver.cc
```diff
@@ -51,4 +51,5 @@ namespace facter { namespace facts { namespace solaris {
 
         return guest_of;
     }
+
```
briancain: Can probably remove this extra line addition

lib/tests/fixtures/facts/linux/cloud/azure-unknown
smcclellan: Fixed.

Magisus:
Have you done any manual testing of this? Does it correctly detect the new cloud stuff when run in the proper environment?

smcclellan:
I haven't tested a proper environment, but I have stubbed out that file with sample contents from the JIRA ticket to see that it works.

lib/inc/internal/facts/linux/virtualization_resolver.hpp
MikaelSmith: Protected functions can be overridden when you subclass. You may need to modify how you inherit by specifying something like `struct my_virt_resolver : public virtualization_resolver`.

MikaelSmith: Or, based on what you're doing below, use a class to inherit this and make protected members public, something like

```
struct peek_resolver : virtualization_resolver
{
    using virtualization_resolver::get_azure;
};
```

lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
```diff
@@ -10,6 +10,36 @@
 namespace facter { namespace facts { namespace resolvers {
 
     /**
+     *  Represents cloud data.
+     */
+    struct cloud_
```
MikaelSmith: Any reason to use `cloud_` rather than `cloud`? We tend to use an underscore suffix for private data members.

lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
```diff
@@ -25,7 +55,7 @@ namespace facter { namespace facts { namespace resolvers {
          */
         virtual void resolve(collection& facts) override;
 
-     protected:
+     public:
```
MikaelSmith: I'd rather use the trick I outlined above to test protected methods.

lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -25,6 +21,31 @@ namespace lth_file = leatherman::file_util;
...
+        // Check for Azure
+        std::string provider = get_azure(facts);
+
+        return provider;
+    }
+
+    string virtualization_resolver::get_azure(collection& facts, string const& leases_file)
+    {
+        std::string provider;
+        if ( boost::filesystem::exists( leases_file ) )
```
MikaelSmith: Spacing here doesn't really match style everywhere else.

lib/src/facts/linux/virtualization_resolver.cc
```diff
@@ -25,6 +21,31 @@ namespace lth_file = leatherman::file_util;
...
+        {
+            lth_file::each_line(leases_file, [&](string& line) {
+                if (line.find("option 245") != std::string::npos ||
```
MikaelSmith: Is there anywhere that documents what this means and why it exists? It'd be nice to include a comment about why we search for this string.

lib/src/facts/osx/virtualization_resolver.cc
```diff
@@ -42,5 +42,4 @@ namespace facter { namespace facts { namespace osx {
         });
         return value;
     }
-
```
MikaelSmith: Removing the line seems unnecessary, and doesn't match a style I see elsewhere. I'd rather just not change it.

lib/inc/facter/facts/fact.hpp
MikaelSmith: Any reason to still change `virtualization` to `virt`? The old one doesn't seem to conflict with anything, and this is still a public header file.

lib/inc/internal/facts/linux/virtualization_resolver.hpp
```diff
@@ -7,20 +7,36 @@
...
+         */
+        std::string get_cloud_provider(collection& facts) override;
+
+        /**
+         * Gets whether the machine is running in Azure.
+         * @param facts The fact collection that is resolving facts.
+         * @param leases_file The location of where the leases file exists.
+         * @return Returns "azure" if running on azure, otherwise an empty string.
+         */
+        static std::string get_azure(collection& facts, string const& leases_file = "/var/lib/dhcp/dhclient.eth0.leases");
```
MikaelSmith: What would it take to implement this for Windows as well?

smcclellan: There is a [Powershell script](https://gallery.technet.microsoft.com/scriptcenter/Detect-Windows-Azure-aed06d51) that exists already; I believe it just needs to be converted into C++.

lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
smcclellan: There was a name conflict with calling the property `cloud` as well.

lib/src/facts/linux/virtualization_resolver.cc
smcclellan: https://social.msdn.microsoft.com/Forums/azure/en-US/f7fbbee6-370a-41c2-a384-d14ab2a0ac12/what-is-the-correct-method-in-linux-on-azure-to-test-if-you-are-an-azure-vm-?forum=WAVirtualMachinesforWindows is the source, plus a few other places.

lib/inc/internal/facts/resolvers/virtualization_resolver.hpp
MikaelSmith: Oh, at line 31. Alright, naming's tricky, I'll ignore it.

lib/inc/internal/facts/linux/virtualization_resolver.hpp
```diff
@@ -7,6 +7,8 @@
 #include "../resolvers/virtualization_resolver.hpp"
 #include <string>
 
+using namespace std;
```
MikaelSmith: Please don't do this in header files. It can cause surprises for anyone trying to use the header file.

lib/schema/facter.yaml
```diff
@@ -184,6 +184,15 @@ chassistype:
     caveats: |
         Linux: kernel 2.6+ is required due to the reliance on sysfs.
 
+cloud:
+    type: map
+    description: Information about the cloud instance of the node.
```
MikaelSmith: Can we make it clearer that this only supports Azure at the moment? This becomes the documentation for the fact.

lib/schema/facter.yaml
```diff
@@ -184,6 +184,15 @@ chassistype:
     caveats: |
         Linux: kernel 2.6+ is required due to the reliance on sysfs.
 
+cloud:
+    type: map
+    description: Information about the cloud instance of the node. This is currently only populated on linux nodes.
```
smcclellan: Updated.

lib/schema/facter.yaml
```diff
@@ -184,6 +184,15 @@ chassistype:
     caveats: |
         Linux: kernel 2.6+ is required due to the reliance on sysfs.
 
+cloud:
+    type: map
+    description: Information about the cloud instance of the node. This is currently only populated on linux nodes running in Microsoft Azure.
+    validate: false
```
MikaelSmith: We only set `validate` to false on facts where we don't know what the names of their elements will be (ec2 metadata is filled from the metadata end-point). In this case we know exactly what it'll be - `provider` - so we should include that in testing.

lib/schema/facter.yaml
```diff
@@ -184,6 +184,15 @@ chassistype:
     caveats: |
         Linux: kernel 2.6+ is required due to the reliance on sysfs.
 
+cloud:
+    type: map
+    description: Information about the cloud instance of the node. This is currently only populated on linux nodes running in Microsoft Azure.
```
MikaelSmith: @gguillotte as far as I can tell we usually capitalize Linux. Should that also be the case here?

gguillotte: Yes indeedy. Linux is always capped.

lib/src/facts/resolvers/virtualization_resolver.cc
```diff
@@ -16,19 +17,44 @@ namespace facter { namespace facts { namespace resolvers {
...
 
     void virtualization_resolver::resolve(collection& facts)
     {
+        auto data = collect_data(facts);
+
+        facts.add(fact::is_virtual, make_value<boolean_value>(data.is_virtual));
+        facts.add(fact::virtualization, make_value<string_value>(data.hypervisor));
+        if (!data.cloud.provider.empty()) {
+            auto cloud = make_value<map_value>();
+            cloud->add("provider", make_value<string_value>(data.cloud.provider, true));
```
MikaelSmith: Why use `true` here? I think that's labeling it a hidden value.

MikaelSmith:
If it passes CI, I think it's ready.

Magisus:
Agreed! Nice to see new people working on Facter 👍

PR #1468, opened 2016-11-02
(FACT-1531) Add SLES back to non-dhcp platforms

LuvCurves:


Magisus:
This looks like it now preserves the original confine 👍

PR #1469, opened 2016-11-02
(PA-652) Bump Facter version to 3.5.1

Magisus:
This bumps the Facter version to 3.5.1 following the release of 3.5.0 as
part of the puppet-agent 1.8.0 release.


Magisus:
Travis commit checker is failing because it requires a FACT ticket. Do we want to allow to fail, or switch to maint? Might it be worthwhile to add PA tickets to the whitelist?

PR #1470, opened 2016-11-02
(PA-655) Add Windows Server 2016 to the LTS agent

Magisus:
This is a cherry-pick of several commits from master to LTS-1.7, in support of [PA-655](https://tickets.puppetlabs.com/browse/PA-655).

Because the LTS pipeline also includes macOS Sierra as a test target, this PR also includes one test fix for Sierra.

### DIFF
--- a/lib/src/facts/windows/operating_system_resolver.cc
+++ b/lib/src/facts/windows/operating_system_resolver.cc
@@ -100,8 +100,8 @@ namespace facter { namespace facts { namespace windows {
         // Override default release with Windows release names
         auto version = result.release.substr(0, lastDot);
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
-        if (version == "6.4") {
-            result.release = consumerrel ? "10" : result.release;
+        if (version == "10.0") {
+            result.release = consumerrel ? "10" : "2016";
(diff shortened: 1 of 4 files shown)


McdonaldSeanp:
👍

lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -100,8 +100,8 @@ namespace facter { namespace facts { namespace windows {
         // Override default release with Windows release names
         auto version = result.release.substr(0, lastDot);
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
-        if (version == "6.4") {
```
MikaelSmith: Is 6.4 still valid for any Windows 10 versions (say, pre-Anniversary Release)?

MikaelSmith: https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832(v=vs.85).aspx suggests that was just pre-release numbering. So 👍

Magisus: Not that I could find. We looked into this a fair amount when I originally put up the PR, and it seems they switched everything to 10.

PR #1472, opened 2016-11-17
(FACT-1524) disable mount tests on AIX

LuvCurves:


Magisus:
👍 

Do you need this on the release branch of puppet-agent to run your tests effectively?

LuvCurves:
@Magisus Not really. It will be fine to be in for the next time we need to run.

PR #1473, opened 2016-11-17
(FACT-1536) List names of facts that belong to each fact group

Magisus:
The first commit updates the commands using the existing API. 

The second commit refactors the collection's methods to return structured data. The code is much cleaner after the refactor, but it technically breaks Facter's public API (as released in 3.5.0). Is this a problem?

### DIFF
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -117,9 +117,9 @@ void log_queries(set<string> const& queries)
 }
 
 void print_fact_groups(map<string, vector<string>> const& fact_groups) {
-    for (auto group : fact_groups) {
+    for (auto& group : fact_groups) {
         boost::nowide::cout << group.first << endl;
-        for (auto fact : group.second) {
+        for (auto& fact : group.second) {
             boost::nowide::cout << "  - " << fact << endl;
         }
     }
--- a/exe/facter.cc
+++ b/exe/facter.cc
@@ -249,6 +249,9 @@ int main(int argc, char **argv)
             vector<string> fact_groups = facts.get_fact_groups();
             for (auto group : fact_groups) {
                 boost::nowide::cout << group << endl;
+                for (auto fact : facts.get_facts_in_group(group)) {
+                    boost::nowide::cout << "  - " << fact << endl;
+                }
             }
             return EXIT_SUCCESS;
         }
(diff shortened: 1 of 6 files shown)


whopper:
@Magisus if we end up being okay with the breaking change, I'd say that the commit that does it should at least have a ticket describing it so we can specifically call it out in the release notes. People *will* get angry if we don't call it out loudly (as we learned with Facter 2.2.0 :)).

Magisus:
Sounds good, I'll edit the message the reference the same ticket and add a comment in the ticket.

ScottGarman:
I'm :+1: on the code review of the change, but not an authority on whether to include a breaking change, so I can't speak to that.

Magisus:
In standup today @ahpook gave a thumbs up to merging the refactor, given that the odds that _anyone_ was using those methods are basically 0. Perhaps @MikaelSmith has thoughts?

ahpook:
just to put it in the thread, i'm OK with the change here. 💯

MosesMendoza:
@Magisus looks like this PR could use some test coverage, perhaps for the new `print_fact_groups` and changes to `collection`?

Magisus:
I've added unit tests for `collection` and updated the acceptance tests for the command line flags.

MosesMendoza:
👍

exe/facter.cc
```diff
@@ -116,6 +116,15 @@ void log_queries(set<string> const& queries)
     log(level::info, "requested queries: %1%.", output.str());
 }
 
+void print_fact_groups(map<string, vector<string>> const& fact_groups) {
+    for (auto group : fact_groups) {
```
MikaelSmith: I'd suggest using refs, as in `for (auto& group : fact_groups)`. Auto uses value by default unless it's not copyable.

Magisus: Thanks for catching that, updated!

exe/facter.cc
```diff
@@ -116,6 +116,15 @@ void log_queries(set<string> const& queries)
     log(level::info, "requested queries: %1%.", output.str());
 }
 
+void print_fact_groups(map<string, vector<string>> const& fact_groups) {
+    for (auto& group : fact_groups) {
+        boost::nowide::cout << group.first << endl;
+        for (auto fact : group.second) {
```
MikaelSmith: Possibly same here? No need to copy the strings.

PR #1474, opened 2016-11-22
(FACT-1537) Let's fix locales/ and get a good tarball

mckern:
Confirmed that this will fix the inability to build facter 3.5.0 from the release tarball

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -6,12 +6,22 @@ homepage: 'https://github.com/puppetlabs/facter'
 summary: 'Facter, a system inventory tool'
 description: 'You can prove anything with facts!'
 # files to be packaged into a tarball and released with deb/rpm
-files: '[A-Z]* exe ext lib locales man scripts vendor cmake'
+files:
+  - '[A-Z]*'
+  - exe
+  - ext
+  - lib
+  - locales
+  - man
+  - scripts
+  - vendor
+  - cmake
 # space separated list of files to *exclude* from the tarball
 # note that each listing in files, above, is recursively copied into the tarball, so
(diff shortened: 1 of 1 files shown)


ScottGarman:
:+1:

glennsarti:
Should there be an acceptance test for this somewhere?

mckern:
@glennsarti yes, but writing a test for tarball "correctness" is beyond the scope of this PR (and should realistically work for every other product we generate tarballs for). I was kicking around **how** last night, and I'll talk about it with you a bit today if you want?

glennsarti:
Probably a better task for the facter maintainers to take up.  This just seemed to me that it should be testable somewhere so we don't repeat the mistake.  I'd be happy for a merge if it was ticketted elsewhere.

glennsarti:
@mckern On a different note, the commit messages don't seem to follow the Contributing guidelines for this project. Would it be possible to get them reworded as per https://github.com/puppetlabs/facter/blob/master/CONTRIBUTING.md#making-changes ?

mckern:
@glennsarti There was a sarcastic comment in here, but it didn't post. Instead I beefed up commit messages a bit.

mckern:
> Probably a better task for the facter maintainers to take up. This just seemed to me that it should be testable somewhere so we don't repeat the mistake. I'd be happy for a merge if it was ticketted elsewhere.

I think that "did this produce a valid tarball" is a task that should be handled by RE, since it's par tof the functionality provided by the Packaging repo and isn't unique to any one project. I'll work on getting it ticketed out -- there's already some internal conversation around how to build these tarball snapshots and if that's even the right thing to do in the future.

MikaelSmith:
Seems like it should go against stable. Mind rebasing?

mckern:
@MikaelSmith done

PR #1476, opened 2016-11-30
(FACT-1541) javah command runs late in parallel build

MikaelSmith:
Generation of the java header file can happen too late in parallel
builds, causing compilation to fail. Ensure it's included in dependency
resolution.


Magisus:
I have verified that this change does not break currently-working builds or tests.

Magisus:
Did we ever get an answer on licensing for this? Because if it seems legit to use the patch as credited, I'm happy to merge this.

PR #1481, opened 2016-12-09
(FACT-1551) Initialize WSA before loading custom facts

MikaelSmith:
On Windows, Ruby initializes WSA via WSAStartup during its init. However,
that's only performed for the Ruby process, not when hosted as an embedded
runtime. Ensure we initialize WSA while loading custom facts so that facts
using network calls succeed.

### DIFF
--- a/lib/src/ruby/ruby.cc
+++ b/lib/src/ruby/ruby.cc
@@ -4,6 +4,9 @@
 #include <internal/ruby/ruby_value.hpp>
 #include <leatherman/ruby/api.hpp>
 #include <leatherman/logging/logging.hpp>
+#ifdef _WIN32
+#include <internal/util/windows/wsa.hpp>
+#endif
 
 #include <numeric>
 
@@ -47,6 +50,13 @@ namespace facter { namespace ruby {
 
     void load_custom_facts(collection& facts, bool initialize_puppet, vector<string> const& paths)
     {
+#ifdef _WIN32
+        // Initialize WSA before resolving custom facts. The Ruby runtime does this only when running
+        // in a Ruby process, it leaves it up to us when embedding it. See
+        // https://github.com/ruby/ruby/blob/v2_1_9/ruby.c#L2011-L2022 for comments.
+        // The only piece we seem to need out of rb_w32_sysinit is WSAStartup.
+        util::windows::wsa winsocket;
+#endif
         api& ruby = api::instance();
         module mod(facts, {}, !initialize_puppet);
         if (initialize_puppet) {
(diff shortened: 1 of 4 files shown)


MikaelSmith:
@puppetlabs/windows anyone have suggestions on how to test Windows network activity without requiring an internet connection? Or should we just assume network connectivity for unit tests?

glennsarti:
Could you ping localhost?  Not sure if that's enough of a test though?

MikaelSmith:
I'm not necessarily sure how to do that in a way that exercises winsock from within Ruby.

lib/tests/ruby/ruby.cc
```diff
@@ -631,4 +631,10 @@ SCENARIO("custom facts written in Ruby") {
             REQUIRE(ruby_value_to_string(facts.get<ruby_value>("bignum_fact")) == "12345678901");
         }
     }
+    GIVEN("a fact that performs network activity") {
+        REQUIRE(load_custom_fact("expect_network_init.rb", facts));
```
MikaelSmith: This doesn't test `load_custom_facts`, awesome.

MikaelSmith:
Updated, hopefully now it passes CI.

Iristyle:
Thanks for merging @Magisus - this looks sensible to me.

lib/src/ruby/ruby.cc
```diff
@@ -47,6 +50,13 @@ namespace facter { namespace ruby {
 
     void load_custom_facts(collection& facts, bool initialize_puppet, vector<string> const& paths)
     {
+#ifdef _WIN32
+        // Initialize WSA before resolving custom facts. The Ruby runtime does this only when running
+        // in a Ruby process, it leaves it up to us when embedding it. See
+        // https://github.com/ruby/ruby/blob/v2_1_9/ruby.c#L2011-L2022 for comments.
+        // The only piece we seem to need out of rb_w32_sysinit is WSAStartup.
+        util::windows::wsa winsocket;
```
Iristyle: Does `load_custom_facts` only happen once?

MikaelSmith: When the object is destroyed, it'll cleanup the WSA config. load_custom_facts definitely doesn't get called from inside its own execution.

Iristyle: What I meant was... does `load_custom_facts` only happen once in a Facter runs lifetime?

I assumed there was some RAII going on with the `winsocket` object.

MikaelSmith:
I believe so, yes.

lib/src/ruby/ruby.cc
MikaelSmith: I believe so, yes.

PR #1376, opened 2016-07-08
(packaging) Use CMake project to specify version

MikaelSmith:
Switches to using the CMake project macro to specify version, and adds
version to the facter executable.

PR #1483, opened 2016-12-14
Revert "(packaging) Use CMake project to specify version"

MikaelSmith:
Reverts puppetlabs/facter#1376

The original commit missed that the gemspec used the version variables. Also, somehow we have no testing that covers the gem.


Magisus:
:(

MikaelSmith:
I can fix this up, I'd just like to look into whether we should be testing the specfile somehow.

PR #1486, opened 2017-01-06
(FACT-1510) Use absolute rather than canonical

MikaelSmith:
Similar to FACT-1276, it's not necessary to get the full canonical path
for external facts dir - we only need it to be absolute. Attempting to
canonicalize paths is more expensive, and can fail on Windows with
symlinks.

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -297,12 +297,10 @@ namespace facter { namespace ruby {
         for (auto dir : paths) {
             _additional_search_paths.emplace_back(dir);
 
-            // Get the canonical directory name
-            boost::system::error_code ec;
-            path directory = canonical(_additional_search_paths.back(), ec);
-            if (ec) {
-                continue;
-            }
+            // Get the absolute directory name
+            // Absolute is used over canonical because canonical on Windows won't recognize paths as valid if
+            // they resolve to symlinks to non-NTFS volumes.
+            path directory = absolute(_additional_search_paths.back());
 
             _search_paths.push_back(directory.string());
         }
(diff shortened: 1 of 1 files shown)


MikaelSmith:
I'm a little concerned about what logging we lose from this.

peterhuene:
👍I think we can do without that logged message; seems like the only condition when `absolute` throws is when memory allocation fails.

MikaelSmith:
Before it'd warn if the directory didn't exist. Unfortunately sometimes it does that incorrectly.

MikaelSmith:
Also, still verifying that this fixes the problem.

MikaelSmith:
Retargeted at LTS.

lib/src/ruby/module.cc
```diff
@@ -297,12 +297,8 @@ namespace facter { namespace ruby {
         for (auto dir : paths) {
             _additional_search_paths.emplace_back(dir);
 
-            // Get the canonical directory name
-            boost::system::error_code ec;
-            path directory = canonical(_additional_search_paths.back(), ec);
-            if (ec) {
-                continue;
-            }
+            // Get the absolute directory name
```
ferventcoder: Any chance you can add more comments as to why absolute over canonical (in that canonical causes errors with symlinks when the targets are not the same file system type (likely due to stat)

ferventcoder: I just don't want someone else to come along later and not research why this change was made and go back to canonical.

ferventcoder:
> Similar to FACT-1276, it's not necessary to get the full canonical path
for external facts dir - we only need it to be absolute. Attempting to
canonicalize paths is more expensive, and can fail on Windows with
symlinks.

As far as I know, it only fails on Windows when attempting to use a symlink that is not also NTFS. This is likely due to things like stat or checking permissions. We had similar issues in Puppet::FileSystem that needed to change.

lib/src/ruby/module.cc
MikaelSmith: In comments, rather than the commit? I guess I can add it to the first instance.

MikaelSmith:
Updated.

lib/src/ruby/module.cc
ferventcoder: In commit, but also in comments 👍

MikaelSmith:
Ping @ferventcoder, @puppetlabs/facter-maintainers

ferventcoder:
LGTM - thanks for adding those comments!

Magisus:
👍

PR #1491, opened 2017-01-20
(FACT-1559) fix os facts on latest CoreOS

masterzen:
The latest CoreOS stable release 1235.6.0 introduces a renaming of
the CoreOS project to Container Linux CoreOS.
Unfortunately this affects the `/etc/os-release` NAME entry which
is used to discover the os facts in facter.
This patch now also uses the ID entry of `/etc/os-release` which presumably
will be more stable (see https://www.freedesktop.org/software/systemd/man/os-release.html
for more information).


MikaelSmith:
Thanks! Hopefully that is more stable; an alternative seems to be using a partial string match on the name.

PR #1492, opened 2017-01-20
(FACT-1560) Switch to newer Solaris 11 LDOM

LuvCurves:


MikaelSmith:
It would be really cool to abstract this sometime.

joshcooper:
@rick is working on adding hardware support to ABS, so that in the future, CI pipelines can ask for a solaris 11 resource, and ABS will be the single point to map that to a physical or virtual host.

PR #1495, opened 2017-01-24
(FACT-1561) Use top namespace when extending TestCase

tvpartytonight:
A recent change in Beaker renamed one of the module names from
"FacterHelpers" to simply "Facter". This exposed a flaw in the way tests
in this repo were extending the TestCase object. When not specifying the
top namespace, the `extend` method introspects the current module first
before looking out to the `$LOAD_PATH`. Now that `Facter` is a module
included in the DSL, attempting to execute `extend Facter::Lib` will
result in ruby looking into `Beaker::DSL::Facter::Lib`.

This change prepends a `::` to all appropriate module names in testing,
so that ruby knows to look for these helper libraries at the top level
namespace. Fixing the tests here will make them more resilient to future
beaker changes.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -7,7 +7,7 @@ require 'beaker-hostgenerator'
 
 $LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), 'lib'))
 require 'puppet/acceptance/git_utils'
-extend Puppet::Acceptance::GitUtils
+extend ::Puppet::Acceptance::GitUtils
 
 ONE_DAY_IN_SECS = 24 * 60 * 60
 REPO_CONFIGS_DIR = "repo-configs"
--- a/acceptance/setup/aio/pre-suite/010_Install.rb
+++ b/acceptance/setup/aio/pre-suite/010_Install.rb
@@ -1,6 +1,6 @@
 require 'puppet/acceptance/common_utils'
 require 'beaker/dsl/install_utils'
-extend Beaker::DSL::InstallUtils
+extend ::Beaker::DSL::InstallUtils
 
 test_name "Install Packages"
 
--- a/acceptance/setup/common/00_EnvSetup.rb
+++ b/acceptance/setup/common/00_EnvSetup.rb
@@ -3,11 +3,11 @@ test_name "Setup environment"
 step "Ensure Git and Ruby"
 
 require 'puppet/acceptance/install_utils'
-extend Puppet::Acceptance::InstallUtils
+extend ::Puppet::Acceptance::InstallUtils
 require 'puppet/acceptance/git_utils'
-extend Puppet::Acceptance::GitUtils
+extend ::Puppet::Acceptance::GitUtils
 require 'beaker/dsl/install_utils'
-extend Beaker::DSL::InstallUtils
+extend ::Beaker::DSL::InstallUtils
 
 PACKAGES = {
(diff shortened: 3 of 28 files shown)


tvpartytonight:
Do I need to convert this to a facter ticket in order for it to be merged?

Magisus:
Yes, or use (maint) perhaps. Though since this change already slipped through once, it might be nice to have an actual ticket.

tvpartytonight:
Updated, now with rhyming commit title.

ScottGarman:
:+1:

MosesMendoza:
@tvpartytonight Is the source of this trouble https://github.com/puppetlabs/beaker/commit/cf064f13387260db4a064c5a9863171597070f35? 
Do we know if there other users out there using beaker who might be bit by this too? Renaming a public beaker module within a feature series worries me a bit. I'm not sure we should need to make all these changes in facter to pull in a new beaker feature version..

acceptance/Rakefile
```diff
@@ -7,7 +7,7 @@ require 'beaker-hostgenerator'
 
 $LOAD_PATH << File.expand_path(File.join(File.dirname(__FILE__), 'lib'))
 require 'puppet/acceptance/git_utils'
-extend Puppet::Acceptance::GitUtils
+extend ::Puppet::Acceptance::GitUtils
```
Iristyle: Having to prefix an extend namespace with `::` is a really uncommon idiom.  This seems like a code smell to me, but digging in a but further to understand what's being described here.

joshcooper:
I share @MosesMendoza concerns. It's surprising that consumers of beaker would need to change during a Y version update.

tvpartytonight:
@MosesMendoza the beaker link above is indeed the commit that changed the module included in beaker from "FacterHelpers" to "Facter". This repo is the only scenario I have seen where the `extend` method is used inside of TestCase to reference a module named `Facter`. To be clear, I am not aware of any other project trying to reference the module [here](https://github.com/puppetlabs/facter/blob/master/acceptance/tests/runs_external_facts_once.rb) from within TestCase.

I agree that the changing module name is not suitable for a y-release, and will be [reverting](https://github.com/puppetlabs/beaker/pull/1328) this change for the beaker release scheduled for Wednesday Jan 25th.

While that change will fix the facter acceptance tests, I still feel strongly that these tests should be more resilient and guard against future possible collisions with modules included in `TestCase`. The `extend` method will always search the current module in place, and there are no guarantees that beaker will not introduce modules that share names with the helper modules declared and used in this job.

@Iristyle I'm not aware of a more idiomatic way to reference the global scope in ruby...

Iristyle:
@tvpartytonight my comment wasn't with how to reach global scope, it was that global scope usually shouldn't be necessary.

MosesMendoza:
@tvpartytonight definitely I'm always +1 to resiliency, and collisions are a fact of life in ruby - agreed there's no guarantees we won't run into more of them. But within the facter project, which defines the `Facter` namespace, I'm inclined to say reasonable efforts towards resiliency don't include having to explicitly global-scope `Facter`. Given we own all the code here, I suggest when name collisions occur that introduce regression, we try first at renaming to alleviate them.

Iristyle:
Given Beaker has reverted the namespace changes and cut a `3.10.0` as of 1/25 - see https://rubygems.org/gems/beaker/versions/3.10.0 I'm closing this PR.

PR #1496, opened 2017-01-24
(maint) Switch to new Solaris LDOM

LuvCurves:


MikaelSmith:
Does this need to be used for LTS testing? If so, please retarget at LTS-1.7 and I'll merge to stable.

PR #1500, opened 2017-01-30
(FACT-1562) Use stoul instead of stoi

branan:
When parsing AIX filesystem structures, we were previously using stoi
for converting strings to integers. This assumes signed integers, and
would throw an exception for any value greater than 0x7fffffff. Instead,
we use stoul, which won't fail on such integers.


McdonaldSeanp:
👍

PR #1501, opened 2017-02-02
(FACT-1544) Removed unused cache files

Magisus:
Previously, when a fact that used to be cached was removed from the
config file, the corresponding cache file continued to exist on disk,
even though Facter ignored it. This commit ensures that such superfluous
files get removed whenever facts are resolved.

### DIFF
--- a/lib/inc/internal/facts/cache.hpp
+++ b/lib/inc/internal/facts/cache.hpp
@@ -59,4 +59,13 @@ namespace facter { namespace facts { namespace cache {
      * @return the timespan in seconds since last modification
      */
     int64_t get_file_lifetime(boost::filesystem::path file_path);
+
+    /**
+     * Removes the cache file for any resolver that does not appear in the
+     * supplied list of fact groups.
+     * @param facts_to_cache the list of fact groups that should be cached
+     * @param cache_location the absolute path to the fact cache directory
+     */
+    void clean_cache(std::unordered_map<std::string, int64_t> const& facts_to_cache,
+            std::string cache_location = fact_cache_location());
 }}}  // namespace facter::facts::cache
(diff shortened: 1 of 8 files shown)


lib/inc/internal/facts/cache.hpp
```diff
@@ -59,4 +59,13 @@ namespace facter { namespace facts { namespace cache {
...
      */
     int64_t get_file_lifetime(boost::filesystem::path file_path);
+
+    /**
+     * Removes the cache file for any resolver that does not appear in the
+     * supplied list of fact groups.
+     * @param facts_to_cache the list of fact groups that should be cached
+     * @param cache_location the absolute path to the fact cache directory
+     */
+    void clean_cache(std::unordered_map<std::string, int64_t> facts_to_cache,
```
MikaelSmith: Why make a copy? `const&` seems to make sense here.

lib/src/facts/cache.cc
```diff
@@ -16,6 +16,22 @@ namespace boost_file = boost::filesystem;
...
+        if (!boost_file::is_directory(cache_dir)) {
+            return;
+        }
+        for (boost_file::directory_iterator itr(cache_dir);
+                itr != boost_file::directory_iterator();
+                ++itr) {
+            boost_file::path cache_file = itr->path();
+            if (!facts_to_cache.count(cache_file.filename().string())) {
+                LOG_DEBUG("Deleting unused cache file {1}", cache_file.string());
+                boost_file::remove(cache_file);
```
MikaelSmith: This could throw an exception if the directory is deleted between checking for existence and removing it, or if we don't have permission to delete it. I'd pass the error_code argument and LOG_DEBUG if there was an error, but otherwise continue.

PR #1504, opened 2017-02-04
(maint) Do not rely on external networking for WSASetup validation

branan:
For FACT-1551, we added a test to ensure we had called WSASetup before
trying to resolve Ruby facts. As originally implemented, this test
required public network access and a functional external resource.

Most of this is not necessary to validate that WSA has been setup, and
makes the test more fragile. Instead, we can simply ensure that
creating and closing a socket proceeds without error.

### DIFF
--- a/lib/src/ruby/ruby.cc
+++ b/lib/src/ruby/ruby.cc
@@ -4,9 +4,6 @@
 #include <internal/ruby/ruby_value.hpp>
 #include <leatherman/ruby/api.hpp>
 #include <leatherman/logging/logging.hpp>
-#ifdef _WIN32
-#include <internal/util/windows/wsa.hpp>
-#endif
 
 #include <numeric>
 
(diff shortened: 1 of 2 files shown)


branan:
This is targeted at LTS, and should be merged up as we go

MikaelSmith:
Put up another commit on top to comment out the fix related to this test, and let's see if it gets through AppVeyor.

branan:
@MikaelSmith Sounds like a good plan. I want to see this go green as-is in appveyor/travis first, since I had some local failures that I think (hope) are related to our tests not being run on Arch all the time.

branan:
Hmph, that's not good.

branan:
OK, so it looks like despite Microsoft's docs, simply opening a socket isn't enough to fail this stuff. I'm gonna try some more things

branan:
Fingers crossed that take three here will do the trick :fearful:

MikaelSmith:
I think it might specifically be the hostname resolution that requires WSAStartup.

branan:
@MikaelSmith THAT'S NOT WHAT MICROSOFT'S DOCS SAY, but I would totally believe it, and that makes this more of a pain :(

branan:
I'm gonna close this for now - it's not turning out to be the quick fix I hoped it was, and it's not all that important. I'll play with it again another day.

PR #1506, opened 2017-02-09
(maint) fix AIX fact test for AIX 7.2

McdonaldSeanp:
Add the kernel_release and kernel_major_version vars for AIX 7.2


McdonaldSeanp:
This will need to be merged up through master

PR #1514, opened 2017-02-17
Revert "(FACT-1541) javah command runs late in parallel build"

Magisus:
Reverts puppetlabs/facter#1476.

In order to validate puppet-agent for the 1.9.2 release, we need some acceptance changes that landed after the commit this reverts. In order to feel safe pinning back to the 3.6.1 tag in that release, we are reverting all non-acceptance changes from this branch. The work in this commit has already been merged up to the master branch, so it will go out in the next major or minor release. This commit can be itself reverted after the release of 1.9.2 if so desired.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -265,17 +265,9 @@ if (JRUBY_SUPPORT)
     )
 
     include(UseJava)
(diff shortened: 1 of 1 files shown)


MikaelSmith:
You'll have to be careful about merging this to master so that it doesn't apply the revert there as well.

Magisus:
Change of plans, not reverting. Change was deemed no risk to validation followed by pinning to 3.6.1.

PR #1517, opened 2017-03-06
(FACT-1579) Devuan support

faust64:
Adding devuan release file and distribution name strings.


MikaelSmith:
Thanks for adding this!

PR #1518, opened 2017-03-07
(FACT-1583) Output exception message when test fails

joshcooper:
For reasons unknown, the custom fact that relies on networking sometimes
fails on Fedora 24, but the test output only says that "" doesn't equal
"Yay".

If an exception occurs, output the message as the fact resolution, which
won't equal "Yay". That way we can debug the issue if/when it occurs
again.


joshcooper:
Kicked off an adhoc run in https://jenkins.puppetlabs.com/view/puppet-agent/view/ad%20hoc/job/platform_puppet-agent_init-van-int_suite-manual-ad-hoc/505

Magisus:
👍

PR #1519, opened 2017-03-08
(FACT-1583) Use mock server to test HTTP custom fact

MikaelSmith:
Uses a mock server to test an HTTP custom fact. This avoids intermittent
failures when dns resolution fails or the network is unavailable.

### DIFF
--- a/lib/tests/mock_server.cc
+++ b/lib/tests/mock_server.cc
@@ -1,0 +1,44 @@
+#include <cstdlib>
+#include <iostream>
+#include <utility>
+#include "mock_server.hpp"
+
+namespace facter {
+
+static void session(tcp::socket sock) {
+  try {
+      for (;;) {
+          char data[1024];
+
+          boost::system::error_code error;
+          sock.read_some(boost::asio::buffer(data), error);
+          if (error == boost::asio::error::eof) {
+              break; // Connection closed cleanly by peer.
+          } else if (error) {
+              throw boost::system::system_error(error); // Some other error.
+          }
+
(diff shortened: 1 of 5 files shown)


joshcooper:
Looks like a valid appveyor compile error

MikaelSmith:
Ah, strict-aliasing warning. I'll fix up in a minute.

MikaelSmith:
Fixed.

lib/tests/mock_server.cc
```diff
@@ -0,0 +1,44 @@
...
+
+          boost::system::error_code error;
+          sock.read_some(boost::asio::buffer(data), error);
+          if (error == boost::asio::error::eof) {
+              break; // Connection closed cleanly by peer.
+          } else if (error) {
+              throw boost::system::system_error(error); // Some other error.
+          }
+
+          std::string response = "HTTP/1.1 301 Moved Permanently\nContent-length: 0\nLocation: https://puppet.com/\nConnection: close\n\n";
```
joshcooper: Will  `Net::HTTP.get_response` follow the redirect automatically? Just want to make sure the client doesn't try to resolve the name and cause the same transient?

MikaelSmith:
In my testing it simply returns the content, it didn't follow the redirect. I could change the content of this, just went with something easy to produce.

lib/tests/mock_server.cc
MikaelSmith: In my testing it simply returns the content, it didn't follow the redirect. I could change the content of this, just went with something easy to produce.

MikaelSmith:
I think this will compile everywhere, because I modeled it on something similar in cpp-pcp-client. If it doesn't, revert and we can work on fixing up the failing platforms.

PR #1520, opened 2017-03-09
(FACT-1583) Fix compiler issues

MikaelSmith:
Overlooked two issues when sourcing from cpp-pcp-client tests:
- BUILD_SHARED_LIBS in the top-level CMakeLists
- ignoring unused variable warnings from asio

Fix those issues.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -299,17 +299,30 @@ if(AIX)
     set_target_properties(libfacter PROPERTIES LINK_FLAGS "-Wl,-G -eInit_libfacter")
 endif()
 
-# Link in additional libraries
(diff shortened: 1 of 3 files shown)


MikaelSmith:
@joshcooper

joshcooper:
appveyor again :(

MikaelSmith:
Better now than later.

MikaelSmith:
I'll have to continue with this tomorrow :(

MikaelSmith:
Ok, passing here. This got more complicated than I expected, and doesn't completely follow the same issues that cpp-pcp-client dealt with. I'm not quite sure why the tests would segfault if mock-server is included at the end on Linux.

PR #1521, opened 2017-03-09
(FACT-1585) Change metadata host to long hostname

zbikmarc:
In some situations when non google DNS is used, facter can not resolve metadata hostname.

Long hostname is automatically added by google
`/etc/host`
```
169.254.169.254 metadata.google.internal  # Added by Google
```
Moreover google docs are using long hostname in examples. (https://cloud.google.com/compute/docs/storing-retrieving-metadata)

### DIFF
--- a/lib/src/facts/resolvers/gce_resolver.cc
+++ b/lib/src/facts/resolvers/gce_resolver.cc
@@ -240,7 +240,7 @@ namespace facter { namespace facts { namespace resolvers {
 
         try
         {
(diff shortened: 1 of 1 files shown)


branan:
Since this is a bugfix, I'm going to cherry-pick it down to stable. If it lands on master, it won't go out until we do Puppet 5, which is at least a couple more months.

branan:
This was cherry-picked as cd5d2eccdf052b966eaf18d0f6a4faeb15a752af

PR #1522, opened 2017-03-09
(FACT-1582) add non-root user fact test

ericldelaney:
Need to add tests for TestRail C64580 facter as non-root user facter should honor fact paths
$HOME/.facter/facts.d
$HOME/.puppetlabs/opt/facter/facts.d

### DIFF
--- a/acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
+++ b/acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
@@ -1,0 +1,137 @@
+test_name "Non-root default user external facts directory is searched for facts" do
+
+  confine :except, :platform => 'aix'     # bug FACT-1586
+
+  confine :except, :platform => 'windows' # this test is for unix systems
+  confine :except, :platform => 'osx'     # does not support managehome
+  confine :except, :platform => 'solaris' # does not work with managehome on solaris boxes
+  confine :except, :platform => 'eos-'    # does not support user creation ARISTA-37
+
+  # TestRail: C64580
+
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  # Generate an external fact dynamically
+  def ext_user_fact(value='BASIC')
+    "test: '#{value}'"
+  end
+
+  # Retrieve a specific user's home directory $HOME/.facter
+  #
+  def get_user_facter_dir(user_home, platform)
+    File.join(user_home, '.facter')
+  end
+
+  # Retrieve a specific user's home facts directory $HOME/.facter/facts.d
+  #
+  def get_user_facts_dir(user_home, platform)
+    File.join(get_user_facter_dir(user_home, platform), 'facts.d')
+  end
+
+  # Retreive a specific user's home puppetlabs directory $HOME/.puppetlabs
+  #
+  def get_user_puppetlabs_dir(user_home, platform)
+    File.join(user_home, '.puppetlabs')
+  end
+
+  # Retreive a specific user's home puppetlabs facts directory $HOME/.puppetlabs/opt/facter/facts.d
+  #
+  def get_user_puppetlabs_facts_dir(user_home, platform)
+    File.join(get_user_puppetlabs_dir(user_home, platform), 'opt', 'facter', 'facts.d')
+  end
+
+  # retrieve the user's home directory for a host and user
+  #
+  def get_home_dir(host, user_name)
+    home_dir = nil
+    on host, puppet_resource('user', user_name) do |result|
+      home_dir = result.stdout.match(/home\s*=>\s*'([^']+)'/m)[1]
+    end
+    home_dir
+  end
+
+  # retrieve the correct shell path for system under test
+  #
+  def user_shell(agent)
+    if agent['platform'] =~ /aix/
+      '/usr/bin/bash'
+    else
+      '/bin/bash'
+    end
+  end
+
+  agents.each do |agent|
+    non_root_user = "nonroot"
+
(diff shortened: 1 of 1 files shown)


acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
```diff
@@ -0,0 +1,157 @@
...
+
+require 'facter/acceptance/user_fact_utils'
+extend Facter::Acceptance::UserFactUtils
+
+# Generate an external fact dynamically
+def ext_user_fact(value='BASIC')
+  "test: '#{value}'"
+end
+
+# Retrieve a specific users home directory $HOME/.facter
```
Magisus: This comment and all the ones following for the utility methods should have "user's" instead of "users"

Magisus:
The logic makes sense. If this passes on the platforms it should, 👍

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
```diff
@@ -0,0 +1,157 @@
...
+#
+def get_user_facter_dir(user_home, platform)
+  if platform =~ /windows/
```
kurtwall: I'm not sure this guard is necessary (here and below) because you've already confined it at the beginning of the test (`confine :except, :platform => 'windows'`).

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
```diff
@@ -0,0 +1,157 @@
...
+  user_facts_path = "#{user_facts_dir}/test.yaml"
+
+  user_base_puppetlabs_dir = get_user_puppetlabs_dir(user_home, agent['platform'])
+  user_puppetlabs_facts_dir = get_user_puppetlabs_facts_dir(user_home, agent['platform'])
+  user_puppetlabs_facts_path = "#{user_puppetlabs_facts_dir}/test.yaml"
+
+  step "Agent #{agent}: figure out facter program location"
+  facter_path = on(agent, "which facter").stdout.chomp
+
+
```
kurtwall: Extra newline? See also lines 122, 139.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
```diff
@@ -0,0 +1,157 @@
...
+  step "Agent #{agent}: chown and chmod the facts to the user #{non_root_user}"
+  on(agent, "chown -R #{non_root_user} #{user_base_facts_dir} #{user_base_puppetlabs_dir}")
+  on(agent, "chmod -R a+rx #{user_base_facts_dir} #{user_base_puppetlabs_dir}")
+
+  step "Agent #{agent}: run facter as #{non_root_user} and .facter will take precedence over .puppetlabs"
+  on(agent, %Q[su #{non_root_user} -c "#{facter_path} test"]) do
+    assert_match(/USER_PRECEDENCE_FACTER/, stdout, "Fact from #{user_puppetlabs_facts_dir} did not resolve correctly")
+  end
+
+  teardown do
```
kurtwall: Most of the tests here at Puppet place the teardown at the top of the test before the other steps. Makes it easy to see that a test cleans up after itself. Not all do, unfortunately.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
```diff
@@ -0,0 +1,157 @@
...
+  user_base_facts_dir = get_user_facter_dir(user_home, agent['platform'])
+  user_facts_dir = get_user_facts_dir(user_home, agent['platform'])
+  user_facts_path = "#{user_facts_dir}/test.yaml"
+
+  user_base_puppetlabs_dir = get_user_puppetlabs_dir(user_home, agent['platform'])
+  user_puppetlabs_facts_dir = get_user_puppetlabs_facts_dir(user_home, agent['platform'])
+  user_puppetlabs_facts_path = "#{user_puppetlabs_facts_dir}/test.yaml"
+
+  step "Agent #{agent}: figure out facter program location"
+  facter_path = on(agent, "which facter").stdout.chomp
```
kurtwall: Not sure it will matter here, but I'll point out that `which facter` will return the path to the first match. That's probably what we want, but it's no the only possible option. Facter might (wrongly) be installed in two locations, for example.

kurtwall:
Code looks good. A couple of changes and a couple of formatting/structure nits.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
ericldelaney: At the moment, that is true. I was worried that I might end up moving them to user_fact_util.rb when I need them in a different test and then we'll want  the guard. I guess I want until to add them.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
ericldelaney: Yep, but based on the PATH we're always testing the first one found.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
ericldelaney: Yeah I was trying to break up the units of work.

acceptance/tests/external_facts/non_root_users_default_external_fact_directory.rb
ericldelaney: There appears to be a 60/40 split on beginning or end for putting teardown in the tests in facter. I'll work on converting them over as I go.

PR #1526, opened 2017-03-16
(FACT-1591) add TestRail numbers and missing teardown code

ericldelaney:
Add TestRail number to the associated test_name strings
While I was there I added teardown code for a couple missing items
tested on Linux and windows

### DIFF
--- a/acceptance/tests/options/config_file/debug.rb
+++ b/acceptance/tests/options/config_file/debug.rb
@@ -1,6 +1,7 @@
 # This test is intended to demonstrate that setting the cli.debug field to true
 # causes DEBUG information to be printed to stderr.
-test_name "setting the debug config field to true prints debug info to stderr" do
+test_name "C99965: setting the debug config field to true prints debug info to stderr" do
+
   config = <<EOM
 cli : {
     debug : true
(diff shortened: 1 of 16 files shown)


Magisus:
I was under the impression you didn't have to do teardown on directories created with tmpdir?

ericldelaney:
Good question on tmpdir(), looking at the code we create a directory with a random name, but I don't know that anything will clean up the file or directory created. Which I guess on a VM isn't a problem per say. Probably the biggest problem would be on AIX where you might get the same file/directory, but your chances are probably pretty low.

    execute("rndnum=${RANDOM} && mkdir /tmp/#{name}.${rndnum} && echo /tmp/#{name}.${rndnum}")

kurtwall:
@Magisus I delete temp dirs created using `tmpdir` during `teardown` because that's the prior art I encountered when I first started writing beaker tests. I'm sure I just assumed it was required, but I have no idea if that is actually the case. The name `tmpdir` notwithstanding, of course.

Magisus:
Thanks for the info! I was curious because I wrote a lot of these tests originally.

PR #1532, opened 2017-03-23
(FACT-1593) Disable proxy when fetching metadata.

pillarsdotnet:
From [FACT-1593](https://tickets.puppetlabs.com/browse/FACT-1593):

> Facter version 2.4.6 uses the openuri mixin, which is affected by the `ENV['http_proxy']` setting.
> 
> It retrieves EC2 metadata by contacting the metadata server at 169.254.169.254.  If `ENV['http_proxy']` is set, facter will wind up retrieving the metadata of the proxy server rather than its own metadata.
> 
> To avoid this, calls to open must include the `:proxy => nil` modifier.  In fact, some of the metadata calls have this modifier, but not all.
>

### DIFF
--- a/lib/facter/ec2/rest.rb
+++ b/lib/facter/ec2/rest.rb
@@ -123,7 +123,7 @@ module Facter
       end
 
       def fetch
-        open(@baseurl).read
+        open(@baseurl, :proxy => nil).read
       rescue OpenURI::HTTPError => e
         if e.message.match /404 Not Found/i
           return nil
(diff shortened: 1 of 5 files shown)


MikaelSmith:
Some background on this would be appreciated. Do you have a ticket filed about why this is needed?

pillarsdotnet:
Ticket is filed at [FACT-1593](https://tickets.puppetlabs.com/browse/FACT-1593).

When our AWS instances run facter with `ENV['http_proxy']` set, we're seeing requests for http://169.254.169.254/ in the proxy logs.

pillarsdotnet:
Updated description to reference the open ticket.

PR #1536, opened 2017-03-27
(FACT-1595) update debug tests, override cases, and split no_ruby test

ericldelaney:
Updated test for code style, risk tags, command output collection, and TestRail numbers
Split the no_ruby tests into separate tests
Split the command line override tests into separate tests
Renamed the ttls.rb test file which will be split out in a different pull request

Tested on:
arista4-32a, ciscoxr-64a%7Bhypervisor=vmpooler%7D, cisconx-64a%7Bhypervisor=vmpooler%7D, cumulus25-64a, debian8-64a, fedora25-64a, osx1012-64a, redhat7-64a, sles12-64a, solaris11-64a, ubuntu1604-64a, windows10ent-64a, windows2016-64a


acceptance/tests/options/config_file/command_line_override.rb
```diff
@@ -1,9 +1,8 @@
-# This test is intended to demonstrate that all command-line flags take precedence
```
Magisus: Could you give this file a more specific name? I see other tests below whose names have been changed to also reference command line overriding, that include which setting they're testing.

acceptance/tests/options/config_file/debug.rb
```diff
@@ -22,8 +24,8 @@
       end
 
       step "debug output should print when config file is loaded" do
-        on(agent, facter("")) do
-          assert_match(/DEBUG/, stderr, "Expected DEBUG information in stderr")
+        on(agent, facter("")) do |facter_output|
```
Magisus: Out of curiosity, what does using the block argument do differently here? Is it just for readability or is the functionality different?

ericldelaney: John said out that I should be doing it this way so there is not a confusion of where the stdout or stderr is coming from.

Magisus: Cool, thanks

acceptance/tests/options/config_file/command_line_override.rb
ericldelaney: Yeah, I pulled a test out, but didn't come back and rename it. I will

ericldelaney:
@Magisus now updated with the file name change.

Magisus:
👍

PR #1541, opened 2017-04-04
(FACT-1446) Add missing Windows platforms

jcoconnor:
The Windows OS checking doesn't differentiate between Windows-2012 &
2012R2. Similar for Windows-2008/2008R2 and tests for Windows 7 and 8.1
are missing. 

This causes the new Windows/Packer/Imaging facter acceptance testing to fail.

This PR replaces #1540

### DIFF
--- a/acceptance/tests/facts/windows.rb
+++ b/acceptance/tests/facts/windows.rb
@@ -12,19 +12,31 @@ confine :to, :platform => /windows-/
 agents.each do |agent|
 
   # Get expected values based on platform name
-  if agent['platform'] =~ /2003/
+  if agent['platform'] =~ /-2003/
     os_version  = '2003 R2'
     kernel_version = '5.2'
-  elsif agent['platform'] =~ /2008/
+  elsif agent['platform'] =~ /-2008r2/
     os_version  = '2008 R2'
     kernel_version = '6.1'
-  elsif agent['platform'] =~ /2012/
+  elsif agent['platform'] =~ /-2008/
+    os_version  = '2008'
+    kernel_version = '6.0'
+  elsif agent['platform'] =~ /-2012r2/
     os_version  = '2012 R2'
+    kernel_version = '6.3'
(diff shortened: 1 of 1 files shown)


jcoconnor:
@Magisus I have opened this PR under FACT-1446 - are you able to review the changes along with @glennsarti ?

glennsarti:
OS version numbers seem to match up (https://en.wikipedia.org/wiki/Comparison_of_Microsoft_Windows_versions)

acceptance/tests/facts/windows.rb
```diff
@@ -15,12 +15,24 @@
...
     kernel_version = '5.2'
-  elsif agent['platform'] =~ /2008/
+  elsif agent['platform'] =~ /2008r2/
     os_version  = '2008 R2'
     kernel_version = '6.1'
-  elsif agent['platform'] =~ /2012/
+  elsif agent['platform'] =~ /2008/
+    os_version  = '2008'
+    kernel_version = '6.0'
+  elsif agent['platform'] =~ /2012r2/
```
glennsarti: Should the regex's be more specific?

e.g.  `2012r2` will match `/2012/
 perhaps use `/-2012$/` ?

Also `/8.1/`  the `.` should be escaped

jcoconnor: I've escaped the ```8.1``` as suggested. Using ```$``` may not work as ```arch``` follows the 2012/2008.

However, think it would be no harm for consistency to prefix all the OS versions with "-".
E.g. ```elsif agent['platform'] =~ /-2008/```

Magisus: I think because the switch statement will be evaluated in order, the regexes should be fine. So long as we test `2012r2` before `2012` I think it will do the right thing.

Magisus:
👍

McdonaldSeanp:
👍

PR #1544, opened 2017-04-06
(FACT-1599) add --color and --no-color tests

ericldelaney:
Added tests to verify that --debug with --color and --no-color actually
produce escape sequences to change the output color as appropriate

### DIFF
--- a/acceptance/tests/options/color.rb
+++ b/acceptance/tests/options/color.rb
@@ -1,0 +1,16 @@
+# This test is intended to ensure with --debug and --color, facter sends escape sequences to colorize the output
+test_name "C86545: --debug and --color command-line options should print DEBUG messages with color escape sequences" do
+  tag 'risk:high'
+
+  confine :except, :platform => 'windows' # On windows we don't get an escape sequence to detect to color change
+
+  agents.each do |agent|
+    step "Agent #{agent}: retrieve debug info from stderr using --debug and --color option" do
+      # set the TERM type to be a color xterm to help ensure we emit the escape sequence to change the color
+      on(agent, facter('--debug --color'), :environment => { 'TERM' => 'xterm-256color' }) do |facter_output|
+        assert_match(/DEBUG/, facter_output.stderr, "Expected DEBUG information in stderr")
+        assert_match(/\e\[0;/, facter_output.stderr, "Expected to see an escape sequence in the output")
+      end
+    end
+  end
+end
--- a/acceptance/tests/options/no_color.rb
+++ b/acceptance/tests/options/no_color.rb
@@ -1,0 +1,16 @@
+# This test is intended to ensure with --debug and --no-color, facter does not send any escape sequences
+# to colorize the output
+test_name "C99975: --debug and --no-color command-line options should print DEBUG messages without color escape sequences" do
+  tag 'risk:low'
+
+  confine :except, :platform => 'windows' # On windows we don't get an escape sequence so we can't detect a color change
+
+  agents.each do |agent|
+    step "Agent #{agent}: retrieve debug info from stderr using --debug anod --no-color options" do
+      on(agent, facter('--debug --no-color')) do |facter_output|
+        assert_match(/DEBUG/, facter_output.stderr, "Expected DEBUG information in stderr")
+        refute_match(/\e\[0;/, facter_output.stderr, "Expected to output to not contain an escape sequence")
+      end
+    end
+  end
+end


Magisus:
Tried this on a couple of platforms, seems alright in general 👍

acceptance/tests/options/color.rb
```diff
@@ -0,0 +1,15 @@
+# This test is intended to ensure with --debug and --color, facter sends escape sequences to colorize the output
+test_name "C864545: --debug and --color command-line options should print DEBUG messages with color escape sequences" do
+  tag 'risk:high'
```
kurtwall: Couldn't find this test case. Typo?

acceptance/tests/options/color.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  confine :except, :platform => 'windows' # On windows we don't get an escape sequence to detect to color change
```
kurtwall: Is there another way to test for color output in `cmd.exe` session?

acceptance/tests/options/color.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  agents.each do |agent|
+    step "Agent #{agent}: retrieve debug info from stderr using --debug and --color option" do
+      on(agent, facter('--debug --color'), :environment => { 'TERM' => 'xterm-256color' }) do |facter_output|
```
kurtwall: Same results in, say, a vanilla xterm? What about good ole vt100?

acceptance/tests/options/no_color.rb
```diff
@@ -0,0 +1,16 @@
+# This test is intended to ensure with --debug and --no-color, facter does not send any escape sequences
+# to colorize the output
+test_name "C99975: --debug and --no-color command-line options should print DEBUG messages without color escape sequences" do
+  tag 'risk:medium'
+
+  confine :except, :platform => 'windows' # On windows we don't get an escape sequence so we can't detect a color change
+
+  agents.each do |agent|
+    step "Agent #{agent}: retrieve debug info from stderr using --debug anod --no-color options" do
+      on(agent, facter('--debug --no-color'), :environment => { 'TERM' => 'xterm-256color' }) do |facter_output|
```
kurtwall: Same results in, say, a vanilla xterm? What about good ole vt100?

acceptance/tests/options/no_color.rb
```diff
@@ -0,0 +1,16 @@
+# This test is intended to ensure with --debug and --no-color, facter does not send any escape sequences
+# to colorize the output
+test_name "C99975: --debug and --no-color command-line options should print DEBUG messages without color escape sequences" do
```
kurtwall: Is there another way to test for color output in `cmd.exe` session?

acceptance/tests/options/color.rb
ericldelaney: Yes, typo, I'll fix it

acceptance/tests/options/color.rb
ericldelaney: Nope I tried, over ssh the nothing that I did would get colorization to work.

When I logged into the console and facter from the command line, I was able to see the color change, but windows doesn't do rely on the use of an escape sequence being sent by the program that is running to change the color.

acceptance/tests/options/color.rb
ericldelaney: Hmm. Didn't try that. I was trying to give the CI the best chance of passing as the escape sequence would be different if it was a different terminal type (assuming it supports color)

acceptance/tests/options/color.rb
johnduarte: +1 to @kurtwall 's risk assessment comment. This seems both low severity to the user and low probability of regression. I would rate it as 'risk:low'

acceptance/tests/options/no_color.rb
```diff
@@ -0,0 +1,16 @@
...
+  tag 'risk:medium'
```
johnduarte: This risk seems appropriate as is because the user impact is higher if the removal of escape sequences is not honored (e.g. script-based parsing of output).

acceptance/tests/options/color.rb
ericldelaney: I played around a bit more and it appears that we send the escape sequence regardless of what the terminal type setting is. (even dumb). I'm inclined to leave the terminal type set to give the test the best chance of working in CI just in case. I'll add a comment

ericldelaney:
Made changes based on review comments.

Passing on the following platforms:
arista4-32a ciscoxr-64a%7Bhypervisor=vmpooler%7D cisconx-64a%7Bhypervisor=vmpooler%7D cumulus25-64a debian8-64a fedora25-64a osx1012-64a redhat7-64a sles12-64a solaris11-64a ubuntu1604-64a windows10ent-64a windows2016-64a

acceptance/tests/options/no_color.rb
ericldelaney: Nope I tried, over ssh the nothing that I did would get colorization to work.

When I logged into the console and facter from the command line, I was able to see the color change, but windows doesn't do rely on the use of an escape sequence being sent by the program that is running to change the color.

PR #1545, opened 2017-04-07
(FACT-1578) add validation test for mount point byte fields

ericldelaney:
This test validates that the mount point available_bytes, size_bytes, and used_bytes fields are all either 0 or at least positive. (If you cast a number to something of the wrong size you can end up with a negative number, this should give us a chance at detecting this failure)

Tested on
arista4-32a ciscoxr-64a%7Bhypervisor=vmpooler%7D cisconx-64a%7Bhypervisor=vmpooler%7D cumulus25-64a debian8-64a fedora25-64a osx1012-64a redhat7-64a sles12-64a solaris11-64a ubuntu1604-64a windows10ent-64a windows2016-64a`

This test will fail on Solaris until FACT-1578 is fixed.

### DIFF
--- a/acceptance/tests/facts/validate_file_system_size_bytes.rb
+++ b/acceptance/tests/facts/validate_file_system_size_bytes.rb
@@ -1,0 +1,37 @@
+# This test verifies that the numbers for file systems sizes are positive from "facter" and "puppet facts"
+# This is a regression test for FACT-1578
+test_name "C100110: verify that file system sizes are positive" do
+  tag 'risk:high'
+
+  confine :except, :platform => 'windows' # Windows does not list mount points as facts like Unix
+
+  require 'json'
+
+  agents.each do |agent|
+    step("verify that facter returns positive numbers for the mount points byte fields") do
+
+      on(agent, facter("--json")) do |facter_output|
+        facter_results = JSON.parse(facter_output.stdout)
+        facter_results['mountpoints'].each_key do |mount_key|
+          ['available_bytes', 'size_bytes', 'used_bytes'].each do |sub_key|
+            assert_operator(facter_results['mountpoints'][mount_key][sub_key], :>=, 0,
+                            "Expected the #{sub_key} from facter to be positive for #{mount_key}")
+          end
+        end
+      end
+    end
+
+    step("verify that puppet facts returns positive numbers for the mount points byte fields") do
+
+      on(agent, puppet("facts --render-as json")) do |puppet_facts|
+        puppet_results = JSON.parse(puppet_facts.stdout)
+        puppet_results['values']['mountpoints'].each_key do |mount_key|
+          ['available_bytes', 'size_bytes', 'used_bytes'].each do |sub_key|
+            assert_operator(puppet_results['values']['mountpoints'][mount_key][sub_key], :>=, 0,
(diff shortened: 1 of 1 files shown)


Magisus:
Can you update this commit message to describe what the new test is testing? When looking back through the history (e.g. in a `git log`), that information will be helpful in remembering what bug it was designed to track, without having to view the diff. Also seeing "DO NOT MERGE" in a commit message is kind of off-putting :P

ericldelaney:
@Magisus Ok how about that for the commit message?

Magisus:
Great, thanks!

acceptance/tests/facts/validate_file_system_size_bytes.rb
```diff
@@ -0,0 +1,39 @@
...
+  tag 'risk:high'
+
+  require 'json'
+
+  agents.each do |agent|
+    step("verify that facter returns positive numbers for the mount points byte fields") do
+
+      on(agent, facter("--json")) do |facter_output|
+        facter_results = JSON.parse(facter_output.stdout)
+        facter_results['mountpoints'].each_key do |mount_key|
```
johnduarte: Suggestion to enumerate on sub-keys
```
['available_bytes',  'size_bytes', 'used_bytes'].each do |subkey|
   assert_operator(facter_results['mountpoints'][mount_key][subkey], :>=, 0,
                   "Expected the #{subkey} from facter to be positive for #{mount_key}")
end
```

johnduarte: FYI, you can use the `expect_failure` helper to allow you to merge the test prior to having the ticket fixed. Once the issue is addressed, it will go red and can be removed.

Example
```
['available_bytes',  'size_bytes', 'used_bytes'].each do |subkey|
  if agent['platform'] =~ /solaris/ then
    expect_failure('expected to fail due to FACT-1578') do
     assert_operator(facter_results['mountpoints'][mount_key][subkey], :>=, 0,
                     "Expected the #{subkey} from facter to be positive for #{mount_key}")
    end
  else
     assert_operator(facter_results['mountpoints'][mount_key][subkey], :>=, 0,
                     "Expected the #{subkey} from facter to be positive for #{mount_key}")
  end
end
```

ericldelaney: That's good to know, but in this case I can't do that as it will only fail if we happen to have a number that overflows the conversion correctly. Its doesn't fail for all the values.

ericldelaney: @johnduarte Updated

joshcooper:
Reverted in commit ce10993 since the underlying issue has not yet been fixed. See [https://tickets.puppetlabs.com/browse/FACT-1578](FACT-1578).

acceptance/tests/facts/validate_file_system_size_bytes.rb
er0ck: just surfed my way to this PR and was curious what the advantage of using `assert_operator` was, over just `assert(blah >= 0)`

ericldelaney: @er0ck assert_operator was what I found when I went looking

PR #1551, opened 2017-04-14
(FACT-1578) Use rb_ll2inum when passing numbers to Ruby

branan:
`rb_int2inum` takes a `long`, but `long` can still be 32 bits on
some platforms. Instead, we should use `rb_ll2inum`, which takes
a `long long`. This is still not guaranteed to be 64-bits by the C
standard, but it is on the platforms we care about, AFAICT.

This also revs leatherman to expose the new Ruby function.


ericldelaney:
LGTM

PR #1554, opened 2017-04-18
(FACT-1268) Changed the EC2_CONNECTION_TIMEOUT to 600

BramBauwens:
(FACT-1268) Changed the EC2_CONNECTION_TIMEOUT to 600

The default EC2_CONNECTION_TIMEOUT, 200 ms, results in a time-out on a regular basis. The solution for this would be to change this from 200 ms to 600 ms.


BramBauwens:
Made a mistake on this one. Duplicate of #1555.

PR #1555, opened 2017-04-18
(FACT-1268) Changed the EC2_CONNECTION_TIMEOUT to 600

BramBauwens:
The default EC2_CONNECTION_TIMEOUT of 200 is too short. The connection-timeout occurs regularly when spinning up ec2-servers. The fix for this is changing the EC2_CONNECTION_TIMEOUT to 600.


MikaelSmith:
Thanks for the PR! The previous failure was an intermittent failure with our Windows package host, I re-ran the test.

As long as the IP address is unreachable, I don't see a reason this should impact Facter behavior outside of EC2. That would be my primary concern on this PR. Otherwise, increasing the timeout seems reasonable.

farkasmate:
Thank you.

PR #1564, opened 2017-04-27
(FACT-1615) Move and update 2 tests to custom_facts and loading libfacter test

ericldelaney:
Moved 2 tests into the custom_facts directory
Update load_libfacter.rb
Updated tests for code style, TestRail ID, and risk tags


ericldelaney:
Tested on all the Vmpooler platforms

kurtwall:
Are you moving these tests in such a way that we preserve the change history?

ericldelaney:
I'm moving the files with "git mv"

acceptance/tests/custom_facts/using_win32ole_should_not_hang.rb
```diff
@@ -0,0 +1,29 @@
+test_name 'C92060: Custom facts should not hang Facter when using win32ole' do
```
johnduarte: This file did not properly replace newlines and adds whitespace errors. The following warning is evident when applying this patch:
```
warning: 29 lines add whitespace errors
```

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
```diff
@@ -0,0 +1,40 @@
...
+    fact_file = File.join(fact_dir, 'test_facts.rb')
+    create_remote_file(agent, fact_file, fact_content)
+    env = {'FACTERLIB' => fact_dir}
+
+    teardown do
+      on(agent, "rm -rf '#{fact_dir}'")
+    end
+
+    step "Agent: Verify test_fact1 from #{fact_file}" do
+      on(agent, facter('test_fact1', :environment => env)) do |facter_output|
```
johnduarte: nit pick: The return from the Beaker `on` call is not an output object. It is an instance of [Beaker Result](http://www.rubydoc.info/github/puppetlabs/beaker/Beaker/Result) of which output streams are accessible from. So, the variable name `facter_output` is misleading. I am not advocating for changing it in this PR, just pointing it out.

acceptance/tests/custom_facts/using_win32ole_should_not_hang.rb
```diff
@@ -0,0 +1,29 @@
...
+  setcode do
+    require 'win32ole'
+    locator = WIN32OLE.new('WbemScripting.SWbemLocator')
+    locator.ConnectServer('', "root/CIMV2", '', '', nil, nil, nil, nil).to_s
+  end
+end
+EOM
+
+  agents.each do |agent|
+    custom_dir = agent.tmpdir('arbitrary_dir')
```
johnduarte: Recommend a `teardown` to remove this directory as has been done in the other tests.

acceptance/tests/load_libfacter.rb
```diff
@@ -1,31 +1,32 @@
-test_name 'Ruby can load libfacter'
+test_name 'C100161: Ruby can load libfacter' do
+  tag 'risk:high'
 
-agents.each do |agent|
-  on agent, "env PATH=\"#{agent['privatebindir']}:${PATH}\" which ruby"
-  ruby_path = stdout.chomp
+  agents.each do |agent|
+    ruby_path = on(agent, "env PATH=\"#{agent['privatebindir']}:${PATH}\" which ruby").stdout.chomp
```
johnduarte: There is a helper method available for this. It should be used rather than duplicating the code.
https://github.com/puppetlabs/facter/blob/master/acceptance/lib/puppet/acceptance/common_utils.rb#L4

acceptance/tests/load_libfacter.rb
```diff
@@ -1,31 +1,32 @@
...
 
-  root_dir = ruby_path
-  while File.basename(root_dir).downcase != 'puppet'
-    new_root_dir = File.dirname(root_dir)
-    if new_root_dir == root_dir
-      break
-    else
-      root_dir = new_root_dir
+    root_dir = ruby_path
+    while File.basename(root_dir).downcase != 'puppet'
```
johnduarte: The path variable logic here looks brittle to me. If we need to be able to introspect a consistent ruby load path based on installation, then this should probably be added to Beaker.

cc @kevpl , @tvpartytonight

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
```diff
@@ -0,0 +1,40 @@
+test_name 'C14893: Facter should handle multiple facts in a single file' do
+  tag 'risk:high'
+
+  fact_content = <<-EOM
+Facter.add(:test_fact1) do
+  setcode do
+    "test fact 1"
+  end
+end
```
kurtwall: \<rant\>No action required: I realize this is pre-existing code.This style/indentation for here-docs makes my eyes bleed. \</rant\>

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
kurtwall: Is it misleading enough to warrant a TODO so we get back to it one day maybe?

acceptance/tests/load_libfacter.rb
kurtwall: Example of use (I just learned about this yesterday myself): https://github.com/puppetlabs/puppet/pull/5827/files#diff-c2d6fd4eb072d50a0c31158ba3ed4f67

acceptance/tests/load_libfacter.rb
kurtwall: I think this is here to find the puppet-agent ruby before the system ruby. Might not be the best way to do it.

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
ericldelaney: Do you have a suggestion for a different name?  facter_beaker_result?

johnduarte: The most common variable names for the Beaker result object are `result` and `res`. If you wish it to be more specific for this context, my recommendation would be `facter_result`.

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
ericldelaney: ok

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
ericldelaney: ok changed

acceptance/tests/custom_facts/using_win32ole_should_not_hang.rb
ericldelaney: Hmmm not sure why

acceptance/tests/custom_facts/using_win32ole_should_not_hang.rb
ericldelaney: added

acceptance/tests/load_libfacter.rb
ericldelaney: I looked at using ruby_command, but based on what the test appears to be doing it doesn't really work. I've added comments explaining what its doing and why

acceptance/tests/load_libfacter.rb
ericldelaney: I looked at what is available in agent object, and I don't see a better option.

ericldelaney:
@johnduarte @kurtwall  Updated based on your comments

acceptance/tests/custom_facts/having_multiple_facts_in_one_file.rb
```diff
@@ -0,0 +1,40 @@
+test_name 'C14893: Facter should handle multiple facts in a single file' do
+  tag 'risk:high'
+
+  fact_content = <<-EOM
+    Facter.add(:test_fact1) do
+      setcode do
+        "test fact 1"
+      end
+    end
+
```
johnduarte: trailing whitespace on this line

acceptance/tests/custom_facts/using_win32ole_should_not_hang.rb
johnduarte: See the following github article for correcting the line ending issue.
https://help.github.com/articles/dealing-with-line-endings/

ericldelaney: Ah, so the original file that I moved and updated contains carriage returns which is causing you issues. I'll remove them so all the tests are consistent.

acceptance/tests/load_libfacter.rb
```diff
@@ -1,31 +1,44 @@
-test_name 'Ruby can load libfacter'
+test_name 'C100161: Ruby can load libfacter' do
```
johnduarte: After reviewing this again, I do not understand the core intention of this test. It seems concerned with the low level plumbing of how the puppet-agent ruby file are laid out rather than validating functionality. Is there a functional side effect of not being able to load libfacter via ruby? If there is can we test that instead?

johnduarte: ...maybe the ability to use facter from within the puppetlabs vendored ruby?
```
/opt/puppetlabs/puppet/bin/ruby -e "require 'facter'"
```

ericldelaney: So if facter.rb fails to load the libfacter.jar (if its jRuby) or fails to load the libfacter.so, then in both cases we raise an error which you would see in the output from ruby.

    begin
      require 'facter.jar'
    rescue LoadError
      raise LoadError.new('libfacter was not built with JRuby support.')
    end

      require "#{facter_dir}/lib/libfacter.so"
    rescue LoadError
      raise LoadError.new('libfacter was not found. Please make sure it was installed to the expected location.')
    end

johnduarte: So, wouldn't that be validated by testing that `facter` is in the $LOAD_PATH for the puppetlabs vendored ruby? If so, then most of the code in this test is not necessary.
```
step 'Ensure that ruby can load facter from the $LOAD_PATH' do
  on(agent, ruby_command(agent), '-e', '"require \'facter\'"')
end
```
What am I missing?

ericldelaney: Hmmm... Yeah, I see what your saying. Let me make some changes

ericldelaney: Using 'require' appears to work for the Unix's, but not for the Windows boxes as the puppet ruby doesn't have facter.rb in its $LOAD_PATH.

ericldelaney: @johnduarte updated

johnduarte: Not having `facter` in the $LOAD_PATH on Windows sounds like a bug, not something we want to skirt around in the test.

johnduarte: When testing the `require 'facter'` on Windows it works, but due to a different `facter` lib file being in the $LOAD_PATH.
```
Administrator@cl29d7nvck9bawi ~
$ env PATH="/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/bin:${PATH}" ruby -e 'puts $LOAD_PATH'
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/site_ruby/2.1.0
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/site_ruby/2.1.0/x64-msvcrt
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/site_ruby
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby/2.1.0
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby/2.1.0/x64-msvcrt
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.1.0
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.1.0/x64-mingw32

Administrator@cl29d7nvck9bawi ~
$ find "/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby" -name facter.rb -print
/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/vendor_ruby/puppet/indirector/facts/facter.rb

Administrator@cl29d7nvck9bawi ~
$ env PATH="/cygdrive/c/Program Files (x86)/Puppet Labs/Puppet/sys/ruby/bin:/cygdrive/c/Program Files/Puppet Labs/Puppet/sys/ruby/bin:${PATH}" ruby -e 'require "facter"'
C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- facter (LoadError)
        from C:/Program Files/Puppet Labs/Puppet/sys/ruby/lib/ruby/2.1.0/rubygems/core_ext/kernel_require.rb:54:in `require'
        from -e:1:in `<main>'

```

PR #1565, opened 2017-05-02
(FACT-1624) Use `is_integer` instead of `is_fixednum` and `is_bignum`

Magisus:
Ruby 2.4 unified Fixnum and Bignum into Integer. We updated the
Leatherman API to match this. This commit updates Facter to use the new
API.


Magisus:
There is a test that was designed to test Bignums: https://github.com/puppetlabs/facter/blob/master/lib/tests/ruby/ruby.cc#L628-L633. It still passes and it still seems useful to me, even though there is no type distinction anymore.

Magisus:
This needs the Leatherman change to be tagged in order to pass these commit checks... meanwhile, the Leatherman changes will not promote into puppet-agent until these updates land (see failing job [here](https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent%20components/view/leatherman/job/platform_leatherman_pkg-van-ship_master/6/BUILD_TARGET=el-6-x86_64,SLAVE_LABEL=beaker/console) ). @MikaelSmith thoughts on how we should proceed here?

I have done manual builds in several Ruby versions and verified that these changes build and pass tests, at least on my Mac.

MikaelSmith:
Oh right, the joy of breaking changes. And using tagged versions for CI here means that all breaking changes in Leatherman have to be done before we can use any of them.

I have two ideas that I'm not wedded to:
- use release-candidate tags in Leatherman, and then do a PR to puppet-agent to bring in Facter/Leatherman changes simultaneously
- make the changes in Leatherman only additive for now and do a 0.12.1 release with is_integer that deprecates the others; we'd probably end with a pre-1.0.0 "remove everything deprecated" PR right before tagging

Magisus:
I think I like the first idea better. That takes care of the problem now instead of just shoving it off to Puppet 5 release time, where there are going to be a lot of other things going around, and this might get lost in the shuffle.

PR #1566, opened 2017-05-04
Merge stable to master

joshcooper:
* upstream/stable:
  (FACT-1616) update the OS specific facts
  (packaging) Rev version to 3.6.4 for release
  (FACT-1614) Move tests into facts directory and update
  (FACT-1613) Update and split various regression tests
  (FACT-1614) move the files to there new home in the facts directory
  (FACT-1607) split the default external fact directory into 2 test
  (FACT-1611) update external fact stderr test

Preserved facter 4.0.0 version and locale strings in master. Acceptance
tests only had indentation conflicts.


joshcooper:
I noticed the facter version in master is 4.0.0. Not sure if we need to bump that down to 3.6.z since we're not shipping 4 as part of puppet 5?

ericldelaney:
👍

Magisus:
@joshcooper yes I think we should update it. Go for 3.6.5 for now I guess?

PR #1567, opened 2017-05-05
(FACT-1627) Use oslevel for AIX kernel release

branan:
In the initial port of cfacter to AIX, we had tried to avoid using
oslevel, since it is much slower than we'd like for
Facter. Unfortunately, our implementation doesn't handle all the edge
cases that oslevel does, and fixing it is incredibly complex.

Instead, we should just call out to oslevel, and deal with the
slowdown.


ericldelaney:
👍

Magisus:
Do we have any tests around this that could be updated to verify some of the edge cases?

PR #1568, opened 2017-05-18
(FACT-1616) merge & split identity, os, processor, and kernel fact tests

ericldelaney:
Merge the identity fact tests into a single test file for all OS'
Merge the os, processor, and kernel fact tests into a single test for all OS'
The test will now fail if it doesn't support an operating system or version with an appropriate failure error.
This also updates our supported platforms:
+win7, +win8.1, +fedora23,24,25, +huawei, +cisco, +arista, +cumulus,
-ubuntu10.04,12.04, -centos4, -win2003, -macosx11.9


ericldelaney:
tested on everything, including the hardware platforms (here is a subset) arista5-32a, centos7-64a, ciscoxr-64a%7Bhypervisor=vmpooler%7D, cisconx-64a%7Bhypervisor=vmpooler%7D, cumulus25-64a, debian8-32a, fedora25-32a, oracle7-64a, osx1012-64a, redhat7-64a, scientific7-64a, sles12-64a, solaris10u8-64a, ubuntu1610-32a, windows10ent-32a, windows2016-6432a

acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -0,0 +1,613 @@
...
+        elsif agent['platform'] =~ /sles-/
+          sles_expected_facts(agent)
+        elsif agent['platform'] =~ /solaris-/
+          solaris_expected_facts(agent)
+        elsif agent['platform'] =~ /ubuntu-/
+          ubuntu_expected_facts(agent)
+        elsif agent['platform'] =~ /windows-/
+          windows_expected_facts(agent)
+        else
+          fail_test("unknown agent type being tested #{agent['platform']}")
```
johnduarte: It seems weird to me to use the `fail_test` method for the Ruby library methods. I would recommend raising an exception with `raise` and allowing Beaker to capture it as an error.

acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -0,0 +1,613 @@
...
+        end
+      end
+
+      # AIX
+      def aix_expected_facts(agent)
+        case agent['platform']
+          when /aix-5.3/
```
johnduarte: I am not a fan of hard-coding all of the OS versions. At the expense of getting the correct kernel version, this creates an ongoing maintenance burden. Is it worth validating the kernel version for each OS version? I would argue that a regex to ensure that the kernel version matches a reasonable pattern guards us against bad output without introducing the burden of managing all 'supported' operating systems at this level of detail.

acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -0,0 +1,613 @@
...
+        case agent['platform']
+          when /(el|centos)-5/
+            os_version     = '5'
+            kernel_version = '2.6'
+          when /(el|centos)-6/
+            os_version     = '6'
+            kernel_version = '2.6'
+          when /(el|centos)-7/
+            os_version = '7'
+            agent['template'] =~ /oracle/ ? kernel_version = '3.8' : kernel_version = '3.1'
```
johnduarte: This is an example of my earlier point. Managing kernel versions for derivatives is going make these tests brittle and error prone. As an example, the validation here for RHEL 7 is incorrect, the kernel version that currently ships with RHEL 7 is '3.10', not '3.1'

acceptance/tests/facts/identity.rb
```diff
@@ -0,0 +1,53 @@
...
+      on(agent, facter("--json")) do |facter_result|
+        results = JSON.parse(facter_result.stdout)
+        expected_identity.each do |fact, value|
+          assert_match(value, results['identity'][fact].to_s, "Incorrect fact value for identity.#{fact}")
+        end
+      end
+    end
+  end
+end
+
```
johnduarte: This empty line should be removed

acceptance/tests/facts/os_processors_and_kernel.rb
```diff
@@ -0,0 +1,21 @@
...
+        results = JSON.parse(facter_result.stdout)
+        expected_facts.each do |fact, value|
+          actual_fact = json_result_fact_by_key_path(results, fact)
+          assert_match(value, actual_fact.to_s, "Incorrect fact value for #{fact}")
+        end
+      end
+    end
+  end
+end
+
```
johnduarte: This empty line should be removed.

acceptance/lib/facter/acceptance/base_fact_utils.rb
ericldelaney: Ok

acceptance/lib/facter/acceptance/base_fact_utils.rb
ericldelaney: If you don't want to have to update per release then we should collect the real value and test  it.

ericldelaney:
@johnduarte it now figures out values based on how facter figures them out.

johnduarte:
@ericldelaney I fail to see the value in reimplementing facter in the test helper adds. If you are testing that facter is calling the proper underlying system commands, then it seems like this could be implemented at the spec layer mocking out the specified commands in the spec tests.

ericldelaney:
John,

Isn't that what you ask for when you complained that the tests would have
to be updated with each new release?

Eric

johnduarte:
@ericldelaney my previous comments were only concerned with the kernel fact.

PR #1569, opened 2017-05-22
(FACT-1633) Relax version constraint of CFPropertyList gem

Magisus:
We want to pull in the newest version of CFPropertyList gem, 2.3.5,
which supports Ruby 2.4. This commit relaxes the version constraint to
allow that to be pulled in, while also preventing us from needed to do
further releases of Facter 2 if it updates again.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -16,7 +16,7 @@ gem_default_executables: 'facter'
 gem_platform_dependencies:
   universal-darwin:
     gem_runtime_dependencies:
-      CFPropertyList: '~> 2.3.3'
+      CFPropertyList: '~> 2.2'
   x86-mingw32:
     gem_runtime_dependencies:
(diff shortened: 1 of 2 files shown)


Magisus:
Where is the rake task changed in this PR used? What will break if we pull in a different CFPropertyList than 2.3.5?

ext/project_data.yaml
```diff
@@ -16,7 +16,7 @@ gem_default_executables: 'facter'
 gem_platform_dependencies:
   universal-darwin:
     gem_runtime_dependencies:
-      CFPropertyList: '~> 2.3.3'
+      CFPropertyList: '~> 2.2'
```
Iristyle: Facter 2.0.1 shipped against CFPropertyList 2.2.6 ... so this will cover from that oldest version through to newest 2.3.5.

👍

PR #1570, opened 2017-05-23
PUP-6164: Tech-debt: links.puppetlabs.com -> links.puppet.com

angrox:


johnduarte:
Thanks for fixing these links @angrox . The commit validator expects commits to facter to begin with one of the following strings:
```
(FACT-<digits>) # this is most common and should be a ticket at tickets.puppetlabs.com
(doc)
(docs)
(maint)
(packaging)
```

You could open a facter ticket that is analogous to the PUP-6164 ticket you are referencing. However, given the trivial nature of this fix, I would suggest using the `(maint)` prefix. This would allow the Travis check to run. Once that is green, I am happy to merge.

Thanks again.

PR #1571, opened 2017-05-23
(FACT-1637) Remove win32-security gem

Iristyle:
- 2b9a6ff41ce77f2d60606015b2f4410d01955fa6 moved relevant security
   code into Facter, and removed any usage of win32/security gem code

 - Note that the usage was removed on 2/24/2016 but was never released
   as Facter 2.4.7 never saw the light of day.

   The next Puppet version 5 will remove all of these win32-*** gems
   which are considered internal implementation details (but which
   may could plausibly be externally consumed).  Facter will be
   released as 2.5.0 with some of these removals.

   The only impacted workflows for Facter should be strictly gem
   based - i.e. local development.  Packages will generally not be
   impacted as their dependencies are expressed differently (and
   Puppet itself carries a superset of what Facter carries).


Magisus:
Looks good, nice to be ripping out unused things :)

PR #1572, opened 2017-05-23
(FACT-1638) Remove win32-dir from Facter

Iristyle:
- Facter was only using Dir::COMMON_APPDATA so replace that with FFI
   calls to the SHGetFolderPathW API which can return this value.

   Note that SHGetFolderPathW has been replaced with a newer API, but
   the newer API is much more complex and requires COM memory
   management, so instead rely on this simpler variant given it should
   not be going away anytime soon.

 - There may be custom fact consumers in the wild relying on the
   existence of various Dir::XXXX constants provided by the win32-dir
   gem. This should not be a problem in practice for a few reasons:

   * Puppet agent packages still include win32-dir and adorn Dir::
     until work is completed to remove the gem - this will likely be
     done in a backward compatible way to avoice breakage
   * Puppet gem still depends on win32-dir, so even without Facter
     explicitly requiring it, it's still present in Rubys global state
   * Only Facter only gem workflows should be impacted, and I don't
     believe there are any

### DIFF
--- a/lib/facter/util/windows/api_types.rb
+++ b/lib/facter/util/windows/api_types.rb
@@ -37,6 +37,27 @@ module Facter::Util::Windows::ApiTypes
       str.encode(dst_encoding)
     end
 
+    # @param max_char_length [Integer] Maximum number of wide chars to return (typically excluding NULLs), *not* bytes
+    # @param null_terminator [Symbol] Number of number of null wchar characters, *not* bytes, that determine the end of the string
+    #   null_terminator = :single_null, then the terminating sequence is two bytes of zero.   This is UNIT16 = 0
+    #   null_terminator = :double_null, then the terminating sequence is four bytes of zero.  This is UNIT32 = 0
+    def read_arbitrary_wide_string_up_to(max_char_length = 512, null_terminator = :single_null)
+      if null_terminator != :single_null && null_terminator != :double_null
+        raise _("Unable to read wide strings with %{null_terminator} terminal nulls") % { null_terminator: null_terminator }
+      end
+
+      terminator_width = null_terminator == :single_null ? 1 : 2
+      reader_method = null_terminator == :single_null ? :get_uint16 : :get_uint32
+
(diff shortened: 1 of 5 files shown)


ericldelaney:
LGTM

Iristyle:
Ran tests locally on:

* OSX - 1734 examples / 0 fails / 4 pending
* Windows 2008R2 - 1665 examples / 0 fails / 4 pending

On Windows the `windows_data_dir` code path is loaded as soon as Facter is require'd, so it's easy to tell if something is awry.

I also validated in pry that the value returned is correct:

```
[1] pry(main)> require 'facter'
=> true
[2] pry(main)> Facter::Util::Windows::Dir.get_common_appdata
=> "C:\\ProgramData"
```

lib/facter/util/windows/api_types.rb
```diff
@@ -37,6 +37,27 @@ def read_wide_string(char_length, dst_encoding = Encoding::UTF_8)
       str.encode(dst_encoding)
     end
 
+    # @param max_char_length [Integer] Maximum number of wide chars to return (typically excluding NULLs), *not* bytes
+    # @param null_terminator [Symbol] Number of number of null wchar characters, *not* bytes, that determine the end of the string
+    #   null_terminator = :single_null, then the terminating sequence is two bytes of zero.   This is UNIT16 = 0
+    #   null_terminator = :double_null, then the terminating sequence is four bytes of zero.  This is UNIT32 = 0
+    def read_arbitrary_wide_string_up_to(max_char_length = 512, null_terminator = :single_null)
```
Iristyle: Code that lives in Puppet as well...

Magisus: Does this still exist in Facter 3? If not, I don't see much point in trying to consolidate this...

Iristyle: Yeah, I'm not worried about the little bit of duplication here inside the Facter 2 gem - this is an internal Win32 API helper built around FFI.

Magisus:
👍

PR #1573, opened 2017-05-24
(FACT-1418) Nano fix - deprecate WMI COM monikers

Iristyle:
- To support Nano server, COM monikers cannot be used when
   establishing the connection to WMI.

   Create a new private method that behaves, by default, identically
   to the previous connect() behavior called connect2()

   Change the connect() signature slightly to provide no default
   value to uri so that the behavior can be transparently swapped by
   default to use actual COM classes instead of monikers.

   To cleanly support the prior behavior in a backwards-compatible way
   allow for passing a non-nil uri directly as a moniker - there
   should not be any such cases, but there's a possibility that
   custom facts may be using this interface.

   Note that such calls with not work on Nano.

 - This change only impacts running Puppet spec tests in a gem based
   workflow until native Facter 3+ is bundled into a gem.  This has
   the possibility to impact AppVeyor or other module workflows that
   consume gems.  This will not impact currently shipping puppet-agent
   packages.

### DIFF
--- a/lib/facter/util/wmi.rb
+++ b/lib/facter/util/wmi.rb
@@ -1,10 +1,28 @@
 module Facter::Util::WMI
   class << self
-    def connect(uri = wmi_resource_uri)
+
+    # Impersonation Level Constants
+    # https://msdn.microsoft.com/en-us/library/windows/desktop/ms693790%28v=vs.85%29.aspx
+    RPC_C_IMP_LEVEL_DEFAULT       = 0
+    RPC_C_IMP_LEVEL_ANONYMOUS     = 1
+    RPC_C_IMP_LEVEL_IDENTIFY      = 2
+    RPC_C_IMP_LEVEL_IMPERSONATE   = 3
+    RPC_C_IMP_LEVEL_DELEGATE      = 4
+
+    # returns a COM class implementing ISWbemServicesEx
+    # prior to Facter 2.5.0, this defaulted to using a moniker
+    # but now defaults to using COM classes directly to support Nano
+    # backward compatibility is maintained in case custom facts specified a moniker
+    # but note that passing in the uri parameter can never work on Nano
+    def connect(uri = nil)
       require 'win32ole'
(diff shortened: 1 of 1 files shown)


MosesMendoza:
Should this PR have tests?

Iristyle:
@MosesMendoza no new for new tests, there are already existing tests that would have failed were anything wrong.

lib/facter/util/wmi.rb
```diff
@@ -1,16 +1,49 @@
...
+
+    # this mimics the previous behavior of using the COM moniker
+    # winmgmts:{impersonationLevel=impersonate}!//./root/cimv2
+    # which is not supported on Nano Server
+    def connect2( server = '.', namespace = 'root\\cimv2', impersonation_level = RPC_C_IMP_LEVEL_IMPERSONATE )
+      locator = WIN32OLE.new("WbemScripting.SWbemLocator")
+      # https://msdn.microsoft.com/en-us/library/aa393720%28v=vs.85%29.aspx
+      # ConnectServer returns an ISWbemServicesEx
+      conn = locator.ConnectServer(server, namespace)
+      conn.Security_.ImpersonationLevel = impersonation_level
```
MosesMendoza: I was tripped up at first because the MSDN docs for the `Security_` function state that its read-only[1] and it appeared initially like we were trying to mutate it, but I see we're actually operating on the SWbemSecurity object[2] returned by `Security_`. So you can't modify _which_ security object you're pointing to, but you can modify the security object _that_ you're pointing to. Indeed.

[1] https://msdn.microsoft.com/en-us/library/aa393874(v=vs.85).aspx
[2] https://msdn.microsoft.com/en-us/library/aa393850(v=vs.85).aspx

Iristyle: Yup, already ran it through (the impersonation level change code was suggested through an MSDN article I read)

Magisus:
👍

PR #1574, opened 2017-05-25
(FACT-1610) Parse 64-bit AIX Volume Group IDs

branan:
Previously, Facter assumed the VGs would have a 128-bit ID. This is
only true for volumes created this century. Older versions of AIX
created groups with 64-bit IDs.

This change updates AIX volume ID parsing to support these shorter
IDs.


Magisus:
Looks right to me. I don't suppose there's a meaningful way to add tests for this?

branan:
@Magisus nah. We really should have structured facter better so that logic like this is easily unit-tested. Unfortunately, it's not :(

Magisus:
Yeah that's what I saw too, looking at the tests. Oh well, LGTM then 👍

ScottGarman:
:+1:

er0ck:
i love the 90s

PR #1575, opened 2017-05-25
(FACT-1641) Remove win32console gem

Iristyle:
- win32console gem was only needed for Ruby 1.9.3 running on 32-bit
   Windows, which is no longer a supported platform combination.

 - Remove the reference from ext/project_data.yaml, which changes what
   a shipped Facter 2.5.0 gem will reference from the x86 Windows
   gem.

 - Update the Gemfile so that source based workflows will no longer
   include win32console if testing Ruby 1.9.3 on 32-bit windows
   locally from source.

 - Remove the win32console related hacks that modify streams when
   running specs introduced 4 years ago in
   6ebe9d1d244749b5875be21f7e4705065aeef612

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,10 +33,6 @@ data['gem_platform_dependencies'].each_pair do |gem_platform, info|
   end
 end
 
-platform(:mingw_19) do
-  gem 'win32console', '~> 1.3.2', :require => false
-end
-
 mingw = [:mingw]
 mingw << :x64_mingw if Bundler::Dsl::VALID_PLATFORMS.include?(:x64_mingw)
 
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -20,7 +20,6 @@ gem_platform_dependencies:
   x86-mingw32:
     gem_runtime_dependencies:
       ffi: '~> 1.9.5'
-      win32console: '~> 1.3.2'
   x64-mingw32:
     gem_runtime_dependencies:
       ffi: '~> 1.9.5'
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -38,13 +38,6 @@ LogSpecOrder.log_spec_order
 RSpec.configure do |config|
   config.mock_with :mocha
 
-  if Facter::Util::Config.is_windows? && RUBY_VERSION =~ /^1\./
-    require 'win32console'
-    config.output_stream = $stdout
-    config.error_stream = $stderr
-    config.formatters.each { |f| f.instance_variable_set(:@output, $stdout) }
-  end
-
   config.before :each do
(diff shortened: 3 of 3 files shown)


joshcooper:
> Ruby 1.9.3 running on 32-bit Windows, which is no longer a supported platform combination.

I thought we still supported 32-bit Windows, e.g. Windows 8.

glennsarti:
> > Ruby 1.9.3 running on 32-bit Windows, which is no longer a supported platform combination.
>
> I thought we still supported 32-bit Windows, e.g. Windows 8, just not 32-bit ruby on 64-bit Windows (in Puppet 5).

We do support 32bit, just not Puppet 3 (Ruby 1.9.3) on 32bit

Magisus:
👍 

But yeah to that point, is it specifically 1.9.3 we don't support on those platforms? Because I know we still test against those platforms generally.

joshcooper:
> We do support 32bit, just not Puppet 3 (Ruby 1.9.3) on 32bit

I'm still confused. If you're testing puppet 4.x as a gem on ruby 1.9.3 and 32-bit Windows, then won't this cause incorrect console unicode output, e.g. `puppet module list`? I imagine the set of affected users would be very small, but still seems like a breaking change.

Iristyle:
I believe that would be a problem, yes, @joshcoope.  But I'm not aware of any place that combo is in use.

Do you know of situations where the Puppet 4 gem is run on top on Ruby 1.9.3 / Windows 32-bit?  Our module pipelines run atop the supported Ruby 2.1.9 that we use in packages, for instance.

joshcooper:
> Do you know of situations where the Puppet 4 gem is run on top on Ruby 1.9.3 / Windows 32-bit?

I don't.. but it seems like the kind of thing that could be done to make sure a module pipeline (outside of Puppet) doesn't break puppet 3.x agents prior to upgrading to 4. Given that we don't know who might be affected, I'm 👎  for removing the gem in a .z release.

Magisus:
Same for a Y release @joshcooper ? We are planning a Facter 2.5 to be release in support of the introduction of Ruby 2.4 in Puppet 5, and this would get released as part of that.

joshcooper:
ffi 1.9.18 requires ruby 2, so I'm ok merging this.

Iristyle:
Thanks for the merge - although it's not great to remove the `win32console` dep from 2.4.6 to 2.5.0, given we don't have the latitude to release anything in the 3 series, our hands are a bit tied until we can compile Facter as a gem.  We probably could have *not* made this change and still been ok, but since no one ever reported the FFI issue as you mentioned, we're probably in the clear pragmatically, despite being in the wrong, technically... :man_shrugging:

PR #1579, opened 2017-05-30
(maint) Use older FFI version under Ruby 1.9.3

Magisus:
FFI dropped support for Ruby 1.9.3 in version 1.9.16. However, 1.9.18 is
needed to support Ruby 2.4. Since we need to support both of these
versions, this commit adds a check in the Gemfile for the Ruby version
when installing FFI, choosing an old version for 1.9.3 and the latest
for anything 2.0.0 and above.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,13 @@ mingw = [:mingw]
 mingw << :x64_mingw if Bundler::Dsl::VALID_PLATFORMS.include?(:x64_mingw)
 
 platform(*mingw) do
-  gem 'ffi', '~> 1.9.5', :require => false
+  # FFI dropped 1.9.3 support in 1.9.16, and 1.9.15 was an incomplete release.
+  # 1.9.18 is required to support Ruby 2.4
+  if RUBY_VERSION < '2.0.0'
+    gem 'ffi', '<= 1.9.14', :require => false
+  else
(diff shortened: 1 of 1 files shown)


Magisus:
Note this does not update the version specification in `project_data.yaml`, ensuring that gem-based workflows will still pull in the most recent FFI. This is important for Ruby 2.4 support.

Gemfile
```diff
@@ -37,7 +37,13 @@ mingw = [:mingw]
 mingw << :x64_mingw if Bundler::Dsl::VALID_PLATFORMS.include?(:x64_mingw)
 
 platform(*mingw) do
-  gem 'ffi', '~> 1.9.5', :require => false
+  # FFI dropped 1.9.3 support in 1.9.16, and 1.9.15 was an incomplete release.
+  # 1.9.18 is required to support Ruby 2.4
+  if RUBY_VERSION < '2.0.0'
+    gem 'ffi', '<= 1.9.14', :require => false
+  else
+    gem 'ffi', '~> 1.9.18', :require => false
```
Iristyle: `1.9.18` is ok on 2.1 as well?

Magisus: It should be. I'm pretty sure the gem requirements said > 2.0.0.

PR #1581, opened 2017-06-02
(FACT-1588) Fix for OpenSSL 1.1.0

MikaelSmith:
Building with OpenSSL 1.1.0 failed due to a change in const-ness of the
`BIO_f_base64` return value. Update functions to accept a const value
when appropriate. Should be backwards compatible with non-const version.

### DIFF
--- a/lib/src/util/posix/scoped_bio.cc
+++ b/lib/src/util/posix/scoped_bio.cc
@@ -6,8 +6,11 @@ using namespace leatherman::util;
 
 namespace facter { namespace util { namespace posix {
 
-    scoped_bio::scoped_bio(BIO_METHOD* method) :
-        scoped_resource(BIO_new(method), free)
+    // Remove const-ness before calling BIO_new. This is "unsafe",
+    // but in isolation here will not cause issues. Allows the code to work
+    // with both OpenSSL 1.0 and 1.1.
+    scoped_bio::scoped_bio(const BIO_METHOD* method) :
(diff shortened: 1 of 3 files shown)


lib/src/util/posix/scoped_bio.cc
```diff
@@ -6,8 +6,11 @@ using namespace leatherman::util;
 
 namespace facter { namespace util { namespace posix {
 
-    scoped_bio::scoped_bio(BIO_METHOD* method) :
-        scoped_resource(BIO_new(method), free)
+    // Remove const-ness before calling BIO_new. This is "unsafe",
+    // but in isolation here will not cause issues. Allows the code to work
+    // with both OpenSSL 1.0 and 1.1.
+    scoped_bio::scoped_bio(const BIO_METHOD* method) :
+        scoped_resource(BIO_new(const_cast<BIO_METHOD*>(method)), free)
```
MikaelSmith: This is messy, but I feel more maintainable than ifdefing based on OpenSSL version (because it doesn't branch code).

Magisus:
Mostly unrelated side note, I reeeally want to get updating the POT file out of these PRs and into CI like it is for Ruby projects.

MikaelSmith:
I can drop the POT portion, it may not be relevant anymore.

Magisus:
No you should keep it for now. We're not updating this any other way right now. I just want to get it automated and in line so we have one story around how and when we do it. Also want to stop relying on whatever tool a dev has installed (because different gettext versions sometimes spit out different metadata).

PR #1582, opened 2017-06-05
(FACT-1646) Add fact specfics tests and TestRail ID's

ericldelaney:
Added fact specific tests for facterversion and productname
Added structured fact specific tests for dmi and partitions
Added a test for not getting errors when facter is run as a non-root user
Added TestRail ID's and risk tags

### DIFF
--- a/acceptance/lib/facter/acceptance/user_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/user_fact_utils.rb
@@ -117,6 +117,16 @@ EOM
   end
         EOM
       end
+
+      # Return the correct shell path for Unix system under test
+      #
+      def user_shell(agent)
+        if agent['platform'] =~ /aix/
+          '/usr/bin/bash'
+        else
+          '/bin/bash'
+        end
+      end
     end
   end
 end
--- a/acceptance/tests/facts/non_root_users_without_errors.rb
+++ b/acceptance/tests/facts/non_root_users_without_errors.rb
@@ -1,0 +1,32 @@
+test_name "C59196: running facter as a non-root user should not produce permission errors" do
+  tag 'risk:high'
+
+  confine :except, :platform => 'windows' # this test currently only supported on unix systems FACT-1647
+  confine :except, :platform => 'aix' # system su(1) command prints errors cannot access parent directories and ticket FACT-1586
+  confine :except, :platform => 'cisco' # system su(1) command prints errors cannot access parent directories
+  confine :except, :platform => 'osx' # system su(1) command prints errors cannot access parent directories
+  confine :except, :platform => 'solaris' # system su(1) command prints errors cannot access parent directories
+  confine :except, :platform => 'eos-' # does not support user creation ARISTA-37
+
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  agents.each do |agent|
+    non_root_user = "nonroot"
+    facter_path = on(agent, "which facter").stdout.chomp
+
+    step "Agent #{agent}: create a #{non_root_user} user to run facter with" do
+      on(agent, "puppet resource user #{non_root_user} ensure=present shell='#{user_shell(agent)}'")
+    end
+
+    teardown do
+      on(agent, puppet("resource user #{non_root_user} ensure=absent"))
+    end
+
+    step "Agent #{agent}: run facter as #{non_root_user} and get no errors" do
+      on(agent, %Q[su #{non_root_user} -c "'#{facter_path}'"]) do |facter_results|
+        assert_empty(facter_results.stderr.chomp, "Expected no errors from facter when run as user #{non_root_user}")
+      end
+    end
+  end
+end
--- a/acceptance/tests/facts/partitions.rb
+++ b/acceptance/tests/facts/partitions.rb
@@ -1,0 +1,37 @@
+test_name "C96148: verify partitions facts" do
+  tag 'risk:high'
+
+  confine :except, :platform => 'osx' # no partitions on osx
+  confine :except, :platform => 'windows' # no partitions on windows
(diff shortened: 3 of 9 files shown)


ericldelaney:
Tested on arista4-32a, centos7-64a, ciscoxr-64a%7Bhypervisor=vmpooler%7D, cisconx-64a%7Bhypervisor=vmpooler%7D, cumulus25-64a, debian8-32a, fedora25-32a, oracle7-64a, osx1012-64a, redhat7-64a, scientific7-64a, sles12-64a, solaris11-64a, ubuntu1610-32a, windows10ent-32a, windows2016-6432a + Sparc, AIX, Huawei

acceptance/tests/facts/non_root_users_without_errors.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  agents.each do |agent|
+    non_root_user = "nonroot"
+    facter_path = on(agent, "which facter").stdout.chomp
+
+    step "Agent #{agent}: create a #{non_root_user} user to run facter with" do
+      on(agent, "puppet resource user #{non_root_user} ensure=present managehome=true shell='#{user_shell(agent)}'")
```
johnduarte: Why do you have to set `managehome=true` for the user resource. I would expect issuing the `facter` command as an existing user to work regardless of whether `$HOME` existed. And not having this restriction would allow you to include solaris and osx.

Example on RHEL7
```
[root@wlg20bt1x5bngc5 ~]# puppet resource user  foo ensure=present
Notice: /User[foo]/ensure: created
user { 'foo':
  ensure => 'present',
}
[root@wlg20bt1x5bngc5 ~]# ls /home/
[root@wlg20bt1x5bngc5 ~]#
[root@wlg20bt1x5bngc5 ~]# su foo -c /opt/puppetlabs/bin/facter
aio_agent_version => 1.10.1.46
...
virtual => vmware
[root@wlg20bt1x5bngc5 ~]# echo $?
0
```

acceptance/tests/facts/non_root_users_without_errors.rb
```diff
@@ -0,0 +1,34 @@
+test_name "C59196: running facter as a non-root user should not produce permission errors" do
+  tag 'risk:high'
+
+  # confine :except, :platform => 'aix' # bug FACT-1586
+
+  confine :except, :platform => 'windows' # this test is for unix systems
```
johnduarte: Why does this only apply to unix systems? Don't we support running `facter` as a non-privileged user on Windows?

acceptance/tests/facts/partitions.rb
```diff
@@ -0,0 +1,37 @@
+test_name "C96148: verify partitions facts" do
+  tag 'risk:high'
+
+  confine :except, :platform => 'osx' # no partitions on osx
+  confine :except, :platform => 'windows' # no partitions on windows
+  confine :except, :platform => 'solaris' # no partitions on solaris
+
+  require 'json'
+
+  possible_facts = [
```
johnduarte: Is there a reason that the expected facts in this test are implemented as arrays rather than as hashes like they are in the other facter tests. Should they be moved into `acceptance/lib/facter/acceptance/base_fact_utils.rb` with the other expected facts?

acceptance/lib/facter/acceptance/user_fact_utils.rb
```diff
@@ -117,6 +117,16 @@ def custom_fact_content(key='custom_fact', value='custom_value', *args)
   end
         EOM
       end
+
+      # Return the correct shell path for Unix system under test
+      #
+      def user_shell(agent)
```
ericldelaney: Because I copied the code from somewhere else. Although now that I look it better this way otherwise you might assume your getting shell, when you really getting one of the following: ksh93, dash, or bash depending on the system your on.

acceptance/tests/facts/non_root_users_without_errors.rb
```diff
@@ -0,0 +1,34 @@
+test_name "C59196: running facter as a non-root user should not produce permission errors" do
+  tag 'risk:high'
+
+  # confine :except, :platform => 'aix' # bug FACT-1586
```
ericldelaney: Its excluded below, I'll merge this into the other one.

acceptance/tests/facts/non_root_users_without_errors.rb
ericldelaney: I'll file a ticket to write a windows version of this.

acceptance/tests/facts/non_root_users_without_errors.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  confine :except, :platform => 'windows' # this test is for unix systems
+  confine :except, :platform => 'aix' # has errors from the su command
```
ericldelaney: No, the system's su(1) command is printing an error message to stderr, (nothing to do with us). I'll update the comment to make that more clear

acceptance/tests/facts/partitions.rb
ericldelaney: Because when I started writing it as a bunch of "unless .nil? ... end" blocks it made sense to refacter it into an array and use a common loop over the ones that might or might not be set to reduce the code.  I can make it a hash if you'd prefer.

I only put the other code in base_fact_util.rb because of how large it was. Given that this tests all fits on a screen, and no one else is going to needs these facts, I'd ready see it in the test so you have one place to look.

acceptance/tests/facts/non_root_users_without_errors.rb
ericldelaney: You would think that, but nope. The systems that fail to manage the home have a version of su(1) that complains that it can't access the parent directories of the working directory. I've changed it but you don't gain any coverage.

PR #1583, opened 2017-06-09
(FACT-1646) fix label regex to allow for '/' on aix

ericldelaney:
On AIX the label can contain just '/', so allow zero or more words
after the '/'

### DIFF
--- a/acceptance/tests/facts/partitions.rb
+++ b/acceptance/tests/facts/partitions.rb
@@ -13,7 +13,7 @@ test_name "C96148: verify partitions facts" do
       ['uuid', /^[-a-zA-Z0-9]+$/],
       ['partuuid', /^[-a-f0-9]+$/],
       ['mount', /^\/.*/],
-      ['label', /\w+/],
+      ['label', /.*/],
       ['partlabel', /\w+/],
   ]
 


ericldelaney:
Tested on aix 7.2

acceptance/tests/facts/partitions.rb
```diff
@@ -13,7 +13,7 @@
       ['uuid', /^[-a-zA-Z0-9]+$/],
       ['partuuid', /^[-a-f0-9]+$/],
       ['mount', /^\/.*/],
-      ['label', /\w+/],
```
branan: This will match anything now ("zero or more word characters anywhere in the string"). Do we want to instead just validate that the fact exists, and not try to enforce any format on it (since volume labels are freeform anyway)

ericldelaney: The problem is that for this array of entries, they may or may not bet set depending on the different operating systems and how they are configured. So rather adding a bunch of do this for this operating system, we instead if the label is set then the regex should match it

ericldelaney: Now that I think about it, it should probably just be /.*/, as it could be set to just about anything when its set.

PR #1584, opened 2017-06-09
(packaging) Bump version to 3.6.5 for release

ScottGarman:


Magisus:
There is also a place in `lib/Doxyfile` that needs to be updated for Facter.

ScottGarman:
@Magisus thanks for catching that! PR updated.

PR #1585, opened 2017-06-15
(FACT-1455) Handle ERANGE for getpwuid_r() and getgrgid_r() calls

riton:
This PR should fix the issue [FACT-1455](https://tickets.puppetlabs.com/browse/FACT-1455) that is raised when `ERANGE` is triggered by `getpwuid_r()` or `getgrgid_r()`

### DIFF
--- a/lib/src/facts/posix/identity_resolver.cc
+++ b/lib/src/facts/posix/identity_resolver.cc
@@ -25,7 +25,15 @@ namespace facter { namespace facts { namespace posix {
         uid_t uid = geteuid();
         struct passwd pwd;
         struct passwd *pwd_ptr;
-        int err = getpwuid_r(uid, &pwd, buffer.data(), buffer.size(), &pwd_ptr);
+        int err = -1;
+
+        do {
+            err = getpwuid_r(uid, &pwd, buffer.data(), buffer.size(), &pwd_ptr);
+            if (err == ERANGE) {
+                buffer.resize(buffer.size() + 1024);
+            }
+        } while (err == EINTR || err == ERANGE);
+
 
         if (err != 0) {
             LOG_WARNING("getpwuid_r failed: {1} ({2})", strerror(err), err);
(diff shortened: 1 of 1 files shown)


riton:
Maybe I should have reference the issue FACT-1455 in my git commit message..
Tell me if you want me to amend my commit message.

branan:
@riton Yeah, please do. Our commit checker wants the format "(FACT-####) <Message>", and we have other internal tooling that relies on that as well.

riton:
@branan should be better now

lib/src/facts/posix/identity_resolver.cc
```diff
@@ -25,7 +25,15 @@ namespace facter { namespace facts { namespace posix {
         uid_t uid = geteuid();
         struct passwd pwd;
         struct passwd *pwd_ptr;
-        int err = getpwuid_r(uid, &pwd, buffer.data(), buffer.size(), &pwd_ptr);
+        int err = -1;
+
+        do {
+            err = getpwuid_r(uid, &pwd, buffer.data(), buffer.size(), &pwd_ptr);
+            if (err == ERANGE) {
+                buffer.resize(buffer.size() + 1024);
```
branan: *Typically* I'd prefer doubling the buffer space each time, since that will approach the needed space in logarithmic time (at the cost of using extra memory when it inevitably overshoots).

In this case, I'm fine with the linear growth. The difference in complexity only affects pwd entries that are greater than 3KiB - and probably not measurably until 7KiB, where this version (theoretically) takes twice as long as a doubling one would.

Since even entries that are greater than 1KiB appear to be pretty rare, I'm fine with the linear algorithm here. No sense creating busy work for you over an optimization that won't matter in the real world.

riton: I do agree that such large entries should be pretty rare.
Thanks for the review @branan !

Magisus:
👍

branan:
Cherry-picked to stable as f852c3a22c0e981bd9c535087cf0dbdb52e222ee. Thanks!

PR #1586, opened 2017-06-15
(FACT-1650) change fqdn test to loop waiting for the hostname to update

ericldelaney:
On Solaris 11 the hostname(1) command returns before its completely changed in the
system and a call to hostname again can return the old name. This change queries
the hostname until its confirmed changed or tells you that it failed to change.

### DIFF
--- a/acceptance/tests/ticket_1238_hostname_fqdn.rb
+++ b/acceptance/tests/ticket_1238_hostname_fqdn.rb
@@ -1,5 +1,6 @@
 test_name 'C93827: facter fqdn should return the hostname when its a fully qualified domain name' do
   tag 'risk:high'
+  require 'timeout'
 
   confine :except, :platform => 'windows'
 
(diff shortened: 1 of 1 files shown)


ericldelaney:
tested on: arista4-32a, centos7-64a, ciscoxr-64a%7Bhypervisor=vmpooler%7D, cisconx-64a%7Bhypervisor=vmpooler%7D, cumulus25-64a, debian8-32a, fedora25-32a, oracle7-64a, osx1012-64a, redhat7-64a, scientific7-64a, sles12-64a, solaris11-64a, ubuntu1610-32a, windows10ent-32a, windows2016-6432a

Magisus:
We're okay doing this loop on all systems even though only Solaris 11 has a problem? Doesn't bother me, just checking.

branan:
I'd prefer to keep the code simpler by not gating it. On most systems it'll exit the loop after the first check anyway

ericldelaney:
@Magisus it works as written on all platforms. I thought about just having it do something just for Solaris 11, but by going simple it tells you should any OS have problems changing the name and in general and we're probably .5 to .75 seconds faster because we're not doing the 1 second sleep on all the platforms regardless

Magisus:
Sounds good.

PR #1588, opened 2017-06-21
(packaging) Update version to 3.7.0

Magisus:
This commit updates Facter's version to 3.7.0 in preparation for the
Puppet Agent 5.0.0 release.

We ended up not having any breaking changes to ship, so we are not
shipping at Facter 4 at this time.


McdonaldSeanp:
👍

PR #1589, opened 2017-06-26
(FACT-1651) Update specific fact tests to support Linux on Power8

ericldelaney:


acceptance/tests/facts/ruby.rb
```diff
@@ -1,49 +1,59 @@
...
+      ruby_version   = /2\.\d+\.\d+/
+      expected_facts = {
+          'ruby.platform' => ruby_platform,
+          'ruby.sitedir'  => /\/site_ruby/,
+          'ruby.version'  => ruby_version
+      }
 
-    expected_ruby.each do |fact, value|
-      assert_match(value, fact_on(agent, fact))
+      step("verify that ruby structured fact contains facts") do
```
Magisus: Did this step get changed so we only talk to the agent once? Or was there some additional reason?

Magisus: Answered in hipchat, yes, just cleanup.

PR #1591, opened 2017-06-30
(FACT-1675) fix virtual, and is_virtual facts for OpenBSD VMs

buzzdeee:
Without the patch, OpenBSD VMs running within OpenBSD's hypervisor
vmm will wrongly report the fact virtual -> physical, and is_virtual -> false
With the patch, the following will be reported: virtual -> vmm, is_virtual -> true

It's unfortunate, but hw.product doesn't return anything in such
cases, therefore in openbsd/dmi_resolver, set the make use of
hw.vendor to set the product name.
The rest is rather mechanic, add a string for the hypervisor to facts/vm
as well as a mapping from the sysctl value to the string to
virtualization_resolver.

### DIFF
--- a/lib/src/facts/openbsd/dmi_resolver.cc
+++ b/lib/src/facts/openbsd/dmi_resolver.cc
@@ -13,7 +13,14 @@ namespace facter { namespace facts { namespace openbsd {
         result.bios_vendor = sysctl_lookup(HW_VENDOR);
         result.uuid = sysctl_lookup(HW_UUID);
         result.serial_number = sysctl_lookup(HW_SERIALNO);
+        // OpenBSD running as virtual machine within
+        // OpenBSD vmm don't return HW_PRODUCT. For that
+        // case use the HW_VENDOR, to please the
+        // virtualization_resolver
         result.product_name = sysctl_lookup(HW_PRODUCT);
+        if (result.product_name.length() == 0) {
+            result.product_name = result.bios_vendor;
+        }
         result.bios_version = sysctl_lookup(HW_VERSION);
 
         return result;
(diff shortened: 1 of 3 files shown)


Magisus:
We are in the process of changing the way we do virtualization detection in Facter, see https://tickets.puppetlabs.com/browse/FACT-1654. However, before we transition to that, this seems reasonable. @caseywilliams thoughts on this?

caseywilliams:
This does seem reasonable, and it gives us some useful information for our new detection library. It would be nice not to call `sysctl_lookup(HW_PRODUCT)` twice, though - it could be stored in a variable and reused. Otherwise this looks good to me 👍

Magisus:
@buzzdeee would you also mind rebasing this to fix those conflicts? We can't merge as is.

buzzdeee:
@Magisus rebased as requested.

lib/src/facts/openbsd/dmi_resolver.cc
```diff
@@ -13,7 +13,15 @@ namespace facter { namespace facts { namespace openbsd {
         result.bios_vendor = sysctl_lookup(HW_VENDOR);
         result.uuid = sysctl_lookup(HW_UUID);
         result.serial_number = sysctl_lookup(HW_SERIALNO);
-        result.product_name = sysctl_lookup(HW_PRODUCT);
+        // OpenBSD running as virtual machine within
+        // OpenBSD vmm don't return HW_PRODUCT. For that
+        // case use the HW_VENDOR, to please the
+        // virtualization_resolver
+        if (sysctl_lookup(HW_PRODUCT).length() > 0) {
```
Magisus: Per @caseywilliams 's suggestion, can you refactor this to only call `sysctl_lookup` once?

buzzdeee:
@Magisus so, the last version also saves first in the variable, and only if it is empty, does the other lookup. hope that's OK then.

buzzdeee:
oh, wait, I just saw something stupid

buzzdeee:
this should now be better

PR #1592, opened 2017-06-30
(FACT-1676) Update el test for Amazon values

johnduarte:
This commit updates the acceptance/tests/facts/el.rb test to
include os.name, os.release, kernelversion specific for Amazon Linux.


johnduarte:
Tested on Amazon Linux 6 x86_64

ericldelaney:
👍 

Looks good

Magisus:
Is this OS available as a pooler image? Or how are we running tests on it?

PR #1595, opened 2017-07-10
(FACT-1710) Support zfs/zpool facts on FreeBSD

smortex:
These changes where stolen from the solaris implementation.


Magisus:
Since these are exactly the same as the Solaris implementation, might there be a way to use the same implementation for both? I'm not sure what the common denominator would be and how to fit that into our directory structure? /cc @MikaelSmith 

On the flip side, @smortex can you think of any reason why the two implementations might diverge at some point?

PR #1598, opened 2017-07-10
(FACT-1713) Add support for FreeBSD jails virtualization resolution

smortex:


Magisus:
We are doing work around changing the way that Facter does virtualization detection (see https://github.com/puppetlabs/libwhereami). Eventually we might be able to update this to use that library, but for now this looks fine to me.

PR #1607, opened 2017-07-20
Distinguish FreeBSD kernelrelease and os.release

smortex:
FreeBSD release (userland) version (returned by `freebsd-version(1)`) might be different from the kernel version (returned by `uname(1)`).  This happens for example after a vulnerability is fixed in a userland program, the kernel is unaffected by the fix.  Fox example, we can have:

```
g4% uname -r
11.0-RELEASE-p9
g4% freebsd-version
11.0-RELEASE-p11
```

Currently, the same information is returned twice. It would be better to return the appropriate information in order to detect e.g. machines with pending security updates:

```
g4% facter os kernelrelease
kernelrelease => 11.0-RELEASE-p9
os => {
  architecture => "amd64",
  family => "FreeBSD",
  hardware => "amd64",
  name => "FreeBSD",
  release => {
    full => "11.0-RELEASE-p9",         <--- should be "11.0-RELEASE-p11"
    major => "11",
    minor => "0-RELEASE-p9"            <--- should be "0-RELEASE-p11"
  }
}
```

This PR makes it possible to override the kernel and release facts resolution, and introduce a FreeBSD resolver that rely on `freebsd-version(1)` to return the right release information.

### DIFF
--- a/lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
+++ b/lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
@@ -216,6 +216,20 @@ namespace facter { namespace facts { namespace resolvers {
          * @return Returns the resolver data.
          */
         virtual data collect_data(collection& facts);
+
+        /**
+         * Collects the resolver's kernel data.
+         * @param facts The fact collection that is resolving facts.
+         * @param result The current resolver data.
+         */
+        virtual void collect_kernel_data(collection& facts, data &result);
+
+        /**
+         * Collects the resolver's release data.
+         * @param facts The fact collection that is resolving facts.
+         * @param result The current resolver data.
+         */
+        virtual void collect_release_data(collection& facts, data &result);
     };
 
 }}}  // namespace facter::facts::resolvers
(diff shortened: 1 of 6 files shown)


smortex:
Hello!  I just rebased this branch on top of master to fix conflicts.

lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
```diff
@@ -216,6 +216,20 @@ namespace facter { namespace facts { namespace resolvers {
          * @return Returns the resolver data.
          */
         virtual data collect_data(collection& facts);
+
+        /**
+         * Collects the resolver's kernel data.
+         * @param facts The fact collection that is resolving facts.
+         * @param result The current resolver data.
+         */
+        virtual void collect_kernel_data(collection& facts, data &result);
```
Magisus: I was trying to find examples but it's still unclear to me, @branan or @MikaelSmith do we want auxiliary methods like these to be private like in https://github.com/puppetlabs/facter/blob/master/lib/inc/internal/facts/resolvers/networking_resolver.hpp or just protected like in https://github.com/puppetlabs/facter/blob/master/lib/inc/internal/facts/resolvers/virtualization_resolver.hpp? I feel like these could easily be private given how they're used.

MikaelSmith: They have to be protected for the inheriting class to override them. The usage here makes sense.

PR #1600, opened 2017-07-10
(maint) Support static compilation

MikaelSmith:
Support compiling libfacter.a, allowing creation of a single statically
compiled binary that uses facter.

PR #1608, opened 2017-07-21
(FACT-1723) Go bindings

parisiale:
- Adding C wrapper that provides a function to collect default facts in JSON format;
- adding a simple Go library that offers a function that returns default facts in JSON format as a string;
- adding unit tests for both C and Go functions;
- adding new CMake option for building a static libfacter;
- setting up building for ~~OS X~~ fedora.

TODO:

- [x] integrate building configuration for linux; 
- [x] consider updating README with Go lib / static building instructions (NB: facter.go contains instructions for dealing with libiconv, required by Boost.Log);
- [x] fix commit prefixes.


parisiale:
Hmmm, I got an immediate appveyor "non-m" failure...

parisiale:
Merged and updated to #1600.
Added a few cpplint fixes.

parisiale:
We still need to fix the commit title requirement, as Travis is not happy with (DI-***).
What do you say @MikaelSmith ?

go_wrapper/facter.go
```diff
@@ -0,0 +1,40 @@
...
+// - edit include/iconv.h to define the LIBICONV_PLUG to some non null value
+//   before it gets checked; that will ensure that the 'lib' prefix is not
+//   appended to some symbols needed by Boost.Locale
+// - make
+// - make install
+
+// NB(ale): getting few ld warnings (direct access) for Boost.Log (Boost v1.64)
+
+/*
+#cgo LDFLAGS: -fPIC -lfacter -laugeas -lcpp-hocon -lyaml-cpp -l:leatherman_execution.a -l:leatherman_logging.a -l:leatherman_locale.a -l:leatherman_ruby.a -l:leatherman_dynamic_library.a -l:leatherman_util.a -l:leatherman_file_util.a -l:leatherman_curl.a -lboost_log-mt -lboost_log_setup-mt -lboost_thread-mt -lboost_date_time -lboost_filesystem-mt -lboost_system-mt -lboost_chrono-mt -lboost_regex-mt -lboost_atomic-mt -lboost_program_options-mt -lboost_locale-mt -liconv -lcurl -lcrypto -ldl -lz -lstdc++ -lm
```
johnmccabe: should remove the `-mt` they're a hangover from the initial work on OSX and not relevant on Linux

CMakeLists.txt
```diff
@@ -130,6 +131,10 @@ set(FACTER_CXX_FLAGS "${LEATHERMAN_CXX_FLAGS}")
 
 add_definitions(${LEATHERMAN_DEFINITIONS})
 
+if (STATIC_FACTER)
```
MikaelSmith: This is usually set by how Leatherman is built, so doesn't make a lot of sense to include here.

MikaelSmith:
I'd opt for a FACT ticket for adding a go wrapper.

parisiale:
@MikaelSmith 👍 
I'll make one and update the commits.
Then I'll pass the ownership of this PR to @davejohnston ; I'll be ready to jump in as needed

CMakeLists.txt
parisiale: Oh, I missed that when I updated to your 'static' commit; will remove.

locales/FACTER.pot
```diff
@@ -6,7 +6,7 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: FACTER 4.0.0\n"
```
johnmccabe: Was this version change intentional?

Magisus: No, we had temporarily thought we were going to ship a Facter 4, then decided against it. It appears this file missed getting updated.

parisiale:
Rebased to master.
Corrected the use of STATIC_FACTER.
Updated commit titles to have the (FACT-1723) label.

lib/src/cwrapper.cc
```diff
@@ -0,0 +1,63 @@
...
+        facts.write(stream,
+                    facter::facts::format::json,
+                    {},     // queries vector
+                    true,   // show_legacy flag
+                    true);  // strict_errors flag
+
+        auto json_facts = stream.str();
+        auto l = json_facts.length()+1;
+        *result = new char [l];
+        auto rc = std::snprintf(*result, l, "%s", json_facts.c_str());
```
MikaelSmith: I think `strncpy` would also work.

parisiale:
Rebased with master.
Replaced `snprintf` with `strncpy`.

parisiale:
This is ready to be merged for our team.

lib/src/cwrapper.cc
```diff
@@ -0,0 +1,59 @@
...
+#include <stdio.h>
+#include <sstream>
+
+namespace po = boost::program_options;
+
+uint8_t get_default_facts(char **result) {
+    try {
+        po::variables_map vm {};
+        hocon::shared_config hocon_conf;
+        hocon_conf = facter::util::config::load_default_config_file();
```
Magisus: We don't seem to be using any of the settings loaded here. Do we plan to in the future? If not, do we need to load it? (I could be not remembering how it's used somewhere)

parisiale: 👍  we don't need that now.
Tested on OS X and Fedora.

parisiale:
Rebased to master.
Removed config calls.

lib/src/cwrapper.cc
```diff
@@ -0,0 +1,40 @@
...
+
+uint8_t get_default_facts(char **result) {
+    try {
+        // NB: ssh resolver cannot be blocked
+        facter::facts::collection facts {{},      // blocklist set
+                                         {},      // ttls map (resolvers - ttl)
+                                         true};   // ignore_cache flag
+
+        // The boolean arg is meant to avoid including ruby facts
+        facts.add_default_facts(false);
```
branan: Why are we skipping external, environment, and ruby facts in this wrapper? If this is meant to be a generic wrapper, I'd expect it to either follow Facter's default behavior (which is to collect all of those), or to provide option flags for which things to collect.

lib/src/cwrapper.cc
```diff
@@ -0,0 +1,40 @@
...
+        std::ostringstream stream;
+        facts.write(stream,
+                    facter::facts::format::json,
+                    {},     // queries vector
+                    true,   // show_legacy flag
+                    true);  // strict_errors flag
+
+        auto json_facts = stream.str();
+        auto l = json_facts.length()+1;
+        *result = new char [l];
```
branan: This is a C wrapper, but you're returning memory allocated with `new`. This makes it impossible for C code to free it. We should either use `malloc` here, or provide our own function for releasing the allocated memory.

MikaelSmith: Good point, I missed that. Presumably the `defer C.free` is what's handling freeing it, and seems unlikely that works correctly with `new`.

branan: I think that works with glibc... no idea about OSX, and it is definitely going to be busted on Windows.

johnmccabe: @branan we're specifically looking only at Linux at the moment, I've added a build constraint in #1615 that'll prevent it being used on non-Linux OS's until we address compatibility.

PR #1610, opened 2017-07-25
(maint) Fix memory allocation for default facts C wrapper

MikaelSmith:


MikaelSmith:
I think this is right, not sure how we test the wrapper though.

lib/src/cwrapper.cc
```diff
@@ -30,7 +30,7 @@ uint8_t get_default_facts(char **result) {
 
         auto json_facts = stream.str();
         auto l = json_facts.length()+1;
-        *result = new char [l];
+        *result = (char*)malloc(sizeof(char)*l);
```
branan: should use a C++-style case here (this is still C++ code).

branan:
LGTM once the cast is fixed up

lib/src/cwrapper.cc
MikaelSmith: Done.

lib/src/cwrapper.cc
```diff
@@ -30,7 +30,7 @@ uint8_t get_default_facts(char **result) {
 
         auto json_facts = stream.str();
         auto l = json_facts.length()+1;
-        *result = new char [l];
+        *result = static_cast<char*>(malloc(sizeof(char)*l));
```
ericldelaney: Don't you need to check that the malloc didn't return NULL?

MikaelSmith: Urgh, yes.

PR #1611, opened 2017-07-26
(maint) Report zpool_version=5000 for feature flags

smortex:
This is a follow-up to #1597.

The OpenZFS folks on IRC confirmed that reporting *5000* instead of *1000*
would be better: « on-disk the fields show version number 5000 »; « some
material is old, 5000 is what's used now. 1000 may have been a sort of
plan that changed early in openzfs creation ».

### DIFF
--- a/lib/src/facts/resolvers/zpool_resolver.cc
+++ b/lib/src/facts/resolvers/zpool_resolver.cc
@@ -62,7 +62,7 @@ namespace facter { namespace facts { namespace resolvers {
             case UNKNOWN:
                 if (re_search(line, zpool_version, &result.version)) {
                 } else if (re_search(line, zpool_feature_flags)) {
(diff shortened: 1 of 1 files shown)


Magisus:
Okay, great, thanks for following up on this 👍

smortex:
@Magisus yes, it's where I found the *5000* reference, and since it was not in-lined with the previous reference of "1000" found in [the slides](http://web.archive.org/web/20160419064650/http://blog.delphix.com/csiden/files/2012/01/ZFS_Feature_Flags.pdf) linked from the [Fature Flags](http://www.open-zfs.org/wiki/Feature_Flags) wiki page (1st line in the *Reference materials* section), I head-out to IRC and asked there.

Thanks for merging!

PR #1614, opened 2017-08-02
make system fact paths available to all. implement FACT-42

tb3088:
FACT-42 was 100% valid and should have been honored. There is never cause to prohibit non-root users from being able to utilize facts that the sysadmin deliberately put in the system-wide directory(s). If some of the external facts can't return valid values without elevated privileges, it's up to the external fact writer to handle it properly (eg. use sudo or return a blank value). Puppet has no business interfering in perfectly reasonable and expected system behavior. Even if Puppet-agent is run as root most (all?) of the time, Facter is extremely useful in non-Puppet contexts and deserves to work correctly.

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -157,7 +157,7 @@ namespace facter { namespace facts {
             } else {
                 LOG_DEBUG("skipping external facts for \"{1}\": {2}", dir, msg);
             }
-            return found;
+            return false;
         }
 
         LOG_DEBUG("searching {1} for external facts.", search_dir);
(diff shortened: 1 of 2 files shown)


lib/src/facts/collection.cc
```diff
@@ -166,15 +166,15 @@ namespace facter { namespace facts {
             for (auto const& res : resolvers) {
                 if (res->can_resolve(path)) {
                     try {
-                        found = true;
                         res->resolve(path, *this);
+                        found = true;
```
MikaelSmith: This is definitely a bug, thanks for finding it.

MikaelSmith:
https://docs.puppet.com/facter/3.7/configuring_facter.html#global can be used to restore system-wide directories.

FACT-42 was definitely necessary to add support for non-root directories. It could have preserved using system-wide directories as well. However, at this point I'd be wary of changing it as it could start causing errors for directories that are unreadable by the non-root user.

tb3088:
a cursory review of the various can_resolve() does seem to indicate the checking is rather lacking.

MikaelSmith:
Can you be more specific about what checking you think is missing?

branan:
This is an intentional behavior to isolate non-root puppet environments from root ones. Facter now supports configuring the fact paths (and has for quite a while), so I recommend going that route if you have specific needs.

PR #1615, opened 2017-08-02
(maint) Fix Go package name/path, update linked flags and add simple example app

johnmccabe:
This PR addresses issues encountered during DI-136.

It renames the go package to `factergo` and updates the directory name accordingly (the previous name/path weren't resolvable), the package will now be importable/go-gettable via:
```
github.com/puppetlabs/facter/factergo
```
A small example app that demonstrates the packages usage is added.

A build constraint is added to restrict the packages usage to Linux for the moment.

And finally it updates the linker flags to allow cgo to produce a static binary, glibc remains dynamically linked. Building on a distro with an old glibc will produce a widely portable binary, for example CentOS 6 will produce a binary that works with glibc 2.12-1.209+.

*Note* any resulting binary can be reduced in size significantly by stripping debug symbols and packing with `upx`.


MikaelSmith:
Please use a FACT project for this. I'm also ok if you just want to use `(maint)` and mention DI-136 in the description, since it doesn't really have any external impact.

johnmccabe:
Ah, cool will do.

johnmccabe:
@MikaelSmith thats the PR/commits updated, not sure why the DEBUG and RELEASE builds jobs are failing tho..

MikaelSmith:
Ruby problems, ugh.

johnmccabe:
@MikaelSmith any objection to merging this prior to the CI Ruby fix?

MikaelSmith:
These changes should have no impact on CI. Local tests confirm it still works generally.

johnmccabe:
@MikaelSmith ty

PR #1616, opened 2017-08-03
(FACT-1720) Resolve environmnt facts after all other sources

branan:
Previous versions of facter would resolve environment facts last,
allowing users to override both built-in and external facts from
the command line. CFacter resolved in a different order, which
would cause external facts to take precedence over those set in
the environment.

This brings Facter 3 in line with earlier facter releases by
resolving environment facts last.


branan:
It looks like our travis config for 3.6.x is pulling in Ruby 2.4, but we don't have the necessary fixes in that version to run against it.

ericldelaney:
👍

PR #1618, opened 2017-08-03
Mergeup 3.6.x to 3.7.x

ScottGarman:
Please take a close look at these changes, I haven't done a merge-up in a while, and I've not included the os-based acceptance test files from 3.6.x, and created FACT-1725 to track the refactoring work required there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: cpp
 compiler:
   - gcc
+dist: precise
 sudo: false
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


branan:
I think we intentionally removed the os-specific test files in favor of fact-specific ones. We shouldn't be re-adding those files, and instead whatever changes were made in 3.6.x need to be made to the fact-specific tests in 3.7.x

ScottGarman:
Ok, I've redone the merge-up and not included the os-specific acceptance test files. FACT-1725 was filed to track that work.

PR #1621, opened 2017-08-11
(FACT-1729) Refactor FFI declarations

Iristyle:
Facter redefines FFI constants in the same area as Puppet.  This can cause Ruby warnings, and has the potential to redefine methods that already exist in Puppet with older / errant code.

To insulate Facter from Puppet, move these similar definitions to new namespaces and update callsites accordingly.


Magisus:
Was able to verify that this gets rid of the errors that have been appearing on Windows when running puppet unit tests 👍

PR #1622, opened 2017-08-11
(PE-21053) Enable nxcompat and dynamicbase flags in Windows builds

jsane:
Being done to meet customer's security audit requirements.

There is a blocker customer escalation on it and is a priority. Pl consider merging at the earliest.


jsane:
Re-submitted with correct ticket number. https://github.com/puppetlabs/facter/pull/1624.

PR #1625, opened 2017-08-14
(packaging) Prepare for 3.8.0 release

branan:


McdonaldSeanp:
👍

Magisus:
I think there might also be a spot in the POT file that will change. If you have gettext tools installed, you should try compiling this to catch any other needed updates to the POT file.

branan:
@Magisus shouldn't Jenkins be doing that for me? I've been explicitly excluding POT changes from my commits generally

geoffnichols:
Is this the place in the POT file? https://github.com/puppetlabs/facter/blob/3.8.x/locales/FACTER.pot#L9

branan:
@Magisus @geoffnichols updated the POT file for now. Would still like a bit better guidance on when we need to manually make changes to it vs. when we can expect Jenkins to do the right thing

geoffnichols:
👍

Magisus:
@branan it doesn't commit them for C++ right now :/

PR #1628, opened 2017-08-23
(maint) Restructure AppVeyor

Iristyle:
- Environment setup can be moved to the 'init' job in preparation of
   supporting Visual Studio builds

   Use multiline scripts where it makes sense.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,8 +7,8 @@ environment:
 init:
   - |
       choco install mingw-w64 -y -Version 4.8.3 -source https://www.myget.org/F/puppetlabs
-      choco install cmake -y -Version 3.2.2 -source https://www.myget.org/F/puppetlabs
       choco install -y gettext -Version 0.19.6 -source https://www.myget.org/F/puppetlabs
+      cmake --version
   - ps: |
       $env:PATH = $env:PATH.Replace("Git\bin", "Git\cmd")
       $env:PATH = $env:PATH.Replace("Git\usr\bin", "Git\cmd")
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,9 +33,11 @@ install:
   - bundle install --jobs 4 --retry 2 --gemfile=lib/Gemfile --quiet
 
 build_script:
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Separately, I plan to apply https://github.com/puppetlabs/pxp-agent/pull/629/files to Facter eventually.

appveyor.yml
```diff
@@ -3,31 +3,34 @@ clone_depth: 10
...
+      wget "https://github.com/puppetlabs/leatherman/releases/download/$env:LEATHERMAN_VERSION/leatherman.7z" -OutFile "$env:temp\leatherman.7z"
+      7z.exe x $env:temp\leatherman.7z -oC:\tools | FIND /V "ing "
 
-  - ps: wget "https://github.com/puppetlabs/cpp-hocon/releases/download/$env:CPPHOCON_VERSION/cpp-hocon.7z" -OutFile "$pwd\cpp-hocon.7z"
-  - ps: 7z.exe x cpp-hocon.7z -oC:\tools | FIND /V "ing "
+      wget "https://github.com/puppetlabs/cpp-hocon/releases/download/$env:CPPHOCON_VERSION/cpp-hocon.7z" -OutFile "$env:temp\cpp-hocon.7z"
+      7z.exe x $env:temp\cpp-hocon.7z -oC:\tools | FIND /V "ing "
 
+install:
+  - git submodule update --init --recursive
```
MikaelSmith: This shouldn't be needed anymore.

Iristyle:
Cool - will try removing the submodule bit....

May stick to a first PR with some clean-up / similar version bumps to what you have in pxp-agent, and then toss up a second PR

Should we look at the list of AppVeyor available tools to see if we even need to install pieces of the toolchain? https://www.appveyor.com/docs/build-environment/

I don't know if we have hard deps on CMake version, etc, etc.

MikaelSmith:
CMake 3.8.2 would work fine, we just require 3.2.2+. Libraries I'd be wary of because there's so many ways they can be built. So primarily CMake.

PR #1629, opened 2017-08-23
(FACT-1737) Add "disks" and "partitions" facts support for FreeBSD

smortex:
Managing FreeBSD storage devices is usually done through a modular disk transformation framework, [GEOM].  The system ships with [libgeom], a library allowing userland interaction with this system.

The commit adds a bunch of `facter::util::freebsd::geom_*` classes, wrapping the API exposed by FreeBSD for an easy use from Facter's code, and use them to build two missing facts on FreeBSD: `disks` and `partitions`.

[GEOM]: https://www.freebsd.org/doc/handbook/geom.html
[libgeom]: https://www.freebsd.org/cgi/man.cgi?query=libgeom

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -201,6 +201,7 @@ elseif ("${CMAKE_SYSTEM_NAME}" MATCHES "FreeBSD")
         "src/facts/freebsd/collection.cc"
         "src/facts/glib/load_average_resolver.cc"
         "src/facts/bsd/filesystem_resolver.cc"
+        "src/facts/freebsd/filesystem_resolver.cc"
         "src/facts/bsd/networking_resolver.cc"
         "src/facts/bsd/uptime_resolver.cc"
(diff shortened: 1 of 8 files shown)


smortex:
I rebased this PR on top of master.  No conflicts and still working as expected.

smortex:
@branan thanks for the review!

While creating this pull request, [I added missing documentation of the `geom_*` functions](https://reviews.freebsd.org/D12679) in `libgeom(3)` but the man pages have only been updated in the *-CURRENT* branch of FreeBSD and are not (yet) visible by following the generic link to the man page I provided in the first comment.  [This specific link](https://www.freebsd.org/cgi/man.cgi?query=libgeom&manpath=FreeBSD+12-current)  will show the updated documentation but will break when *-CURRENT* will be FreeBSD 13, so I am not changing the link in the first comment.

ScottGarman:
LGTM :+1:

PR #1631, opened 2017-08-24
(packaging) Update facter version to 3.6.7

McdonaldSeanp:


branan:
Need to update the translations file as well

Magisus:
I think there's also a reference to the version in the POT file too.

McdonaldSeanp:
@Magisus @branan updated

PR #1632, opened 2017-08-28
(FACT-1740) Update the filesystem resolver to skip AHAFS

McdonaldSeanp:
The /aha filesystem is a for the AIX event infrastructure, and the filesystem
reports 0 size, but several blocks in use. This commit filters out AHAFS in
the same way we filter out PROCFS.


McdonaldSeanp:
I will run a build to ensure this will not produce errors

branan:
Typo in commit message: "...is a for..."

McdonaldSeanp:
Updated

branan:
I'm :+1: once you get a build through

McdonaldSeanp:
Build completed successfully on AIX 6.1

PR #1634, opened 2017-08-29
(FACT-1740) Update aix filesystem resolver change

McdonaldSeanp:
This commit updates the filesystem resolver change to filter out MNT_AHAFS.
Aix 5.3 does not define that variable, so this commit updates the definition
to not check MNT_AHAFS on 5.3


branan:
@McdonaldSeanp did you confirm that this is still working as expected on AIX 6.1? It looks right, but I'd like to know for sure we found the right preprocessor define :)

McdonaldSeanp:
@branan I checked and that preprocessor define exists on 6.1, did you want me to test the package?

PR #1635, opened 2017-08-29
(FACT-1742) Add hypervisors fact

caseywilliams:
This commit introduces a new fact, `hypervisors`, which contains a list
of hypervisors detected by libwhereami, plus any collected metadata for
these hypervisors. The libwhereami dependency is optional, so this
change will have no effect unless it is installed.

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -7,6 +7,7 @@
 #include <facter/ruby/ruby.hpp>
 #include <facter/util/string.hpp>
 #include <facter/version.h>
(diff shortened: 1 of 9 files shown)


lib/src/facts/collection.cc
```diff
@@ -25,6 +25,10 @@
 #include <yaml-cpp/yaml.h>
 #include <algorithm>
 
+#ifdef USE_WHEREAMI
```
branan: Since this header no longer requires the libwhereami headers, I think we can remove this guard

PR #1637, opened 2017-09-06
(FACT-1744) Use LOG_INFO instead of LOG_WARNING on failed lookup

elofu17:
See https://tickets.puppetlabs.com/browse/FACT-1744

If the kenv() lookup fails, only give an INFOrmational error message, not a WARNing.

When I run puppet I don't want to see an error in each run!

When I run 'puppet agent -t -d' or 'facter -p -d' in debug mode, I see that all other errors of the same significance all use INFO, not WARN, so I think we should simply use WARN here as well, and the problem is solved. :-)

### DIFF
--- a/lib/src/facts/freebsd/dmi_resolver.cc
+++ b/lib/src/facts/freebsd/dmi_resolver.cc
@@ -30,7 +30,7 @@ namespace facter { namespace facts { namespace freebsd {
 
         LOG_DEBUG("kenv lookup for {1}", file);
         if (kenv(KENV_GET, file, buffer, sizeof(buffer) - 1) == -1) {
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Thanks for the PR! This change does seem to be in accord with our logging standards for failures in Facter.

Travis is erroring because the actual commit title doesn't match the expected pattern. You updated the PR title, but to modify the commit you could do something like
```
git commit --amend
git commit push -f elofu17/patch-1
```

elofu17:
Hi Michael!
Sorry, I'm no developer so I simply used the web-gui to create the PR.

So I don't know how to amend and re-push it. :-/

MikaelSmith:
Ok, I'll fix it with git.

PR #1639, opened 2017-09-09
(FACT-1746) Fix virtual and is_virtual facts for Proxmox VMs

elofu17:
See https://tickets.puppetlabs.com/browse/FACT-1746

A FreeBSD VM running under a Proxmox host will get:
result.product_name = kenv lookup for smbios.system.product = "Standard PC (i440FX + PIIX, 1996)".
This results in the incorrect facts virtual=physical and is_virtual=false.
 
By simply overriding result.product_name to "KVM", the facts are set correctly to virtual=kvm and is_virtual=true.

Patch was tested on a FreeBSD 10.3 VM.


MikaelSmith:
This seems safe enough. I don't have a way to test it.

PR #1640, opened 2017-09-13
(FACT-1750) Catch exceptions in resolvers and report as warnings

MikaelSmith:
Resolver failures should not prevent Facter continuing to function. If
we find failures that prevent the resolver from working, report them as
warnings rather than stopping Facter.

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -364,7 +364,11 @@ namespace facter { namespace facts {
 
         // Resolve normally
         LOG_DEBUG("resolving {1} facts.", res->name());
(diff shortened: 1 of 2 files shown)


MikaelSmith:
It'd be nice to also fix whatever's throwing the exception. This is useful on its own.

branan:
Should this target 3.6.x so it can land in the LTS (not that we have any *known* leaking exceptions in that version, but...)

MikaelSmith:
That doesn't sound crazy.

branan:
Per hipchat: Just gonna leave this targeted as-is, and pull back the change if we actually half to.

PR #1642, opened 2017-09-14
(FACT-1753) Update sles fact acceptance test for SLES 12 Power8

ScottGarman:
This allows me to run the facter acceptance tests with a SLES 12 Power8 target.

### DIFF
--- a/acceptance/tests/facts/sles.rb
+++ b/acceptance/tests/facts/sles.rb
@@ -27,6 +27,10 @@ test_name "Facts should resolve as expected on SLES" do
       os_arch                 = 's390x'
       os_hardware             = 's390x'
       processor_model_pattern = '' # s390x does not populate a model value in /proc/cpuinfo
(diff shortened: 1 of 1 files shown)


acceptance/tests/facts/sles.rb
```diff
@@ -83,6 +87,7 @@
         kernel_version = '3.0'
       elsif os_version == '12'
         kernel_version = '3.12'
+        kernel_version = '4.4' if os_arch == 'ppc64le'
```
branan: What's up with this wildly different kernel version?

ScottGarman: The Power8 hosts uses SLES 12 SP2, whereas our x86 ones are based on SP1.

branan: That is a hilarious kernel bump for a service pack O_O

ericldelaney:
👍

PR #1644, opened 2017-09-15
Mergeup 3.6.x to 3.9.x

ScottGarman:
Since the acceptance tests were refactored, this also includes a separate commit to update the base_fact_utils test to work with SLES 12 Power8. Tests pass against the agent when run manually.


McdonaldSeanp:
👍

PR #1647, opened 2017-09-22
(FACT-1642) Fix /proc/cpuinfo parsing on Power for Linux machines

ekinanp:


ekinanp:
Tested this on a Power machine:
processors => {
  count => 160,
  isa => "ppc64le",
  models => [
    "POWER8 (raw), altivec supported",
    "POWER8 (raw), altivec supported",
    "POWER8 (raw), altivec supported",
    "POWER8 (raw), altivec supported"
  ],
  physicalcount => 160,
  speed => "2.93 kHz"
}
^ I fixed the bug in the speed just now (forgot to multiply by 1000 to convert from MHz to kHZ), but other than that, I think that output looks good.

I'll fix linting issues later today after some feedback.

lib/inc/internal/facts/linux/processor_resolver.hpp
```diff
@@ -14,12 +14,40 @@ namespace facter { namespace facts { namespace linux {
     struct processor_resolver : posix::processor_resolver
     {
      protected:
+        enum ArchitectureType {POWER, GENERIC};
```
caseywilliams: Perhaps use an `enum class`? I might also call this `X86` or something instead of `GENERIC`

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -13,65 +14,155 @@ using namespace std;
...
+            if (id.empty() || cpus.emplace(move(id)).second) {
+                // Haven't seen this processor before
+                ++data.physical_count;
+            }
+            return true;
+        }, "^cpu\\d+$");
+
+        bool have_counts = data.logical_count > 0;
+        
+        if (architecture_type(data, root) == GENERIC) {
```
caseywilliams: It looks like both `architecture_type` and `add_*_cpu_data` read the contents of `/proc/cpuinfo` - might be nice to read it only once and pass in the output string.

lib/inc/internal/facts/linux/processor_resolver.hpp
```diff
@@ -14,12 +14,40 @@ namespace facter { namespace facts { namespace linux {
...
         /**
          * Collects the resolver data.
          * @param facts The fact collection that is resolving facts.
          * @return Returns the resolver data.
          */
         virtual data collect_data(collection& facts) override;
+
+     private:
+        void add_generic_cpu_data(data& data, bool have_counts, std::string const& root);
+        void add_power_cpu_data(data& data, bool have_counts, std::string const& root);
```
caseywilliams: If you continue to read from `/proc/cpuinfo` in these two functions, it seems consistent to just add an empty default `root` argument like in the `add_cpu_data`

lib/inc/internal/facts/linux/processor_resolver.hpp
ekinanp: Nice! I didn't know about enum classes. I chose GENERIC to capture all architectures that aren't POWER8 -- you can have x86, and i386. Maybe NOTPOWER? I don't want to use a boolean for this either, because in the future we could have different architectures that could be added to this enum.

lib/src/facts/linux/processor_resolver.cc
ekinanp: Yes, I'm going to use that instead of root at the end since that's the whole reason the root is passed in there. Good idea!

lib/inc/internal/facts/linux/processor_resolver.hpp
caseywilliams: Well, x86 as a [family](https://en.wikipedia.org/wiki/X86) would cover x86_64, i386, etc

ekinanp:
@caseywilliams I addressed your style comments. One thing -- I kept the root parameter in the add_*_cpu_data routines b/c in general, architecture_type will have an "isa" fact that is already computed, so you will probably not have to read /proc/cpuinfo there. This would mean that /proc/cpuinfo would be only read once for most cases and that is in the add_*_cpu_data routines.

In the edge case where you read it twice, I don't think it'll be a huge issue because that file is pretty small. So I think leaving root there is OK -- otherwise, the code won't be as clean if I'm going to ensure that /proc/cpuinfo is read once.

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -13,65 +14,155 @@ using namespace std;
...
-            if (id.empty() || cpus.emplace(move(id)).second) {
-                // Haven't seen this processor before
-                ++result.physical_count;
+        // use /proc/cpuinfo
+        unordered_set<string> to_be_seen;
+        bool seen_all = false;
+        lth_file::each_line(root + "/proc/cpuinfo", [&](string& line) {
+            // if we already know that we're on a Power machine, we can just skip
+            // the remaining lines
+            if (seen_all) {
```
caseywilliams: I think you could `return false` here to skip the remaining lines without continuing to loop over them

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -13,65 +14,155 @@ using namespace std;
...
         });
 
+    }
+
+    void processor_resolver::add_cpu_data(data& data, std::string const& root)
+    {
+        unordered_set<string> cpus;
+        lth_file::each_subdirectory(root + "/sys/devices/system/cpu", [&](string const& cpu_directory) {
+            ++data.logical_count;
+            string id = lth_file::read((path(cpu_directory) / "/topology/physical_package_id").string());
```
caseywilliams: Just for my own edification, what does the middle slash in `path(cpu_directory) / "/topology/..."` for here?

lib/inc/internal/facts/linux/processor_resolver.hpp
```diff
@@ -14,12 +14,40 @@ namespace facter { namespace facts { namespace linux {
...
+
         /**
          * Collects the resolver data.
          * @param facts The fact collection that is resolving facts.
          * @return Returns the resolver data.
          */
         virtual data collect_data(collection& facts) override;
+
+     private:
+        void add_generic_cpu_data(data& data, bool have_counts, std::string const& root = "");
```
caseywilliams: Since you renamed the enum member from `GENERIC` to `X86`, I'd rename this function too, myself

lib/src/facts/linux/processor_resolver.cc
ekinanp: That takes two parameters: <dir> and <path> and forms a new path <dir> / <path>. So if I had dir = "foo" and path = "bar" (or "/bar"), I'd get "foo/bar". It's a portable way of doing that.

ekinanp:
Retargeted it to 3.6.x and fixed linting issues.

ekinanp:
Rebased

ekinanp:
After a quick grep, it seems like some acceptance tests do use the model fact on power linux machines. I'm going to update them and add them to this PR

ekinanp:
Running this through an ad-hoc pipeline using rhel, ubuntu and sles w/ their x86 and power variants.

ekinanp:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental/job/experimental_puppet-agent_intn-van-sys_suite-daily-facter-FACT-1642/

All of the acceptance tests passed save for the x86 ubuntu machine. That failure, however, is not something related to the modifications made in this PR as the error message itself states that "libfacter was not found" so it is probably something that went wrong earlier in the build process.

So, I think that this PR is ready for merge once another reviewer approves of the changes.

PR #1648, opened 2017-09-26
(FACT-1764) zpool_version should return nothing when zfs-fuse is installed but service not running

lweller:
fixed issue https://tickets.puppetlabs.com/browse/FACT-1764


branan:
Facter 2 (the ruby implementation) is more or less dead, and we have no plans for a release that could include this fix.

If there are similar issues with Facter 3, please feel free to open a PR against that version.

PR #1649, opened 2017-09-26
(FACT-1763) Fix acceptance tests on power linux machines

ekinanp:
With the changes introduced in PA-1466, the acceptance tests fail
on power linux because stderr will report the missing dmidecode
component when it is supposed to be empty on normal linux
platforms. This commit fixes that.

### DIFF
--- a/acceptance/lib/facter/acceptance/user_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/user_fact_utils.rb
@@ -117,6 +117,13 @@ EOM
   end
         EOM
       end
+
+      # Returns a boolean indicating whether or not the host
+      # is a power linux machine
+      def power_linux?(host)
+        variant, _, arch, __ = host['platform'].to_array
+        variant =~ /el|sles|ubuntu/ && arch =~ /ppc64/
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


ekinanp:
Ran these tests locally on using
`beaker-hostgenerator redhat7-64a-redhat7-POWERa-sles12-POWERa-ubuntu1604-POWERa-aix72-POWERfa --hypervisor abs --disable-default-role --osinfo 1`

`rake ci:test:aio SHA=cef3673995924ecd48320dd06c24bb7ee25e0679 BEAKER_HOSTS=hosts.yaml ABS_RESOURCE_HOSTS=`cat abs_resource_hosts.json` SUITE_VERSION=5.2.0.156.gcef3673`

where abs_resource_hosts.json contains the vms I used for each OS. Outside of an error in a ruby test (probably because of the version conflict, as I used the SHA obtained from the failing pipeline outlined in FACT-1763 which was for master, while this branch is for 1.10.x), the tests all pass.

McdonaldSeanp:
👍

branan:
I'd rather see us fix Facter to not print that warning on POWER platforms. Since dmidecode doesn't work on that platform, Facter shouldn't even be trying to use it.

ekinanp:
Filed FACT-1765 per @branan 's comment above, and left a comment in the test to clarify the special case.

PR #1650, opened 2017-09-26
(FACT-1763) Merge 3.6.x -> 3.9.x and fix another acceptance test

ekinanp:


ekinanp:
Tested this locally on a redhat power and non-power machine; all the tests passed. I figured that this did not need to be tested with as many platforms as https://github.com/puppetlabs/facter/pull/1649 because the tests there showed the case by case pattern to be correct (specifically the power_linux? helper).

ekinanp:
This should be merged once the 1.10.x pipeline goes green

branan:
Acceptance changes only, so I'm merging this without waiting for Travis

PR #1656, opened 2017-10-04
(maint) Fix tests for Catch 1.10 update in Leatherman

MikaelSmith:


MikaelSmith:
Darn, this won't work without a newer catch version.

MikaelSmith:
Huh, it's only Clang that doesn't like it. Since the newer Leatherman/Catch will be in puppet-agent, I say we merge this. I should get Facter up to a newer build of Leatherman in Travis/AppVeyor CI soon as it moves to container-based builds.

MikaelSmith:
Actually, my issue was a symbol ref. Clean and rebuild fixed it.

PR #1659, opened 2017-10-09
(FACT-1774) Enable testing system openssl linked PA in CI pipelines on rhel7

jsane:
Openssl versions on RHEL7 images being used by CI/VMpooler have an old version
which causes failures when running agent setup and test scripts. They need to be
upgraded to a desired version that is >= that on Centos7 used to build PA.

Related to PA-1580 (part of epic PA-1608). Has been test in PA CI - logs would be attached to PR for PA-1264 (enables linking PA against system openssl).

This does not impact regular agent build using vendored openssl which disregard system
openssl.


jsane:
Closing this PR as this was an interim approach till we have FIPS platform support in vmpooler, CI etc. that would boot in FIPS mode which would obviate the need for these changes in install scripts.

jsane:
Whoops! closing per above note.

PR #1660, opened 2017-10-12
(FACT-1778) Attempt to canonicalize search paths

MikaelSmith:
Attempt to canonicalize search paths, then fall back to using absolute
path. This preserves the fix for FACT-1510, but also ensures we
deduplicate most paths.

This primarily fixes that in Puppet we load `lib/facter` from both
Puppet's search paths and Facter's internal handling of Ruby's LOAD_PATH.
Previously canonicalization would ensure that the resulting file paths for
custom facts were identical, so when we check whether we've previously
loaded a file we would identify duplicates and avoid loading them again.
The fix for FACT-1510 broke that. This is the least-impact approach to
fixing it.

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -174,6 +174,20 @@ extern "C" {
 
 namespace facter { namespace ruby {
 
+    static string canonicalize(string p)
+    {
+        // Get the canonical/absolute directory name
+        // If it can be resolved, use canonical to avoid duplicate search paths.
+        // Fall back to absolute because canonical on Windows won't recognize paths as valid if
+        // they resolve to symlinks to non-NTFS volumes.
+        boost::system::error_code ec;
+        auto directory = canonical(p, ec);
+        if (ec) {
+            return absolute(p).string();
+        }
+        return directory.string();
+    }
+
     map<VALUE, module*> module::_instances;
 
     module::module(collection& facts, vector<string> const& paths, bool logging_hooks) :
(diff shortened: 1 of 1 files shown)


lib/src/ruby/module.cc
```diff
@@ -953,16 +963,16 @@ namespace facter { namespace ruby {
 
         // Look for "facter" subdirectories on the load path
         for (auto const& directory : ruby.get_load_path()) {
-            // Get the absolute directory name
             boost::system::error_code ec;
-            path dir = absolute(directory);
+            auto dir = canonicalize(directory).string();
```
branan: I think we want to keep `dir` as a `path` here, and only make it a `string` down below? (line 979)

branan: Since this is now a string, you need to remove the `dir.string` below on line 979

MikaelSmith: Made a few more changes.

lib/src/ruby/module.cc
```diff
@@ -953,20 +958,19 @@ namespace facter { namespace ruby {
 
         // Look for "facter" subdirectories on the load path
         for (auto const& directory : ruby.get_load_path()) {
-            // Get the absolute directory name
             boost::system::error_code ec;
-            path dir = absolute(directory);
+            // Use forward-slash to keep this consistent with Ruby conventions.
+            auto dir = canonicalize(directory) + "/facter";
```
MikaelSmith: I can try to add a test that adds a Ruby LOAD_PATH for a fixture, then also sets it as a search path, and make sure the custom fact is only loaded once (somehow). Detecting that is the slightly tricky part.

PR #1663, opened 2017-10-19
(FACT-1781) Don't try to print boost::optionals in tests

caseywilliams:
Trying to print a `boost::optional` without its IO header was failing in tests on Solaris and AIX - this commit changes the tests so they don't attempt to print the value. It also backports a Catch 1.10 compatibility fix from master, which reappeared here when I built the agent with this branch.


branan:
:+1:

PR #1670, opened 2017-11-01
(FACT-1800) fix LIBFACTER_INSTALL_DESTINATION to respec the libdir

tampakrap:


MikaelSmith:
That makes sense. I'll amend the commit to fix it up.

PR #1673, opened 2017-11-01
(FACT-1788) More acceptance test fixups for el-7-aarch64

ScottGarman:
These changes weren't addressed in the 3.6.x merge-up to 3.9.x as the code had been significantly re-factored, so I made the necessary changes separately here.


McdonaldSeanp:
👍

PR #1674, opened 2017-11-02
(FACT-1653) Parse JSON/YAML powershell output for external facts

caseywilliams:


caseywilliams:
I still need to write acceptance tests for this change; Please don't merge until then.

lib/tests/fixtures/facts/external/windows/powershell/json.ps1
```diff
@@ -0,0 +1 @@
+Write-Output (@{ 'ps1_json_fact1'='value1'; 'PS1_JSON_fact2' = 'value2' } | ConvertTo-JSON -Depth 1 -Compress)
```
Iristyle: Just an FYI (that we might want to add to docs somewhere) - `ConvertTo-JSON` is only available in PowerShell 3 and up... and there are servers in the wild still running PowerShell 2.

Iristyle: I would suggest making the test a bit more comprehensive, to make sure types are mapped across properly from PowerShell -> JSON:

```powershell
@{ 
  'ps1_json_fact1'='value1'
  'PS1_JSON_fact2' = 2
  'PS1_JSON_fact3' = $true
  'PS1_JSON_fact4' = @('uno', 'dos')
  'PS1_JSON_fact5' = $Null
  'PS1_JSON_fact6' = @{ 'key' = 'value' }
 } | ConvertTo-JSON -Depth 1 -Compress
```

This should produce a JSON string like:

```json
{"PS1_JSON_fact6":{"key":"value"},"PS1_JSON_fact3":true,"PS1_JSON_fact4":["uno","dos"],"ps1_json_fact1":"value1","PS1_JSON_fact2":2,"PS1_JSON_fact5":null}
```

We tested something similar in the JSON -> PS mapping that's done as part of PE tasks.  I have a more comprehensive list of common types available in https://gist.github.com/Iristyle/3b37ad44c93e3e00502d799344294d93 that we vetted.

caseywilliams:
Okay, thanks @Iristyle - I've added some more complexity to the json and yaml outputs and allowed for older versions of powershell in the tests so they don't fail on older windows platforms. Acceptance tests are here now too.

lib/tests/fixtures/facts/external/windows/powershell/json.ps1
```diff
@@ -0,0 +1,22 @@
+if (Get-Command ConvertTo-JSON -errorAction SilentlyContinue)
+{
+    Write-Output(@{
```
Iristyle: I don't believe calls to `Write-Output` should be necessary here... it should happen automatically.

lib/tests/fixtures/facts/external/windows/powershell/yaml.ps1
```diff
@@ -0,0 +1,19 @@
...
+  - one
+  - two
+ps1_yaml_fact5: [first, second, third]
+ps1_yaml_fact6:
+  red: green
+  orange: blue
+  yellow: purple
+ps1_yaml_fact7: { key: value }
+'@
+Write-Output($yaml_content)
```
Iristyle: I think you can do the same here... remove the variable assignment and `Write-Output`, and I believe it should work automatically.

lib/tests/fixtures/facts/external/windows/powershell/json.ps1
```diff
@@ -0,0 +1,19 @@
...
+    @{
+        'PS1_JSON_FACT1' = 'value1'
+        'ps1_json_fact2' = 2
+        'ps1_json_fact3' = $True
+        'ps1_json_fact4' = @('first', 'second')
+        'ps1_json_fact5' = $Null
+        'ps1_json_fact6' = @{ 'a' = 'b'; 'c' = 'd' }
+    } | ConvertTo-JSON -Depth 1 -Compress)
+} else {
+'{
```
Iristyle: Huh, I didn't realize you could do multi-line string literals in PowerShell without making them part of a here-string like:

```powershell
@'
foo
bar
@'
```

Interesting!

lib/src/facts/external/windows/powershell_resolver.cc
```diff
@@ -1,18 +1,24 @@
...
 
 using namespace std;
+using namespace boost::filesystem;
 using namespace leatherman::execution;
+using namespace leatherman::util;
 using namespace leatherman::windows;
-using namespace boost::filesystem;
+using namespace facter::util::yaml;
+
+namespace lth = leatherman::util;
```
MikaelSmith: This is redundant with `using namespace leatherman::util`. A pattern used in pxp-agent is to avoid `using` statements, and do `namespace lth_util = leatherman::util;`. That makes understanding where functions come from easier. Should at least pick one approach.

acceptance/tests/external_facts/structured_powershell_facts.rb
```diff
@@ -0,0 +1,80 @@
+# This test is intended to demonstrate that powershell external facts can return
+# YAML or JSON data.
+test_name "powershell facts can return structured data" do
```
MikaelSmith: Has this been run through an ad-hoc pipeline or anything?

lib/src/facts/external/windows/powershell_resolver.cc
```diff
@@ -49,19 +56,44 @@ namespace facter { namespace facts { namespace external {
...
+            };
+
+            auto result = execute(pwrshell, ps_opts, 0, {
+                execution_options::trim_output,
+                execution_options::merge_environment,
+                execution_options::throw_on_failure
+            });
+
+            try {
+                auto node = YAML::Load(result.output);
```
MikaelSmith: We're relying on the YAML implementation to parse JSON?

acceptance/tests/external_facts/structured_powershell_facts.rb
MikaelSmith: The acceptance test seems a little redundant, the unit tests cover the same behavior.

caseywilliams: FACT-1653 asked for acceptance tests as part of its acceptance criteria, but I think I agree - let's see what @branan thinks - this did pass in an ad-hoc run a while ago.

lib/src/facts/external/windows/powershell_resolver.cc
caseywilliams: Right - this is pretty much a copy-and-paste from the execution resolver, which does the same thing for other types of executables

caseywilliams:
Thanks for the review @MikaelSmith! I opted to rely on `using` instead of namespace aliases here since the rest of the project seems to do that. Also starting a second ad-hoc build; I've made a few edits since the last one.

lib/src/facts/external/windows/powershell_resolver.cc
branan: For the case of static facts when we know the file type, it makes sense to use a fast JSON parser when we can. When we're already testing multiple formats and doing fallback, adding a third parser into the mix seemed like the wrong tradeoff. That was my thinking when I did the initial YAML/JSON for executable facts, at any rate

acceptance/tests/external_facts/structured_powershell_facts.rb
branan: I think I'm in agreement that we want to just test this at the "unit"\* level - the acceptance test seems redundant to me.

* It's really an integration test, but whatever :)

lib/src/facts/external/windows/powershell_resolver.cc
MikaelSmith: That seems fine.

caseywilliams:
Ok, acceptance tests removed!

PR #1677, opened 2017-12-05
(FACT-1802) Add support for fips fact

jsane:
As part of larger FIPS effort, we are adding a new fips fact. It is made available as a single bool (in_fips_mode) to indicate whether the system is running in fips mode or not.
This would be used in puppet acceptance tests to adjust behaviors of some tests to either skip them or replace any md5 with fips approved hash algs like sha256.


lib/inc/internal/facts/linux/fips_resolver.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include "../resolvers/fips_resolver.hpp"
+
+namespace facter { namespace facts { namespace linux {
+
+    /**
+     * Responsible for resolving fips-related facts.
+     */
+    struct fips_resolver : resolvers::fips_resolver
+    {
+     // protected:
```
caseywilliams: This `protected` needs uncommenting (and the comment below can be indented a tad more)

lib/inc/internal/facts/resolvers/fips_resolver.hpp
```diff
@@ -0,0 +1,46 @@
...
+        };
+
+        /**
+         * Called to resolve all facts the resolver is responsible for.
+         * @param facts The fact collection that is resolving facts.
+         */
+        virtual void resolve(collection& facts) override;
+
+        /**
+ 	 * REMIND: Do we need this???
```
caseywilliams: Yes, this should probably stay here

caseywilliams: Ah, and this method and the data struct should probably also be under a `protected:` line - most other resolvers tend to set the constructor and `resolve` method to public and default to protected for other members.

lib/src/facts/linux/fips_resolver.cc
```diff
@@ -0,0 +1,35 @@
...
+
+namespace facter { namespace facts { namespace linux {
+
+    fips_resolver::data fips_resolver::collect_data(collection& facts)
+    {
+        data result;
+
+        // Below file might not exist in which case we need to ensure to not report
+        // anything incorrectly by initializing the result to a safe value
+        result.is_fips_mode_enabled = false;
```
ekinanp: Might be better to specify a way in the data to check whether it is valid or not before adding the fact? I am not familiar with the FIPS mode, but as a user if I read "is_fips_mode_enabled = false" I might be lead to think that FIPS mode is not enabled when it in fact might be, but isn't reported in the /proc/sys/crypto/fips_enabled.

lib/src/facts/linux/fips_resolver.cc
```diff
@@ -0,0 +1,35 @@
...
+
+        lth_file::each_line("/proc/sys/crypto/fips_enabled", [&](string& line) {
```
ekinanp: You could simplify the entire code of this routine to:

result.is_fips_mode_enabled = boost::trim(lth_file::read("/proc/sys/crypto/fips_enabled")) != "0"

Note that if the file does not exist, lth_file::read would return the empty string (per https://github.com/puppetlabs/leatherman/blob/master/file_util/src/file.cc#L37) so your original default would still work.

lib/src/facts/resolvers/fips_resolver.cc
```diff
@@ -0,0 +1,23 @@
...
+
+    fips_resolver::fips_resolver() :
+        resolver("fips", {fact::in_fips_mode})
+    {
+    }
+
+    void fips_resolver::resolve(collection& facts)
+    {
+        auto data = collect_data(facts);
+        facts.add(fact::in_fips_mode, make_value<boolean_value>(data.is_fips_mode_enabled));
```
ekinanp: Follow-up to my above comment, this is where the validation check would go, where the fact would only be added if the data is valid.

lib/schema/facter.yaml
```diff
@@ -369,6 +369,14 @@ filesystems:
         Mac OSX: The usable file systems is limited to the file system of mounted devices.
     blockgroup: file system
 
+in_fips_mode:
```
joshcooper: Is it always boolean or can there be a "learning" mode like SELinux?

If boolean, what do you think about naming it `fips_enabled`?

lib/src/facts/linux/fips_resolver.cc
jsane: Agreeably there was a bit of redundancy particularly after I added the initialization but forgot to removing doing the same down below if fips_enabled read "0"

lib/src/facts/linux/fips_resolver.cc
jsane: Hmm... not sure. Either absence or a false value should be synonymous but don't mind not reporting the fact if not enabled.

lib/schema/facter.yaml
jsane: From what I know it's always boolean. We could name it as but it might have different connotations compared to actually being in fips mode. But no strong opinions.

lib/inc/internal/facts/resolvers/fips_resolver.hpp
jsane: k..

jsane:
Since after this PR learned that FIPS work needs to target 1.10.x so I would be submitting a PR against 3.6.x branch after incorporating above comments.

lib/src/facts/resolvers/fips_resolver.cc
jsane: Also here the way the tests are wired, if we skip adding the fact altogether there seems to be a check that tries to validate the number of facts expected versus actually found and that starts failing. I did not investigate it further in case it is possible to suppress that and have retained the above implementation to always return the fact.

lib/src/facts/linux/fips_resolver.cc
jsane: Actually the above generates compile error perhaps becoz trim expects a reference string type as against a string object returned by read. It could be massaged but am planning to use the earlier check. Will remove the redundant assignment though in case fips is not enabled.

PR #1678, opened 2017-12-08
(FACT-1377) Include key type in ssh fact

danielparks:
Previously the ssh fact returned host keys by their general type, e.g. `ecdsa`, but did not include information about their real type, e.g. `ecdsa-sha2-nistp256`.

This adds a `type` attribute for each host key containing the real type.


danielparks:
Ping @branan; we chatted about this yesterday.

danielparks:
Seems reasonable.

danielparks:
Actually, I'd rather not, since the fact would fail if a new key type is introduced.

PR #1679, opened 2017-12-13
 (FACT-1802) Add support for FIPS fact

jsane:
The fact is made available as a single bool (fips_enabled) to indicate
whether the system is in FIPS mode or not.


lib/inc/internal/facts/resolvers/fips_resolver.hpp
```diff
@@ -0,0 +1,46 @@
...
+             * Stores the is_fips_mode_enabled data.
+             */
+            bool is_fips_mode_enabled;
+        };
+
+        /**
+         * Called to resolve all facts the resolver is responsible for.
+         * @param facts The fact collection that is resolving facts.
+         */
+        virtual void resolve(collection& facts) override;
```
ekinanp: This does not need to be a protected member -- can make it public. See e.g. https://github.com/puppetlabs/facter/blob/master/lib/inc/internal/facts/resolvers/processor_resolver.hpp#L28

lib/src/facts/linux/fips_resolver.cc
```diff
@@ -0,0 +1,32 @@
...
+    {
+        data result;
+
+        // Set a safe default
+        result.is_fips_mode_enabled = false;
+
+        lth_file::each_line("/proc/sys/crypto/fips_enabled", [&](string& line) {
+            boost::trim(line);
+            if (line != "0")
+                result.is_fips_mode_enabled = true;
```
ekinanp: Can just do "results.is_fips_mode_enabled = line != 0" here.

lib/inc/internal/facts/resolvers/fips_resolver.hpp
jsane: yikes.

jsane:
Can this be merged? It is required for the other fips PRs I have on puppet. thanks.

ekinanp:
@branan ^

jsane:
Hi @branan 
Pl let me know if there are any concerns/questions.

joshcooper:
ping @branan

PR #1681, opened 2018-01-04
(maint) Fix facter tests to run on EC2 hosts

mchllweeks:
Fixed the base_fact_utils.rb and dmi.rb to run on ec2 hosts


ScottGarman:
:+1:

PR #1684, opened 2018-01-18
(maint) Fix flaky cached fact tests

caseywilliams:
A few cached fact tests relied on the uptime, which can be unreliable on
some windows systems (see FACT-1504) - This changes those tests to use timezone instead,
which doesn't have the same problems.

### DIFF
--- a/acceptance/tests/options/config_file/ttls_cached_facts_creates_json_cache_file.rb
+++ b/acceptance/tests/options/config_file/ttls_cached_facts_creates_json_cache_file.rb
@@ -8,7 +8,7 @@ test_name "C99973: ttls configured cached facts create json cache files" do
 
   # This fact must be resolvable on ALL platforms
   # Do NOT use the 'kernel' fact as it is used to configure the tests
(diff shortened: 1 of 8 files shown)


caseywilliams:
Travis is persistently failing to download dependencies - this PR only contains acceptance test changes, though

caseywilliams:
jenkins please test this on windows2008-64

caseywilliams:
Hmm, acceptance tests failed on an incorrect `os.release.full` here; My only guess is it might be the result of recent updates to windows VM templates - I don't recognize that failure from any recent CI runs, though. I'll do some digging.

caseywilliams:
jenkins please test this on windows2008r2-64

caseywilliams:
Ok, that makes sense - the tests are meant to be run on 2008 R2 and not regular 2008 🤦‍♀️

caseywilliams:
jenkins please test this on windows2008r2-64

caseywilliams:
Closing this in favor of #1687, which actually fixes the problem at the source

PR #1687, opened 2018-01-19
(FACT-1504) Use GetTickCount64 as an uptime metric for windows platof…

jsane:
…orms

The earlier metric was unreliable and also stopped working altogether
(uptime fact on windows comes as empty) causing CI failures.
We now use GetTickCount64 instead which offers better resolution
and is available on all supported windows platforms.

Let know if there is any preference for value in seconds or just ticks (msecs). Also if this is not a priority right now this can wait but it appeared to be causing some failures.

### DIFF
--- a/lib/inc/internal/facts/windows/uptime_resolver.hpp
+++ b/lib/inc/internal/facts/windows/uptime_resolver.hpp
@@ -18,9 +18,8 @@ namespace facter { namespace facts { namespace windows {
     {
         /**
          * Constructs the uptime_resolver.
-         * @param wmi_conn The WMI connection to use when resolving facts.
          */
-        uptime_resolver(std::shared_ptr<leatherman::windows::wmi> wmi_conn = std::make_shared<leatherman::windows::wmi>());
+        uptime_resolver();
 
      protected:
         /**
@@ -28,9 +27,6 @@ namespace facter { namespace facts { namespace windows {
          * @return Returns the system uptime in seconds.
          */
         virtual int64_t get_uptime() override;
-
-     private:
-        std::shared_ptr<leatherman::windows::wmi> _wmi;
     };
 
 }}}  // namespace facter::facts::windows
--- a/lib/src/facts/windows/collection.cc
+++ b/lib/src/facts/windows/collection.cc
@@ -75,6 +75,7 @@ namespace facter { namespace facts {
         add(make_shared<windows::memory_resolver>());
         add(make_shared<windows::networking_resolver>());
         add(make_shared<windows::timezone_resolver>());
+        add(make_shared<windows::uptime_resolver>());
 
         try {
             shared_ptr<wmi> shared_wmi = make_shared<wmi>();
@@ -82,7 +83,6 @@ namespace facter { namespace facts {
             add(make_shared<windows::operating_system_resolver>(shared_wmi));
             add(make_shared<windows::processor_resolver>(shared_wmi));
             add(make_shared<windows::virtualization_resolver>(shared_wmi));
-            add(make_shared<windows::uptime_resolver>(shared_wmi));
         } catch (wmi_exception &e) {
             LOG_ERROR("failed adding platform facts that require WMI: {1}", e.what());
         }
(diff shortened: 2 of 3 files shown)


caseywilliams:
jenkins please test this on windows2008r2-64

lib/src/facts/windows/uptime_resolver.cc
```diff
@@ -23,28 +24,10 @@ namespace facter { namespace facts { namespace windows {
...
     {
-        auto vals = _wmi->query(wmi::operatingsystem, {wmi::lastbootuptime, wmi::localdatetime});
-        if (vals.empty()) {
-            return -1;
-        }
-
-        ptime boottime = get_ptime(wmi::get(vals, wmi::lastbootuptime));
-        ptime now = get_ptime(wmi::get(vals, wmi::localdatetime));
-        return (now - boottime).total_seconds();
+        uint64_t tickCount = GetTickCount64();
```
MikaelSmith: This wasn't available in Windows Server 2003. Yay for using it now. 👍

lib/src/facts/windows/uptime_resolver.cc
```diff
@@ -1,4 +1,5 @@
 #include <internal/facts/windows/uptime_resolver.hpp>
+#include <leatherman/windows/windows.hpp>
 #include <leatherman/windows/wmi.hpp>
```
MikaelSmith: There's some larger cleanup that could happen here: almost all the includes could be removed, wmi argument to uptime_resolver can be dropped.

lib/inc/internal/facts/windows/uptime_resolver.hpp
```diff
@@ -30,7 +29,7 @@ namespace facter { namespace facts { namespace windows {
         virtual int64_t get_uptime() override;
 
      private:
-        std::shared_ptr<leatherman::windows::wmi> _wmi;
+        // std::shared_ptr<leatherman::windows::wmi> _wmi;
```
MikaelSmith: Would rather remove this than comment it out.

lib/src/facts/windows/collection.cc
```diff
@@ -82,7 +82,7 @@ namespace facter { namespace facts {
             add(make_shared<windows::operating_system_resolver>(shared_wmi));
             add(make_shared<windows::processor_resolver>(shared_wmi));
             add(make_shared<windows::virtualization_resolver>(shared_wmi));
-            add(make_shared<windows::uptime_resolver>(shared_wmi));
+            add(make_shared<windows::uptime_resolver>());
```
MikaelSmith: Let's move this up to line 77 so it's still resolved if WMI is unavailable.

lib/inc/internal/facts/windows/uptime_resolver.hpp
jsane: Yes...I actually wanted to get it thru one quick CI run (primarily to catch any compile issues) before deleting things.

lib/src/facts/windows/collection.cc
jsane: Good pt on the above - will do.

jsane:
@caseywilliams yes it was run on windows 2008 r2-64 (https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/58/)

jsane:
ok latest update passed CI on windows2008 - job #60 (redhat and ubuntu are still chugging but are n/a for this change). So should be good to merge

lib/src/facts/windows/uptime_resolver.cc
Iristyle: 2 ^ 64 = 1.844674407370955e19 converted to seconds by dividing by 1000 = 1.844674407370955e16
2 ^ 63 = 9.223372036854776e18

Therefore count in seconds does not exceed range of allowed values in `int64_t` (2 ^63) ... so the cast should work fine here.

MikaelSmith:
Anything blocking merging this @caseywilliams?

caseywilliams:
Hmm, I'm ready to merge this, but come to think of it - is there a reason not to target this at 3.6.x instead of 3.9.x?

MikaelSmith:
I'm not aware of one.

caseywilliams:
Ok, neither am I - @jsane would you mind retargeting this at 3.6.x? We can run one more test and then I'll merge this.

jsane:
Sorry I was out for a while and just catching up on things. I have re submitted it for 3.6.x

speedofdark:
Merged into 3.6.x as mentioned above.

PR #1692, opened 2018-01-25
(FACT-1810) Fix open timeout for fetching ec2 metadata

jagij:
The open_timeout wasn't specified, which means that it spent 60s
for each call to open (9 times in total). With this extra argument,
the run duration for facter goes down from over 9 minutes to
several seconds.

A regular 'curl 169.254.169.254' also hangs for a long time on
those machines, unless the --connect-timeout switch is specified.

### DIFF
--- a/lib/facter/ec2/rest.rb
+++ b/lib/facter/ec2/rest.rb
@@ -25,7 +25,7 @@ module Facter
         attempts = 0
 
         begin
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Thanks for following up on this!

thde:
There seems to be no release for this fix until now. It would be great to get it via rubygems.org! 😄

PR #1693, opened 2018-01-31
(FACT 1797) Fix Amazon EC2 c5 instance virtualization facts

caseywilliams:
Amazon's more recent c5 instance types report custom SMBIOS values which weren't being detected as virtual. These commits add WMI and DMI checks for those values and reports KVM as their hypervisor (Amazon's hypervisor for c5 is based on KVM). This allows the `ec2_metadata` fact to be resolved, too.


caseywilliams:
I have a test to fix on windows, marking don't merge for now.

ScottGarman:
👍

PR #1694, opened 2018-02-02
(FACT-1797) Fix virtualization resolvers for solaris, openbsd

caseywilliams:
I broke the solaris and openbsd virtualization resolvers when I reorganized the base resolver to detect amazon ec2 c5 instances. This updates them so that they use the updated detection function.

### DIFF
--- a/lib/src/facts/solaris/virtualization_resolver.cc
+++ b/lib/src/facts/solaris/virtualization_resolver.cc
@@ -32,23 +32,7 @@ namespace facter { namespace facts { namespace solaris {
             return vm::zone;
         }
 
-        string guest_of;
-
-        // Use the same timeout as in Facter 2.x
-        const uint32_t timeout = 20;
-        try {
-            each_line(
-                "/usr/sbin/prtdiag",
-                [&](string& line) {
-                    guest_of = get_product_name_vm(line);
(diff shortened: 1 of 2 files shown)


caseywilliams:
Jenkins please test this on solaris10-64,solaris10-SPARC,solaris11-64,solaris11-SPARC

lib/src/facts/solaris/virtualization_resolver.cc
```diff
@@ -32,23 +32,7 @@ namespace facter { namespace facts { namespace solaris {
             return vm::zone;
         }
 
-        string guest_of;
-
-        // Use the same timeout as in Facter 2.x
-        const uint32_t timeout = 20;
-        try {
-            each_line(
-                "/usr/sbin/prtdiag",
```
caseywilliams: The solaris [dmi resolver](https://github.com/puppetlabs/facter/blob/3.6.x/lib/src/facts/solaris/dmi_resolver.cc#L46) also collects prtdiag information, including product name, so I think it's okay to remove this.

branan: :+1: to not calling out to the same command twice. Good catch!

PR #1698, opened 2018-02-18
(FACT-1772) Use CMAKE_INSTALL_LIBDIR for multilib install

ekinanp:
This allows the libfacter library to be installed in e.g.
/usr/lib, /usr/lib64, etc. depending on the platform that it is
being built on.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -251,8 +251,7 @@ endif()
 if (WIN32)
     set(LIBFACTER_INSTALL_DESTINATION bin)
 else()
-    # TODO: lib64 for certain operating systems?
-    set(LIBFACTER_INSTALL_DESTINATION lib)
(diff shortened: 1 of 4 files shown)


lib/CMakeLists.txt
```diff
@@ -251,8 +251,7 @@ endif()
 if (WIN32)
     set(LIBFACTER_INSTALL_DESTINATION bin)
 else()
-    # TODO: lib64 for certain operating systems?
-    set(LIBFACTER_INSTALL_DESTINATION lib)
+    set(LIBFACTER_INSTALL_DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
MikaelSmith: facter.rb.in may also need to be updated for this. At least, Travis builds are failing related to finding libfacter.so

ekinanp:
https://github.com/puppetlabs/leatherman/pull/273 (that's why CI's failing)

PR #1701, opened 2018-02-21
(FACT-1504) Use GetTickCount64 as an uptime metric for windows platof…

jsane:
…orms

The earlier metric was unreliable and also stopped working altogether
(uptime fact on windows comes as empty) causing CI failures.
We now use GetTickCount64 instead which offers better resolution
and is available on all supported windows platforms.


ekinanp:
jenkins please test this on windows2012r2-64a

ekinanp:
PR testing passed, so I think this should be good to go.

PR #1703, opened 2018-03-01
[FACT-1822] Add support for EC2 instances based on the Nitro Hypervisor

szhuga:
The new Amazon C5 instances are based on the Nitro Hypervisor: https://aws.amazon.com/about-aws/whats-new/2018/01/amazon-ec2-c5-instances-are-now-available-in-us-west--n--california--and-europe--london/

So here is the patch to add support of that instances to the Facter 2.x.

There is another pull-request for this issue (#1700) but it does not cover the Windows instances.


sarajarjoura:
@puppetcla  Can this please get merged as our instances are affected by this patch and we would love to be able to upgrade to a working version?

PR #1709, opened 2018-03-30
(FACT-1816) Run acceptance setup from beaker-puppet

melissa:
This includes removing all the set up logic from facter. This is now
stored in beaker-puppet.


MikaelSmith:
Has the necessary beaker-puppet version been released?

melissa:
Not yet, no. Though this should still work with the latest beaker-puppet release. The new one should be cut early next week, depending on when my PR is merged

melissa:
jenkins please test this

melissa:
@MikaelSmith the beaker-puppet release is now out, and this change has been updated to hit it. I've kicked off pr testing, which hopefully 'does the right thing'

melissa:
jenkins please test this

acceptance/config/aio/options.rb
```diff
@@ -1,9 +1,3 @@
 {
   :type => 'aio',
```
MikaelSmith: Are these really even needed? The type appears to be implied by the file path, which the beaker-puppet changes appear to know how to find.

acceptance/config/aio/options.rb
```diff
@@ -1,9 +1,3 @@
...
-  :pre_suite => [
-    'setup/common/pre-suite/000-delete-puppet-when-none.rb',
-    'setup/aio/pre-suite/010_Install.rb',
-    'setup/aio/pre-suite/021_InstallAristaModule.rb',
-    'setup/aio/pre-suite/022_Remove_LD_PRELOAD.rb',
```
joshcooper: I'm not seeing this step in beaker-puppet. Should we add it there?

melissa: @MikaelSmith and I decided we probably don't need it anymore, but I'm happy to add it to beaker-puppet if that was misguided

acceptance/config/aio/options.rb
joshcooper: They are mostly not needed, since the CI type above is typically the same as the rake task name, e.g. `ci:test:aio`. However, that's not true for the `ci:aio:gem` task (in puppet and eventually in facter), as it relies on the `git` CI type. I think having it explicit makes it cleaner, though more verbose.

acceptance/config/aio/options.rb
joshcooper: 👍 this passes on `ciscoxr-64a`

joshcooper:
jenkins please test this on aix61-POWERf,aix71-POWERf,aix72-POWERf,redhat7-POWER,redhat7-AARCH64,sles12-POWER,solaris10-SPARC,solaris11-SPARC,ubuntu1604-POWER,ciscoxr-64,amazon6-64,amazon7-64 with agent 1.10.x

PR #1714, opened 2018-04-14
(FACT-1844) JDK10 drops support for javah

mlevitt:
Any JDK post 9 requires change of invocation from

   javah -d <dir>
to
   javac -h <dir>

The arguments change too. This PR allows the headers to be built on JDK 9 and 10.


mlevitt:
The CLA click through doesn't seem to have sent the email though I've clicked resend and even tried other email accounts. I believe this is exempt as it has only:
 - Changes in build or installation scripts

lib/CMakeLists.txt
```diff
@@ -283,8 +283,18 @@ if (JRUBY_SUPPORT)
 
     # javah does not atomically write the header file, so parallel builds can
     # read it before it finishes writing if not careful.
+    # JDK versions after 9 don't provide javah. Use javac in these cases
+
+    if(Java_VERSION VERSION_LESS "10")
+	    set(this_JAVAH_COMMAND javah)
```
branan: I'm not really a fan of the `this_` style here - we don't use it anywhere else in the CMake code.

mlevitt:
Done!  ... and a rebase against current master for good measure.

PR #1716, opened 2018-05-04
(FACT-1851) Gather DHCP lease data under systemd-networkd

caseywilliams:
Previously, the BSD/Linux networking resolver expected to find DHCP lease
information only in files created by dhclient. This change checks the
systemd-networkd DHCP lease files as well.

Running applicable acceptance tests for this change in PR testing is blocked on updates to beaker-puppet for Ubuntu 18.04, which we expect to be merged later today.

### DIFF
--- a/lib/src/facts/bsd/networking_resolver.cc
+++ b/lib/src/facts/bsd/networking_resolver.cc
@@ -1,15 +1,21 @@
+#include <facter/util/string.hpp>
 #include <internal/facts/bsd/networking_resolver.hpp>
 #include <internal/util/bsd/scoped_ifaddrs.hpp>
 #include <leatherman/execution/execution.hpp>
 #include <leatherman/file_util/file.hpp>
 #include <leatherman/file_util/directory.hpp>
 #include <leatherman/logging/logging.hpp>
+#include <leatherman/util/regex.hpp>
 #include <boost/algorithm/string.hpp>
+#include <boost/filesystem.hpp>
 #include <netinet/in.h>
+#include <unordered_map>
 
 using namespace std;
 using namespace facter::util::bsd;
 using namespace leatherman::execution;
+using namespace leatherman::util;
+using facter::util::maybe_stoi;
 
 namespace lth_file = leatherman::file_util;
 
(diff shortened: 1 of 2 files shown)


lib/src/facts/bsd/networking_resolver.cc
```diff
@@ -153,6 +155,56 @@ namespace facter { namespace facts { namespace bsd {
                 return true;
             }, "^dhclient.*lease.*$");
         }
+    }
+
+    void networking_resolver::find_networkd_dhcp_servers(std::map<std::string, std::string>& servers) const
+    {
+        static const string networkd_lease_directory = "/run/systemd/netif/leases/";
```
caseywilliams: Files in this directory look something like this:
```
root@k1tyuo9fuykflek:~# cat /run/systemd/netif/leases/2 
# This is private data. Do not parse.
ADDRESS=10.32.119.212
NETMASK=255.255.240.0
ROUTER=10.32.112.1
SERVER_ADDRESS=10.32.22.9
NEXT_SERVER=10.28.22.13
T1=450
T2=787
LIFETIME=900
DNS=10.240.0.10 10.240.1.10
NTP=10.32.22.9 10.0.22.10 10.32.22.10 10.0.22.11
DOMAINNAME=delivery.puppetlabs.net
```
The "do not parse" message is ominous, but I wasn't sure where else to find this information. At least on Ubuntu 18.04, these files are readable by non-root users.

branan: That is, indeed, ominous.

I'd hope they expose this information /somewhere/ as a stable API, but who knows.

caseywilliams:
jenkins please test this on ubuntu1804-64

lib/src/facts/bsd/networking_resolver.cc
```diff
@@ -153,6 +156,56 @@ namespace facter { namespace facts { namespace bsd {
...
+            return true;
+        });
+    }
+
+    map<string, string> networking_resolver::find_dhcp_servers() const
+    {
+        map<string, string> servers;
+
+        find_networkd_dhcp_servers(servers);
+        find_dhclient_dhcp_servers(servers);
```
branan: I think instead of trying to do both, we should figure out which file exists and explicitly choose the more-correct one.

lib/src/facts/bsd/networking_resolver.cc
branan: Can you add a comment here with the info that this isn't public API, but its the best we can find

caseywilliams:
jenkins please test this on ubuntu1804-64

caseywilliams:
jenkins please test this on ubuntu1804-64a

PR #1720, opened 2018-07-09
(FACT-1832) Be smarter with how we use kstat on Solaris

branan:
Previously, there were a couple of issues with our kstat usage:

1. We tried to call `kstat_read` before confirming that an entry was
   one we want
2. We used `kstat_lookup` to find our first entry, but then just
   traversed the entire list from there anyway.

Since we were still manually traversing the list to find additional
entries, we ended up effectively re-implementing `kstat_lookup`
anyway.

This updated version removes the use of `kstat_lookup`, relying
instead entirely on our own list traversal. It also moves the
`kstat_read` call to be conditional on whether the entry looks like
one we want, preventing us from reading unnecessary (and perhaps
unreadable) kstat entries.

### DIFF
--- a/lib/src/util/solaris/k_stat.cc
+++ b/lib/src/util/solaris/k_stat.cc
@@ -30,44 +30,36 @@ namespace facter { namespace util { namespace solaris {
         return lookup(entry.first, -1, entry.second);
     }
 
+// I believe precedence to be sufficiently obvious in this function,
+// but this is a worthwhile diagnostic in most cases. We disable it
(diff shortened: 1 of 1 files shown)


branan:
jenkins please test this on solaris11-32a,solaris11-SPARCa

branan:
jenkins please test this on solaris11-32a,solaris11-SPARCa

branan:
jenkins please test this on solaris11-32a,solaris11-SPARCa

PR #1724, opened 2018-07-19
(FACT-1477) Also check for config. file when resolving selinux.enabled

ekinanp:
Previously, we'd determine if selinux was enabled on a given system by
checking if the selinux filesystem is mounted. This is incorrect. Based
on https://github.com/SELinuxProject/selinux/blob/e93899c8f3c01dfb47e81d3da4b67283fb459f20/libselinux/src/enabled.c#L20, we also need to check if the selinux config. file
/etc/selinux/config also exists. We can have a scenario where the selinux
filesystem is mounted, but the /etc/selinux/config file does not exist. This
is what was happening to the customer, presumably because their kickstart file
was installing some selinux stuff under yum-utils (while our machines weren't).
Here, Facter would erroneously report that selinux is enabled.

This commit fixes our selinux enablement check by also checking for the
existence of the selinux config. file.

### DIFF
--- a/lib/src/facts/linux/operating_system_resolver.cc
+++ b/lib/src/facts/linux/operating_system_resolver.cc
@@ -61,11 +61,13 @@ namespace facter { namespace facts { namespace linux {
 
     operating_system_resolver::selinux_data operating_system_resolver::collect_selinux_data()
     {
+        static string SELINUX_CONFIG_FILE("/etc/selinux/config");
+
         selinux_data result;
         result.supported = true;
 
         string mountpoint = get_selinux_mountpoint();
-        result.enabled = !mountpoint.empty();
+        result.enabled = !mountpoint.empty() && exists(SELINUX_CONFIG_FILE);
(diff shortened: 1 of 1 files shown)


ekinanp:
jenkins please test this on redhat7-64

ekinanp:
jenkins, please test this on redhat7-64

ekinanp:
jenkins please test this on redhat7-64

ekinanp:
Ok, this one's ready for merge. Also tested it out on the following scenarios:

1. IF selinux is enabled, does it return the right info? => Yes

2. IF selinux is disabled on our VMs, does it return the right info? => Yes

Existing behavior is preserved. Finally, I tested it on a VM I built using the customer's kickstart file linked in the support ticket in one of the comments for this ticket. The fix here correctly reports that selinux is disabled on their system.

image

^ Screenshot showing the customer's system and the missing config. file and the fact being correctly reported as false.

PR #1725, opened 2018-07-21
(FACT-1475) Clear out bogus IPV4 data for IPV6 query

ekinanp:
On some platforms, in particular CentOS 5 and 6, facter erroneously
outputs ipv4 values for ipv6 facts when ipv6 is disabled. Specifically,
it outputs an ipv4 address for the ip6 fact. The reason it does this is
because the command `ip -6 route show` returns ipv4 data. Facter uses
this command to figure out the primary network interface and to also
provide at least the network interface's ip address when getifaddr does
not return any bindings for it. Facter fills in the ip6 fact when ipv6
is disabled for exactly this reason -- getifaddr does not return any
ipv6 bindings, but `ip -6 route show` outputs a bogus ipv4 address that
gets filled in instead.

This commit detects when the ip command outputs ipv4 information for an
ipv6 query, correctly erasing all currently collected ipv6 route data when
this occurs.

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -148,7 +148,7 @@ namespace facter { namespace facts { namespace linux {
             "throw"
         };
 
-        auto parse_route_line = [&known_route_types](string& line, std::vector<route>& routes) {
+        auto parse_route_line = [&known_route_types](string& line, int family, std::vector<route>& routes) {
             vector<boost::iterator_range<string::iterator>> parts;
             boost::split(parts, line, boost::is_space(), boost::token_compress_on);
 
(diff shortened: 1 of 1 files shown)


ekinanp:
I've tested this on CentOS 5 and 6 and have verified that the ipv6 facts no longer show up when ipv6 is disabled. Also verified that they do show up when enabled. If you'd like to see some screenshots, I can create another branch with my tests showing the customer module is applied, that I rebooted the machine to get the changes, then ran Facter.

ekinanp:
jenkins please test this on aix71-POWERf,redhat7-64,redhat6-64,solaris11-64,sles12-64,debian8-64

ekinanp:
jenkins please test this on aix71-POWERfa,redhat7-64a,redhat6-64a,solaris11-64a,sles12-64a,debian8-64a

PR #1726, opened 2018-07-30
(FACT-1550) Get os.architecture from processors.models[0] on AIX

ekinanp:
On AIX, we can dynamically add and remove resources from the LPAR.
This includes removing proc0, which is what we originally checked
to obtain the OS architecture. In this case, we will report an
empty value for the OS architecture, which is incorrect.

This commit uses the processors.model[0] fact to obtain the
architecture instead, where the processors fact directly queries the
ODM for all of the available processors on the system and thus is not
affected by any LPAR configuration.

We also clean up the code here a bit by using our internal ODM APIs
instead of shelling out to lsattr and parsing its output to get the
hardware info. It is better to use our internal ODM APIs instead to
keep things consistent with our other AIX fact resolvers and to
avoid shelling out when we don't need to. The resulting code's
also a lot cleaner.

### DIFF
--- a/lib/src/facts/aix/operating_system_resolver.cc
+++ b/lib/src/facts/aix/operating_system_resolver.cc
@@ -1,43 +1,72 @@
 #include <internal/facts/aix/operating_system_resolver.hpp>
+#include <internal/util/aix/odm.hpp>
+#include <facter/facts/collection.hpp>
+#include <facter/facts/fact.hpp>
 #include <facter/facts/os.hpp>
-#include <leatherman/execution/execution.hpp>
-#include <leatherman/file_util/file.hpp>
+#include <facter/facts/array_value.hpp>
+#include <facter/facts/map_value.hpp>
+#include <facter/facts/scalar_value.hpp>
 #include <leatherman/logging/logging.hpp>
-#include <leatherman/util/regex.hpp>
 
 #include <boost/algorithm/string.hpp>
+#include <odmi.h>
+#include <sys/cfgodm.h>
 
 using namespace std;
-using namespace leatherman::util;
-using namespace leatherman::file_util;
-using namespace boost;
-namespace execution = leatherman::execution;
+using namespace facter::util::aix;
 
+// This routine's meant to be a general utility function that replicates the behavior of
+// lsattr -El <object> -a <field>. Although it's only used to get the modelname of the
+// sys0 device, we still would like to have it here in case we ever need to separate it
+// out to an AIX utils file to query other attributes. Part of what lsattr does is check
+// PdAt if we don't have an entry for the object's attribute in CuAt, even though it is very
+// unlikely that sys0.modelname will not have a CuAt entry. That is why we have the extra code
+// in here.
 static string getattr(string object, string field)
 {
-    string result;
+    // High-level logic here is:
+    //   * Check if there's an entry for our object's field attribute in CuAt (the device-specific
+    //   attribute entry).
+    //
+    //   * Else, check for the field attribute's default value in PdAt. We do this by first
+    //   figuring out the PdDv type from the CuDv entry for the object, then use our PdDv type
+    //   to query the field's default value in PdAt.
+    string query = (boost::format("name = %1% AND attribute = %2%") % object % field).str();
+    auto cuat_query = odm_class<CuAt>::open("CuAt").query(query);
 
-    execution::each_line(
-        "/usr/sbin/lsattr", {"-El", object, "-a", field},
-        [&](string& line) {
-            if (!line.empty()) {
-                vector<string> tokens;
-                boost::split(tokens, line, boost::is_space(), boost::token_compress_on);
-                if (tokens.size() < 2) {
-                    return true;
-                }
-                result = tokens[1];
-                return false;
-            }
-            return true;
-        },
-        nullptr,
-        0);
-
-    if (result == "") {
-        LOG_WARNING("Could not get a value from lsattr -El {1} -a {2}", object, field);
+    // This is a more verbose way of saying that we only expect our query to have one element
+    auto cuat_ref = cuat_query.begin();
+    if (cuat_ref != cuat_query.end()) {
+      auto value = string(cuat_ref->value);
+      if (value.empty()) {
+        LOG_DEBUG("Could not get a value from the ODM for {1}'s '{2}' attribute.", object, field);
+      }
+      return value;
     }
-    return result;
+
+    // Get the PdDv type from the CuDv entry
+    query = (boost::format("name = %1%") % object).str();
+    auto cudv_query = odm_class<CuDv>::open("CuDv").query(query);
+    auto cudv_ref = cudv_query.begin();
+    if (cudv_ref == cudv_query.end()) {
+      LOG_DEBUG("Could not get a value from the ODM for {1}'s '{2}' attribute: There is no CuDv entry for {1}.", object, field);
+      return "";
+    }
+    auto pddv_type = cudv_ref->PdDvLn_Lvalue;
+
+    query = (boost::format("uniquetype = %1% AND attribute = %2%") % pddv_type % field).str();
+    auto pdat_query = odm_class<PdAt>::open("PdAt").query(query);
+    auto pdat_ref = pdat_query.begin();
+    if (pdat_ref != pdat_query.end()) {
+      auto value = string(pdat_ref->deflt);
+      if (value.empty()) {
+        LOG_DEBUG("Could not get a value from the ODM for {1}'s '{2}' attribute.", object, field);
+      }
+      return value;
+    }
+
+    LOG_DEBUG("Could not get a value from the ODM for {1}'s '{2}' attribute: There is no PdAt entry for {1} with {2}.", object, field);
+    return "";
 }
 
 namespace facter { namespace facts { namespace aix {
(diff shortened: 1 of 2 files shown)


ekinanp:
jenkins please test this on aix71-POWERfa

ekinanp:
jenkins please test this on aix71-POWERfa

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,31 @@
...
-            }
-            return true;
-        },
-        nullptr,
-        0);
-
-    if (result == "") {
-        LOG_WARNING("Could not get a value from lsattr -El {1} -a {2}", object, field);
+    string query = (boost::format("name = %1% AND attribute = %2%") % object % field).str();
+    auto attr_query = odm_class<CuAt>::open("CuAt").query(query);
```
branan: This doesn't quite match the behavior of shelling out to `lsattr`. We should fall back to checking `PdAt` if the attribute doesn't exist in `CuAt`

To do this, we need to query `CuDv` for the object, then look at its `PdDvLn` field. We can then use that to query PdAt.

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,31 @@
...
+    for (auto& attr : attr_query) {
```
branan: It's a little confusing to use for loop here with an immediate return of the first element. I'd rather see the slightly more verbose manual iterator wrangling that makes it clear we're returning a single element.

```
auto attr = attr_query.begin()
if (attr != attr_query.end()) {
    return attr->value
}
```

branan: Hmm, it might be worthwhile to add a `value` helper to the `query_proxy` when we expect only a single value out of an ODM query?

ekinanp: That seems like a better choice. I'll add the helper.

ekinanp: Hmm, on second maybe it's best to have people use the more verbose iterator. B/c with the helper, we'd have to return a pointer to account for the possibility that the value may not exist. They'd still have to do an if check on that return and then dereference it.

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,84 @@
...
+      return "";
+    }
+    auto cudv_obj = (*cudv_ref);
+
+    // TODO: Fill this in after figuring out PdDvLn
+    auto pddv_type = (*cudv_ref).PdDvLn;
+
+    LOG_DEBUG("NAME = {1}", cudv_obj.name); 
+    LOG_DEBUG("DDINS = {1}", cudv_obj.ddins); 
+    LOG_DEBUG("PDDVLN = {1}", cudv_obj.PdDvLn);
```
ekinanp: @branan Unfortunately, it looks like `odm_get_first` and `odm_get_next` do not fill in the structure pointed to by `PdDvLn`. Based on what I have here, `cudv_obj.PdDvLn` prints 0, which is the null pointer. I looked elsewhere in the codebase, and we never directly reference `PdDvLn` like is done here. We only use it as a query parameter e.g. here: https://github.com/puppetlabs/facter/blob/master/lib/src/facts/aix/filesystem_resolver.cc#L87.

I'm not sure what I'm doing wrong here -- `cudv_obj.PdDvLn` seems to be the right type b/c I got a compile error earlier saying that it was of type `PdDv* const`. 

IBM's docs indicate that we could get things working with https://www.ibm.com/support/knowledgecenter/ssw_aix_72/com.ibm.aix.basetrf1/odm_get_list.htm, which lets you specify the link depth. But this option would require some changes into the internal ODM api.

ekinanp: Worst case, we could have this routine be something like `getattr(object, type, field)` where the type would be the uniquetype (or getattr(object, field, type = "") so we can have the option of skipping a check for PdAtr).

ekinanp: Actually a cleaner approach might be to just create a separate iterator class that uses the struct array returned from `odm_get_list` as the underlying container. We could have the interface for open be something like
```
odm_class<T>::open(class_name, expand_link_descriptors = NO)
```

and conditionally create either our current iterator or the `get_list` iterator. It would probably not be a good idea to use `get_list` off the getgo b/c that returns everything at once, which seems kind of bad (if I'm reading the documentation right).

I can look into this if we want to look at this approach.

branan: I just poked `/usr/include/sys/cfgodm.h` to see how the struct actually stores its link. There are three values associated with it - the struct pointer is the `PdDv` field, but there is also a `PdDvLn_Lvalue` field that is a `char[]` - that may be the string form of the link?

lib/src/facts/aix/operating_system_resolver.cc
branan: :+1: the iterator wrangling isn't that bad, if a helper is a hassle

ekinanp:
@branan Updated the PR. I also added a separate commit for some test cases there that I will revert if you think they look good. I could convert those to unit tests too if you think they might be useful, even though they wouldn't technically be unit tests.

Here's the output from running those tests using `/opt/puppetlabs/puppet/bin/facter --query os --debug`:
```
2018-07-31 12:43:22.385534 DEBUG puppetlabs.facter - Test Case: Can read value from CuAt
2018-07-31 12:43:22.385663 DEBUG puppetlabs.facter - getattr(sys0, modelname) = "IBM,8284-22A"
2018-07-31 12:43:22.385976 DEBUG puppetlabs.facter - Could not get a value from the ODM for no_cudv's 'bad_value' attribute: There is no CuDv entry for no_cudv.
2018-07-31 12:43:22.386111 DEBUG puppetlabs.facter - Test Case: Returns empty if CuDv does not exist
2018-07-31 12:43:22.386233 DEBUG puppetlabs.facter - getattr(no_cudv, bad_value) = ""
2018-07-31 12:43:22.386791 DEBUG puppetlabs.facter - Test Case: Can read value from PdAt
2018-07-31 12:43:22.386906 DEBUG puppetlabs.facter - getattr(sys0, ghostdev) = "0"
2018-07-31 12:43:22.387362 DEBUG puppetlabs.facter - Could not get a value from the ODM for sys0's 'bad_value' attribute: There is no PdAt entry for sys0 with bad_value.
2018-07-31 12:43:22.387508 DEBUG puppetlabs.facter - Test Case: Returns empty if PdAt does not exist
2018-07-31 12:43:22.387612 DEBUG puppetlabs.facter - getattr(sys0, bad_value) = ""
```

Also, here is a direct ODM query showing that the `ghostdev` attribute for `sys0` does not have a CuAt entry.
```
[0] [AIX] root@aix71-4:~ # odmget -q "name = sys0 AND attribute = ghostdev" CuAt
[0] [AIX] root@aix71-4:~ #
```

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,72 @@
...
-using namespace leatherman::file_util;
-using namespace boost;
-namespace execution = leatherman::execution;
+using namespace facter::util::aix;
 
+// This routine's meant to be a general utility function that replicates the behavior of
+// lsattr -El <object> -a <field>. Although it's only used to get the modelname of the
+// sys0 device, we still would like to have it here in case we ever need to separate it
+// out to an AIX utils file to query other attributes. Part of what lsattr does is check
+// PdAt if we don't have an entry for the object's attribute in CdAt, even though it is very
```
branan: typo: CuAt

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,72 @@
...
-        LOG_WARNING("Could not get a value from lsattr -El {1} -a {2}", object, field);
+    // High-level logic here is:
+    //   * Check if there's an entry for our object's field attribute in CuAt (the device-specific
+    //   attribute entry).
+    //
+    //   * Else, check for the field attribute's default value in PdAt. We do this by first
+    //   figuring out the PdDv type from the CuDv entry for the object, then use our PdDv type
+    //   to query the field's default value in PdAt.
+    string query = (boost::format("name = %1% AND attribute = %2%") % object % field).str();
+    auto cdat_query = odm_class<CuAt>::open("CuAt").query(query);
```
branan: typo: cuat again

lib/src/facts/aix/operating_system_resolver.cc
```diff
@@ -1,43 +1,72 @@
...
+
+    // Get the PdDv type from the CuDv entry
+    query = (boost::format("name = %1%") % object).str();
+    auto cudv_query = odm_class<CuDv>::open("CuDv").query(query);
+    auto cudv_ref = cudv_query.begin();
+    if (cudv_ref == cudv_query.end()) {
+      LOG_DEBUG("Could not get a value from the ODM for {1}'s '{2}' attribute: There is no CuDv entry for {1}.", object, field);
+      return "";
+    }
+    auto pddv_type = (*cudv_ref).PdDvLn_Lvalue;
```
branan: should be `cudv_ref->PdDvLn_Lvalue`?

ekinanp: Unfortunately no, GCC complains b/c `cudv_ref` is a different type. Here's the error.
```
/root/facter/lib/src/facts/aix/operating_system_resolver.cc: In function 'std::__cxx11::string getattr(std::__cxx11::string, std::__cxx11::string)':
/root/facter/lib/src/facts/aix/operating_system_resolver.cc:55:30: error: base operand of '->' has non-pointer type 'facter::util::aix::odm_class<CuDv>::iterator'
```

branan: we might just need to implement `operator->` for the iterator type?

ekinanp:
@branan Updated the PR. Removed the test commit, implemented `->` operator for the odm_class iterator and used it in `getattr`, and also fixed the typos.

Tested out the fact locally and things still work OK.

McdonaldSeanp:
jenkins please test this on aix61-POWERa

PR #1729, opened 2018-08-02
(FACT-1623) Implement load average resolver for AIX

ekinanp:
This uses the getkerninfo system call to get the current load
averages. AIX's uptime command also uses the same system call
to obtain this information.

### DIFF
--- a/lib/src/facts/aix/load_average_resolver.cc
+++ b/lib/src/facts/aix/load_average_resolver.cc
@@ -1,0 +1,50 @@
+#include <internal/facts/aix/load_average_resolver.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <leatherman/locale/locale.hpp>
+
+#include <sys/inttypes.h>
+#include <sys/kinfo.h>
+
+// Mark string for translation (alias for leatherman::locale::format)
+using leatherman::locale::_;
+
+using namespace std;
+
+/**
+ * This system call lets us query the kernel directly for system
+ * information. We use it to get our current load averages.
+ *
+ * @param info the info we're retrieving from the kernel.
+ * @param buf the buffer that we'll store the information in
+ * @param buf_size a pointer to the variable containing the size of the buffer in bytes
+ * @param arg no idea what this param. represents. we will usually set this to 0.
+ * @return 0 if we successfully retrieve the information, else a negative value
+ */
+extern "C" int getkerninfo(int info, char* buf, int* buf_size, int32long64_t arg);
+
+// Converts the given integer average into a load average.
+static double to_load_avg(double average) {
+  // 65536 is the load average scale on AIX machines.
+  return average / 65536;
+}
+
+namespace facter { namespace facts { namespace aix {
+
+    boost::optional<tuple<double, double, double> > load_average_resolver::get_load_averages()
+    {
+        // This approach was adapted from screen-4.6.2's loadav.c file. See
+        // https://www.mail-archive.com/opensuse-commit@opensuse.org/msg122486.html
+        array<long long, 3> averages;
+        int buf_size = averages.size() * sizeof(long long);
+        int rc = getkerninfo(KINFO_GET_AVENRUN, reinterpret_cast<char*>(averages.data()), &buf_size, 0);
+        if (rc < 0) {
+            LOG_DEBUG(_("failed to retrieve the load averages"));
+            return boost::none;
+        }
+
+        return make_tuple(
+            to_load_avg(averages[0]),
+            to_load_avg(averages[1]),
+            to_load_avg(averages[2]));
+    }
+}}}  // namespace facter::facts::aix
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -120,6 +120,7 @@ if (AIX)
         "src/facts/aix/disk_resolver.cc"
         "src/facts/aix/filesystem_resolver.cc"
         "src/facts/aix/kernel_resolver.cc"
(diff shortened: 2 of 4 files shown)


lib/src/facts/aix/load_average_resolver.cc
```diff
@@ -0,0 +1,51 @@
...
+ * This system call lets us query the kernel directly for system
+ * information. We use it to get our current load averages.
+ *
+ * @param info the info we're retrieving from the kernel.
+ * @param buf the buffer that we'll store the information in
+ * @param buf_size a pointer to the variable containing the size of the buffer in bytes
+ * @param arg no idea what this param. represents. we will usually set this to 0.
+ * @return 0 if we successfully retrieve the information, else a negative value
+ */
+extern "C" int getkerninfo(int info, char* buf, int* buf_size, int32long64_t arg);
```
ekinanp: I see this declaration in https://github.com/puppetlabs/facter/blob/3.11.x/lib/src/facts/aix/networking_resolver.cc#L34. Should we move it over to a separate .h file? E.g. something like `util/aix/getkerninfo.hpp`? Also, I tried using https://github.com/puppetlabs/facter/blob/3.11.x/lib/src/facts/aix/networking_resolver.cc#L37 here but it throws a bad alloc exception b/c `getkerninfo(KINFO_GET_AVENRUN, nullptr, nullptr, 0)` returns -1. Probably b/c there is no load averages struct from what I could see in the IBM header files while the networking stuff has their own structs.

ekinanp:
On an AIX 7.1 VM:
```
[0] [AIX] root@aix71-4:~ # /opt/puppetlabs/puppet/bin/facter --query load_averages && uptime
{
  15m => 0.907669,
  1m => 1.0027,
  5m => 0.916275
}
  11:12PM   up 2 days,  13:52,  3 users,  load average: 1.00, 0.92, 0.91
```

You can see that the values match the uptime command.

Should I have an acceptance test here? We can use uptime to compare the values. I didn't add one here b/c I couldn't see one for the glib resolver that the other platforms use.

lib/src/facts/aix/load_average_resolver.cc
```diff
@@ -0,0 +1,51 @@
...
+
+    boost::optional<tuple<double, double, double> > load_average_resolver::get_load_averages()
+    {
+        // This approach was adapted from screen-4.6.2's loadav.c file. See
+        // https://www.mail-archive.com/opensuse-commit@opensuse.org/msg122486.html
+        array<long long, 3> averages;
+        int buf_size = averages.size() * sizeof(long long);
+        int rc = getkerninfo(KINFO_GET_AVENRUN, reinterpret_cast<char*>(averages.data()), &buf_size, 0);
+        if (rc < 0) {
+            LOG_DEBUG(_("failed to retrieve the load averages"));
```
ekinanp: The `glib` resolver logs a message too (see here: https://github.com/puppetlabs/facter/blob/master/lib/src/facts/glib/load_average_resolver.cc#L17), which is why I'm doing the same here. But the networking resolver throws an exception if `getkerninfo` fails. Which should I go with?

lib/src/facts/aix/load_average_resolver.cc
```diff
@@ -0,0 +1,50 @@
...
+ * @param buf_size a pointer to the variable containing the size of the buffer in bytes
+ * @param arg no idea what this param. represents. we will usually set this to 0.
+ * @return 0 if we successfully retrieve the information, else a negative value
+ */
+extern "C" int getkerninfo(int info, char* buf, int* buf_size, int32long64_t arg);
+
+// Converts the given integer average into a load average.
+static double to_load_avg(double average) {
+  // 65536 is the load average scale on AIX machines.
+  return average / 65536;
```
ekinanp: I have no idea how the screen source code figured out that `65536` is the load scale. I've grepped through all of the include files on my AIX machine and couldn't find this declaration. IBM's docs don't point to anything either. I don't know how often this will change from AIX machine to AIX machine, but if it does change it might be worthwhile to try and get this info. from the `uptime` command instead.

For what's it's worth, the screen code defines it like this (in their configure.ac file).
```
# if defined(apollo) || defined(_IBMR2) || defined(_AUX_SOURCE) || defined(AUX)
loadscale=65536
```

ekinanp: I could take a crack at disassembling the `uptime` command. Maybe I'll try that.

ekinanp:
@caseywilliams Agreed. I spent a while sifting through their docs trying to figure that out and I couldn't find anything. @branan Do you happen to know?

lib/src/facts/aix/load_average_resolver.cc
ekinanp: Nope. The disassemble doesn't really give any useful info.

lib/src/facts/aix/load_average_resolver.cc
branan: I think it depends on how critical the fact is whether we want to raise an exception or just log. Duplicating the behavior of other load average implementations seems fine to me :+1:

lib/src/facts/aix/load_average_resolver.cc
branan: I think your usage below is fine, since there's no struct for this KINFO value.

Regarding moving it to a header, I don't think we have a good place to put it right now? Seems silly to add a new header for one declaration

branan:
I have no idea what the last arg is. Everything I know about getkerninfo is captured in the comment in the network resolver

PR #1733, opened 2018-08-09
(FACT-1821) Fix AIX memory fact

smcclellan:
To determine the AIX memory fact, several devices are queried. In some
cases, an error was thrown (ENXIO) that was not caught. This handles
that error when it occurs.

Additionally, this will now debug-log devices that cannot be used, and
the error number from trying to use it.

### DIFF
--- a/lib/src/facts/aix/memory_resolver.cc
+++ b/lib/src/facts/aix/memory_resolver.cc
@@ -1,6 +1,7 @@
 #include <internal/facts/aix/memory_resolver.hpp>
 #include <internal/util/aix/odm.hpp>
 
+#include <leatherman/logging/logging.hpp>
 #include <sys/vminfo.h>
 #include <sys/cfgodm.h>
 #include <sys/limits.h>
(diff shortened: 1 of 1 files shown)


smcclellan:
@branan Is this something we can test in the memory_resolver.cc unit tests?

caseywilliams:
jenkins please test this on aix61-POWERa,aix71-POWERa,aix72-POWERa

caseywilliams:
jenkins please test this on aix61-POWERa

lib/src/facts/aix/memory_resolver.cc
```diff
@@ -33,10 +34,14 @@ namespace facter { namespace facts { namespace aix {
                 // swapqry the things that have an attribute we
                 // expect, but ignore any errno values that look like
                 // "this just wasn't a good device to query"
+                // ENXIO: No such device address.
                 if (errno != ENODEV &&
                     errno != ENOENT &&
-                    errno != ENOTBLK) {
+                    errno != ENOTBLK &&
+                    errno != ENXIO) {
                     throw system_error(errno, system_category());
```
branan: I think it'd be good to add the failing device to this exception while we're in here making changes. `system_error`'s constructor can [take a third argument](https://en.cppreference.com/w/cpp/error/system_error/system_error) to add additional context. Probably just `system_error(errno, system_category(), device)` is fine.

We probably use the two-argument form all over the place, but we can fix those up as we encounter them.

smcclellan: Updated.

branan:
jenkins please test this on aix61-POWERa

smcclellan:
Note: Re-file against 3.11.x.

PR #1734, opened 2018-08-16
(FACT-1821) Fix AIX memory fact

smcclellan:
Cherry-picked from https://github.com/puppetlabs/facter/pull/1733.

To determine the AIX memory fact, several devices are queried. In some
cases, an error was thrown (ENXIO) that was not caught. This handles
that error when it occurs.

Additionally, this will now debug-log devices that cannot be used, and
the error number from trying to use it.


branan:
Merging pre-travis because:
* This is a cherry-pick of something that's already been linted etc.
* This is AIX-specific code, so build/test won't run it

PR #1735, opened 2018-08-16
3.11.x master mergeup

smcclellan:


branan:
There's a couple more new conflicts, but they're just version bumps so should be resolvable automatically once this is in. Gonna go ahead and land this

PR #1736, opened 2018-08-21
(FACT-1862) Add support for Suse v15 os detection

ScottGarman:
This refactors how we handle platforms that rely on Freedesktop.org's
/etc/os-release to determine os facts. This includes CoreOS, Cumulus, and
adds Suse v15 releases, which have removed the /etc/SuSE-release file in
favor of os-release.


ScottGarman:
Here's an ad-hoc run with passing facter tests: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/puppet-agent/view/ad-hoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/

The puppet test failures are I believe a known issue I saw discussed in hipchat earlier today.

ScottGarman:
Typo fixed. :)

PR #1740, opened 2018-08-29
(maint) Add back package:bootstrap and package:implode for compatibility

underscorgan:


smortex:
This seems to be a dupe of #1741?  CI which failed _here_ did not failed _there_, so maybe this PR can be closed and progress continue on #1741?

mwaggett:
This and https://github.com/puppetlabs/facter/pull/1741 are targeting different branches, that's why there's two of them.

But we don't actually need either of these anymore, so will close both.

PR #1741, opened 2018-08-29
(maint) Add back package:bootstrap and package:implode for compatibility

underscorgan:


mwaggett:
we don't need this anymore, now that the packaging-as-a-gem work is complete

PR #1744, opened 2018-09-05
(FACT-1415) Use RtlGetVersion to get the Windows OS version

ekinanp:
Previously, we would use the version for kernel32.dll as the OS
version. However, there were some uncertainties around whether
this was the best path forward.

Facter 2.x gets the OS version using the kernel call RtlGetVersion
from ntoskrnl.exe, which will return the correct versions of the
kernel. This commit modifies the Windows kernel resolver to also
use RtlGetVersion.

### DIFF
--- a/lib/src/facts/windows/kernel_resolver.cc
+++ b/lib/src/facts/windows/kernel_resolver.cc
@@ -1,60 +1,42 @@
+#include <leatherman/dynamic_library/dynamic_library.hpp>
 #include <internal/facts/windows/kernel_resolver.hpp>
-#include <leatherman/windows/system_error.hpp>
-#include <leatherman/windows/windows.hpp>
 #include <facter/facts/os.hpp>
 #include <leatherman/logging/logging.hpp>
+
 #include <boost/optional.hpp>
-#include <boost/algorithm/string/trim.hpp>
 #include <boost/format.hpp>
-#include <boost/nowide/convert.hpp>
+#include <windows.h>
+#include <ntstatus.h>
 
 using namespace std;
-using namespace leatherman::windows;
+using namespace leatherman::dynamic_library;
+using RtlGetVersionPtr = NTSTATUS (WINAPI *)(PRTL_OSVERSIONINFOW);
 
 namespace facter { namespace facts { namespace windows {
 
     static boost::optional<string> get_release()
     {
-        // GetVersionEx requires the manifest is correct, and is essentially deprecated.
-        // Another method of getting the OS version can be found at
-        // http://msdn.microsoft.com/en-us/library/windows/desktop/ms724429(v=vs.85).aspx
-        // and is what we use here.
-        auto fileName = L"Kernel32.dll";
-        auto fileVerSize = GetFileVersionInfoSizeW(fileName, nullptr);
-        if (fileVerSize == 0) {
-            return boost::none;
+        dynamic_library ntoskrnl;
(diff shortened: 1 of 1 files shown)


ekinanp:
Tested this out on Win-10 pro x86_64. The result (10.0.16299) matches what we previously got when using the kernel32.dll method. Also, note that Facter 2.x gets the version fact the same way -- https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/windows/process.rb#L137. I will run PR tests once the basic CI passes on all of our Windows platforms to be safe though.

lib/src/facts/windows/kernel_resolver.cc
```diff
@@ -1,60 +1,42 @@
...
     {
-        // GetVersionEx requires the manifest is correct, and is essentially deprecated.
-        // Another method of getting the OS version can be found at
-        // http://msdn.microsoft.com/en-us/library/windows/desktop/ms724429(v=vs.85).aspx
-        // and is what we use here.
-        auto fileName = L"Kernel32.dll";
-        auto fileVerSize = GetFileVersionInfoSizeW(fileName, nullptr);
-        if (fileVerSize == 0) {
-            return boost::none;
+        dynamic_library ntoskrnl;
```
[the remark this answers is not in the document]
ekinanp: Nevermind, I answered my own question. It looks like it marks the operation as having returned successfully even if it fails, so we can't log errors with it anymore. I updated the PR to add appropriate logging in the code.

ekinanp:
jenkins please test this on windows10ent-64a,windows10ent-32a,windows2008r2-64a,windows2012r2-64a,windows2012r2_ja-64a%7Blocale=ja%7D,windows2016-64a

ekinanp:
PR tests passed, this should be good to go.

lib/src/facts/windows/kernel_resolver.cc
```diff
@@ -1,60 +1,42 @@
...
-        wstring subBlock = str(boost::wformat(L"\\StringFileInfo\\%04x%04x\\ProductVersion")
-            % lpTranslate->wLanguage % lpTranslate->wCodePage);
-
-        wchar_t *version;
-        UINT versionLen;
-        if (!VerQueryValueW(buffer.data(), subBlock.c_str(), reinterpret_cast<LPVOID*>(&version), &versionLen)) {
-            return boost::none;
+        OSVERSIONINFOW versionInfo;
+        if (rtlGetVersion(&versionInfo) != STATUS_SUCCESS) {
+          LOG_DEBUG("failed to get the OS version info. from RtlGetVersion");
```
branan: I think the full spelling "information" will translate better?

lib/src/facts/windows/kernel_resolver.cc
```diff
@@ -1,60 +1,42 @@
...
+          return boost::none;
         }
 
-        vector<wchar_t> buffer(fileVerSize);
-        if (!GetFileVersionInfoW(fileName, 0, fileVerSize, buffer.data())) {
-            return boost::none;
+        auto rtlGetVersion = reinterpret_cast<RtlGetVersionPtr>(
+            ntoskrnl.find_symbol("RtlGetVersion"));
+        if (! rtlGetVersion) {
+          return boost::none;
```
branan: we should log a debug something here

lib/src/facts/windows/kernel_resolver.cc
ekinanp: Updated the PR with this change.

PR #1745, opened 2018-09-05
(FACT-1880) Ensure os.release.minor returns a value

ScottGarman:
New releases (e.g, SLES 15) return a VERSION_ID of a simple integer
without a point-release. Since a lot of code (e.g, versioncmp) may
depend on making use of os.release.minor, ensure that this fact at least
returns a '0' for theses cases rather than nothing.

### DIFF
--- a/lib/inc/internal/facts/linux/os_osrelease.hpp
+++ b/lib/inc/internal/facts/linux/os_osrelease.hpp
@@ -7,6 +7,7 @@
 #include <internal/facts/linux/os_linux.hpp>
 #include <facter/facts/os.hpp>
 #include <facter/facts/os_family.hpp>
+#include <boost/regex.hpp>
 #include <iostream>
 using namespace std;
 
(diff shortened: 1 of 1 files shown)


lib/inc/internal/facts/linux/os_osrelease.hpp
```diff
@@ -81,7 +82,17 @@ namespace facter { namespace facts { namespace linux {
         virtual std::string get_release(std::string const& name, std::string const& distro_release) const override
         {
             auto val = _release_info.find("VERSION_ID");
-            return (val != _release_info.end()) ? val->second : std::string();
+            if (val != _release_info.end()) {
+                string major, minor;
```
caseywilliams: These two variables look unused - otherwise this makes sense and looks fine to me

caseywilliams: Ah, and I believe this should be targeted at 3.11.x for agent 5.5.x

speedofdark: Nice catch, Casey.

ScottGarman:
@caseywilliams I removed the redundant variable declaration. As for targeting the wrong branch - I've already had the previous work merged to 3.9.x, and this fix impacts all platforms that parse /etc/os-release (including Cumulus and CoreOS), so I think this fix should still land here and get merged up - unless it's preferred to revert the work in 3.9.x instead.

PR #1748, opened 2018-09-18
(maint) Fix expected os_version fact to distinguish 2012 and 2012 R2

ekinanp:
Previously, the acceptance tests would assume that the os_version fact
resolved to '2012 R2' for both Windows 2012 and Windows 2012 R2. This is
incorrect because on Windows 2012, the os_version fact resolves to '2012'
instead, which is correct. This failure was not detected in the agent CI
pipelines because we only test with 2012 R2.

This commit fixes the expected os_version fact in the tests to correctly
distinguish between Windows 2012 and Windows 2012 R2. [no-promote]


ekinanp:
jenkins please test this on windows2012r2-64a,windows2012-64a

PR #1751, opened 2018-09-27
(FACT-1870) Redirect Ruby stdout to stderr for structured output

branan:
Previously, any output to `stdout` in a custom fact would be echoed to
the terminal as-is, since Facter did nothing special with its output
streams. This caused problems when outputting structured data (json or
yaml), since anything printed to stdout would end up in the structure,
possibly corrupting it.

Facter now redirects the Ruby interpreter's stdout to stderr when
resolving custom facts for structured output. This means that the
output isn't lost, just put someplace less annoying.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,54 +1,33 @@
-language: cpp
-compiler:
-  - gcc
-dist: precise
-sudo: false
-addons:
-  apt:
-    sources:
-      - ubuntu-toolchain-r-test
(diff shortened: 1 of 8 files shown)


branan:
This is gonna be red until we get leatherman tagged for the new ruby method

branan:
Fun, for some reason the unpacked leatherman library can't be found by cmake now >_<

PR #1753, opened 2018-10-07
(FACT-1873) Use builtin fact if custom facts don't resolve or have 0 wt.

ekinanp:
Previously, if any builtin facts have conflicting custom facts that all
fail to resolve, and at least one of these custom facts has a weight >
0, Facter will not report a value for that fact. This is undesirable
behavior because we always want to fall back to the builtin value if
all custom facts fail to resolve. Note that the "weight > 0" part is
important because Facter _will_ use the builtin value if all custom
facts have 0 weight.

This commit falls back to the builtin value if all custom facts fail
to resolve, which fixes the reported bug in the ticket.

### DIFF
--- a/acceptance/tests/custom_facts/conflicts_with_builtin_fact.rb
+++ b/acceptance/tests/custom_facts/conflicts_with_builtin_fact.rb
@@ -1,0 +1,106 @@
+test_name 'Facter should appropriately resolve a custom fact when it conflicts with a builtin fact' do
+  tag 'risk:medium'
+
+  def create_custom_fact_on(host, custom_fact_dir, fact_file_name, fact)
+    fact_file_contents = <<-CUSTOM_FACT
+Facter.add(:#{fact[:name]}) do
+  has_weight #{fact[:weight]}
+  setcode do
+    #{fact[:value]}
+  end
+end
+CUSTOM_FACT
+
+    fact_file_path = File.join(custom_fact_dir, fact_file_name) 
+    create_remote_file(host, fact_file_path, fact_file_contents)
+  end
+
+  def clear_custom_facts_on(host, custom_fact_dir)
+    step "Clean-up the previous test's custom facts" do
+      on(agent, "rm -f #{custom_fact_dir}/*")
+    end
+  end
+
+  agents.each do |agent|
+    custom_fact_dir = agent.tmpdir('facter')
+    teardown do
+      on(agent, "rm -rf '#{custom_fact_dir}'")
+    end
+
+    fact_name = 'timezone'
+    builtin_value = on(agent, facter('timezone')).stdout.chomp
+
+    step "Verify that Facter uses the custom fact's value when its weight is > 0" do
+      custom_fact_value = "custom_timezone"
+      create_custom_fact_on(
+        agent,
+        custom_fact_dir,
+        'custom_timezone.rb',
+        name: fact_name,
+        weight: 10,
+        value: "'#{custom_fact_value}'"
+      )
+
+      on(agent, facter("--custom-dir=#{custom_fact_dir} timezone")) do |result|
+        assert_match(/#{custom_fact_value}/, result.stdout.chomp, "Facter does not use the custom fact's value when its weight is > 0")
+      end
+    end
+
+    clear_custom_facts_on(agent, custom_fact_dir)
+
+    step "Verify that Facter uses the builtin fact's value when all conflicting custom facts fail to resolve" do
+      [ 'timezone_one.rb', 'timezone_two.rb'].each do |fact_file|
+        create_custom_fact_on(
+          agent,
+          custom_fact_dir,
+          fact_file,
+          { name: fact_name, weight: 10, value: nil }
+        )
+      end
+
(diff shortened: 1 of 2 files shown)


lib/src/ruby/fact.cc
```diff
@@ -120,6 +120,21 @@ namespace facter { namespace ruby {
                 // Set the value to what was resolved
                 _value = value;
                 _weight = weight;
+
+                if (! ruby.is_nil(_value)) {
```
ekinanp: Some questions/comments on this:
* Currently when all custom facts have weight 0, we use the builtin fact value (see [here](https://github.com/puppetlabs/facter/blob/master/lib/src/ruby/fact.cc#L87-L99). Is this desired behavior?
* If ^ is the desired behavior, then this code will break that consistency when all of our custom facts with weight > 0 fail to resolve and one custom fact with weight == 0 resolves. Would that be bad?

ekinanp:
jenkins please test this on windows10ent-64a,redhat7-64a

lib/src/ruby/fact.cc
McdonaldSeanp: According to the docs facts should always have a positive integer value (i.e. not 0): https://docs.puppet.com/facter/3.0/fact_overview.html

So as a user I wouldn't expect a facter with weight == 0 to resolve over the builtin fact.

branan: I'd say any fact with weight 0 should be considered "disabled". If that ends up surprising users in the future it's easy to fix :man_shrugging:

ekinanp:
@McdonaldSeanp Thanks for double-checking that, will update the PR

lib/src/ruby/fact.cc
```diff
@@ -120,6 +120,21 @@ namespace facter { namespace ruby {
...
+                  return 0;
+                }
+
+                // None of our resolvers were able to resolve the value.
+                // Thus, we attempt to use the "built-in" fact's value.
+                auto builtin_value = facts[ruby.to_string(_name)];
```
branan: Does it make more sense to do this at https://github.com/puppetlabs/facter/blob/b1960ad78d78616cf71a44164e2169801b15db27/lib/src/ruby/module.cc#L360 ?

lib/src/ruby/fact.cc
```diff
@@ -120,6 +120,21 @@ namespace facter { namespace ruby {
...
+                if (builtin_value) {
+                  // Already in collection, do not add
+                  add = false;
+                  _value = facter->to_ruby(builtin_value);
```
branan: I don't really like wrapping the builtin fact as a ruby one - I'd much rather handle the fallback somewhere else, if we can

branan: Ah, nevermind, I see we already do this wrapping for other cases when we don't add. Weird.

ekinanp:
jenkins please test this on windows10ent-64a,redhat7-64a

ekinanp:
It looks like we can't disable 0 weight facts because Facer will default custom facts that have no weight to have weight 0 (i.e. stuff breaks if 0 weight facts are disabled). I have, however, updated the PR so that built-in facts now have precedence over 0 weight facts, and have also updated the acceptance tests with the new behavior.

acceptance/tests/custom_facts/conflicts_with_builtin_fact.rb
```diff
@@ -0,0 +1,106 @@
...
+      )
+
+      on(agent, facter("--custom-dir=#{custom_fact_dir} timezone")) do |result|
+        assert_match(/#{custom_fact_value}/, result.stdout.chomp, "Facter does not use the custom fact's value when its weight is > 0")
+      end
+    end
+
+    clear_custom_facts_on(agent, custom_fact_dir)
+
+    step "Verify that Facter uses the builtin fact's value when all conflicting custom facts fail to resolve" do
```
ekinanp: This test fails without the changes here, as expected.

acceptance/tests/custom_facts/conflicts_with_builtin_fact.rb
```diff
@@ -0,0 +1,106 @@
...
+        end
+
+        on(agent, facter("--custom-dir=#{custom_fact_dir} timezone")) do |result|
+          assert_match(/#{builtin_value}/, result.stdout.chomp, "Facter does not give precedence to the builtin fact when all custom facts have zero weight")
+        end
+      end
+
+      clear_custom_facts_on(agent, custom_fact_dir)
+
+      step "when some custom facts have zero weight" do
```
ekinanp: This test also fails without the changes here, as expected. All other tests pass, as expected.

ekinanp:
@McdonaldSeanp Builtin facts already took precedence over custom facts that didn't specify a weight (because those facts defaulted to having a weight of 0) -- see https://github.com/puppetlabs/facter/blob/master/lib/src/ruby/fact.cc#L90. However, if one of those custom facts did have a weight, then builtin facts would not take precedence over the weightless custom facts. There should be some acceptance tests outlining these two cases. This change would break the latter scenario.

PR #1754, opened 2018-10-15
(FACT-1889) Update os release fact to include Windows Server 2019

ekinanp:


ekinanp:
jenkins please test this on windows10ent-64a,windows2016-64a,windows2019-64a

lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -101,7 +104,25 @@ namespace facter { namespace facts { namespace windows {
         auto version = result.release.substr(0, lastDot);
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
         if (version == "10.0") {
-            result.release = consumerrel ? "10" : "2016";
+            // Calculate the build number to distinguish between
+            // Windows Server 2016 and 2019. Note that the kernel
+            // version is written as <major>.<minor>.<build_number>
+            auto kernel_version_fact = facts.get<string_value>(fact::kernel_version);
```
ekinanp: I figured calculating the build number from the kernel version fact is easier because calling RtlGetVersion requires [several steps](https://github.com/puppetlabs/facter/blob/master/lib/src/facts/windows/kernel_resolver.cc#L19-L34) and I didn't see a need to create a common function when kernel version gets this info. for us (and the string parsing isn't complicated).

lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -101,7 +104,25 @@ namespace facter { namespace facts { namespace windows {
...
+            if (! kernel_version_fact) {
+                return result;
```
ekinanp: Should a warning/debug message be outputted here? We don't do this in [earlier steps](https://github.com/puppetlabs/facter/blob/master/lib/src/facts/windows/operating_system_resolver.cc#L90-L98) which is why I didn't do it here.

ekinanp:
jenkins please test this on windows10ent-64a,windows2016-64a,windows2019-64a

ekinanp:
https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_pr/759/ is the passing acceptance test job showing that the changes here worked.. Going to re-run PR tests on all the components just to make sure nothing else needs to be done for Win Server 2019 enablement.

ekinanp:
jenkins please test all on windows10ent-64a,windows2016-64a,windows2019-64a

lib/src/facts/windows/operating_system_resolver.cc
glennsarti: I feel a Debug message is required at a minimum.

lib/src/facts/windows/operating_system_resolver.cc
glennsarti: I can't think of a reason why this is not-good.  Is there a precedence for dependent core facts?

ekinanp:
The pxp-agent failures in https://jenkins-platform.delivery.puppetlabs.net/job/platform_puppet-agent-extra_puppet-agent-integration-suite_pr/761/RMM_COMPONENT_TO_TEST_NAME=pxp_agent,SLAVE_LABEL=beaker,TEST_TARGET=windows2019-64a/ are the same ones that's been blocking 5.5.x. It looks like this PR's all that's required to enable Windows Server 2019.

lib/src/facts/windows/operating_system_resolver.cc
ekinanp: No, we use dependent core facts elsewhere too like e.g. [here](https://github.com/puppetlabs/facter/blob/master/lib/src/facts/aix/operating_system_resolver.cc#L89). My only concern with doing it this way was b/c it isn't a direct C API call to `RtlGetVersion` (which is the general pattern for obtaining these facts in cfacter).

lib/src/facts/windows/operating_system_resolver.cc
ekinanp: Good point. I'll go ahead and add one.

lib/src/facts/windows/operating_system_resolver.cc
```diff
@@ -101,7 +104,26 @@ namespace facter { namespace facts { namespace windows {
...
         bool consumerrel = (wmi::get(vals, wmi::producttype) == "1");
         if (version == "10.0") {
-            result.release = consumerrel ? "10" : "2016";
+            // Calculate the build number to distinguish between
+            // Windows Server 2016 and 2019. Note that the kernel
+            // version is written as <major>.<minor>.<build_number>
+            auto kernel_version_fact = facts.get<string_value>(fact::kernel_version);
+            if (! kernel_version_fact) {
+                LOG_DEBUG("Could not resolve the OS release and OS major version facts from the kernel version fact");
+                return result;
```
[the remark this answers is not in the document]
ekinanp: Yes because these types of changes query system calls and those are tough to mock out in C++. For example, there are no unit tests for the Windows operating system resolver.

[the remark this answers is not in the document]
ekinanp: This code is covered by the acceptance tests.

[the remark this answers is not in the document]
speedofdark: Coolness.

speedofdark:
This seems good to me. I think we need @glennsarti to approve and we're good to merge.

lib/src/facts/windows/operating_system_resolver.cc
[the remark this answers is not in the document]
smcclellan: @ekinanp Is that to say the acceptance tests are failing until this goes in?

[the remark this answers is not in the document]
ekinanp: No, the acceptance tests are passing for our suite pipelines. However once we add Server 2019 to agent CI, then the acceptance tests will fail without this change.

PR #1761, opened 2018-11-26
Explicitly require hypervisor beaker gems, for non-CI contexts

caseywilliams:
These are no longer built into beaker 4


caseywilliams:
Appveyor failure is unrelated

PR #1763, opened 2018-11-29
Explicitly require beaker hypervisor gems for acceptance

caseywilliams:
These are development dependencies in Beaker 4, but they should be explicitly required


caseywilliams:
Just realized I had already made a PR for this and forgot! Closing.

PR #1764, opened 2018-12-18
Added Alt Linux configuration to detect

majioa:


GabrielNagy:
Hi @majioa,

Thanks for your contribution! Unfortunately, Facter 2 has been EOL'd for some time now.
Let me know if you want to rebase rebase your changes on a newer version, or close the PR altogether.

Thanks!

majioa:
ok, i ll fix it

majioa:
@GabrielNagy  The PR is rebased on 2.x

lib/facter/operatingsystem/linux.rb
```diff
@@ -247,7 +251,6 @@ def get_operatingsystem_with_release_files
           "AristaEOS"   => "/etc/Eos-release",
           "Debian"      => "/etc/debian_version",
           "Gentoo"      => "/etc/gentoo-release",
-          "Fedora"      => "/etc/fedora-release",
```
GabrielNagy: Is this intended?

lib/facter/operatingsystem/linux.rb
```diff
@@ -276,10 +279,14 @@ def get_operatingsystem_with_release_files
...
               operatingsystem = "OEL"
             end
+          elsif FileTest.exists?("/etc/altlinux-release")
+            operatingsystem = get_alt_operatingsystem_name
           elsif FileTest.exists?("/etc/redhat-release")
             operatingsystem = get_redhat_operatingsystem_name
           elsif FileTest.exists?("/etc/SuSE-release")
             operatingsystem = get_suse_operatingsystem_name
+          elsif FileTest.exists?("/etc/fedora-release")
+            operatingsystem = "Fedora"
```
GabrielNagy: Is this intended?

lib/facter/operatingsystem/linux.rb
majioa: What do you mean international change?

GabrielNagy: What is the purpose of the Fedora-related changes?

majioa: Fedora related change was done because all of ALT linux releases have fedora-release, redhat-release files in /etc, and since the code determines the os by the table firstly it will detect the OS as Fedora not as the ALT, so Fedora line was cropped out of the table, and inserted below the ALT Linux detection.

I've updated the PR, along the slightly changed and validated return strings.

GabrielNagy: Thanks for the clarification, thought that might be the case :).

majioa:
done

PR #1765, opened 2018-12-19
(FACT-1899) Set OS facts for ubuntu from os-release

donoghuc:
This commit handles the case where `/etc/os-release` contains the OS {name,family,release} for `ubuntu` targets. Previously if a target system did not contain the `lsb_release` executable the OS facts for a 16.04 machine would be Debain stretch/sid (from `/etc/debian_release`). The base ubuntu {16,18}.04 docker images do not ship with the `lsb_release` executable but contain the relevant details in `/etc/os-release`.

### DIFF
--- a/lib/inc/internal/facts/linux/os_osrelease.hpp
+++ b/lib/inc/internal/facts/linux/os_osrelease.hpp
@@ -45,6 +45,8 @@ namespace facter { namespace facts { namespace linux {
                     return os::suse_enterprise_desktop;
                 } else if (id == "sles") {
                     return os::suse_enterprise_server;
+                } else if (id == "ubuntu") {
+                    return os::ubuntu;
                 }
             }
             return std::string();
@@ -68,6 +70,8 @@ namespace facter { namespace facts { namespace linux {
                     return os_family::debian;
                 } else if (id == "opensuse" || id == "opensuse-leap" || id == "sled" || id == "sles") {
                     return os_family::suse;
+                } else if (id == "ubuntu") {
+                    return os_family::debian;
                 }
             }
             return std::string();
--- a/lib/src/facts/linux/operating_system_resolver.cc
+++ b/lib/src/facts/linux/operating_system_resolver.cc
@@ -31,7 +31,7 @@ namespace facter { namespace facts { namespace linux {
         auto release_info = os_linux::key_value_file(release_file::os, {"ID", "CISCO_RELEASE_INFO"});
         auto const& id = release_info["ID"];
         if (id == "coreos" || id == "cumulus-linux" || id == "opensuse" ||
-            id == "opensuse-leap" || id== "sled" || id == "sles") {
+            id == "opensuse-leap" || id== "sled" || id == "sles" || id == "ubuntu") {
             return unique_ptr<os_linux>(new os_osrelease());
         } else {
             auto const& cisco = release_info["CISCO_RELEASE_INFO"];


branan:
The issue with lightweight container images missing lsb_release is common across many distros. Adding ubuntu here works around your problem, but it's just adding another special case and doesn't solve the general issue.

Really the right fix here is to switch to using os-release as the main/initial source of data, and only use our "legacy" OS detection paths when os-release is missing. I believe we have a ticket for that floating around, but we haven't tackled that particular mountain of tech debt yet.

branan:
Two other smaller things:

* We generally prefer FACT- tickets, since Jira doesn't always play nice with fixversions across multiple trackers
* This probably should be targeting an existing branch as a bugfix? If it's targeted at master it will end up in puppet-agent 6.2, and won't hit PE until sometime in the spring.

MikaelSmith:
What other distros do we run into this on? My memory is fuzzy why we didn't trust os-release more broadly. Maybe that doesn't tell you about the various RedHat flavors?

branan:
https://tickets.puppetlabs.com/browse/FACT-1897 describes the same thing on a RHEL 7.

WRT why we don't use os-release - I think it's just that we copied Facter 2's OS detection, and that all predates os-release being common. We didn't ever get around to refactoring for using that file as the primary data source.

MikaelSmith:
I'm open to fixing it if it seems low-risk, otherwise I think the Ubuntu fix is valuable on its own.

donoghuc:
Thanks Branan an Michael. I can do a bit more exploration to see if this stop-gap measure would help any other common target systems. 
Was not able to reproduce on a vmpooler redhat-7-x86_64.
```
[root@g0y3cffyy6ji020 ~]# /opt/puppetlabs/bin/facter os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "RedHat",
  release => {
    full => "7.2",
    major => "7",
    minor => "2"
  },
  selinux => {
    enabled => false
  }
}
[root@g0y3cffyy6ji020 ~]# /opt/puppetlabs/bin/facter -v
3.12.2 (commit 5dc9a805ac42ae292cdb4d2c5447ea55c5d77272)
[root@g0y3cffyy6ji020 ~]# cd /
[root@g0y3cffyy6ji020 /]# find . -name lsb_release
[root@g0y3cffyy6ji020 /]# lsb_release -a
-bash: lsb_release: command not found
```
Or centos-7-x86_64
```
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "7.2.1511",
    major => "7",
    minor => "2"
  },
  selinux => {
    enabled => false
  }
}
```

donoghuc:
@branan We have minimized the impact of this in Bolt. Do you want to move forward with this or close FACT-1897 in favor of refactor described in conversation?

tphoney:
nice work everyone, looking forward to this fix going in

donoghuc:
Thanks for the review @ScottGarman. What are the next steps for getting this fix in? I targeted it against the `3.11.x` branch. Once this is merged will it propagate up to the `3.12.x` and master branches in CI? Or will I need to do that manually? Essentially my question is what else can I do to see this through? Thanks.

ScottGarman:
Once the PR is merged, our CI has automation to make the merge-ups happen to the other branches. As far as getting this merged, I'd just like to make sure @branan has no further objections - he can perform the merge once he gets a chance to look at this.

PR #1766, opened 2018-12-20
Fixed timeout

majioa:
Fixed timeout with the timeout param, instead of read_timeout for newrr ruby versions


majioa:
But I think may be better to use common **timeout** param in this method.

PR #1767, opened 2019-01-02
(FACT-1900) Only check for `sysnative` if we are running under Wow64

branan:
Previously, facter would check for the existence of
`%sysroot%\sysnative` and assume it was the correct system32 directory
if it exists. This is incorrect if that directory has somehow been
created as a normal directory in the normal filesystem.

Instead, we should just check for whether we are running under WOW64,
and spit out the correct directory based on that query. We don't use
IsWow64Process2, since we need to support versions of Windows which do
not have it available.

### DIFF
--- a/lib/src/facts/windows/operating_system_resolver.cc
+++ b/lib/src/facts/windows/operating_system_resolver.cc
@@ -63,15 +63,17 @@ namespace facter { namespace facts { namespace windows {
             LOG_DEBUG("error finding SYSTEMROOT: {1}", leatherman::windows::system_error());
         }
 
-        auto pathNative = path(szPath) / "sysnative";
-        boost::system::error_code ec;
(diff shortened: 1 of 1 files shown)


branan:
I'm totally relying on appveyor to build/test this on Windows, so let's see what happens.

branan:
I guess I could also just hardcode `system32` here, since we no longer support 32-bit agents on 64-bit Windows? But I figure if anyone is building this code themselves I don't wanna just break the semantics of this.

ekinanp:
Also, should this be targeted at 3.11.x? Seems like a bugfix.

branan:
@ekinanp ah, yup. I had based it on 3.11.x, just targeted the PR at the wrong branch. I've also pushed up a fix for the lints

PR #1768, opened 2019-01-17
FACT-1905 do not test dhcp on rhel-8

gimmyxd:
Since redhat-8 is not using dhcp, this test was failing
https://github.com/sebastian-miclea/puppetlabs-packer/blob/eea826c428ebabf17a2bdb2d5d2ff61ad3ccf169/manifests/modules/packer/manifests/vsphere/networking.pp#L22-L34

Skipped dhcp test on rhel-8

### DIFF
--- a/acceptance/tests/facts/networking_facts.rb
+++ b/acceptance/tests/facts/networking_facts.rb
@@ -46,7 +46,7 @@ test_name 'C59029: networking facts should be fully populated' do
       expected_bindings.delete("networking.interfaces.#{primary_interface}.bindings6.0.network")
     end
 
-    if agent['platform'] =~ /aix|sparc|cisco|huawei|sles|s390x/
+    if agent['platform'] =~ /aix|sparc|cisco|huawei|sles|s390x|el-8/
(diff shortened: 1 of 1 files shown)


gimmyxd:
passing build: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/491/

caseywilliams:
I suspect that redhat 8 uses systemd-networkd to manage dhcp instead of dhclient, and that there is some other way to find this inforamtion, but I don't have time to investigate right this second - I hope we can get this fact to work instead of skipping the test.

McdonaldSeanp:
@gimmyxd can you open a facter ticket to solve the actual bug?

McdonaldSeanp:
@gimmyxd nevermind, I just realized this PR already linked to a facter ticket

caseywilliams:
This deserves another ticket, since FACT-1905 was related to CI failures. I've created FACT-1906 to track this.

PR #1772, opened 2019-03-08
(FACT-1910) Remove check that mount point is physical/tmpfs

josephvoss:
While running puppet I noticed that overlayfs mountpoints are not being picked up by facter. Looking through the code it appears that only physical and tmpfs mounts are currently being loaded into the mountpoint facts (https://github.com/puppetlabs/facter/blob/master/lib/src/facts/linux/filesystem_resolver.cc#L79-L82).

 

For me, the expected return for the mountpoint facts are all mount points, not just physical or tmpfs mounts. It appears that others have a similar desire (FACT-1519). Is it possible to remove this check? It was necessary when the partition data was populated within the filesystem_resolver::collect_mountpoint_data function, but it has been since moved to a separate function and is no longer necessary. This was also the original suggestion for FACT-1418.

### DIFF
--- a/lib/src/facts/linux/filesystem_resolver.cc
+++ b/lib/src/facts/linux/filesystem_resolver.cc
@@ -77,8 +77,8 @@ namespace facter { namespace facts { namespace linux {
             string device = ptr->mnt_fsname;
             string mtype = ptr->mnt_type;
 
-            // Skip over anything that doesn't map to a device and is not tmpfs
-            if (!boost::starts_with(device, "/dev/") && mtype != "tmpfs") {
+            // Skip over any non-tmpfs mount under /proc or /sys
+            if (mtype != "tmpfs" && (boost::starts_with(ptr->mnt_dir, "/proc") || boost::starts_with(ptr->mnt_dir, "/sys"))) {
                 continue;
             }
 
--- a/lib/tests/facts/linux/filesystem_resolver.cc
+++ b/lib/tests/facts/linux/filesystem_resolver.cc
@@ -1,8 +1,14 @@
 #include <catch.hpp>
+#include <facter/facts/fact.hpp>
+#include <facter/facts/map_value.hpp>
+#include <facter/facts/scalar_value.hpp>
 #include <internal/facts/linux/filesystem_resolver.hpp>
+#include "../../collection_fixture.hpp"
+#include <sstream>
 
 using namespace std;
 using namespace facter::facts::linux;
+using namespace facter::testing;
 
 SCENARIO("blkid output with non-printable ASCII characters") {
     REQUIRE(filesystem_resolver::safe_convert("") == "");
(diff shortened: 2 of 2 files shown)


josephvoss:
> Waiting for CLA signature by @josephvoss
> 
> @josephvoss - We require a Contributor License Agreement (CLA) for people who contribute to Puppet, but we have an easy click-through license with instructions, which is available at https://cla.puppet.com/
> 
> Note: if your contribution is trivial and you think it may be exempt from the CLA, please post a short reply to this comment with details. http://docs.puppet.com/community/trivial_patch_exemption.html

> A patch containing the following items would be trivial if less than 10 lines, and are likely to be trivial even above that threshold:
>
> ...
> * Patches that are purely deletions, such as removal of duplicate information or code that never executes.


I think this change falls under the trivial patch exemption policy. It is purely a deletion to remove a filter and allow the default code path to populate additional facts. Please let me know if you disagree and will require the CLA to be signed. 

Also, the posted link for this policy is broken. Can it be changed to print https://puppet.com/community/trivial-patch-exemption-policy?

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -75,12 +75,6 @@ namespace facter { namespace facts { namespace linux {
         char buffer[4096];
         while (mntent *ptr = getmntent_r(file, &entry, buffer, sizeof(buffer))) {
             string device = ptr->mnt_fsname;
-            string mtype = ptr->mnt_type;
-
-            // Skip over anything that doesn't map to a device and is not tmpfs
-            if (!boost::starts_with(device, "/dev/") && mtype != "tmpfs") {
-                continue;
```
oanatmaria: Hi,
Thank you for your contribution. This change will display the nfs server mounting point but will display things from /sys, /sys/kernel, /proc and this is not really a good thing as you don't want to be flooded will 20 or more mount points related to the kernel. I think is better to change the "if" statement in order to skip things from /sys and /proc directory.

oanatmaria: Also, please rebase your changes to 3.11.x not on master.

josephvoss: I agree, the noise from /proc and /sys are unintentional consequences of this change. I've pushed a new change and rebased this branch against 3.11.x to remove this noise. Please let me know if there are any other changes you would suggest. Thank you for taking the time to review this.

lib/src/facts/linux/filesystem_resolver.cc
```diff
@@ -75,10 +75,9 @@ namespace facter { namespace facts { namespace linux {
         char buffer[4096];
         while (mntent *ptr = getmntent_r(file, &entry, buffer, sizeof(buffer))) {
             string device = ptr->mnt_fsname;
-            string mtype = ptr->mnt_type;
 
-            // Skip over anything that doesn't map to a device and is not tmpfs
-            if (!boost::starts_with(device, "/dev/") && mtype != "tmpfs") {
+            // Skip over any mount under /proc or /sys
+            if (boost::starts_with(device, "/proc") || boost::starts_with(device, "/sys")) {
```
oanatmaria: This won't work also because you are trying to  skip if the device matches the /sys or /proc directory. You can see what fields getmntent_r returns here: http://ftp.linuxfoundation.org/pub/lsb/lsb-openjdk/openjdk7-lsb-overrides/include/mntent.h 

i thinks in order that this fix to work the statement should be : 
`if (!boost::starts_with(device, "/dev/") && mtype != "tmpfs" && mtype.find("nfs") == string::npos) {
                continue;
  }`

Also please rebase on 3.11.x as you are still on puppetlabs:master

josephvoss: This change is to have facter return non-tmpfs mounts not under /dev. Adding an specific exception for nfs type mounts won't address all the problems.

josephvoss:
I've modified the check to not return mounts under /proc or /sys. I had to add a check to still allow tmpfs mounts under these directories to avoid changing current behavior of facter. Please let me know if there are any other changes you'd like to this check.

I've also changed this MR to merge into 3.11.x. Apologies, previously I rebased the commits but didn't change the MR.

oanatmaria:
jenkins please test this

oanatmaria:
Hi Joseph,

I tested your last change and it worked fine. Could you also write a unit test for this change?

josephvoss:
Happy to hear that! I've added some tests to verify the filesystem resolver is functional and that the logic in the mountpoint collector is working okay. Please let me know if you would like any changes to these tests.

josephvoss:
Glad to help, thank you all for taking the time to review this and work with me.

PR #1773, opened 2019-04-03
(maint) Fix build with GCC on FreeBSD

smortex:
We need to include stdexcept to build with GCC on FreeBSD.  Most architecture default to clang where the problem is not shown, but some tier-2 architectures are still built with GCC.

Reference:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235601

### DIFF
--- a/lib/inc/internal/util/freebsd/geom.hpp
+++ b/lib/inc/internal/util/freebsd/geom.hpp
@@ -4,6 +4,7 @@
  */
 #pragma once
 
+#include <stdexcept>
 #include <string>
 #include <vector>
 


smortex:
I see that the Travis-CI tests have never been executed.  Is this blocking the merge of this change?  It seems like a few other PR in the repo has a similar issue.  Is there anything we can do to help with this?

smortex:
@gimmyxd closing/reopening seems to only partially fix the issue: the invalid travis-ci hook (continuous-integration/travis-ci) is still clogged, but a new valid one (Travis CI - Pull Request) was added and run successfully…

I guess the project configuration was changed but these changes are not totally reflected to open issues.  I think I saw similar issues when automating automation through GitLab API on my projects.

PR #1775, opened 2019-05-02
(FACT-1916): fix route parsing on Linux

b4ldr:
This PR removes the assumption that all values in the output of
`ip route show` includes a key value pairs.  #1283 first introduced this
assumption fixing where the `ip` command will optionally show the route type.  

This PR also removes the patch for FACT-1405 introduced
in #1372 as it is no longer needed.  

one could further remove the code from #1464 to address FACT-1394 
as that seems to be caused by the same issue, however i have left that 
in as one dosen't need to parse linkdown routes.

Currently the known cases of values which are not key value pairs are
 * flags: linkdown, pervasive, onlink
 * parameters with optional values:  mtu [lock] value

 We could add something like the following to the loop at
https://github.com/puppetlabs/facter/blob/master/lib/src/facts/linux/networking_resolver.cc#L198
however this may be overkill as the flags and mtu always come after the
`dev` and `src` parameters.

```c++
unordered_set<string> route_flags {
  "linkdown",
  "pervasive",
  "onlink"
}
size_t step = 2;
for (size_t i = dst_idx+1; i < parts.size(); i += step) {
    std::string key(parts[i].begin(), parts[i].end());
    std::string value(parts[i+1].begin(), parts[i+1].end());
    step = 2;
    if route_flags.find(key) {
      step = 1;
      continue;
    } else if (key == "mtu" and value == "lock"){
      step = 3;
      continue;
    }
    if (key == "dev") {
        r.interface.assign(value);
    }
    if (key == "src") {
        r.source.assign(value);
    }
}
```

I also wonder if its worth always passing `-d` to `ip route show` to
ensure the route_type is always included but i'm unsure if that flag is
available on all linux

### DIFF
--- a/lib/src/facts/linux/networking_resolver.cc
+++ b/lib/src/facts/linux/networking_resolver.cc
@@ -159,28 +159,14 @@ namespace facter { namespace facts { namespace linux {
                 return true;
             }
 
-            // remove trailing "onlink" or "pervasive" flags
-            while (parts.size() > 0) {
-                std::string last_token(parts.back().begin(), parts.back().end());
-                if (last_token == "onlink" || last_token == "pervasive")
-                    parts.pop_back();
-                else
-                    break;
-            }
-
-            size_t dst_idx = 0;
-            if (parts.size() % 2 == 0) {
-                std::string route_type(parts[0].begin(), parts[0].end());
-                if (known_route_types.find(route_type) == known_route_types.end()) {
-                    LOG_WARNING("Could not process routing table entry: Expected a destination followed by key/value pairs, got '{1}'", line);
-                    return true;
-                } else {
-                    dst_idx = 1;
-                }
+            // FACT-1282
+            std::string route_type(parts[0].begin(), parts[0].end());
+            if (known_route_types.find(route_type) != known_route_types.end()) {
+                parts.erase(parts.begin());
             }
 
             route r;
-            r.destination.assign(parts[dst_idx].begin(), parts[dst_idx].end());
+            r.destination.assign(parts[0].begin(), parts[0].end());
 
             // Check if we queried for the IPV6 routing tables. If yes, then check if our
             // destination address is missing a ':'. If yes, then IPV6 is disabled since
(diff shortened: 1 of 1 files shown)


GabrielNagy:
jenkins please test this on debian8-64a,debian8-32a,debian9-64a,debian9-32a,fedora28-64a,fedora29-64a,redhat5-64a,centos6-64a,centos6-32a,redhat7-64a,redhat8-64a,redhatfips7-64a,sles11-64a,sles11-32a,sles12-64a,sles15-64a,ubuntu1404-64a,ubuntu1404-32a,ubuntu1604-64a,ubuntu1604-32a,ubuntu1804-64a

ema:
Hi! Any update on this?

GabrielNagy:
Hi @ema, @b4ldr,
Thanks for your contribution and for digging into this. This seems fine to me, though I can't say I fully understand the intricacies of the Linux `ip` command.

Do you want this to get into master? If you use puppet-agent, facter 3.11.x goes into PA 5.5.x, 3.12.x -> 6.0.x, 3.13.x -> 6.4.x and master -> master.

We're probably going to merge this sometime next week either way.

b4ldr:
> Do you want this to get into master? 
Please 

> If you use puppet-agent, facter 3.11.x goes into PA 5.5.x, 3.12.x -> 6.0.x, 3.13.x -> 6.4.x and master -> master.
I'm not sure i fully parse this however we will probably backport the fix once merged and manage our own package

Thanks

GabrielNagy:
@b4ldr cool!

Please amend your commit to have the ticket number enclosed in round brackets (i.e. `(FACT-1916) Fix route parsing on Linux`). We have some automated jobs that care about this type of formatting.

Thanks!

b4ldr:
> Please amend your commit to have the ticket number enclosed in round brackets 
done also rebased

b4ldr:
> If you use puppet-agent, facter 3.11.x goes into PA 5.5.x, 3.12.x -> 6.0.x, 3.13.x -> 6.4.x and master -> master.

after more coffee i think i get what you are saying here. I think it would be be useful and safe to have this fix in all branches.  i guess you have different processes for that, im happy to create separate PR's  against every branch if thats whats needed?

GabrielNagy:
> i guess you have different processes for that, im happy to create separate PR's against every branch if that helps?

It should be fine if you rebase the commit and change the PR base branch to 3.11.x. The commit will automatically promote to master when acceptance suites pass.

GabrielNagy:
:+1:

PR #1777, opened 2019-05-07
(FACT-1921) Fix usage of FACTERDIR env var in java class

montaguethomas:
https://tickets.puppetlabs.com/browse/FACT-1921

### DIFF
--- a/lib/Facter.java.in
+++ b/lib/Facter.java.in
@@ -11,11 +11,11 @@ public class Facter {
         // Look for libfacter in the same place as facter.rb does
         String facterDir = System.getenv("FACTERDIR");
         if (facterDir == null) {
-            facterDir = "${CMAKE_INSTALL_PREFIX}/${LIBFACTER_INSTALL_DESTINATION}";
+            facterDir = "${CMAKE_INSTALL_PREFIX}";
         }
 
         // Load the library
-        System.load(facterDir.replaceAll("/$", "") + "/libfacter.so");
+        System.load(facterDir.replaceAll("/$", "") + "/${LIBFACTER_INSTALL_DESTINATION}/libfacter.so");
     }
 
     /**


MikaelSmith:
Please just submit this to the oldest branch. It can then be merged up to all newer branches.

MikaelSmith:
Thanks for submitting a fix. I guess our merge-up policy isn't listed anywhere, so thanks for trying to apply it everywhere that seemed relevant.

I'm trying to wrap my head around why this didn't work before, because I thought our packaging relied on setting FACTERDIR already.

MikaelSmith:
@RandomNoun7 mind taking a quick look at this? I don't believe it will break any of our existing use, and seems like a more sensible default. Although I think we should only merge the version targeted at Facter 3.6.x and merge up from there.

montaguethomas:
@MikaelSmith the fallback here was to default to the compile time `INSTALL_PREFIX` value and since package I'm betting is normally expecting everything to be located at defined install prefix, this would rarely be hit.

Yea, I didn't see anything about backporting and the CONTRIBUTING doc says to branch from master. I closed all the other PRs but left the 3.6.x PR open.

MikaelSmith:
Ah, and this file isn't really used on Windows where we wouldn't rely on `CMAKE_INSTALL_PREFIX`.

RandomNoun7:
@MikaelSmith Am I reading this right that you don't really need a Windows reviewer anymore?

branan:
AFAIK we don't rely on FACTERDIR *at all* in our normal packages/deployments. Based on what facter.rb does, I /believe/ this patch is correct to make the same FACTERDIR value work for both JVM and MRI users of facter (I totally haven't spun up an environment to test it, though)

MikaelSmith:
jenkins please test this

MikaelSmith:
It looks like a test validates the FACTERDIR behavior. I think lib/tests/fixtures.hpp.in needs a change to how it's setup, and will need special handling when testing on Windows similar to what's done in facter.rb.in.

PR CI isn't testing java, I'm going to see if I can fix that.

I think I'm against changing this in a z-release; it works now, just is inconsistent. So I'm going to close the 3.6.x PR.

MikaelSmith:
Opened with tests fixed at https://github.com/puppetlabs/facter/pull/1785.

PR #1783, opened 2019-05-09
(FACT-1918) Use more permissive ruby binding for facter gem

underscorgan:
Puppet-agent starting with 5.0.0 use ruby >= 2.4.1, so we need to update
the binding to work with newer rubies.

### DIFF
--- a/lib/gemspec.in
+++ b/lib/gemspec.in
@@ -16,6 +16,6 @@ Gem::Specification.new do |s|
   s.require_paths = ["lib"]
   s.summary = "Facter, a system inventory tool"
   s.specification_version = 3
-  s.required_ruby_version = '~> 2.1.7'
+  s.required_ruby_version = '~> 2.1'
 
 end


logicminds:
This still needs to be 2.1 since this would work on anything 2.1-2.7 instead of just 2.4+

underscorgan:
ah, ok. I mostly went for 2.4+ since the version of agent with this facter has ruby 2.4, but I don't know that more restrictive bindings are actually important here.

underscorgan:
@branan @McdonaldSeanp do either of you know if there's a reason for more restrictive ruby bindings for the facter gemspec? Mostly wanting to be permissive if possible so we don't forget to update this again.

logicminds:
Just in case it isn't clear.  `~> 2.1` makes it work with ruby >= 2.1 < 3.0.  Since we are years away from 3.0.  Very safe to bind to ~> 2.1

PR #1785, opened 2019-05-10
(FACT-1921) Fix usage of FACTERDIR env var in java class

MikaelSmith:
The Ruby integration for Facter expects to find `libfacter.so` at `$FACTERDIR/${LIBFACTER_INSTALL_DESTINATION}`. Previously the Java integration expected it at `$FACTERDIR`, so they were inconsistent and could not use the same environment variable value.

Change the java integration to expect libfacter at `$FACTERDIR/${LIBFACTER_INSTALL_DESTINATION}` so the Ruby and Java integration can co-exist.


MikaelSmith:
jenkins please test this

MikaelSmith:
jenkins please test this

PR #1791, opened 2019-06-13
(FACT-1930) Relax ffi dependency for Windows Ruby 2.6

joshcooper:
Ruby facter cannot be installed on Windows Ruby 2.6 because it is pinned to
ffi 1.9.x, and ffi didn't add ruby 2.6 support until 1.10.

Update the Gemfile to require ffi 1.10 or greater, but less than 2. The Gemfile
is used when running facter from source. Also use `to_f` when comparing version
numbers to avoid issues like "10" < "2".

Update `project_data.yaml` to >= 1.9, < 2 for all ruby versions on Windows.
This is used when building facter as a gem.


gimmyxd:
Hi @joshcooper, the change looks good to me but i'm not sure how to build facter 2.x gem. Seems that the last release was some time ago, i'll ask around to see.

PR #1809, opened 2019-07-23
(FACT-1964) Multiple mountpoints on same directory

oanatmaria:


oanatmaria:
Jenkins please test this on redhat7-64a

GabrielNagy:
For this specific issue, I would just do this honestly:

```cpp
// Skip over rootfs
if (mtype == "rootfs") {
    continue;
}
```

Even though the Linux kernel allows multiple mounts on the same mountpoint, this is not something you'd often see in the wild... because there are virtually no reasons one should do it, since data from the first mount is shadowed by subsequent mounts.

```console
$ mkdir first_mount
$ mkdir second_mount
$ touch first_mount/1
$ touch second_mount/2
$ mkdir mountpoint
$ mount --bind first_mount mountpoint
$ ls mountpoint
1
$ mount --bind second_mount mountpoint
$ ls mountpoint
2
```

oanatmaria:
Jenkins please test this

oanatmaria:
Jenkins please test this

PR #1810, opened 2019-07-24
(maint) Facter reports wrong minor version for Debian 10

GeorgeMrejea:
Use /etc/debian_version for release major minor instead of lsb_release starting with Debian 10

### DIFF
--- a/lib/src/facts/linux/operating_system_resolver.cc
+++ b/lib/src/facts/linux/operating_system_resolver.cc
@@ -144,6 +144,9 @@ namespace facter { namespace facts { namespace linux {
 
         auto release = implementation->get_release(result.name, result.distro.release);
         if (!release.empty()) {
+            if (result.name == os::debian) {  // on debian 10 final iso, versions from lsb_release and /etc/debian_version are different
+                result.distro.release = release;
+            }
             result.release = move(release);
             tie(result.major, result.minor) = implementation->parse_release(result.name, result.release);
         }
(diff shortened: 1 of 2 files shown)


GeorgeMrejea:
https://jenkins-master-prod-1.delivery.puppetlabs.net/view/Adhoc/job/platform_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/

lib/src/facts/linux/operating_system_resolver.cc
```diff
@@ -143,6 +143,9 @@ namespace facter { namespace facts { namespace linux {
         }
 
         auto release = implementation->get_release(result.name, result.distro.release);
+        if(result.name == os::debian && !release.empty()){
```
GabrielNagy: You can move this under the `if` below, since we already check for `!release.empty()`.

```cpp
if (!release.empty()) {
	if (result.name == os::debian) {  // on debian 10, lsb_release version is different from /etc/debian_version
		result.distro.release = release;
    }
    result.release = move(release);
    tie(result.major, result.minor) = implementation->parse_release(result.name, result.release);
}
```

GabrielNagy:
Looks good! I think we need to rebase this on 3.11.x, since we want this to get into puppet-agent 5.5.x.

GeorgeMrejea:
https://github.com/puppetlabs/facter/pull/1811

PR #1816, opened 2019-07-31
(FACT-1935) facter ipaddress6 without interface id

Dorin-Pleava:
Prior to this fix, when "facter ipaddress6" was run on windows servers,
it returned the ipv6 and the interface identifier.
Now it correctly returns only the ipaddress without the interface id.

### DIFF
--- a/lib/src/facts/windows/networking_resolver.cc
+++ b/lib/src/facts/windows/networking_resolver.cc
@@ -172,7 +172,9 @@ namespace facter { namespace facts { namespace windows {
                     bool ipv6 = it->Address.lpSockaddr->sa_family == AF_INET6;
 
                     binding b;
-                    b.address = addr;
+
+                    // IpAddress6 contains interface identifier, and should be removed when returning the ipaddress6
+                    b.address = ipv6 ? addr.substr(0, addr.find('%')) : addr;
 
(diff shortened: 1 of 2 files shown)


lib/src/facts/windows/networking_resolver.cc
```diff
@@ -172,7 +172,13 @@ namespace facter { namespace facts { namespace windows {
                     bool ipv6 = it->Address.lpSockaddr->sa_family == AF_INET6;
 
                     binding b;
-                    b.address = addr;
+
+                    // IpAddress6 contains interface identifier, and should be removed when returning the ipaddress6
+                    if (ipv6){
+                        b.address = addr.substr(0, addr.find('%'));
```
GabrielNagy: you could use the ternary operator here: `b.address = ipv6 ? addr.substr(0, addr.find('%')) : addr;` or similar

oanatmaria:
network6 => "fe80::%27" on Windows contains interface identifier as well and it shouldn't

Dorin-Pleava:
jenkins please test this on windows10ent-64a,windows10ent-32a,windows2008r2-64a windows2012r2-64a,windows2012r2_ja-64a%7Blocale=ja%7D,windows2016-64a,windows2019-64a

PR #1817, opened 2019-08-05
Fix virtual and is_virtual facts for Nutanix (AHV) VMs

simonfuhrer:


GabrielNagy:
Hi @simonfuhrer,

Thanks for your contribution. Please prepend your commit title with `(maint)` and then we can merge this.

PR #1818, opened 2019-08-08
(FACT-1968) unpin rake dependency for tests

bastelfreak:
Currently rake is pinned to version 10, which is quite old. This PR
unpins it so it works with the latest ruby version.

### DIFF
--- a/lib/Gemfile
+++ b/lib/Gemfile
@@ -1,9 +1,9 @@
 source ENV['GEM_SOURCE'] || "https://rubygems.org"
 
 group :development, :test do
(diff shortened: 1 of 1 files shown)


bastelfreak:
I didn't execute the testsuite locally. But I verified that I can still compile facter with this patch. I guess that the pinning isn't required anymore.

GabrielNagy:
This is fine since we don't run this in our CI, so it shouldn't break anything.
You might as well unpin all dependencies from there.

Also, please prepend your commit with the ticket ID enclosed in round brackets, like `(FACT-1968)`.

Thanks!

PR #1820, opened 2019-08-09
(FACT-1968) add cmake config for UDEV

bastelfreak:
I'm not sure if this is a bug or not. UDEV is a optional dependency,
like libwhereami. cmake searches for suitable libs if it has a proper
configuration file. This file is missing for UDEV so it will never work.
I don't know much about cmake, but the new config file works perfectly
fine for me.

### DIFF
--- a/cmake/FindUDEV.cmake
+++ b/cmake/FindUDEV.cmake
@@ -1,0 +1,10 @@
+# based on https://gist.github.com/GabrielNagy/09a25f06a431684bbfa07a736f2191cb
+# `find_dependency()` is provided by the FindDependency.cmake file.
+# Facter can build against libudev.h to gather disk facts:
+# https://github.com/puppetlabs/facter/blob/d5507926aca43adcbcc39f8cc5eeddcd81bc241f/lib/src/facts/linux/disk_resolver.cc#L83-L94
+include(FindDependency)
+find_dependency(UDEV DISPLAY "udev" HEADERS "libudev.h" LIBRARIES "udev")
+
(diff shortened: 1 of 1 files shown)


GabrielNagy:
@bastelfreak, it should be enough to have something like this inside of the `FindUDEV` file: https://gist.github.com/GabrielNagy/09a25f06a431684bbfa07a736f2191cb

We already have a `FindDependency.cmake` helper which does most of the job for us.

WIth your commit the library gets detected but the `USE_UDEV` definition is never added, so the [code involving it](https://github.com/puppetlabs/facter/blob/master/lib/src/facts/linux/disk_resolver.cc#L83) is never compiled. `FindDependency.cmake` [adds the definition](https://github.com/puppetlabs/facter/blob/master/cmake/FindDependency.cmake#L49) for us.

bastelfreak:
I tested your patch and updated the commit description:

```
I'm not sure if this is a bug or not. UDEV is a optional dependency,
like libwhereami. cmake searches for suitable libs if it has a proper
configuration file. This file is missing for UDEV so it will never work.

I tested this successfully. Without this change (fact disks.sda):

{
  model => "TOSHIBA DT01ACA3",
  size => "2.73 TiB",
  size_bytes => 3000592982016,
  vendor => "ATA"
}

With this change:

{
  model => "TOSHIBA DT01ACA3",
  serial_number => "532SD5***",
  size => "2.73 TiB",
  size_bytes => 3000592982016,
  vendor => "ATA"
}
```

PR #1833, opened 2019-09-25
(FACT-2004) use also nm lease files for azure fact

ciprianbadescu:
Beside /var/lib/dhcp/dhclient.eth0.leases, now facter will look for
azure specific dhcp option in all dhclient.*lease.* files from
/var/lib/dhcp and /var/lib/NetworkManager directories

### DIFF
--- a/lib/inc/internal/facts/linux/virtualization_resolver.hpp
+++ b/lib/inc/internal/facts/linux/virtualization_resolver.hpp
@@ -30,11 +30,16 @@ namespace facter { namespace facts { namespace linux {
 
         /**
          * Gets whether the machine is running in Azure.
-         * @param facts The fact collection that is resolving facts.
-         * @param leases_file The location of where the leases file exists.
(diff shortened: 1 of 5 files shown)


lib/inc/internal/facts/linux/virtualization_resolver.hpp
```diff
@@ -34,7 +34,8 @@ namespace facter { namespace facts { namespace linux {
          * @param leases_file The location of where the leases file exists.
          * @return Returns "azure" if running on azure, otherwise an empty string.
          */
-        static std::string get_azure(collection& facts, std::string const& leases_file = "/var/lib/dhcp/dhclient.eth0.leases");
+        static std::string get_azure();
```
GabrielNagy: can you update the documentation to reflect the updated `get_azure()` function, and the newly added `get_azure_from_leases_file` function

GabrielNagy: don't we need the `collection& facts` param anymore?

PR #1838, opened 2019-10-15
(FACT-2072) Update the regex to get Infiniband MAC and IP address on CentOS 7.x platform

raj76:
This pr updates the regular expression used to find and replace the mac address as shown by the ifconfig command.The ifconfig output changed on CentOS 7 platforms and the old regular expression did not work anymore. I've changed the conditional logic to look for ib* instead of just ib.

### DIFF
--- a/lib/facter/util/ip.rb
+++ b/lib/facter/util/ip.rb
@@ -161,7 +161,7 @@ module Facter::Util::IP
       output = Facter::Util::IP.ifconfig_interface(interface)
     when 'Linux'
       ifconfig_output = Facter::Util::IP.ifconfig_interface(interface)
-      if interface =~ /^ib.*/ then
+      if interface =~ /^ib/ then
         real_mac_address = get_infiniband_macaddress(interface)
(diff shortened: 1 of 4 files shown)


lib/facter/util/ip.rb
```diff
@@ -161,9 +161,9 @@ def self.get_single_interface_output(interface)
       output = Facter::Util::IP.ifconfig_interface(interface)
     when 'Linux'
       ifconfig_output = Facter::Util::IP.ifconfig_interface(interface)
-      if interface =~ /^ib/ then
+      if interface =~ /^ib.*/ then
```
ciprianbadescu: /^ib/ is equivalent to /^ib.*/ as used here, why is this change needed?

raj76: It was a change I made when trying to figure out the problem. Since they are equivalent this line can be reverted. Thanks!

GabrielNagy: You can revert this line since it makes no difference

raj76: I've reverted this change.

PR #1841, opened 2019-10-18
(FACT-2065) Add FIPS enabled fact for Windows

GabrielNagy:
This commit adds a `fips_enabled` fact for Windows. To get this information, Facter queries the
`HKEY_LOCAL_MACHINE/System/CurrentControlSet/Control/Lsa/FipsAlgorithmPolicy/Enabled`
registry key using Leatherman's `get_registry_dword` function.

A compile guard was also added to ensure we don't compile the file if we don't have the function in our Leatherman build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,6 +153,30 @@ enable_cppcheck()
 # Pull in helper macros for working with leatherman libraries
 include(leatherman)
 
+# Guard to make sure we don't compile what we don't have in Leatherman
+if (WIN32)
+    set(CMAKE_REQUIRED_INCLUDES ${LEATHERMAN_INCLUDE_DIRS})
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -153,6 +153,26 @@ enable_cppcheck()
 # Pull in helper macros for working with leatherman libraries
 include(leatherman)
 
+# Guard to make sure we don't compile what we don't have in Leatherman
```
ciprianbadescu: why is guard enabled only in Windows?

GabrielNagy: we only include the source file if `if ("${CMAKE_SYSTEM_NAME}" MATCHES "Windows")`, so it's redundant to have it on other platforms

PR #1846, opened 2019-11-22
(FACT-2128) Fix mountpoints fact

valia0906:
https://tickets.puppetlabs.com/browse/FACT-2128

### DIFF
--- a/lib/tests/facts/resolvers/filesystem_resolver.cc
+++ b/lib/tests/facts/resolvers/filesystem_resolver.cc
@@ -14,7 +14,7 @@ using namespace facter::testing;
 
 struct test_filesystem_resolver : filesystem_resolver
 {
-    void add_mountpoint(string name, string device, string filesystem, uint64_t size, uint64_t available, reserved, vector<string> options)
+    void add_mountpoint(string name, string device, string filesystem, uint64_t size, uint64_t available, uint64_t reserved, vector<string> options)
     {
         mountpoint mp;
         mp.name = move(name);
--- a/lib/tests/facts/resolvers/filesystem_resolver.cc
+++ b/lib/tests/facts/resolvers/filesystem_resolver.cc
@@ -14,7 +14,7 @@ using namespace facter::testing;
 
 struct test_filesystem_resolver : filesystem_resolver
 {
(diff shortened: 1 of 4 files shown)


GabrielNagy:
Hi @valia0906, thanks for your contribution!

Looks like we were reading `f_bfree` while `df` uses `f_bavail` to report the available space, which takes into account the root-reserved space (see https://stackoverflow.com/a/51107001).

However, the tests are failing because the `free` field of the `mountpoint` struct is not mocked.

I would suggest the following:
* add a `reserved` parameter to the `add_mountpoint` function here https://github.com/puppetlabs/facter/blob/master/lib/tests/facts/resolvers/filesystem_resolver.cc#L17
* set `mp.free` to `available + reserved`
* add the new parameter to the function call here https://github.com/puppetlabs/facter/blob/master/lib/tests/facts/resolvers/filesystem_resolver.cc#L78 (call it with 0 for the tests to pass)

Would you be able to work on this?

valia0906:
I`ve alrady add t opass tests but they failed. Maybe I`ve done something wrong. Can you see?

lib/tests/facts/resolvers/filesystem_resolver.cc
```diff
@@ -14,14 +14,15 @@ using namespace facter::testing;
 
 struct test_filesystem_resolver : filesystem_resolver
 {
-    void add_mountpoint(string name, string device, string filesystem, uint64_t size, uint64_t available, vector<string> options)
+    void add_mountpoint(string name, string device, string filesystem, uint64_t size, uint64_t available, reserved, vector<string> options)
```
GabrielNagy:
```suggestion
    void add_mountpoint(string name, string device, string filesystem, uint64_t size, uint64_t available, uint64_t reserved, vector<string> options)
```

This fails because it's missing the variable type

valia0906:
All tests were successful, thank you for your help.

GabrielNagy:
Looks good! Can you please squash your work into a single commit? Also, please include the ticket number in the commit message (i.e.: `(FACT-2128) Fix mountpoints fact`). We have some automated checks that rely on this format.

valia0906:
I`ve created new pull request with squashed commits.

GabrielNagy:
Continued in https://github.com/puppetlabs/facter/pull/1857

PR #1847, opened 2019-11-22
(FACT-2102) Update GCE metadata endpoint for 2.x

erjohnso:
The Google Compute Engine team will be deprecating the older `v1beta1` and `v0.1` metadata endpoints in April 2020. This is an update to use the correct `v1` endpoint and includes the required custom HTTP header.

Once merged, please update and release a new gem https://rubygems.org/gems/facter/ so that users have time to upgrade before April.

_NOTE: This fix has already been applied to the 3.x version of Facter in [FACT-2018](https://tickets.puppetlabs.com/browse/FACT-2018) and released in version 3.14.5._

### DIFF
--- a/lib/facter/gce/metadata.rb
+++ b/lib/facter/gce/metadata.rb
@@ -17,7 +17,7 @@ module Facter
         Timeout::Error,
       ]
 
-      METADATA_URL = "http://metadata/computeMetadata/v1beta1/?recursive=true&alt=json"
+      METADATA_URL = "http://metadata/computeMetadata/v1/?recursive=true&alt=json"
 
       def initialize(url = METADATA_URL)
         @url = url
@@ -45,7 +45,7 @@ module Facter
 
         begin
           Timeout.timeout(timeout) do
-            body = open(@url, :proxy => nil).read
+            body = open(@url, options={:proxy => nil, "Metadata-Flavor" => "Google"}).read
           end
         rescue *CONNECTION_ERRORS => e
           attempts = attempts + 1


joshcooper:
We'll need to update puppet as well, see https://github.com/puppetlabs/puppet/blob/6.4.x/spec/spec_helper.rb#L166

erjohnso:
Ooh, missed that! Thanks Josh!

I'm guessing that would be a separate Jira ticket to track? Is that something you/Puppet will take on, or is the expectation that I'll take a crack at it?

joshcooper:
@erjohnso it would be great if you could submit a PR to puppet. Just prefix the commit with `(maint)` and target the 6.4.x branch. Thanks!

joshcooper:
@erjohnso I went ahead and submitted a PR here https://github.com/puppetlabs/puppet/pull/7860

erjohnso:
You are awesome, thank you @joshcooper!

lib/facter/gce/metadata.rb
```diff
@@ -45,7 +45,7 @@ def with_metadata_server
 
         begin
           Timeout.timeout(timeout) do
-            body = open(@url, :proxy => nil).read
+            body = open(@url, options={:proxy => nil, "Metadata-Flavor" => "Google"}).read
```
gimmyxd: I think `option=` is unnecessary, as the `open` command accepts a hash as second argument.
https://ruby-doc.org/stdlib-2.6.3/libdoc/open-uri/rdoc/OpenURI.html

erjohnso:
Thank you @gimmyxd - very sorry to have not gotten to your feedback in a timely manner!

To be honest, I'm not a day-to-day ruby developer. When I was reading through OpenURI and testing, I was under the impression that `open` used the Mixin (https://ruby-doc.org/stdlib-2.6.3/libdoc/open-uri/rdoc/OpenURI/OpenRead.html) and thought being explicit with the argument name `options` would be better. I guess since you've merged, it wasn't a big deal to leave it as-is.

Thanks again!

If I could pester - what is the process for releasing a new rubygem that includes this fix? I'm hoping it goes out in plenty of time for users to upgrade before GCP deprecates the old metadata endpoints in April.

gimmyxd:
@erjohnso no worries,  it was mostly a stylish remark, how parameters are treated was not impacted.

We still have a sort of process in place for releasing this version of gem, we'll make sure the gem is released in time.

erjohnso:
Fantastic!

Thank you @gimmyxd and @joshcooper for helping us make sure puppet users are unaffected with this change in GCE.

erjohnso:
@gimmyxd - I hate to pester, but ideally a new 2.x release and gem could be pushed out soon so that users have time to update before GCE's turndown of the old metadata endpoints. Do you have a rough idea of when the new gem will be released?

gimmyxd:
@erjohnso - No worries, i created https://tickets.puppetlabs.com/browse/FACT-2299 to not loose track of this, it will get into our next sprint.

PR #1858, opened 2019-12-04
(FACT-1824) Offline CPUs count as physical CPUs

Dorin-Pleava:
If a CPU is set to offline, the /sys/devices/system/cpu/cpu/topology
folder will be removed, and the topology folder contains informations
needed for processor fact.

Now facter will check if the file that contains informations is readable
before reading it. As the file does not exist for offline CPUs, the file
will not be readable and will not cause the bug.

### DIFF
--- a/lib/src/facts/linux/processor_resolver.cc
+++ b/lib/src/facts/linux/processor_resolver.cc
@@ -74,15 +74,18 @@ namespace facter { namespace facts { namespace linux {
         bool cpu0_valid = false;
 
         lth_file::each_subdirectory(root + "/sys/devices/system/cpu", [&](string const& cpu_directory) {
-            bool at_cpu0 = data.logical_count == 0;
-            data.logical_count++;
-            string id = lth_file::read((path(cpu_directory) / "/topology/physical_package_id").string());
-            boost::trim(id);
-            if (id.empty() || (is_valid_id(id) && cpus.emplace(move(id)).second)) {
-                // Haven't seen this processor before
-                ++data.physical_count;
-                if (at_cpu0) {
-                    cpu0_valid = true;
+            string physical_id_path = (path(cpu_directory) / "/topology/physical_package_id").string();
+            if (lth_file::file_readable(physical_id_path)) {
+                bool at_cpu0 = data.logical_count == 0;
+                data.logical_count++;
+                string id = lth_file::read(physical_id_path);
+                boost::trim(id);
+                if ((is_valid_id(id) && cpus.emplace(move(id)).second)) {
+                    // Haven't seen this processor before
+                    ++data.physical_count;
+                    if (at_cpu0) {
+                        cpu0_valid = true;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


gimmyxd:
@Dorin-Pleava Would be good to have a test for this, since the functionality was broken

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -74,16 +74,19 @@ namespace facter { namespace facts { namespace linux {
...
-            string id = lth_file::read((path(cpu_directory) / "/topology/physical_package_id").string());
-            boost::trim(id);
-            if (id.empty() || (is_valid_id(id) && cpus.emplace(move(id)).second)) {
+            string physical_id_path = (path(cpu_directory) / "/topology/physical_package_id").string();
+            if (lth_file::file_readable(physical_id_path)) {
+              bool at_cpu0 = data.logical_count == 0;
+              data.logical_count++;
+              string id = lth_file::read(physical_id_path);
+              boost::trim(id);
+              if ((is_valid_id(id) && cpus.emplace(move(id)).second)) {
```
luchihoratiu: Indentation here seems to be 4 spaces (instead of the 2 used in your fix).

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -74,16 +74,19 @@ namespace facter { namespace facts { namespace linux {
...
                 // Haven't seen this processor before
                 ++data.physical_count;
                 if (at_cpu0) {
-                    cpu0_valid = true;
+                  cpu0_valid = true;
```
luchihoratiu: Indentation here seems to be 4 spaces (instead of the 2 used in your fix).

lib/src/facts/linux/processor_resolver.cc
```diff
@@ -74,16 +74,19 @@ namespace facter { namespace facts { namespace linux {
...
                 }
+              }
```
luchihoratiu: Indentation here seems to be 4 spaces (instead of the 2 used in your fix).

lib/tests/facts/linux/processor_resolver.cc
```diff
@@ -201,6 +201,42 @@ SCENARIO("resolving processor-specific facts for linux machines") {
...
+
+                // for this test, we want to ensure that Model D and Model E cpus are disabled, meaning the last 2 cpus will not have the topology folder
+                int number_cpu_to_disable = 2;
+                vector<cpu_param> disabled_cpu_params(cpu_params.end() - number_cpu_to_disable, cpu_params.end());
+
+                for (auto& cpu_param : disabled_cpu_params) {
+                  string logical_id = get<1>(cpu_param);
+                  // Remove the topology foder for offline cpus
+                  string topology_folder = root_dir + "/sys/devices/system/cpu"+ "/cpu"+logical_id + "/topology";
+                  fs::remove_all(topology_folder);
```
luchihoratiu: Indentation here seems to be 4 spaces (instead of the 2 used in your fix).

PR #1860, opened 2019-12-11
(maint) fix appveyour

mihaibuzgau:
Since there are multiple commands for the test phase of appveyour run, if a test does not pass, appveyour does not correctly report this.
This PR attempts to solve this.


mihaibuzgau:
fixed in this PR: https://github.com/puppetlabs/facter/pull/1863

PR #1864, opened 2019-12-12
(maint) throw unsupported feature

GeorgeMrejea:


lib/src/ruby/module.cc
```diff
@@ -1,4 +1,4 @@
-#include <internal/ruby/module.hpp>
+918#include <internal/ruby/module.hpp>
```
mihaibuzgau: nit

acceptance/tests/custom_facts/windows_not_expand_command.rb
```diff
@@ -0,0 +1,34 @@
+test_name 'FACT-2054: Custom facts that execute on windows should raise an error' do
```
GabrielNagy:
```suggestion
test_name 'FACT-2054: Execute on Windows with expand => false should raise an error' do
```

a more suggestive test name

acceptance/tests/custom_facts/windows_not_expand_command.rb
```diff
@@ -0,0 +1,34 @@
...
+      on(agent, "rm -rf '#{fact_dir}'")
+    end
+
+    step "Agent: Verify that exception is raised" do
+      begin
+        on(agent, facter('foo', :environment => env))
+      rescue Beaker::Host::CommandFailure => ex
+        assert(ex.message.include?("Unsupported argument on Windows expand"))
+      ensure
+        assert(ex != nil, "I should always raise an error")
```
GabrielNagy: instead of a begin..rescue you can pass `:acceptable_exit_codes = [0, 1]` to `on`, then assert the output of stderr

i.e.

```ruby
on(agent, facter('foo', :environment => env), :acceptable_exit_codes => [0, 1]) do |output|
  assert_match(/Unsupported argument on Windows expand/, output.stderr, 'expand => false on Windows should raise')
end
```

acceptance/tests/custom_facts/windows_not_expand_command.rb
GeorgeMrejea: excelent

acceptance/tests/custom_facts/windows_not_expand_command.rb
GeorgeMrejea: excelent

PR #1867, opened 2020-01-15
(maint) Enhance FreeBSD os.release facts

smortex:
So far, facter reported the following for FreeBSD `os.release` fact:

```
{
  full => "12.1-RELEASE-p1",
  major => "12",
  minor => "1-RELEASE-p1"
}
```

The `full` and `major` information definitively makes sense, but the `minor` field is a bit confusing because if contains 3 different information:
 1. The actual minor release;
 2. The "branch" this FreeBSD release commes from (one of "CURRENT", "STABLE" or "RELEASE-pX");
 3. The "patchlevel" if the branch is a RELEASE one.

Add two new fields to store the branch and patchlevel so that all valid FreeBSD versions report nice and consistent data.

```
{
  branch => "RELEASE-p1",
  full => "12.1-RELEASE-p1",
  major => "12",
  minor => "1",
  patchlevel => "1"
}
```

### DIFF
--- a/lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
+++ b/lib/inc/internal/facts/resolvers/operating_system_resolver.hpp
@@ -61,6 +61,22 @@ namespace facter { namespace facts { namespace resolvers {
         };
 
         /**
+         * Represents information about FreeBSD.
+         */
+        struct freebsd_data
+        {
+            /**
+             * Stores FreeBSD branch.
+             */
+            std::string branch;
+
+            /**
+             * Stores FreeBSD patchlevel.
+             */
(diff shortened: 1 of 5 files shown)


smortex:
I am not sure about where to store these new facts.  In a way, `os.release` makes sense, but some OS provide `os.macosx.*` / `os.windows.*`, so we can also imagine keeping `os.release` unchnaged and provide these facts as `os.freebsd.*`…

Your feedback is appreciated :wink:

branan:
We generally consider the os-specific fact sets to be a bad idea these days.

I'm :+1: to what you have here.

PR #1869, opened 2020-01-16
(maint) Install bundle deps before running specs

GabrielNagy:
Facter has some spec tests which it decides to run based on whether it has ruby and bundler installed.

If bundler is installed, Facter goes ahead and runs `bundle exec rspec` which fails if dependencies were not installed beforehand.

Stick a `bundle install` before running the tests to ensure we have the necessary dependencies installed.


GabrielNagy:
```
    Start 2: libfacter specs

2: Test command: /bin/sh "-c" "/usr/bin/bundle" "install" "&&" "/usr/bin/bundle" "exec" "rspec"
2: Test timeout computed to be: 10000000
2: Don't run Bundler as root. Bundler can ask for sudo if it is needed, and installing your bundle as root will break this application for all non-root users on this machine.
2: Using rake 10.1.1
2: Using bundler 1.17.2
2: Using diff-lcs 1.1.3
2: Using metaclass 0.0.4
2: Using mocha 0.10.5
2: Using rspec-core 2.11.1
2: Using rspec-expectations 2.11.3
2: Using rspec-mocks 2.11.3
2: Using rspec 2.11.0
2: Bundle complete! 3 Gemfile dependencies, 9 gems now installed.
2: Use `bundle info [gemname]` to see where a bundled gem is installed.
2/3 Test #2: libfacter specs ..................   Passed    0.96 sec
```

PR #1871, opened 2020-01-20
(maint) Merge 3.11.x to 3.13.x

luchihoratiu:


GabrielNagy:
Those are valid failures, we need to conditionally chain the commands for Windows also...

GabrielNagy:
https://github.com/puppetlabs/facter/pull/1872 should fix the problem

PR #1877, opened 2020-01-30
(FACT-1934) Add ssh fact for Windows OpenSSH feature

GeorgeMrejea:


lib/src/facts/windows/ssh_resolver.cc
```diff
@@ -0,0 +1,128 @@
...
+        // Since there is no user associated with the sshd service, the host keys are stored under \ProgramData\ssh.
+        string sshPath = ((path(dataPath) / "ssh").string());
+        static vector<string> const search_directories = {
+                sshPath
+        };
+
+        // Search the directories for the fact's key file
+        path key_file;
+        for (auto const& directory : search_directories) {
+            key_file = directory;
```
GabrielNagy: I know you copied the code from the posix ssh_resolver, but we don't really need to iterate over a single directory here...

lib/src/facts/windows/ssh_resolver.cc
```diff
@@ -0,0 +1,128 @@
...
+namespace facter { namespace facts { namespace windows {
+
+    ssh_resolver::data ssh_resolver::collect_data(collection& facts)  {
+            data result;
+            populate_key("ssh_host_rsa_key.pub", 1, result.rsa);
+            populate_key("ssh_host_dsa_key.pub", 2, result.dsa);
+            populate_key("ssh_host_ecdsa_key.pub", 3, result.ecdsa);
+            populate_key("ssh_host_ed25519_key.pub", 4, result.ed25519);
+            return result;
+        }
```
GabrielNagy: please use 4 spaces for indentation, that seems to be the standard here

lib/src/facts/windows/ssh_resolver.cc
```diff
@@ -0,0 +1,128 @@
...
+
+    void ssh_resolver::populate_key(std::string const& filename, int type, ssh_key& key)
+    {
+        string dataPath;
+        if (!environment::get("programdata", dataPath)) {
+                   LOG_DEBUG("error finding programdata: {1}",
+                           leatherman::windows::system_error());
+               }
```
GabrielNagy: please use 4 spaces for indentation, that seems to be the standard here

lib/src/facts/windows/ssh_resolver.cc
GeorgeMrejea: Agree, for moment Windows allows for this feature only one directory, will remove iteration

lib/src/facts/windows/ssh_resolver.cc
GeorgeMrejea: done

lib/src/facts/windows/ssh_resolver.cc
GeorgeMrejea: done

lib/src/facts/windows/ssh_resolver.cc
```diff
@@ -0,0 +1,122 @@
...
+        return result;
+    }
+
+    void ssh_resolver::populate_key(std::string const& filename, int type, ssh_key& key)
+    {
+        string dataPath;
+        if (!environment::get("programdata", dataPath)) {
+                LOG_DEBUG("error finding programdata: {1}",
+                leatherman::windows::system_error());
+        }
```
GabrielNagy: still have extra spaces here. also, you can fit the contents in one line:
`LOG_DEBUG("error finding programdata: {1}", leatherman::windows::system_error());`

lib/inc/internal/facts/windows/ssh_resolver.hpp
```diff
@@ -0,0 +1,24 @@
+#pragma once
```
ciprianbadescu: This file is identical to the one in posix namespace. Can both me merge in a single file in parent directory?

lib/CMakeLists.txt
```diff
@@ -117,6 +117,9 @@ if (WIN32)
         "src/facts/windows/collection.cc"
         "src/util/windows/wsa.cc"
     )
+    if (OPENSSL_FOUND)
```
ciprianbadescu: this is now duplicate with condition at line 103, could this be executed once outside of UNIX/WIN32 checks?

lib/src/facts/windows/ssh_resolver.cc
```diff
@@ -0,0 +1,122 @@
...
+using namespace boost::filesystem;
+
+namespace bs = boost::system;
+namespace lth_file = leatherman::file_util;
+
+using namespace leatherman::util;
+
+namespace facter { namespace facts { namespace windows {
+
+    ssh_resolver::data ssh_resolver::collect_data(collection& facts)
```
ciprianbadescu: check if common code can be extracted in a file in parent dir instead duplicate it, similar to collection.cc class

lib/tests/CMakeLists.txt
```diff
@@ -75,7 +75,7 @@ if (UNIX)
         "util/posix/scoped_descriptor.cc"
     )
     if (OPENSSL_FOUND)
```
ciprianbadescu: same as src CMakeLists.txt

lib/inc/internal/facts/windows/ssh_resolver.hpp
GeorgeMrejea: I have took it under consideration, but refactoring to have single namespace might introduce some problem. I would avoid modify existing posix code

lib/CMakeLists.txt
luchihoratiu: Agreed and done.

lib/tests/CMakeLists.txt
luchihoratiu: Done.

lib/inc/internal/facts/windows/ssh_resolver.hpp
luchihoratiu: Managed to do so, no problems seen so far, but I'm honestly not happy about the end result. This, combined with the extraction of common code in base class, made an ugly diff (also noisy due to changing paths/namespaces).
@ciprianbadescu : Please have a look at the 2nd commit and let me know if this is what you had in mind. Other suggestions/approaches are very welcomed! 😄

lib/src/facts/windows/ssh_resolver.cc
luchihoratiu: Done.

PR #1888, opened 2020-02-12
(FACT-2346) Check for `ai_canonname` before comparing

GabrielNagy:
On OS X (and probably more BSD-based systems), calling `getaddrinfo` with a numeric value does not fill the `ai_canonname` field of the `addrinfo` struct, causing any facter call that queries the networking resolver to segfault. This happens if the user has a fully numeric hostname.

From the `getaddrinfo` manpage:

> If the AI_CANONNAME bit is set, a successful call to getaddrinfo()
> will return a NUL-terminated string containing the canonical name of
> the specified hostname in the ai_canonname element of the first
> addrinfo structure returned.

Side-note: `getaddrinfo` will treat any numeric value as an IP, so numeric hostnames cannot be reliably looked up. The GNU libc implementation of `getaddrinfo` always fills the `ai_canonname` field, so this hasn't been a problem on Linux systems.

### DIFF
--- a/acceptance/tests/facts/osx_numeric_hostname.rb
+++ b/acceptance/tests/facts/osx_numeric_hostname.rb
@@ -1,0 +1,23 @@
+test_name 'Querying the fqdn with a numeric hostname should not fail' do
+
+  # calling getaddrinfo with a numeric value on OS X does not fill the
+  # ai_canonname field of the addrinfo structure
+  confine :to, :platform => /^osx-/
+
+  agents.each do |agent|
+    original_hostname = agent.hostname.split('.').first
+    numeric_hostname = 42
+
+    teardown do
+      on(agent, "scutil --set HostName #{original_hostname}")
+    end
+
+    step "Change hostname from '#{original_hostname}' to '#{numeric_hostname}'" do
+      on(agent, "scutil --set HostName #{numeric_hostname}")
+    end
+
+    step 'Verify fqdn fact does not fail' do
+      on(agent, facter('fqdn'))
+    end
+  end
+end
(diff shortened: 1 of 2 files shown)


GabrielNagy:
jenkins please test this on aix61-POWERf,amazon6-64,amazon7-64,debian9-32,debian10-64,fedora31-64,osx1014-64,osx1015-64,redhat5-64,centos6-64,centos6-32,centos8-64,redhat7-64,redhat8-64,redhatfips7-64,sles12-64,sles12-POWER,sles15-64,solaris10-64,solaris11-64,solaris114-64,ubuntu1404-64,ubuntu1804-64,windows10ent-64a

acceptance/tests/facts/osx_numeric_hostname.rb
```diff
@@ -0,0 +1,23 @@
+test_name 'Querying the fqdn with a numeric hostname should not fail' do
+
+  # calling getaddrinfo with a numeric value on OS X does not fill the
+  # ai_canonname field of the addrinfo structure
+  confine :to, :platform => /^osx-/
+
+  agents.each do |agent|
+    original_hostname = agent.hostname.split('.').first
+    numeric_hostname = rand(999999)
```
ciprianbadescu: I would recommend to have a fixed value here, and nightly/CI tests should be deterministic

acceptance/tests/facts/osx_numeric_hostname.rb
```diff
@@ -0,0 +1,23 @@
+test_name 'Querying the fqdn with a numeric hostname should not fail' do
+
+  # calling getaddrinfo with a numeric value on OS X does not fill the
+  # ai_canonname field of the addrinfo structure
+  confine :to, :platform => /^osx-/
+
+  agents.each do |agent|
+    original_hostname = agent.hostname.split('.').first
```
ciprianbadescu: `agent.hostname` is only hostname or full fqdn?

GabrielNagy: Thought it was hostname at first, but after testing I found out that it's actually a fqdn.

PR #1890, opened 2020-02-20
(FACT-2390) Use thread local to check recursion

pcarlisle:


lib/facter/util/fact.rb
```diff
@@ -26,13 +26,13 @@ class Facter::Util::Fact
   # @api private
   def initialize(name, options = {})
     @name = name.to_s.downcase.intern
+    @recursion_key = "facter_searching_#{@name}".intern
```
mwaggett: Why is this not a boolean to start?

Magisus: Because we use it as a key to a value in the map of current thread vars. We don't change it to a boolean later, but rather the value it points to.

mwaggett: oh oh oh wow, reading is hard 🤦‍♀

PR #1900, opened 2020-03-24
(maint) Ubuntu 20.04 adaptation

ciprianbadescu:
On Ubuntu 20.04, os.distro.description returns codename instead
'Ubuntu <version>' on older versions

### DIFF
--- a/acceptance/lib/facter/acceptance/base_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/base_fact_utils.rb
@@ -492,7 +492,7 @@ module Facter
         expected_facts = {
             'os.architecture'          => os_arch,
             'os.distro.codename'       => /\w+/,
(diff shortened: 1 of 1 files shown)


acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -492,7 +492,7 @@ def ubuntu_expected_facts(agent)
         expected_facts = {
             'os.architecture'          => os_arch,
             'os.distro.codename'       => /\w+/,
-            'os.distro.description'    => /Ubuntu #{os_version}/,
+            'os.distro.description'    => agent['platform'] == 'ubuntu-20.04-amd64' ? 'Ubuntu Focal Fossa (development branch)' : /Ubuntu #{os_version}/,
```
gimmyxd: I think something similar happened with Debian 10 beta, but it was changed at release.

PR #1901, opened 2020-03-25
(FACT-2489) external facts files log update

ciprianbadescu:
When there is no appropriate external facts resolver for a file
in `puppet/cache/facts.d` or other external facts dir, a warning
message is logged. This leads to many warning messages in case
multi OS external facts files are in the same directory.

With this change, a debug log message is  logged instead a warning
in case there is no appropriate external facts resolver selected.

### DIFF
--- a/lib/src/facts/collection.cc
+++ b/lib/src/facts/collection.cc
@@ -174,11 +174,7 @@ namespace facter { namespace facts {
                 auto resolver = erf.get_resolver(path);
                 files.push_back(make_pair(path, resolver));
             } catch (external::external_fact_no_resolver& e) {
-                if (warn) {
-                    LOG_WARNING("skipping file \"{1}\": {2}", path, e.what());
-                } else {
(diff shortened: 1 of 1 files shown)


GabrielNagy:
Might wanna target this to 3.11.x, otherwise this is :+1:

lib/src/facts/collection.cc
```diff
@@ -174,11 +174,7 @@ namespace facter { namespace facts {
                 auto resolver = erf.get_resolver(path);
                 files.push_back(make_pair(path, resolver));
             } catch (external::external_fact_no_resolver& e) {
-                if (warn) {
-                    LOG_WARNING("skipping file \"{1}\": {2}", path, e.what());
-                } else {
-                    LOG_DEBUG("skipping file \"{1}\": {2}", path, e.what());
-                }
+              LOG_DEBUG("skipping file \"{1}\": {2}", path, e.what());
```
GabrielNagy: minor nit: code here seems to be indented by 4 spaces

PR #1902, opened 2020-03-26
(maint) Run tests on facter-ng

Filipovici-Andrei:


b4ldr:
I'm curious why would you want to do this.  this would mean that acceptance test for this project depend on the master branch of an  external resource

mihaibuzgau:
@b4ldr its just for testing, this PR will probably never get merged.

PR #1905, opened 2020-04-10
(FACT-2545) execute `oslevel` with default options

ciprianbadescu:
redirecting `stderr` to `stdout` in not working for `oslevel` command
executed on old AIX boxes

other commands are redirecting `stderr` to `null`, oslevel` should
do the same

### DIFF
--- a/lib/src/facts/aix/kernel_resolver.cc
+++ b/lib/src/facts/aix/kernel_resolver.cc
@@ -15,7 +15,7 @@ namespace facter { namespace facts { namespace aix   {
     {
         data result;
 
-        auto exec = execute("/usr/bin/oslevel", {"-s"}, 0, { execution_options::trim_output, execution_options::redirect_stderr_to_stdout, execution_options::merge_environment });
+        auto exec = execute("/usr/bin/oslevel", {"-s"});
 
         result.name = "AIX";
         result.release = exec.output;


lib/src/facts/aix/kernel_resolver.cc
```diff
@@ -15,7 +15,7 @@ namespace facter { namespace facts { namespace aix   {
     {
         data result;
 
-        auto exec = execute("/usr/bin/oslevel", {"-s"}, 0, { execution_options::trim_output, execution_options::redirect_stderr_to_stdout, execution_options::merge_environment });
+        auto exec = execute("/usr/bin/oslevel", {"-s"});
```
gimmyxd, line 18: Should we keep `trim_output` and `:merge_environment`?

ciprianbadescu: They should a be [set by default](https://github.com/puppetlabs/leatherman/blob/1.4.x/execution/inc/leatherman/execution/execution.hpp#L276):
```
    /**
     * Executes the given program.
     * @param file The name or path of the program to execute.
     * @param arguments The arguments to pass to the program. On Windows they will be quoted as needed for spaces.
     * @param timeout The timeout, in seconds. Defaults to no timeout.
     * @param options The execution options.  Defaults to trimming output, merging the environment, and redirecting stderr to null.
     * @return Returns a result struct.
     */
    result execute(
        std::string const& file,
        std::vector<std::string> const& arguments,
        uint32_t timeout = 0,
        lth_util::option_set<execution_options> const& options = { execution_options::trim_output, execution_options::merge_environment, execution_options::redirect_stderr_to_null });

```

ciprianbadescu: @mihaibuzgau, leatherman always shows both `stdout` and `stderr` in debug:
```
2020-04-13 10:42:08.870846 DEBUG leatherman.execution:93 - executing command: /usr/bin/uname -p
2020-04-13 10:42:08.871315 DEBUG !!! - stderr message
2020-04-13 10:42:08.871353 DEBUG | - stdout message
```

PR #1907, opened 2020-04-16
(FACT-2537) Add guard for Leatherman function

GabrielNagy:
Add guard for `leatherman::execution::expand_command`.

Fix readme typo and update minimum required version of Leatherman. 1.4.3 is needed due to the change in puppetlabs/leatherman#281. Bump severity of related acceptance tests, so they run in CI.

Passing acceptance:
image

### DIFF
--- a/lib/src/ruby/module.cc
+++ b/lib/src/ruby/module.cc
@@ -946,8 +946,13 @@ namespace facter { namespace ruby {
 
         // Expand the command only if expand is true,
         std::string expanded;
-        try{
+        try {
+#ifdef HAS_LTH_EXPAND
             expanded = expand_command(command, leatherman::util::environment::search_paths(), expand);
+#else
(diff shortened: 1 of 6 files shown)


lib/src/ruby/module.cc
```diff
@@ -946,8 +946,13 @@ namespace facter { namespace ruby {
 
         // Expand the command only if expand is true,
         std::string expanded;
-        try{
+        try {
+#ifdef HAS_LTH_EXPAND
             expanded = expand_command(command, leatherman::util::environment::search_paths(), expand);
+#else
+            expanded = expand_command(command);
+            LOG_DEBUG("This version of Leatherman cannot expand shell builtins. Falling back to the default behavior.");
```
ciprianbadescu, line 954: We could point that is the Leatherman version used while compiling, just an update of Leatherman will not help.

PR #1909, opened 2020-04-22
(BKR-1639) Updated to use specific methods for chmod, rm_rf, mkdir_p

sebastian-miclea:
Updated tests to use specific methods defined in beaker.
mkdir_p - creating directories
rm_rf - deleting files/dirs
chmod - changing permissions
modified_at - updating the modified_date on a file (this will also create the file if does not exist)


sebastian-miclea:
Moved here https://github.com/puppetlabs/facter/pull/1910

PR #1929, opened 2020-06-19
(FACT-2212) Networking facts for OSX 

Filipovici-Andrei:


lib/resolvers/utils/networking.rb
```diff
@@ -0,0 +1,70 @@
...
+        end
+
+        def find_valid_binding(bindings)
+          bindings.each do |binding|
+            return binding unless ignored_ip_address(binding[:address])
+          end
+          bindings.empty? ? nil : bindings.first
+        end
+
+        def ignored_ip_address(addr)
```
IrimieBogdan, line 52: Should this be private?

lib/resolvers/macosx/networking.rb
```diff
@@ -0,0 +1,109 @@
...
+          end
+
+          def clean_up_interfaces_response(response)
+            # convert ip ranges into single ip. eg. 10.16.132.213 -->  10.16.132.213 is converted to 10.16.132.213
+            response.gsub!(/(\d+(\.\d+)*)\s+-->\s+\d+(\.\d+)*/, '\\1')
+          end
+
+          def parse_interfaces_response(response)
+            parsed_interfaces_data = {}
+            interfaces_data = Hash[*response.split(/^([A-Za-z0-9_]+): /)[1..-1]].sort
```
IrimieBogdan, line 56: The `sort` is not needed. The sorting is done when building the fact collection for all facts.

lib/resolvers/macosx/networking.rb
```diff
@@ -0,0 +1,109 @@
...
+              extract_mac(properties, values)
+              extract_ip_data(properties, values)
+
+              parsed_interfaces_data[interface] = values
+            end
+            @fact_list[:interfaces] = parsed_interfaces_data unless parsed_interfaces_data.empty?
+          end
+
+          def extract_mtu(properties, values)
+            mtu = properties.match(/mtu (\d+)/)&.captures&.first&.to_i
```
IrimieBogdan, line 70: `mtu\s+(\d+)` from https://github.com/puppetlabs/facter/blob/f81a7066c98e2c47e0f9f9f17b709ccf34cbf52e/lib/facter/util/ip.rb#L20

lib/resolvers/macosx/networking.rb
```diff
@@ -0,0 +1,109 @@
...
+            values[:mtu] = mtu unless mtu.nil?
+          end
+
+          def extract_mac(properties, values)
+            mac = properties.match(/ether (\S+)/)&.captures&.first
```
IrimieBogdan, line 75: https://github.com/puppetlabs/facter/blob/f81a7066c98e2c47e0f9f9f17b709ccf34cbf52e/lib/facter/util/ip.rb#L18

lib/resolvers/utils/networking.rb
Filipovici-Andrei: This method is used in the Windows networking resolver when retrieving the primary interface.

lib/resolvers/macosx/networking.rb
Filipovici-Andrei: I'm using the sort to give the same sorted output as facter 3. It doesn't happen by default, or am I doing something wrong?

lib/resolvers/macosx/networking.rb
Filipovici-Andrei: Cool, didn't know about them.

PR #1944, opened 2020-06-29
(FACT-2676) fix os identifier for opensuse-leap

Filipovici-Andrei:
Opensuse-leap is incorrectly is incorrectly identified as linux, thus the os family name was wrong


lib/resolvers/os_release_resolver.rb
```diff
@@ -61,6 +62,10 @@ def process_name
                                 @fact_list[:name].split(' ')[0].strip
                               end
         end
+
+        def normalize_opensuse_identifier
+          @fact_list[:identifier] = 'opensuse' if @fact_list[:identifier] =~ /opensuse/
```
oanatmaria, line 67: ``/opensuse/i`` as the regex to not be case sensitive

Filipovici-Andrei: fixed

Filipovici-Andrei:
No other acceptance tests failed for this PR for any of the supported platforms. Link to Jenkins test results: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/__experimental%20automatic/job/experimental_auto_puppet-agent-extra_puppet-agent-integration-suite_adhoc-ad_hoc/1/testReport/

PR #1945, opened 2020-06-29
(FACT-2218) virtual fact for OSX

IrimieBogdan:
Added `virtual` fact. Refactored `system_profiler` resolver so it only reads partial data.


lib/facts/macosx/virtual.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      def check_virtualbox
+        boot_rom_version = Facter::Resolvers::Macosx::SystemProfiler.resolve(:boot_rom_version)
+        return 'virtualbox' if boot_rom_version&.start_with?('VirtualBox')
+      end
+
+      def check_parallels
+        subsystem_vendor_id = Facter::Resolvers::Macosx::SystemProfiler.resolve(:subsystem_vendor_id)
+        return 'parallels' if subsystem_vendor_id&.start_with?('0x1ab8')
```
sebastian-miclea, line 28: Is **0x1ab8** constant? Won't it change on other versions?

IrimieBogdan: This is a constant used since Facter 2.x https://github.com/puppetlabs/facter/blob/f81a7066c98e2c47e0f9f9f17b709ccf34cbf52e/lib/facter/virtual.rb#L59

It is the same for Facter 3.x https://github.com/puppetlabs/facter/blob/56309e95024a5617c95296a309a80770e2ae178e/lib/src/facts/osx/virtualization_resolver.cc#L37

PR #1948, opened 2020-06-30
(FACT-2682) wisable stdout buffering while loading custom facts

ciprianbadescu:
Disable stdout buffering while loading custom facts, similar to `stderr` in `init_stdhandle`
https://github.com/ruby/ruby/blob/9e41a75255d15765648279629fd3134cae076398/win32/win32.c#L2655
This is needed in a specific case:
- run facter from ruby with backticks
- have a custom fact executing external command with backticks

In this case, `\x00` character will be shown on stdout instead of fact output
We suppose that somwhere between ruby(`facter my_fact`)<->c(rb_load)<->ruby(Facter.add)<->c(rb_funcall_passing_block)<->ruby(`echo test`)
stdout gets the wchar end of string that will break it

### DIFF
--- a/lib/src/ruby/ruby.cc
+++ b/lib/src/ruby/ruby.cc
@@ -77,6 +77,16 @@ namespace facter { namespace ruby {
         // https://github.com/ruby/ruby/blob/v2_1_9/ruby.c#L2011-L2022 for comments.
         // The only piece we seem to need out of rb_w32_sysinit is WSAStartup.
         util::windows::wsa winsocket;
+
+        // Disable stdout buffering while loading custom facts, similar to `stderr` in `init_stdhandle`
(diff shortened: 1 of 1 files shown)


ciprianbadescu:
jenkins please test this on windows10ent-64a,windows2012r2-64a,windows2016-64a,windows2019-64a

mihaibuzgau:
Tested this with the following custom fact:
```
Facter.add(:myf) do
  setcode do
    `cmd /c echo bar`.inspect
  end
end
```
The custom fact is correctly displayed:
```
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin>irb
irb(main):001:0> `facter -p myf`
=> "\n\"bar\\n\"\n"
irb(main):002:0> `facter -p myf`
=> "\"bar\\n\"\n"
irb(main):003:0> `facter -p myf --json`
=> "{\n  \"myf\": \"\\\"bar\\\\n\\\"\"\n}\n"
```

PR #1952, opened 2020-07-02
(FACT-2683) Added remaining legacy networking facts for OSX

Filipovici-Andrei:
Also fixed ssh fact which showed an empty hash if the resolver returned an empty list

### DIFF
--- a/lib/facts/macosx/scope6_interfaces.rb
+++ b/lib/facts/macosx/scope6_interfaces.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Facts
+  module Macosx
+    class Scope6Interfaces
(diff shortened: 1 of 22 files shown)


lib/facts/macosx/scope6_interfaces.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Facts
+  module Macosx
+    class Scope6Interfaces
+      FACT_NAME = 'scope6_.*'
```
oanatmaria, line 6: please add some tests for this class

Filipovici-Andrei: Done

PR #1953, opened 2020-07-02
(FACT-2692) Add hypervisors lxc fact

oanatmaria:


spec/facter/facts/linux/hypervisors/lxc_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      it 'returns virtual fact' do
+        expect(fact.call_the_resolver).to be_an_instance_of(Facter::ResolvedFact).and \
+          have_attributes(name: 'hypervisors.lxc', value: value)
+      end
+    end
+
+    context 'when resolver returns docker' do
+      let(:hv) { { docker: { 'id' => 'testid' } } }
+
+      it 'returns virtual fact' do
```
Filipovici-Andrei, line 30: test name should be 'returns nil' or 'returns nil fact'

oanatmaria: Done

PR #1956, opened 2020-07-07
(FACT-2695) implementation for virtualbox hypervisor fact 

Filipovici-Andrei:
Implementation for virtual box fact and dmidecode resolver to retrieve the virtualbox version and revision

### DIFF
--- a/lib/resolvers/dmi_decode.rb
+++ b/lib/resolvers/dmi_decode.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class DmiDecode < BaseResolver
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+
+      class << self
+        private
+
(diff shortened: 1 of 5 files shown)


lib/resolvers/dmi_decode.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Resolvers
+    class DmiDecode < BaseResolver
+      @semaphore = Mutex.new
+      @fact_list ||= {}
+
+      class << self
+        private
+
+        def post_resolve(fact_name)
+          @fact_list.fetch(fact_name) { virtual_box_version(fact_name) }
```
oanatmaria, line 13: This resolver will be used for vmware info discovery as well. I recommend changing the name of this method to something more general like: execute_dmidecode

Filipovici-Andrei: got it

PR #1961, opened 2020-07-10
(maint) Read version from version.rb

florindragos:


lib/facter/custom_facts/version.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
```
gimmyxd, line 1: Do we need this file?

PR #1962, opened 2020-07-10
(FACT-2696) Add vmware fact

IrimieBogdan:
Add vmware fact and update the `dmidecode` resolver to detect vmware version.


oanatmaria:
Please squash your commits.

PR #1966, opened 2020-07-10
(FACT-2703) Added unit test for when the product name is not recognized

Filipovici-Andrei:


Filipovici-Andrei:
jenkins please test this

PR #1969, opened 2020-07-13
(FACT-2703) Return nil when product name is not recognized

Filipovici-Andrei:
When product name is not a known hypervisor, nil is returned.


Filipovici-Andrei:
jenkins please test this

PR #1972, opened 2020-07-14
(FACT-2666) Add load_external ruby method

Dorin-Pleava:
The Facter.load_external(<boolean>) will set a flag in facter that will
cause the run to skip loading of external facts.

This method is usefull when we only want to get core facts, and we
do not want external facts to be loaded.
Puppet uses this setting when evaluating default settings that require
Facter to provide the core fact value. Loading the external facts is not
needed.

### DIFF
--- a/lib/spec/unit/facter_spec.rb
+++ b/lib/spec/unit/facter_spec.rb
@@ -133,4 +133,27 @@ describe Facter do
     end
   end
 
+  describe '.load_external' do
+    context 'when set to false' do
+      it 'skips resolving external facts' do
+        Facter.search_external([
+          File.expand_path('../../../lib/tests/fixtures/facts/external/yaml'),
+        ])
+        Facter.load_external(false);
+        facts = Facter.to_hash
+        expect(facts['yaml_fact1']).to be_nil
+      end
+    end
+
+    context 'when set to true' do
+      it 'resolves external facts' do
+        Facter.search_external([
+          File.expand_path('../../../lib/tests/fixtures/facts/external/yaml'),
+        ])
+        Facter.load_external(true);
+        facts = Facter.to_hash
+        expect(facts['yaml_fact1'] == "foo")
+      end
+    end
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/src/ruby/module.cc
```diff
@@ -475,15 +477,21 @@ namespace facter { namespace ruby {
         if (_collection.empty()) {
             bool include_ruby_facts = true;
             _collection.add_default_facts(include_ruby_facts);
-            _collection.add_external_facts(_external_search_paths);
-
+            if (_load_external && !_external_facts_loaded) {
+                _collection.add_external_facts(_external_search_paths);
+                _external_facts_loaded = true;
+            }
```
gimmyxd, line 483: is this needed, as it will always get evaluated in :491?

lib/src/ruby/module.cc
```diff
@@ -475,15 +477,21 @@ namespace facter { namespace ruby {
         if (_collection.empty()) {
```
gimmyxd, line 477: Some debug messages would be useful to know if external facts loading was disabled on purpose

lib/src/ruby/module.cc
Dorin-Pleava: Yep, it was not needed.

lib/src/ruby/module.cc
Dorin-Pleava: Added some debug messages in the `module::ruby_load_external` method as the `collection& module::facts()` method could spam the debug log.

lib/spec/unit/facter_spec.rb
```diff
@@ -131,6 +131,18 @@
...
+      Facter.search_external([
+        File.expand_path('../../../lib/tests/fixtures/facts/external/yaml'),
+      ])
+      Facter.load_external(false);
+      facts = Facter.to_hash
+      Facter.load_external(true);
+      expect(facts['yaml_fact1']).to be_nil
+      facts = Facter.to_hash
+      expect(facts['yaml_fact1'] == "foo")
+    end
```
gimmyxd, line 145: Can you split this in two examples? each having its own expect? One to test `Facter.load_external(false)` and the other `Facter.load_external(true)`

```ruby

describe '.load_external' do
  context 'when set to false'
	it '' do
    end
  end
  
  context 'when set to true'
	it '' do
    end
  end
end
```

PR #1977, opened 2020-07-16
(maint) handle Time and Symbol in executable facts

gimmyxd:
Executable facts can produce structured facts if they
produce a YAML output, when parsing the output script,
YAML.safe_load does not know to handle Time and Symbol.

This commit allow YAML to load Time and Symbol from output.


florindragos:

florindragos:

PR #1984, opened 2020-07-21
(FACT-2708) Add man pages

florindragos:


lib/facter/framework/cli/cli.rb
```diff
@@ -103,6 +108,20 @@ class Cli < Thor
...
 
+
+    desc '--man', 'Manual', hide: true
+    map ['--man'] => :man
+    def man(*args)
+      require 'erb'
+      template_name = @options['ronn'] ? 'man.erb' : 'help.erb'
+      negate_options = [ 'block', 'custom_facts', 'external_facts']
+
+      template = Pathname(__FILE__).dirname + "../../templates/man.erb"
```
gimmyxd, line 119: use `File.join` to handle paths correctly on all platforms

Gemfile
```diff
@@ -10,3 +10,9 @@ end
 
 local_gemfile = File.expand_path('Gemfile.local', __dir__)
 eval_gemfile(local_gemfile) if File.exist?(local_gemfile)
+
+group(:documentation) do
+  gem 'ronn', '~> 0.7.3', require: false, platforms: [:ruby]
+end
+
+gem 'pry-byebug'
```
gimmyxd, line 18: forgot pry

tasks/manpages.rake
```diff
@@ -0,0 +1,25 @@
...
+    require 'ronn'
+  rescue LoadError
+    abort('Run `bundle install --with documentation` to install the `ronn` gem.')
+  end
+
+  ronn = `which ronn`.chomp
+
+  abort('Ronn does not appear to be installed') unless File.executable?(ronn)
+
+  `mkdir -p ./man/man8`
```
IrimieBogdan, line 20: We should use `ruby` code to create directories.

PR #1986, opened 2020-07-21
(FACT-2687) --trace shows backtrace for core facts

sebastian-miclea:


Filipovici-Andrei:
jenkins please test this

lib/facter/framework/logging/logger.rb
```diff
@@ -3,10 +3,11 @@
 require 'logger'
 
 module Facter
-  RED = 31
-  GREEN = 32
-  YELLOW = 33
-  CYAN = 36
+  RED     = "\e[31m"
+  GREEN   = "\e[32m"
+  YELLOW  = "\e[33m"
```
oanatmaria, line 8: \e[0;{COLOR_CODE}m https://www.cyberciti.biz/faq/bash-shell-change-the-color-of-my-shell-prompt-under-linux-or-unix/

sebastian-miclea: The test is quite bad, it doesn't actually test that the output has color. It tests that the output text is not formatted.
I'd rather fix the test.

lib/facter/framework/core/options/option_store.rb
```diff
@@ -132,6 +132,7 @@ def show_legacy=(bool)
       end
 
       def trace=(bool)
+        @trace = bool
         Facter.trace(bool)
```
IrimieBogdan, line 136: We should remove `Facter.trace(bool)` and use the options in `facter.rb`

PR #1993, opened 2020-07-23
(FACT-2699) augeas fact for OSX is not resolved

IrimieBogdan:
Detect augeas from gem if augparse is not available.

### DIFF
--- a/lib/facter/resolvers/augeas_resolver.rb
+++ b/lib/facter/resolvers/augeas_resolver.rb
@@ -14,9 +14,28 @@ module Facter
         end
 
         def read_augeas_version(fact_name)
-          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
-          @fact_list[:augeas_version] = Regexp.last_match(1) if output =~ /^augparse (\d+\.\d+\.\d+)/
+          @fact_list[:augeas_version] = read_augeas_from_cli
+          @fact_list[:augeas_version] ||= read_augeas_from_gem
+
           @fact_list[fact_name]
+        end
+
+        def read_augeas_from_cli
+          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
+          Regexp.last_match(1) if output =~ /^augparse (\d+\.\d+\.\d+)/
+        end
+
+        def read_augeas_from_gem
+          require 'augeas'
+
+          return ::Augeas.create { |aug| aug.get('/augeas/version') } if ::Augeas.respond_to?(:create)
+
+          # it is used for legacy augeas <= 0.5.0
+          return ::Augeas.open { |aug| aug.get('/augeas/version') } if ::Augeas.respond_to?(:open)
+        rescue StandardError => e
+          log.debug('ruby-augeas not available')
+          log.log_exception(e)
+          nil
         end
       end
     end
--- a/lib/facter/resolvers/augeas_resolver.rb
+++ b/lib/facter/resolvers/augeas_resolver.rb
@@ -28,14 +28,12 @@ module Facter
         def read_augeas_from_gem
           require 'augeas'
 
-          return ::Augeas.create { |aug| aug.get('/augeas/version') } if ::Augeas.respond_to?(:create)
-
-          # it is used for legacy augeas <= 0.5.0
-          return ::Augeas.open { |aug| aug.get('/augeas/version') } if ::Augeas.respond_to?(:open)
-        rescue StandardError => e
-          log.debug('ruby-augeas not available')
-          log.log_exception(e)
-          nil
+          if Gem.loaded_specs['augeas']
+            ::Augeas.create { |aug| aug.get('/augeas/version') }
+          else
+            # it is used for legacy augeas <= 0.5.0 (ruby-augeas gem)
+            ::Augeas.open { |aug| aug.get('/augeas/version') }
+          end
         end
       end
     end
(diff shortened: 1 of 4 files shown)


lib/facter/resolvers/augeas_resolver.rb
```diff
@@ -14,10 +14,28 @@ def post_resolve(fact_name)
...
+          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
+          Regexp.last_match(1) if output =~ /^augparse (\d+\.\d+\.\d+)/
+        end
+
+        def read_augeas_from_gem
+          require 'augeas'
+          aug = ::Augeas.open('/', nil, ::Augeas::NO_MODL_AUTOLOAD)
+          ver = aug.get('/augeas/version')
+          aug.close
+          ver
```
gimmyxd, line 33: you should use open with a block so the close is always ensured.
Also there is another gem that provides `Augeas` interface: https://rubygems.org/gems/augeas/versions/0.6.4, in which the open methods is private and it was replaced with `.create`.
This will not work on systems where users install facter as a gem and have augeas > 0.5.

facter.conf
```diff
@@ -5,7 +5,7 @@ facts : {
   ]
 }
 global : {
-  external-dir     : [ "./external_facts"],
+  external-dir     : [ "/Users/bogdan.irimie/projects/facter/external_facts"],
```
Filipovici-Andrei, line 8: probably you should revert that

IrimieBogdan: Reverted tha change

lib/facter/resolvers/augeas_resolver.rb
IrimieBogdan: Used `Augeas.create` when `Augeas` responds to `create` (version > 0.5.0) and `Augeas.open` when `Auges` responds to `open` (version <= 0.5.0)

spec/facter/resolvers/augeas_resolver_spec.rb
```diff
@@ -7,12 +7,82 @@
...
+        .with('augparse --version 2>&1', logger: log_spy)
+        .and_return('sh: augparse: command not found')
+    end
+
+    context 'when augeas gem > 0.5.0 is installed' do
+      let(:augeas_mock) { instance_spy(Augeas) }
+
+      before do
+        allow(Augeas).to receive(:respond_to?).with(:create).and_return(true)
+        allow(Augeas).to receive(:respond_to?).with(:create, true).and_return(true)
```
gimmyxd, line 40: Why is this stub needed? `respond_to? :create` is already stubbed above

spec/facter/resolvers/augeas_resolver_spec.rb
```diff
@@ -7,12 +7,82 @@
...
+      end
+    end
+
+    context 'when augeas gem <= 0.5.0 is installed' do
+      let(:augeas_mock) { instance_spy(Augeas) }
+
+      before do
+        allow(Augeas).to receive(:respond_to?).with(:create).and_return(false)
+        allow(Augeas).to receive(:respond_to?).with(:open).and_return(true)
+        allow(Augeas).to receive(:respond_to?).with(:open, true).and_return(true)
```
gimmyxd, line 55: Why is this stub needed? respond_to? :open is already stubbed above

spec/facter/resolvers/augeas_resolver_spec.rb
```diff
@@ -7,12 +7,69 @@
 
   before do
     augeas.instance_variable_set(:@log, log_spy)
-    allow(Facter::Core::Execution).to receive(:execute)
-      .with('augparse --version 2>&1', logger: log_spy)
-      .and_return('augparse 1.12.0 <http://augeas.net/>')
+    allow(Facter::Resolvers::Augeas).to receive(:require).with('augeas')
```
oanatmaria, line 10: do you still need this if you mocked the require from kernel?

lib/facter/resolvers/augeas_resolver.rb
```diff
@@ -14,10 +14,28 @@ def post_resolve(fact_name)
...
+
+        def read_augeas_from_cli
+          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
+          Regexp.last_match(1) if output =~ /^augparse (\d+\.\d+\.\d+)/
+        end
+
+        def read_augeas_from_gem
+          require 'augeas'
+
+          return ::Augeas.create { |aug| aug.get('/augeas/version') } if ::Augeas.singleton_method(:create)
```
gimmyxd, line 31: This will raise error if augeas 0.5.0 is installed.
```ruby
require 'augeas'
>> Gem.loaded_specs['ruby-augeas'].version   #=> #<Gem::Version "0.5.0">
>> ::Augeas.singleton_method(:create)
NameError (undefined singleton method `create' for `Augeas')
```

gimmyxd: You can branch on `Gem.loaded_specs` as there are two gems that can be installed and provide `Augeas`

ruby-augeas(has .open methdod) - https://rubygems.org/gems/ruby-augeas 
augeas(has create methods) - https://rubygems.org/gems/augeas

```ruby
if Gem.loaded_specs['augeas']
  create
else
  open
```

spec/facter/resolvers/augeas_resolver_spec.rb
IrimieBogdan: You are right, this is no longer needed.

lib/facter/resolvers/augeas_resolver.rb
IrimieBogdan: Changed to `Gem.loaded_specs`

PR #1998, opened 2020-07-28
(FACT-2717) Block external facts

florindragos:


lib/facter/custom_facts/util/directory_loader.rb
```diff
@@ -56,6 +56,8 @@ def load_directory_entries(_collection)
         facts = []
         entries.each do |file|
           basename = File.basename(file)
+          next if file_blocked?(basename)
```
IrimieBogdan, line 59: We should have a unit test for this.

acceptance/tests/external_facts/block_external_facts.rb
```diff
@@ -0,0 +1,38 @@
+test_name 'custom facts included in blocklist will not be displayed' do
+  tag 'risk:high'
+
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  agents.each do |agent|
+    os_version = on(agent, facter('kernelmajversion')).stdout.chomp.to_f
```
IrimieBogdan, line 8: `os_version` does not seam to be used

PR #1999, opened 2020-07-28
(FACT-2040) Added solaris memory resolver

sebastian-miclea:


lib/facter/resolvers/solaris/memory.rb
```diff
@@ -0,0 +1,72 @@
...
+
+          def sys
+            sys = {}
+            output = Facter::Core::Execution.execute('/usr/bin/kstat -m unix -n system_pages', logger: log).strip
+            kstats = output.scan(/(physmem|pagesfree)\s+(\d+)/)
+            kstats = kstats.to_h
+            return unless kstats['physmem'] || kstats['pagesfree']
+
+            total = kstats['physmem'].to_i * pagesize
+            free = kstats['pagesfree'].to_i * pagesize
```
oanatmaria, line 32: if pagesize is 0 (empty string - means that pagesize call failed) all the values will be 0. You should add a check for this case.

spec/facter/resolvers/solaris/memory_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        used_bytes: 0
+      )
+    end
+  end
+
+  context 'when there is no output from swap' do
+    let(:kstat_output) { load_fixture('kstat_sys_pages').read }
+    let(:swap_output) { '' }
+
+    it 'returns nil for system' do
```
oanatmaria, line 50: returns info for system and nil for swap

lib/facter/resolvers/solaris/memory.rb
```diff
@@ -0,0 +1,72 @@
...
+
+            output.each_line do |line|
+              swap_sizes = line.match(/(\d+)\s+(\d+)$/)
+              next if swap_sizes.nil?
+
+              total += swap_sizes[1].to_i
+              free += swap_sizes[2].to_i
+            end
+            total *= 512
+            free *= 512
```
oanatmaria, line 56: the magic number (512) should be a constant with a meaningful name

lib/facter/resolvers/solaris/memory.rb
```diff
@@ -0,0 +1,91 @@
...
+          end
+
+          def parse_sys_output(output)
+            kstats = output.scan(/(physmem|pagesfree)\s+(\d+)/)
+            kstats = kstats.to_h
+            return unless kstats['physmem'] || kstats['pagesfree']
+            return unless pagesize != 0
+
+            total = kstats['physmem'].to_i * pagesize
+            free = kstats['pagesfree'].to_i * pagesize
```
Filipovici-Andrei, line 60: I think this will log the pagesize 3 times. Probably 
`unless @fact_list[:pagesize]
     @fact_list[:pagesize] ||= Facter::Core::Execution.execute('pagesize', logger: log).strip.to_i
     log.debug("Pagesize: #{@pagesize}")
@fact_list[:pagesize]
` in the pagesize method will work better

lib/facter/resolvers/solaris/memory.rb
```diff
@@ -0,0 +1,91 @@
...
+
+          def calculate_memory(fact_name)
+            @fact_list = { system: sys, swap: swap }
+
+            @fact_list[fact_name]
+          end
+
+          def sys
+            sys = {}
+            output = Facter::Core::Execution.execute('/usr/bin/kstat -m unix -n system_pages', logger: log).strip
```
igalic, line 27: given that freebsd doesn't have this utility

lib/facter/resolvers/solaris/memory.rb
```diff
@@ -0,0 +1,91 @@
...
+            sys[:available_bytes] = free
+            sys[:used_bytes] = total - free
+            sys[:capacity] = FilesystemHelper.compute_capacity(total - free, total)
+
+            sys
+          end
+
+          def swap
+            swap_hash = {}
+            output = Facter::Core::Execution.execute('/usr/sbin/swap -l', logger: log).strip
```
igalic, line 42: nor this one

we need a freebsd version of this resolver now too

smortex:
@igalic Yes, adding memory and dmi fact is high on my Puppet todo-list: it's the only piece missing so that Facter 4 can be used as a drop-in replacement to Facter 3, at least on my computer :stuck_out_tongue:

PR #2004, opened 2020-07-30
(FACT-2735) virtual not working on EXADATA baremetal

IrimieBogdan:
`each_line` returns the initial string, in our case the output of `lspci`.

```ruby 
 "example string".each_line{ |line| }
 => "example string"
```

In case no hypervisor match is found, we want to return `nil`.


IrimieBogdan:
jenkins please test this on windows2012r2-64a

PR #2016, opened 2020-08-06
(FACT-2739) Extend os hierarchy to consider multiple os families

IrimieBogdan:
When there are multiple families detected, `facter` tries to determine a hierarchy for each of them sequentially. If it finds a hierarchy for one of the families, it stops the search and uses that hierarchy to load facts.


IrimieBogdan:

IrimieBogdan:

PR #2019, opened 2020-08-10
Make *BSD independent from Solaris

smortex:
Facter 4 historically used the following OS hierarchy:

* Solaris
  * BSD
    * FreeBSD

While this is away from reality from an [historical point of view](https://www.levenez.com/unix/), it helped to provide all ZFS facts on FreeBSD (The ZFS code from Sun was made public and borrowed to be included into FreeBSD) in the early days of Facter 4.  However, more and more this weirdness is causing trouble: BSD is in no way a Solaris derivative, and some facts that only make sense on Solaris [break FreeBSD support](db0756f7bd80b434608bb608266285535668997d).  Recently a similar problem occured with the addition of the [Solaris networking support](69fe7701390fb9ad55a8b7a2b788b28d1cca35cc).  For this later case, it is unfortunately not enough to override the Solaris code with BSD one: the FFI failure because of the Solaris specific `uint_t` cannot be easily worked around.

So it's maybe time to make the BSD code independent from the Solaris one…

This PR rework the hierarchy; add back the facts that disappeared during that operation; fix the networking facts and remove obsolete workarounds.

The reintroduced facts are attached to the FreeBSD OS since I currently only run FreeBSD.  I think that most of them could be attached to the BSD OS, but though that since we do not have explicit support for other BSD variants, when someone will add support to another *BSD OS, they will see that the FreeBSD code is relevant and move it to the BSD os.


IrimieBogdan:
@smortex The changes look good, great job on detaching `Solaris` from `BSD`.

.rubocop_todo.yml
```diff
@@ -134,6 +134,10 @@ RSpec/MessageSpies:
     - 'spec/custom_facts/util/resolution_spec.rb'
     - 'spec/facter/facts/aix/os/name_spec.rb'
     - 'spec/facter/facts/aix/os/release_spec.rb'
+    - 'spec/facter/facts/freebsd/os/architecture_spec.rb'
+    - 'spec/facter/facts/freebsd/os/name_spec.rb'
+    - 'spec/facter/facts/freebsd/ruby/platform_spec.rb'
+    - 'spec/facter/facts/freebsd/ruby/sitedir_spec.rb'
```
IrimieBogdan, line 140: These offences should be fixes.

e.g.
```ruby
    it 'calls Facter::Resolvers::Uname' do
      fact.call_the_resolver
      expect(Facter::Resolvers::Uname).to have_received(:resolve).with(:machine)
    end
```

smortex: I amended the relevant commits accordingly, thanks!

lib/facter/resolvers/freebsd/networking_resolver.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Freebsd
+      class Networking < BaseResolver
```
Filipovici-Andrei, line 6: This is the networking resolver for OSX. Could you please move it to lib/facter/resolvers/networking.rb and delete the Freebsd module. Leave only Facter::Resolvers::Networking. Probably it's best to move the one from the resolvers/osx and benefit from the IDE's refactoring capabilities, because this change will break all OSX networking facts.

lib/facter/facts/freebsd/mountpoints.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Facts
+  module Freebsd
+    class Mountpoints
+      FACT_NAME = 'mountpoints'
+
+      def call_the_resolver
+        mountpoints = Facter::Resolvers::Linux::Mountpoints.resolve(FACT_NAME.to_sym)
```
Filipovici-Andrei, line 9: Could you please remove the Linux module from the mountpoints resolver's namespace, because now this resolver is used on two platforms.

lib/facter/facts/freebsd/zfs_featurenumbers.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Facts
+  module Freebsd
+    class ZfsFeaturenumbers
+      FACT_NAME = 'zfs_featurenumbers'
+
+      def call_the_resolver
+        require 'facter/resolvers/solaris/zfs_resolver'
```
Filipovici-Andrei, line 9: Could you please delete this require and move the zfs and zpool resolvers to /lib/facter/resolvers/zfs.rb and /lib/facter/resolvers/zpool.rb. The resolvers in lib/facter/resolvers are loaded for all platforms. The resolvers which reside in different folders inside  /lib/facter/resolvers are loaded only for specific platforms.

lib/facter/resolvers/freebsd/networking_resolver.rb
smortex: Gotcha, `lib/facter/resolvers/windows/networking_resolver.rb` already defines `Facter::Resolvers::Networking`… I'll rename this `Facter::Resolvers::Windows::Networking` for consistency!

smortex:
@Filipovici-Andrei I added commits to fix the issues you raised on top of the old one for easier reviewing.  I can merge them in the relevant commit to cleanup history on request. Thanks!

Filipovici-Andrei:
@smortex Excellent work. You may merge everything into one commit. Then I'll merge your PR.

smortex:
@Filipovici-Andrei I just squash all commits and rebased on top of 4.x

igalic:
either that, or github's review ui is… fuckier than usual
(or i am too tired to be doing this)

smortex:
@igalic it looks like you added comments to the commit without doing a review, so your comments disappeared while I addressed them.  For the record, the topic was about mocked values that did not make sense for FreeBSD but where not adjusted when copying unit tests from another OS to FreeBSD…

I adjusted the values to make them more FreeBSD plausible, thanks for this.

PR #2020, opened 2020-08-10
Add FreeBSD memory facts

smortex:
Collect system and swap memory information on FreeBSD

:gift: @igalic

The memory facts where not provided by Facter 4.  With this PR, Facter returns the same facts as Facter 3 does.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -180,7 +180,7 @@ RSpec/SubjectStub:
     - 'spec/custom_facts/util/fact_spec.rb'
     - 'spec/custom_facts/util/resolution_spec.rb'
 
-# Offense count: 134
+# Offense count: 138
 # Configuration parameters: IgnoreNameless, IgnoreSymbolicNames.
 RSpec/VerifiedDoubles:
   Exclude:
@@ -191,6 +191,10 @@ RSpec/VerifiedDoubles:
     - 'spec/custom_facts/util/fact_spec.rb'
     - 'spec/custom_facts/util/resolution_spec.rb'
     - 'spec/facter/facts/aix/ssh_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/swap/capacity_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/swap/used_bytes_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/system/capacity_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/system/used_bytes_spec.rb'
     - 'spec/facter/facts/macosx/memory/swap/capacity_spec.rb'
     - 'spec/facter/facts/macosx/memory/swap/used_bytes_spec.rb'
     - 'spec/facter/facts/macosx/memory/system/capacity_spec.rb'
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -180,7 +180,7 @@ RSpec/SubjectStub:
     - 'spec/custom_facts/util/fact_spec.rb'
     - 'spec/custom_facts/util/resolution_spec.rb'
 
-# Offense count: 138
+# Offense count: 134
 # Configuration parameters: IgnoreNameless, IgnoreSymbolicNames.
 RSpec/VerifiedDoubles:
   Exclude:
@@ -191,10 +191,6 @@ RSpec/VerifiedDoubles:
     - 'spec/custom_facts/util/fact_spec.rb'
     - 'spec/custom_facts/util/resolution_spec.rb'
     - 'spec/facter/facts/aix/ssh_spec.rb'
-    - 'spec/facter/facts/freebsd/memory/swap/capacity_spec.rb'
-    - 'spec/facter/facts/freebsd/memory/swap/used_bytes_spec.rb'
-    - 'spec/facter/facts/freebsd/memory/system/capacity_spec.rb'
-    - 'spec/facter/facts/freebsd/memory/system/used_bytes_spec.rb'
     - 'spec/facter/facts/macosx/memory/swap/capacity_spec.rb'
     - 'spec/facter/facts/macosx/memory/swap/used_bytes_spec.rb'
     - 'spec/facter/facts/macosx/memory/system/capacity_spec.rb'
--- a/lib/facter/resolvers/freebsd/swap_memory_resolver.rb
+++ b/lib/facter/resolvers/freebsd/swap_memory_resolver.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Freebsd
+      class SwapMemory < BaseResolver
+        @semaphore = Mutex.new
+        @fact_list ||= {}
+        class << self
+          private
+
+          def post_resolve(fact_name)
+            @fact_list.fetch(fact_name) { read_swap_memory(fact_name) }
+          end
+
+          def read_swap_memory(fact_name) # rubocop:disable Metrics/AbcSize
+            output = Facter::Core::Execution.execute('swapinfo -k', logger: log)
+            data = output.split("\n")[1..-1].map { |line| line.split(/\s+/) }
+
+            unless data.empty?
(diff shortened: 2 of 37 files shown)


lib/facter/resolvers/freebsd/system_memory_resolver.rb
```diff
@@ -0,0 +1,48 @@
...
+            read_available_memory_in_bytes
+
+            @fact_list[:used_bytes] = @fact_list[:total_bytes] - @fact_list[:available_bytes]
+            @fact_list[:capacity] = compute_capacity(@fact_list[:used_bytes], @fact_list[:total_bytes])
+
+            @fact_list[fact_name]
+          end
+
+          def read_available_memory_in_bytes
+            output = Facter::Core::Execution.execute('vmstat -H', logger: log)
```
igalic, line 27: briefly wondered if it'd make sense to use `--libxo`

lib/facter/resolvers/freebsd/system_memory_resolver.rb
```diff
@@ -0,0 +1,48 @@
...
+            if (data = output.split("\n")[-1].match(/^\s*\d+\s*\d+\s*\d+\s*\d+\s*(\d+)/))
```
igalic, line 28: but since we're not doing much parsing here, it wouldn't make much sense

anyway, `\s` should also get a ``+` instead of the `*`

lib/facter/resolvers/freebsd/system_memory_resolver.rb
smortex: Makes sense, yes!  Let's do this :wink:

igalic: what hath i wrought?? 😅

spec/facter/facts/freebsd/memory/swap/capacity_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+describe Facts::Freebsd::Memory::Swap::Capacity do
+  describe '#call_the_resolver' do
+    it 'returns a fact' do
+      expected_fact = double(Facter::ResolvedFact, name: 'memory.swap.capacity', value: 1024)
```
IrimieBogdan, line 6: We should use verifying doubles over normal doubles.

```suggestion
      expected_fact = instance_double(Facter::ResolvedFact, name: 'memory.swap.capacity', value: 1024)
```

.rubocop_todo.yml
```diff
@@ -191,6 +191,10 @@ RSpec/VerifiedDoubles:
     - 'spec/custom_facts/util/fact_spec.rb'
     - 'spec/custom_facts/util/resolution_spec.rb'
     - 'spec/facter/facts/aix/ssh_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/swap/capacity_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/swap/used_bytes_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/system/capacity_spec.rb'
+    - 'spec/facter/facts/freebsd/memory/system/used_bytes_spec.rb'
```
IrimieBogdan, line 197: We should fix these offences. An example for `spec/facter/facts/freebsd/memory/swap/capacity_spec.rb`
```ruby
      expected_fact = instance_double(Facter::ResolvedFact, name: 'memory.swap.capacity', value: 1024)
```

spec/facter/facts/freebsd/memory/swap/capacity_spec.rb
IrimieBogdan: Just noticed, this is the old model for writing unit test for facts. We should use the new model, like you have used in `spec/facter/facts/freebsd/memory/swap/encrypted_spec.rb`

spec/facter/facts/freebsd/memory/swap/used_bytes_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expected_fact = double(Facter::ResolvedFact, name: 'memory.swap.used_bytes', value: 1024)
+
+      allow(Facter::Resolvers::Freebsd::SwapMemory).to receive(:resolve).with(:used_bytes).and_return(1024)
+      allow(Facter::ResolvedFact).to receive(:new).with('memory.swap.used_bytes', 1024).and_return(expected_fact)
+
+      fact = Facts::Freebsd::Memory::Swap::UsedBytes.new
+      expect(fact.call_the_resolver).to eq(expected_fact)
+    end
+  end
+end
```
IrimieBogdan, line 15: This test should be written using the new template for writing unit test for facts.

spec/facter/facts/freebsd/memory/system/capacity_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expected_fact = double(Facter::ResolvedFact, name: 'memory.system.capacity', value: '15.53%')
+
+      allow(Facter::Resolvers::Freebsd::SystemMemory).to receive(:resolve).with(:capacity).and_return('15.53%')
+      allow(Facter::ResolvedFact).to receive(:new).with('memory.system.capacity', '15.53%').and_return(expected_fact)
+
+      fact = Facts::Freebsd::Memory::System::Capacity.new
+      expect(fact.call_the_resolver).to eq(expected_fact)
+    end
+  end
+end
```
IrimieBogdan, line 15: This test should be written using the new template for writing unit test for facts.

spec/facter/facts/freebsd/memory/system/used_bytes_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expected_fact = double(Facter::ResolvedFact, name: 'memory.system.used_bytes', value: 1024)
+
+      allow(Facter::Resolvers::Freebsd::SystemMemory).to receive(:resolve).with(:used_bytes).and_return(1024)
+      allow(Facter::ResolvedFact).to receive(:new).with('memory.system.used_bytes', 1024).and_return(expected_fact)
+
+      fact = Facts::Freebsd::Memory::System::UsedBytes.new
+      expect(fact.call_the_resolver).to eq(expected_fact)
+    end
+  end
+end
```
IrimieBogdan, line 15: This test should be written using the new template for writing unit test for facts.

smortex:
@IrimieBogdan I added a commit that fix the issues you listed, thank you for catching them, I mostly copied the Solaris facts which mix old and new style specs.  Should these be updated too (I mean in this PR, or do we keep the focus on FreeBSD memory facts here)?

lib/facter/resolvers/freebsd/system_memory_resolver.rb
igalic: btw, this means we're dropping support for FreeBSD 10.4… https://www.freebsd.org/cgi/man.cgi?query=vmstat&apropos=0&sektion=0&manpath=FreeBSD+10.4-RELEASE&arch=default&format=html

then again, that's no longer supported, so… https://www.freebsd.org/security/security.html#sup

igalic:
let's clean up solaris tests in a next round

lib/facter/resolvers/freebsd/swap_memory_resolver.rb
```diff
@@ -0,0 +1,42 @@
...
+
+            @fact_list[fact_name]
+          end
+
+          def kilobytes_to_bytes(quantity)
+            (quantity.to_f * 1024).to_i
+          end
+
+          def compute_capacity(used, total)
+            "#{format('%<value>.2f', value: (used / total.to_f * 100))}%"
```
oanatmaria, line 36: Can you please use https://github.com/puppetlabs/facter/blob/d5c658b9c8e9f74ca2afd68bc2dd2195de156e50/lib/facter/resolvers/utils/filesystem_helper.rb#L19 ?

oanatmaria: Same for system_memory_resolver

smortex: I just fixed both of them

PR #2021, opened 2020-08-11
Add FreeBSD dmi facts

smortex:
The dmi facts where not provided by Facter 4.  With this PR, Facter returns the same facts as Facter 3 does.


IrimieBogdan:
Excellent work!

PR #2023, opened 2020-08-11
(FACT-2727) add load averages for Solaris

Filipovici-Andrei:


Filipovici-Andrei:

igalic:
am i reading this right, that bsd doesn't need a specialisation, but Solaris does?

Filipovici-Andrei:
Solaris and BSD need the same load_averages resolver. So I moved it to /lib/facter/resolvers where we have all the common resolvers, which are shared by multiple platforms. These resolvers are loaded for every platform.

PR #2024, opened 2020-08-11
(FACT-2723) --list-*-groups also displays external facts

sebastian-miclea:
When using --list-cache-groups or --list-block-groups, both custom and external facts are displayed


lib/facter/framework/config/fact_groups.rb
```diff
@@ -13,6 +13,7 @@ def initialize(group_list_path = nil)
       @groups_file_path = group_list_path || default_path
       @groups ||= File.readable?(@groups_file_path) ? Hocon.load(@groups_file_path) : {}
       load_groups
+      load_groups_from_options
```
Filipovici-Andrei, line 16: I think more meaningful names are required. Something like load_core_groups and load_external_and_custom_groups.

IrimieBogdan:

PR #2027, opened 2020-08-12
(FACT-2722) Add disks fact for Solaris

Filipovici-Andrei:


Filipovici-Andrei:

Filipovici-Andrei:
Sure @igalic. That's a reasonable request.

lib/facter/resolvers/load_averages.rb
```diff
@@ -3,6 +3,7 @@
 module Facter
   module Resolvers
     class LoadAverages < BaseResolver
+      @log = Facter::Log.new(self)
```
IrimieBogdan, line 6: You don't need to define a logger, the base resolver will automatically create one for you. You can use it directly via 
`log.debug(...)`

lib/facter/resolvers/solaris/disks.rb
```diff
@@ -0,0 +1,52 @@
...
+
+            @log.debug('loading disks info')
+
+            kstat_output = Facter::Core::Execution.execute('/usr/bin/kstat sderr', logger: log)
+            return if kstat_output.empty?
+
+            @fact_list[fact_name] = parse(kstat_output)
+          end
+
+          def parse(input)
```
IrimieBogdan, line 29: `input` is generic for a variable name, maybe `kstat_output `

PR #2034, opened 2020-08-18
(FACT-2747) Add missing legacy facts on all platforms

IrimieBogdan:


lib/facter/resolvers/processors_resolver.rb
```diff
@@ -7,10 +7,12 @@ class Processors < BaseResolver
         @log = Facter::Log.new(self)
         @semaphore = Mutex.new
         @fact_list ||= {}
+        MGHERTZ_TO_HERTZ = 1_000_000
```
sebastian-miclea, line 10: Can you rename this to MHZ_TO_HZ?

IrimieBogdan: Changed

lib/facter/facts/aix/disks.rb
```diff
@@ -4,13 +4,28 @@ module Facts
...
+        facts.push(Facter::ResolvedFact.new(FACT_NAME, disks))
+        facts.push(Facter::ResolvedFact.new('blockdevices', blockdevices, :legacy))
+        add_regex_facts(disks, facts)
+
+        facts
+      end
+
+      private
+
+      def add_regex_facts(disks, facts)
```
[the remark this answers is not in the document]
IrimieBogdan: Updated the method name.

lib/facter/facts/aix/disks.rb
```diff
@@ -4,13 +4,28 @@ module Facts
...
     class Disks
       FACT_NAME = 'disks'
+      ALIASES = %w[blockdevices blockdevice_.*_size'].freeze
 
       def call_the_resolver
+        facts = []
         disks = Facter::Resolvers::Aix::Disks.resolve(:disks)
 
         disks = disks&.empty? ? nil : disks
+        blockdevices = disks&.keys&.join(',')
```
oanatmaria, line 14: why don't you return a resolved fact with value nil if it's nil or empty in order to avoid checking twice if it's nil (once at line 13 and twice here with &)? This is valid for the other disk facts as well.

sebastian-miclea (approved):

lib/facter/facts/aix/disks.rb
IrimieBogdan: You are suggestion something like
```
return Facter::ResolvedFact.new(FACT_NAME, disks) if disks.nil?
```

oanatmaria: yes

oanatmaria (approved):

PR #2033, opened 2020-08-18
(FACT-2721) Added Solaris virtual fact

sebastian-miclea:

PR #2035, opened 2020-08-19
(FACT-2740) Add Gce fact

Filipovici-Andrei:
Added Gce fact for Windows and Linux platforms


lib/facter/facts/linux/gce.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Facts
+  module Linux
+    class Gce
+      FACT_NAME = 'gce'
+
+      def call_the_resolver
+        bios_vendor = Facter::Resolvers::Linux::DmiBios.resolve(:bios_vendor)
+
+        fact_value = bios_vendor&.include?('Google') ? Facter::Resolvers::Gce.resolve(:metadata) : nil
```
igalic, line 11: this reads like it should be a general Unix fact

igalic (commented):
🙋🏻‍♀️

lib/facter/facts/linux/gce.rb
Filipovici-Andrei: And it is, because Linux is the root facts platform for Debian, Rhel and Sles flavors. This relation is exemplified in lib/facter/os_hierarchy.json. I know it's confusing with the way facts files are structured now, but we'll move all the common ones into the Linux facts folder.

igalic: sorry, what i meant is: this looks like it would work the same for Solaris and FreeBSD, without a change

Filipovici-Andrei: I didn't add this fact on Solaris and FreeBSD because the virtual fact is not yet implemented on them. So, I'm not sure if it would work as it is.

igalic: aye, 
see also https://github.com/puppetlabs/facter/pull/2033#discussion_r472756510

Filipovici-Andrei: And also I want to add that the DmiBios fact for Linux doesn't work on Solaris.

PR #2038, opened 2020-08-20
(FACT-2454) fix how used memory is calculated

Filipovici-Andrei:
On Linux platforms we calculated the used system memory by subtracting the free memory from the total memory. 
Now we'll calculate it by subtracting the free memory, buffers, caches and SReclaimable memory from the total memory.


oanatmaria:

oanatmaria (approved):

PR #2042, opened 2020-08-24
(maint) Add nil check for ec2 facts.

IrimieBogdan:


oanatmaria (approved):

lib/facter/facts/linux/ec2_metadata.rb
```diff
@@ -6,14 +6,16 @@ class Ec2Metadata
...
         fact_value = Facter::Resolvers::Ec2.resolve(:metadata)
 
-        Facter::ResolvedFact.new(FACT_NAME, fact_value.empty? ? nil : fact_value)
+        Facter::ResolvedFact.new(FACT_NAME, fact_value&.empty? ? nil : fact_value)
       end
 
-      def aws?
+      private
+
+      def aws_hypervisors?
```
gimmyxd, line 18: Is this the correct way to fix it? as this will return true for non-aws machines, so even if we are not an AWS we'll continue to resolve the fact knowing we're not on the right path.

Was facter 3 using dmidecode for this check?

oanatmaria: Unfortunately there isn't a clean way to find out if it's aws. We can only check if the vm is on one of aws's hypervisors. Facter 3  does the same thing : https://github.com/puppetlabs/facter/blob/914d92373e4304ef24d1ce4b0a7f9a5cf46b215e/lib/src/facts/resolvers/ec2_resolver.cc#L122

PR #2045, opened 2020-08-25
(FACT-2728) Added hypervisors fact for Solaris

sebastian-miclea:


lib/facter/facts/solaris/hypervisors/ldom.rb
```diff
@@ -0,0 +1,34 @@
...
+            role_io: resolve(:role_io),
+            role_root: resolve(:role_root),
+            role_service: resolve(:role_service)
+          }
+
+          Facter::ResolvedFact.new(FACT_NAME, fact_value)
+        end
+
+        def resolve(key)
+          Facter::Resolvers::Solaris::Ldom.resolve(key)
```
gimmyxd, line 29: i think you can replace this with something like:

```ruby
%i[
  chassis_serial control_domain domain_name
  domain_uuid role_control role_io role_root role_service
].map! { |key| [key, Facter::Resolvers::Solaris::Ldom.resolve(key)] }.to_h
```

lib/facter/facts/solaris/hypervisors/zone.rb
```diff
@@ -0,0 +1,39 @@
...
+
+          current_zone = zones.find { |r| r[:name] == current_zone_name }
+
+          {
+            brand: current_zone[:brand],
+            id: current_zone[:id],
+            ip_type: current_zone[:iptype],
+            name: current_zone[:name],
+            uuid: current_zone[:uuid]
+          }
```
gimmyxd, line 34: Alternatively yo this you could also use Hash#select or something similar to keep only the needed keys.

sebastian-miclea: The problem here is that one key **(ip_type)** needs to be changed, so no general rule to map keys would work.

lib/facter/facts/solaris/hypervisors/zone.rb
```diff
@@ -0,0 +1,39 @@
...
+
+          Facter::ResolvedFact.new(FACT_NAME, fact_value)
+        end
+
+        def current_zone
+          current_zone_name = Facter::Resolvers::Solaris::ZoneName.resolve(:current_zone_name)
+          return unless current_zone_name
+
+          zones = Facter::Resolvers::Solaris::Zone.resolve(:zone)
+          return Facter::ResolvedFact.new(FACT_NAME, nil) unless zones
```
oanatmaria, line 24: here should be **return nil** otherwise the value for this fact will be another ResolvedFact object

spec/facter/facts/solaris/hypervisors/zone_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      it 'returns current zone details' do
+        expect(fact.call_the_resolver).to be_an_instance_of(Facter::ResolvedFact).and \
+          have_attributes(name: 'hypervisors.zone', value: {
+                            'brand' => 'solaris',
+                            'id' => 0,
+                            'ip_type' => 'shared',
+                            'name' => 'global',
+                            'uuid' => '1234'
+                          })
```
oanatmaria, line 69: could you please add a test for the scenario when the result for this call : Facter::Resolvers::Solaris::Zone.resolve(:zone) is nil ?

sebastian-miclea: Updated. This also caught the case in the other comment. 
Nice catch!

lib/facter/facts/solaris/hypervisors/zone.rb
sebastian-miclea: Updated.

oanatmaria (approved):

PR #2046, opened 2020-08-25
(FACT-2561) Fix blocking mechanism

oanatmaria:
**Description of the problem** Facts with aliases can be blocked only if all aliases are blocked
**Description of the solution** Blocking one core fact should block all legacy facts (aliases) that are equivalent to that fact. When an alias (legacy fact) is blocked, only the legacy fact is blocked without blocking the equivalent core fact.


oanatmaria:

oanatmaria:

sebastian-miclea (approved):

IrimieBogdan (approved):

PR #2048, opened 2020-08-27
(maint) Fix processor's speed fact for linux ppc

oanatmaria:
**Description of the problem:** On linux power pc processors.speed fact is wrong because the order of magnitude is incorrect (MHz instead of GHz).
**Description of the fix:** Made the method that calculate the processor's speed  more configurable in order to accept the magnitude as a parameter.


oanatmaria:

GabrielNagy (approved):

PR #2051, opened 2020-08-31
(FACT-2753-parallel) Resolve facts concurrently

IrimieBogdan:
Resolve facts by default in parallel (concurrently with MRI ruby). Sequential resolution of facts can be enables via the `--sequential` cli flag or via `facter.conf` (via de global.sequantial)


IrimieBogdan:

IrimieBogdan:

lib/facter/resolvers/aio_agent_version.rb
```diff
@@ -3,6 +3,7 @@
 module Facter
   module Resolvers
     class AioAgentVersion < BaseResolver
+      @semaphore = Mutex.new
```
ciprianbadescu, line 6: could we have the semaphore in a base class?

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -47,5 +56,59 @@ def resolve(searched_facts)
...
+
+      resolved_facts.flatten!
+
+      FactAugmenter.augment_resolved_facts(searched_facts, resolved_facts)
+    end
+
+    def thread_pool(searched_facts)
+      require 'concurrent'
+
+      pool = Concurrent::FixedThreadPool.new(12)
```
ciprianbadescu, line 90: number of threads could be configurable

lib/facter/resolvers/aio_agent_version.rb
IrimieBogdan: We need to have concurrency between resolvers, but no concurrency inside a resolver, thus we need a semaphore for each resolver. 

The tricky part comes when we need to have static methods (because we don't want to create a new resolver in each fact). The semaphore need to be accessible in the `BaseResolver`, but cannot be instantiated in it as class variables (@semaphore) and `initialise` methods are never called when calling static methods from "BaseResolver" children. 

Here is a small example
```ruby
class Base
  def initialize
    puts "baser init"
    @semaphore = Mutex.new
  end

  class << self
    def work
      puts @semaphore.inspect
      do_work
    end
  end
end


class A < Base
  def initialize
    puts "A init"
  end

  class << self
    def do_work
      puts "work in A"
    end
  end
end

class B < Base
  def intialize
    puts "B init"
  end

  class << self
    def do_work
      puts "work in B"
    end
  end
end

A.work
B.work
```

results
```
nil
work in A
nil
work in B
```

An option would be to create a mechanism similar to the one used for [logger](https://github.com/puppetlabs/facter/blob/1f90f3f4fad599eb4f4dadbb61aca18381078de6/lib/facter/resolvers/base_resolver.rb#L6) in `BaseResolver`, but this would require an extra method call in each resolver.

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: Removed the thread pool as it was the slowest implementation.

gimmyxd: I think this depends on the system that it runs on. Not sure that removing the thread poll entirely is a good idea.

lib/facter/resolvers/aio_agent_version.rb
gimmyxd: I think that could be achieved with modules and some Ruby magic:

https://gist.github.com/gimmyxd/d0685c7bdc2fe2c6f92bcf55e953520e

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: The code with the thread pool was never used and could not be enabled. If we want to try the thread pool in the future, we can revert commit: 3115202ecdc26cffed1abe3b30e1f596081e4a27

IrimieBogdan:

IrimieBogdan:

IrimieBogdan:

ciprianbadescu (approved):

IrimieBogdan:
Will close this as a new version is available at #2162

PR #2054, opened 2020-08-31
(FACT-2753) Extend facter API with parse_and_resolve.

IrimieBogdan:


lib/facter/framework/cli/cli.rb
```diff
@@ -165,6 +184,57 @@ def list_cache_groups(*args)
...
+
+      help_string
+    end
+
+    no_commands do
+      def help_header(args)
+        path = File.join(File.dirname(__FILE__), '../../')
+
+        if args.include?(:puppet)
+          Util::FileHelper.safe_read("#{path}fixtures/puppet_help_header")
```
joshcooper, line 203: I left my comments on the puppet PR, but I'm not a fan of this approach. I think this example shows how we're adding more circular dependencies between puppet and facter.

IrimieBogdan: Agree, removed most (if not all) of the puppet dependencies.

IrimieBogdan:
Jenkins please test this

IrimieBogdan:

IrimieBogdan:

sebastian-miclea (approved):

PR #2058, opened 2020-09-01
(FACT-2457) Display newlines in values

florindragos:


lib/facter/framework/formatters/legacy_fact_formatter.rb
```diff
@@ -84,6 +86,11 @@ def remove_enclosing_accolades(pretty_fact_json)
       pretty_fact_json.gsub(/^},/, '}')
     end
 
+    def remove_newlines(pretty_fact_json)
```
oanatmaria, line 89: i think this name is not very accurate, you are not removing newlines, you are replacing them with actual newlines. Please rename this to something more descriptive.

oanatmaria (approved):

PR #2061, opened 2020-09-02
(FACT-1380) Restore --timing option to native facter

IrimieBogdan:
Restore `--timing` argument. By using this argument, users can see the time it took to resolve each fact. The implementation works for core, custom and external facts.


igalic:
since there's no (obviously visible) link between github and puppet's jira, can you please be a little more descriptive in your PR?

sebastian-miclea (approved):

PR #2067, opened 2020-09-03
(FACT-2551) Defaulting resolver keys to nil after no resolution

sebastian-miclea:


sebastian-miclea:

spec/facter/resolvers/base_resolver_spec.rb
```diff
@@ -92,6 +92,24 @@ def self.post_resolve(fact_name)
...
+    before do
+      allow(resolver).to receive(:validate_resolution)
+    end
+
+    it 'sets the fact to nil if undefined' do
+      resolver.validate_resolution('unresolved_fact')
+      expect(resolver.resolve('unresolved_fact')).to be_nil
+    end
+
+    it 'does not overwirte values' do
```
Filipovici-Andrei, line 105: typo in overwrite

lib/facter/resolvers/base_resolver.rb
```diff
@@ -18,13 +18,21 @@ def self.subscribe_to_manager
       def self.resolve(fact_name)
         @semaphore.synchronize do
           subscribe_to_manager
+
           post_resolve(fact_name)
```
Filipovici-Andrei, line 22: Would it be ok to call post_resolve only if @fact_list is empty?

sebastian-miclea: Not really, you loose control on your resolver call if the execution stops in base_resolver. Compared to exceptions that are caught in base_resolver, this is a more common flow that allows you to 'cache' successful resolver calls that have no output.

IrimieBogdan: We can have partial results in @fact_list so we cannot decide in the base resolver if we want to call the resolver or not. We should always call the resolver.

IrimieBogdan (approved):

PR #2071, opened 2020-09-07
(maint) On El5 some info are missing from meminfo

oanatmaria:
**Description of the problem:** On Redhat 5 SReclaimable field is missing from /proc/meminfo file and this result in some facter errors and some memory facts are missing.

**Description of the fix:** Added an additional check for this field before parsing it.


lib/facter/resolvers/memory_resolver.rb
```diff
@@ -48,7 +48,12 @@ def kilobytes_to_bytes(quantity)
           def reclaimable_memory(output)
             buffers = kilobytes_to_bytes(output.match(/Buffers:\s+(\d+)\s/)[1])
             cached = kilobytes_to_bytes(output.match(/Cached:\s+(\d+)\s/)[1])
-            s_reclaimable = kilobytes_to_bytes(output.match(/SReclaimable:\s+(\d+)\s/)[1])
+            s_reclaimable = output.match(/SReclaimable:\s+(\d+)\s/)
```
Filipovici-Andrei, line 51: I think that we could use something like this: s_reclaimable = output.match(/SReclaimable:\s+(\d+)\s/).fetch(1) { 0 }

oanatmaria: will not solve the problem : `undefined` method 'fetch' for nil:NilClass`.
Also if we do : `&.fetch(1) { 0 }` the result will be nil

Filipovici-Andrei: yes, you are right

Filipovici-Andrei (approved):

PR #2075, opened 2020-09-08
(FACT-2788) Remove semaphore from all resolvers

IrimieBogdan:


IrimieBogdan:

MikaelSmith:
Why?

igalic:
Good question, and it's not answered by
https://tickets.puppetlabs.com/browse/FACT-2788 either

IrimieBogdan:
Until #2050 we were creating a thread for each fact. As multiple facts could use the same resolver we needed to ensure that there is no concurrency inside a resolver. This was achieved with a `Mutex` in each resolver. One of the disadvantages of this approach was that we created many threads and the performance gain was inconsistent (one some platforms the parallel version was faster and on others the sequential one was faster). A step was made to improve the multi threaded implementation with https://github.com/puppetlabs/facter/compare/main...IrimieBogdan:FACT-2753-parallel. This new implementation used a thread pool from `concurrent-ruby`. The thread pool implementation was the slowest of them all. 

With MRI ruby we can obtain concurrency but not parallelism, and it seams that concurrency alone does not offer a consistent performance advantage over the sequential approach.

Taking this into account we decided to resolve facts sequentially for the time being and to experiment with parallelism on https://github.com/IrimieBogdan/facter/tree/FACT-2753-parallel

sebastian-miclea (approved):

igalic:
that would've made for a pretty good commit message / PR description

IrimieBogdan:
@igalic agreed, will try to add a description for further PRs

PR #2080, opened 2020-09-11
(FACT-2793) Time limit for Facter::Core::Execute

oanatmaria:
**Description of the problem:** `Open3.capture3` sometimes can hang for indefinite time. 
**Description of the fix:** `Facter::Core::Execute` can be configured with a time limit after which the pid for the process that executes the command is killed. If no time limit is specified the command will time out after 1.5 seconds.


oanatmaria:
Jenkins please test this on windows2012r2-64a,windowsfips2012r2-64a

Filipovici-Andrei:
jenkins please test this on windows2019-64a

oanatmaria:

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -77,12 +79,31 @@ def builtin_command?(command)
...
+                Timeout.timeout(time_limit) do
+                  output << stdout.read
+                  err << stderr.read
+                end
+              rescue Timeout::Error
+                @log.debug("Timeout encounter after #{time_limit}s, killing process with pid: #{pid}")
+                Process.kill('KILL', pid)
+              end
+              [output, err]
+            end
```
gimmyxd, line 106: Just a heads-up that `popen3` can end-up in a deadlock(as code is written now) if the command that gets executed generates to much output on stderr: https://docs.ruby-lang.org/en/2.0.0/Open3.html#method-i-popen3

> You should be careful to avoid deadlocks. Since pipes are fixed length buffer, #popen3(“prog”) {|i, o, e, t| o.read } deadlocks if the program generates many output on stderr. You should be read stdout and stderr simultaneously (using thread or IO.select). However if you don't need stderr output, #popen2 can be used. If merged stdout and stderr output is not a problem, you can use #popen2e. If you really needs stdout and stderr output as separate strings, you can consider #capture3.


I put a quick gist which i used to replicate the issue. It just outputs a lot of text to stderr: https://gist.github.com/gimmyxd/92ae2e849ecb9b1865f215853359b760

I guess this is a pretty uncommon case but it can happen.

IrimieBogdan (approved):

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -35,10 +35,12 @@ def with_env(values)
           rv
         end
 
-        def execute(command, options = {})
+        def execute(command, options = {}) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           on_fail = options.fetch(:on_fail, :raise)
           expand = options.fetch(:expand, true)
           logger = options[:logger]
+          time_limit = options[:limit].to_i
+          time_limit = time_limit.positive? ? time_limit : nil
```
IrimieBogdan, line 43: We can extract this block of code in another private method e.g.

```
         def extract_options(options)
          on_fail = options.fetch(:on_fail, :raise)
          expand = options.fetch(:expand, true)
          logger = options[:logger]
          time_limit = options[:limit].to_i
          time_limit = time_limit.positive? ? time_limit : nil

          [on_fail, expand, logger, time_limit]
        end
```

and use 
```
on_fail, expand, logger, time_limit = extract_options(options)
```

to get rid of the rubocop exceptions.

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -77,12 +79,31 @@ def builtin_command?(command)
           output.chomp =~ /builtin/ ? true : false
         end
 
-        def execute_command(command, on_fail, logger = nil)
+        def execute_command(command, on_fail, logger = nil, time_limit = nil)
+          @log ||= Log.new(self)
```
Filipovici-Andrei, line 83: shouldn't the log be initialised outside the method?

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -77,12 +79,31 @@ def builtin_command?(command)
...
+          time_limit ||= 1.5
           begin
             # Set LC_ALL and LANG to force i18n to C for the duration of this exec;
             # this ensures that any code that parses the
             # output of the command can expect it to be in a consistent / predictable format / locale
-            out, stderr, _status_ = Open3.capture3({ 'LC_ALL' => 'C', 'LANG' => 'C' }, command.to_s)
+            opts = { 'LC_ALL' => 'C', 'LANG' => 'C' }
+            require 'timeout'
+            @log.debug("Executing command: #{command}")
```
Filipovici-Andrei, line 91: Why do you use 2 loggers? One initialised in this method and one received as parameter

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -77,12 +79,31 @@ def builtin_command?(command)
...
+                  err << stderr.read
+                end
+              rescue Timeout::Error
+                @log.debug("Timeout encounter after #{time_limit}s, killing process with pid: #{pid}")
+                Process.kill('KILL', pid)
+              end
+              [output, err]
+            end
             log_stderr(stderr, command, logger)
           rescue StandardError => e
```
Filipovici-Andrei, line 108: you have two begin..rescue blocks. Could you separate at least the begin blocks into separate methods?

Filipovici-Andrei (changes requested):

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: Done

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: Done

oanatmaria:

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: There are 2 faulty scenarios that we want to be logged different: 
1st when the command timeout because it takes more than the time limit to execute (this will be logged with Facter::Core::Execution specific logger)
2nd the command doesn't exceed the time limit but the command still fails and it produces some output on stderr (this will be logged with the logger from the specific resolver that the command came from )

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: I wouldn't separate this in another method because the rubocop is not complaining and there is already a method called `execute_command` that calls a method called `execute`.

oanatmaria:

Filipovici-Andrei (approved):

lib/facter/custom_facts/core/execution/base.rb
joshcooper: I don't think this is safe. If stdout/stderr are N bytes, and the child process writes N+1 bytes to `stderr`, but nothing to `stdout`, then facter will block waiting to read from `stdout`, hit the timeout, and never read from `stderr`. Meanwhile, the child process will be blocked trying to write the `N+1`th byte, which it can't until facter drains the `stderr` pipe. I think you're better off creating two reader threads, and have each append to their respective `output` and `err` buffers.

PR #2082, opened 2020-09-14
(FACT-2746) Added cloud resolver

sebastian-miclea:


lib/facter/resolvers/cloud.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class Cloud < BaseResolver
+      # cloud_provider
+
+      @semaphore = Mutex.new
```
Filipovici-Andrei, line 8: We are no longer using the @semaphore

spec/facter/resolvers/cloud_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    before do
+      allow(File).to receive(:directory?).with('/var/lib/dhcp').and_return(true)
+      allow(Facter::Util::FileHelper)
+        .to receive(:safe_readlines)
+        .with('/var/lib/dhcp/dhclient.rhel.lease.8')
+        .and_return(content)
+      allow(Dir).to receive(:entries).with('/var/lib/dhcp').and_return(['.', 'dhclient.rhel.lease.8'])
+    end
+
+    it 'returns azure' do
```
Filipovici-Andrei, line 58: it 'returns nil'

spec/facter/resolvers/cloud_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      expect(cloud_resolver.resolve(:cloud_provider)).to eq('azure')
+    end
+  end
+
+  context 'when lease file is found without option 245' do
```
Filipovici-Andrei, line 46: Could you add a test when the lease file is found but it can't be read?

Filipovici-Andrei (changes requested):

spec/facter/resolvers/cloud_spec.rb
sebastian-miclea: Yes, updated.

Filipovici-Andrei (approved):

PR #2087, opened 2020-09-15
(FACT-2748) Fix fact types on all platforms-Part 1

oanatmaria:
**Fixed output for:**
"fips_enabled" 
"identity.gid" 
"identity.privileged"
"identity.uid"
"is_virtual"
load_averages.15m" 
"load_averages.1m" 
"load_averages.5m"
"memory.swap.available_bytes"
"memory.swap.capacity"
"memory.swap.total_bytes"
"memory.swap.used_bytes"
**On linux**
"hypervisors.lpar.partition_number"
**On Aix**
"swapfree_mb"
"swapsize_mb"
**On all platforms**


gimmyxd:
Maybe this could make use of a better title and some description in the commit message to understand why this is needed.

Filipovici-Andrei (approved):

PR #2089, opened 2020-09-15
(FACT-2786) Fix fact caching if fact is defined in multiple groups

florindragos:
This PR also enables fact caching by only specifying individual facts in ttls configuration, rather than needing a cache group.

### DIFF
--- a/lib/facter/framework/config/fact_groups.rb
+++ b/lib/facter/framework/config/fact_groups.rb
@@ -2,7 +2,7 @@
 
 module Facter
   class FactGroups
-    attr_reader :groups, :block_list
+    attr_reader :groups, :block_list, :facts_ttls
 
     @groups_ttls = []
 
@@ -14,6 +14,10 @@ module Facter
       @groups ||= File.readable?(@groups_file_path) ? Hocon.load(@groups_file_path) : {}
       load_groups
       load_groups_from_options
+      load_facts_ttls
+
+      # Reverse sort facts so that children have precedence when caching. eg: os.macosx vs os
+      @facts_ttls = @facts_ttls.sort.reverse.to_h
     end
 
     # Breakes down blocked groups in blocked facts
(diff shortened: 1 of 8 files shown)


florindragos:

lib/facter/framework/config/fact_groups.rb
```diff
@@ -41,6 +45,15 @@ def get_group_ttls(group_name)
       ttls_to_seconds(ttls[group_name])
     end
 
+    def get_fact(fact_name)
+      return @facts_ttls[fact_name] if @facts_ttls[fact_name]
+
+      fact = @facts_ttls.select { |k, v| break v if fact_name =~ /^#{k}\..*/ }
```
gimmyxd, line 51: `k, v` are not very suggestive, maybe use `key`, `value`.
Also that break there is a bit hard to read overall. Do you need just the first value that matches `/^#{k}\..*/`?

florindragos:

florindragos:

lib/facter/framework/config/fact_groups.rb
florindragos: Yes, we take the first match. `@facts_ttls` is sorted so the children come first (os.macosx vs os)

florindragos:

lib/facter/framework/config/fact_groups.rb
Filipovici-Andrei: You could use detect instead of select. It returns the first element for which the block returns true.

Filipovici-Andrei:
Is there a unit test that checks if a fact contained in two groups is not cached twice?

florindragos:
> Is there a unit test that checks if a fact contained in two groups is not cached twice?

too complex for a unit test. added acceptance test.

Filipovici-Andrei (approved):

PR #2093, opened 2020-09-18
(FACT-2802) Fix Cloud resolver

Filipovici-Andrei:
When calling the cloud fact using a non root user, 
an error is logged, because we attempt to access a privileged folder.

The fix skips folders which don't have a read permission.


Filipovici-Andrei:

gimmyxd:
This should have a ticket associated, that way we can track the changes.
Even more if this is a bug that was introduced in a previously released facter version

Filipovici-Andrei:

Filipovici-Andrei:
@gimmyxd I'll create a ticket for this bug

oanatmaria (approved):

PR #2095, opened 2020-09-18
(FACT-2803) Detect hypervisors as amazon if virtwhat detects aws.

IrimieBogdan:
On `Redhat8` `arm` architecture, ec2_metadata fact was not detecting correctly that we are on AWS and we need to resolve the fact. 

The fix considers we are on AWS if `virt_what` resolver reports `aws`

### DIFF
--- a/lib/facter/facts/linux/ec2_metadata.rb
+++ b/lib/facter/facts/linux/ec2_metadata.rb
@@ -16,9 +16,11 @@ module Facts
       private
 
       def aws_hypervisors?
-        virtual = check_virt_what || check_xen || check_product_name || check_bios_vendor || check_lspci
+        virtual =~ /kvm|xen|aws/
+      end
 
-        virtual == 'kvm' || virtual =~ /xen/
+      def virtual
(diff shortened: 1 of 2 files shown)


oanatmaria (approved):

lib/facter/facts/linux/ec2_metadata.rb
```diff
@@ -16,9 +16,13 @@ def call_the_resolver
       private
 
       def aws_hypervisors?
-        virtual = check_virt_what || check_xen || check_product_name || check_bios_vendor || check_lspci
+        virt = virtual
 
-        virtual == 'kvm' || virtual =~ /xen/
+        virt == 'kvm' || virt =~ /xen/ || virt == 'aws'
```
sebastian-miclea, line 21: I think that writing this way `virt.match?(/kvm|xen|aws/)` would be a bit more easy to understand

IrimieBogdan: Nice suggestion, changed!

Filipovici-Andrei (approved):

sebastian-miclea (approved):

PR #2094, opened 2020-09-18
(maint) add support for redhat-8-arm64

ciprianbadescu:
redhat-8-arm64 ami is using NetworkManager 1.22 which is using
nettools plugin by default and provides incomplete dhcp information
(https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426)

PR #2097, opened 2020-09-21
(maint) add support for redhat-8-arm64

ciprianbadescu:
redhat-8-arm64 ami is using NetworkManager 1.22 which is using
nettools plugin by default and provides incomplete dhcp information
(https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426)


ciprianbadescu:
Same as: https://github.com/puppetlabs/facter/pull/2094(which was closed by github after rebasing to different branch)

gimmyxd (approved):

gimmyxd:
@ciprianbadescu i think a PR for main branch is also needed as there is no merge-up set form master to main due to c++ to ruby tranzition.

PR #2100, opened 2020-09-21
FACT(2326) Add Facter.each method

florindragos:
This method will execute a given block for each fact name and value pair

### DIFF
--- a/lib/facter.rb
+++ b/lib/facter.rb
@@ -138,6 +138,26 @@ module Facter
       Facter::Options[:debug] = debug_bool
     end
 
+    # Iterates over fact names and values
+    #
+    # @yieldparam [String] name the fact name
+    # @yieldparam [String] value the current value of the fact
+    #
+    # @return [Facter]
+    #
+    # @api public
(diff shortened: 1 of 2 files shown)


oanatmaria (approved):

lib/facter.rb
```diff
@@ -125,6 +125,26 @@ def debugging(debug_bool)
       Facter::Options[:debug] = debug_bool
     end
 
+    # Iterates over fact names and values
+    #
+    # @yieldparam [String] name the fact name
+    # @yieldparam [String] value the current value of the fact
+    #
+    # @return [void]
```
Filipovici-Andrei, line 133: It says that you return void but the method returns self.
What is facter 3 returning?

Filipovici-Andrei (changes requested):

lib/facter.rb
florindragos: Facter 3 is also returning `Facter`

Filipovici-Andrei (approved):

PR #2103, opened 2020-09-22
(FACT-2324) Add loadfacts API method

sebastian-miclea:
Added public API method **loadfacts**. Using this method, all facts can be loaded on demand.

### DIFF
--- a/lib/facter.rb
+++ b/lib/facter.rb
@@ -168,6 +168,16 @@ module Facter
       nil
     end
 
+    # Loads all facts
+    #
+    # @return [nil]
+    #
+    # @api public
+    def loadfacts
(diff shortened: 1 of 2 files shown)


lib/facter.rb
```diff
@@ -168,6 +168,15 @@ def reset
       nil
     end
 
+    # Loads all facts
+    #
+    # @return [nil]
```
Filipovici-Andrei, line 173: I did a test and it returns an empty array. I think you need to add a nil at the end of the method.

Filipovici-Andrei (changes requested):

Filipovici-Andrei (approved):

PR #2107, opened 2020-09-27
(FACT-2432) Fixed environment dependant tests

sebastian-miclea:


spec/framework/core/options/option_store_spec.rb
```diff
@@ -307,4 +307,18 @@
         .from(true).to(false)
     end
   end
+
+  describe '#reset' do
+    it 'resets to default' do
+      default = option_store.all
+
+      option_store.show_legacy = true
```
Filipovici-Andrei, line 315: I think this line is not necessary.

sebastian-miclea: Good catch, I wanted to change the default values so the test will catch resetting the values.

Filipovici-Andrei (approved):

PR #2109, opened 2020-09-29
(FACT-2805) Read available memory from MemAvailable

florindragos:
Available memory should use the new MemAvailable info in `/proc/meminfo` when available. When not, fall back to the formula used in facter 3 (MemFree + Cached + Buffers)
For capacity, just use Total - Available

### DIFF
--- a/spec/facter/resolvers/memory_resolver_spec.rb
+++ b/spec/facter/resolvers/memory_resolver_spec.rb
@@ -16,11 +16,8 @@ describe Facter::Resolvers::Linux::Memory do
 
     context 'when there is swap memory' do
       let(:total) { 4_036_680 * 1024 }
-      let(:free) { 3_547_792 * 1024 }
-      let(:buffers) { 4_288 * 1024 }
-      let(:cached) { 298_624 * 1024 }
-      let(:s_reclaimable) { 29_072 * 1024 }
-      let(:used) { total - free - buffers - cached - s_reclaimable }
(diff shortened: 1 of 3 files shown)


spec/facter/resolvers/memory_resolver_spec.rb
```diff
@@ -109,22 +103,68 @@
...
-      let(:s_reclaimable) { 0 }
-      let(:used) { total - free - buffers - cached - s_reclaimable }
+      let(:free) { 3_547_792 * 1024 }
+      let(:available) { free + buffers + cached }
+      let(:used) { total - available }
       let(:swap_total) { 2_097_148 * 1024 }
       let(:swap_free) { 2_097_148 * 1024 }
       let(:swap_used) { swap_total - swap_free }
-      let(:fixture_name) { 'rhel5_memory' }
+      let(:fixture_name) { 'meminfo_no_available' }
```
Filipovici-Andrei, line 160: I think that meminfo_not_available is better than meminfo_no_available

sebastian-miclea (approved):

spec/facter/resolvers/memory_resolver_spec.rb
florindragos: meminfo is available, `available` is not available

Filipovici-Andrei: then meminfo_missing_available?

Filipovici-Andrei (approved):

PR #2113, opened 2020-10-01
(FACT-2824) Facter make ec2 metadata requests when on gce

IrimieBogdan:
`ec2_metadata` fact tries to determine on what hypervisor or on what cloud provider it is running on. If it detecters aws, it will call the `ec2` fact.

Problem:
- `ec2_metadata` fact calls the `ec2` resolver when it detects `kvm` hypervisor, but there are other providers that use `kvm`, one example is `gce`. 

Solution:
- improve the detection of hypervisor or cloud provider by adding the logic from virtual and dmi decoder


sebastian-miclea (approved):

IrimieBogdan:

PR #2114, opened 2020-10-01
(maint) Avoid deadlock of Facter::Core::Execution.execute

oanatmaria:
**Description of the problem:** If stdout/stderr are N bytes, and the child process writes N+1 bytes to stderr, but nothing to stdout, then facter will block waiting to read from stdout, hit the timeout, and never read from stderr.
**Description of the fix** Facter will create two reader threads to avoid the deadlock

### DIFF
--- a/acceptance/tests/custom_facts/long_stderr_with_time_limit.rb
+++ b/acceptance/tests/custom_facts/long_stderr_with_time_limit.rb
@@ -1,0 +1,34 @@
+test_name "Facter::Core::Execution doesn't kill process with long stderr message" do
+  tag 'risk:high'
+
+  confine :except, :platform => /windows/
+
+  long_output = "This is a very long error message. " * 4096
+  file_content = <<-EOM
+   #!/bin/sh
+   echo 'newfact=value_of_fact'
+   1>&2 echo #{long_output}
+   exit 1
+  EOM
+
+
+  agents.each do |agent|
+
+    external_dir = agent.tmpdir('external_dir')
+    fact_file = File.join(external_dir, 'test.sh')
+    create_remote_file(agent, fact_file, file_content)
+    agent.chmod('+x', fact_file)
+
+    teardown do
+      agent.rm_rf(external_dir)
+    end
+
+    step "Facter: should resolve the external fact and print as warning script's stderr message" do
+      on agent, facter('--external-dir', external_dir, 'newfact') do |facter_output|
+        assert_match(/value_of_fact/, facter_output.stdout.chomp)
+        assert_match(/WARN test.sh .*test.sh resulted with the following stderr message: This is a very long error message./, facter_output.stderr.chomp)
+      end
+    end
+  end
+end
+
--- a/lib/facter/custom_facts/core/execution/base.rb
+++ b/lib/facter/custom_facts/core/execution/base.rb
@@ -90,7 +90,7 @@ module Facter
         end
 
         def execute_command(command, on_fail, logger = nil, time_limit = nil)
-          time_limit ||= 1.5
+          time_limit ||= 300
           begin
             # Set LC_ALL and LANG to force i18n to C for the duration of this exec;
             # this ensures that any code that parses the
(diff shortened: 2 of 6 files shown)


acceptance/tests/custom_facts/long_stderr_with_time_limit.rb
```diff
@@ -0,0 +1,155 @@
+test_name "Facter::Core::Execution doesn't kill process with long stderr message" do
+  tag 'risk:high'
+
+  file_content = <<-EOM
```
gimmyxd, line 6: You can make use of ruby here, to generate the needed text, eg:
```ruby
length = 1000
file_content = <<-EOM
   #!/bin/sh
   long_output=#{"Lorem ipsum dolor sit amet" * length }
   echo 'newfact=value_of_fact'
   1>&2 echo $long_output
   exit 1
EOM
```

And play with the `length` to achieve the expected output.

Also i think would be useful to add some description to this scenario and what is this test doing

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -100,18 +100,22 @@ def execute_command(command, on_fail, logger = nil, time_limit = nil)
             @log.debug("Executing command: #{command}")
             out, stderr = Open3.popen3(opts, command.to_s) do |_, stdout, stderr, wait_thr|
               pid = wait_thr.pid
-              output = +''
-              err = +''
+              stdout_messages = +''
+              stderr_messages = +''
```
gimmyxd, line 104: why not use a simple initialize?
```ruby
stdout_messages = ''
stderr_messages = ''
```

oanatmaria: they will be treated as frozen strings and this will generate an error

joshcooper: I've started using `String.new` to make it clearer that I'm creating a string that isn't frozen

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -100,18 +100,22 @@ def execute_command(command, on_fail, logger = nil, time_limit = nil)
...
+              out_reader = Thread.new { stdout.read }
+              err_reader = Thread.new { stderr.read }
               begin
                 Timeout.timeout(time_limit) do
-                  output << stdout.read
-                  err << stderr.read
+                  stdout_messages << out_reader.value
+                  stderr_messages << err_reader.value
```
joshcooper, line 110: If `stdout.read` (on line 105) raises then ruby will raise that exception when `out_reader.value` is called. So it's possible for line 109 to raise something like `Errno::EIO`, in which case we never call `err_reader.value`. As a result we may never join that thread and fully read `stderr`. I don't know if that's going to be an actual problem though...

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -100,18 +100,22 @@ def execute_command(command, on_fail, logger = nil, time_limit = nil)
...
                 end
               rescue Timeout::Error
                 @log.debug("Timeout encounter after #{time_limit}s, killing process with pid: #{pid}")
                 Process.kill('KILL', pid)
+                out_reader.kill
+                err_reader.kill
```
joshcooper, line 116: Can `Process.kill` raise such as if the child process no longer exists? If so then kill is never called on the readers.

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: Done

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: i wrapped the threads kill in an `ensure` block

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: If `out_reader.value` raises an error, the error message will be logged (if execution is called with `on_fail =>  :raise` line 122). If the stdout will raise an error, the resolvers can't get the information needed for facts to be solved, the stderr message can only be used for debug purposes in this case.

lib/facter/custom_facts/core/execution/base.rb
oanatmaria: Unfortunately i get the following error when i try to use `String.new` : `lib/facter/custom_facts/core/execution/base.rb:104:33: C: Performance/UnfreezeString: Use unary plus to get an unfrozen string literal.`

IrimieBogdan (approved):

PR #2120, opened 2020-10-06
(FACT-2818) (experimental) Enable facter -p with circular dependency

IrimieBogdan:
Enable `facte -p` and `facter --puppet` via a circular dependency. Added new logic that blocks puppet from injecting it's logger in facter.


IrimieBogdan:
Jenkins please test this on osx1014-64a,osx1015-64a

PR #2121, opened 2020-10-06
(FACT-2817) Only invalidate session cache on clear and reset.

IrimieBogdan:
The `SessionCache` is used internal to store resolver values. If a resolver is called multiple times, we only resolve once the data and all subsequent request to that resolver, will hit the cache. 

## Problem
The current implementation invalidates this session cache after each request (to_hash, value, []), but Facter 3 invalidates this cache only on `reset` and `clear`. 

## Solution
Invalidate the session cache only on `reset` and `clear`


IrimieBogdan:

IrimieBogdan:

oanatmaria (approved):

Filipovici-Andrei (approved):

PR #2128, opened 2020-10-08
(maint) Fix virtual_detector

IrimieBogdan:
When we detect Amazon as vendor using `dmidecode` we can conclude that the hypervisor is `kvm` because when it is `xen`, the vendor is `Xen`

As the `virtual_detector` is used in virtual fact, we should return the hypervisor instead of the cloud provider.

### DIFF
--- a/lib/facter/facts_utils/virtual_detector.rb
+++ b/lib/facter/facts_utils/virtual_detector.rb
@@ -22,7 +22,7 @@ module Facter
       @log.debug('Checking DMI')
       vendor = Facter::Resolvers::DmiDecode.resolve(:vendor)
       @log.debug("dmi detected vendor: #{vendor}")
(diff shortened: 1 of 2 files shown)


IrimieBogdan:

lib/facter/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Facter
-  VERSION = '4.0.42' unless defined?(VERSION)
+  VERSION = '4.0.45' unless defined?(VERSION)
```
oanatmaria, line 4: nit

IrimieBogdan: Reverted

oanatmaria (approved):

IrimieBogdan:

PR #2129, opened 2020-10-08
(FACT-2643)  Add test to verify timeout on resolution

oanatmaria:
Facter 3 doesn't support timeout option for custom facts, it logs a warning message when a user tries to set that option. 
We decided to keep the mechanism that sets timeout on resolution on Facter 4 and to add an acceptance test to verify that the behaviour is indeed as expected.


IrimieBogdan (approved):

oanatmaria:

IrimieBogdan (approved):

PR #2133, opened 2020-10-12
(FACT-2806) Fix os.release.minor on amazon 6

florindragos:
We don't need to remove leading 0 for minor versions on rhel.

### DIFF
--- a/lib/facter/facts/rhel/os/release.rb
+++ b/lib/facter/facts/rhel/os/release.rb
@@ -27,7 +27,7 @@ module Facts
           fact_value = {}
           fact_value['full'] = version
           fact_value['major'] = versions[0]
-          fact_value['minor'] = versions[1].gsub(/^0([1-9])/, '\1') if versions[1]
+          fact_value['minor'] = versions[1] if versions[1]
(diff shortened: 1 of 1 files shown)


oanatmaria:

lib/facter/facts/rhel/os/release.rb
```diff
@@ -27,7 +27,7 @@ def determine_release_version
           fact_value = {}
           fact_value['full'] = version
           fact_value['major'] = versions[0]
-          fact_value['minor'] = versions[1].gsub(/^0([1-9])/, '\1') if versions[1]
+          fact_value['minor'] = versions[1] if versions[1]
```
gimmyxd, line 30: Is there any test for the scenario that this change is fixing?

florindragos: no. from what we can tell at this moment, this code was copy-pasted from the debian os.release fact when the rhel fact was first written. it was not suppose to be there in the first place and it appears to be affecting only amazon 6.

Filipovici-Andrei (approved):

PR #2135, opened 2020-10-12
(FACT-2832) Use full path for augparse command 

oanatmaria:
**Description of the problem:** On unix like systems and linux operating systems, the call: `::Augeas.open` is very time consuming. Also `augparse` command is not found because the puppet directories are not in path.
**Description of the fix:** On unix like and linux systems, call augparse with the full path and avoid calling the method `::Augeas.open`.

**Note:** Augeas is shipped in puppet-agent packages except Windows ones (that's is why this change is not affecting Windows behaviour)

### DIFF
--- a/lib/facter/resolvers/augeas_resolver.rb
+++ b/lib/facter/resolvers/augeas_resolver.rb
@@ -20,7 +20,13 @@ module Facter
         end
 
         def read_augeas_from_cli
-          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
+          command = if File.readable?('/opt/puppetlabs/puppet/bin/augparse')
+                      '/opt/puppetlabs/puppet/bin/augparse'
+                    else
+                      'augparse'
+                    end
+
+          output = Facter::Core::Execution.execute("#{command} --version 2>&1", logger: log)
           Regexp.last_match(1) if output =~ /^augparse (\d+\.\d+\.\d+)/
         end
 
--- a/spec/facter/resolvers/augeas_resolver_spec.rb
+++ b/spec/facter/resolvers/augeas_resolver_spec.rb
@@ -4,9 +4,11 @@ describe Facter::Resolvers::Augeas do
   subject(:augeas) { Facter::Resolvers::Augeas }
 
   let(:log_spy) { instance_spy(Facter::Log) }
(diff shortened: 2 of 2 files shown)


lib/facter/resolvers/augeas_resolver.rb
```diff
@@ -20,7 +20,13 @@ def read_augeas_version(fact_name)
         end
 
         def read_augeas_from_cli
-          output = Facter::Core::Execution.execute('augparse --version 2>&1', logger: log)
+          command = if RbConfig::CONFIG['host_os'] =~ /mswin|msys|mingw|cygwin|bccwin|wince|emc/
+                      'augparse'
+                    else
+                      '/opt/puppetlabs/puppet/bin/augparse'
```
gimmyxd, line 26: Do we want to have this fact only in a puppet-agent context? What if i want to use the augeas fact with standalone facter and i install `augeas` to have `augparse` command?

For example installing augeas-tools and cfacter 3.11.0 on ubuntu-20.04 i can have the augeas fact.

oanatmaria: I change the behaviour as follows:
1st: check if the path from puppet-agent's `augparse` is exists and is readable; If not
2st: call 'augparse' without path and hope it is in path; If this command returns empty string
3rd: try requiring augeas gem and calling Augeas cli method to determine it's version.

So if augeas is present on the system, facter will determine it's version, even if it isn't in path or if it was installed manually and not shipped in puppet-agent context.

spec/facter/resolvers/augeas_resolver_spec.rb
```diff
@@ -4,9 +4,11 @@
   subject(:augeas) { Facter::Resolvers::Augeas }
 
   let(:log_spy) { instance_spy(Facter::Log) }
+  let(:executable?) { false }
```
gimmyxd, line 7: usually variables do not have `?` in name, that is a convention only for boolean methods

gimmyxd (approved):

spec/facter/resolvers/augeas_resolver_spec.rb
oanatmaria: changed

PR #2139, opened 2020-10-14
(FACT-2835) Extract external fact stderr logging outside of Facter::Util::Resolution.exec

florindragos:
`Facter::Util::Resolution.exec`should not log anything to stderr.

### DIFF
--- a/lib/facter/custom_facts/core/execution.rb
+++ b/lib/facter/custom_facts/core/execution.rb
@@ -98,7 +98,7 @@ module Facter
       # Execute a command and return the output of that program.
       # @param command [String] Command to run
       #
-      # @param options [Hash] Hash with options for the aggregate fact
+      # @param options [Hash] Hash with options for the command
       #
       # @option options [Object] :on_fail How to behave when the command could
       #   not be run. Specifying :raise will raise an error, anything else will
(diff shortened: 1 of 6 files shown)


oanatmaria (approved):

lib/facter/custom_facts/core/execution.rb
```diff
@@ -115,6 +115,26 @@ def execute(command, options = {})
         @@impl.execute(command, options)
       end
 
+      # Execute a command and return the stdout and stderr of that program.
+      # @param command [String] Command to run
+      #
+      # @param options [Hash] Hash with options for the aggregate fact
```
Filipovici-Andrei, line 121: Is this line accurate?

lib/facter/custom_facts/util/parser.rb
```diff
@@ -8,6 +8,8 @@
 module LegacyFacter
   module Util
     module Parser
+      STDERR_MESSAGE = 'Command %s resulted with the following stderr message: %s'
```
Filipovici-Andrei, line 11: instead of resulted doesn't finished sound better?

Filipovici-Andrei (approved):

PR #2140, opened 2020-10-14
(FACT-2826) Added Facter.flush

sebastian-miclea:
Facter flush sets the resolved value to nil without restarting the resolution. Besides that flushing can be hooked using on_flush in custom facts.

### DIFF
--- a/acceptance/tests/facter_flush.rb
+++ b/acceptance/tests/facter_flush.rb
@@ -1,0 +1,52 @@
+test_name 'facter should flush fact values' do
+  tag 'risk:high'
+
+  fact_content1 = <<-EOM
+    require 'facter'
+
+    Facter.add(:fact1) do
+      'this should be flushed'
+    end
+
+    Facter.flush
+
+    puts "Fact1: \#\{Facter.value(:fact1)\}"
+  EOM
+
+  fact_content2 = <<-EOM
+    require 'facter'
+
+    Facter.add(:fact2) do
+      on_flush do
+        puts 'before flush'
+      end
+    end
+
+    Facter.flush
+  EOM
+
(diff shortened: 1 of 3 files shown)


sebastian-miclea:

acceptance/tests/facter_flush.rb
```diff
@@ -0,0 +1,52 @@
...
+  fact_content1 = <<-EOM
+    require 'facter'
+
+    Facter.add(:fact1) do
+      'this should be flushed'
+    end
+
+    Facter.flush
+
+    puts "Fact1: \#\{Facter.value(:fact)\}"
```
Filipovici-Andrei, line 13: you are asking for a fact that doesn't exist

spec/facter/facter_spec.rb
```diff
@@ -403,6 +403,20 @@ def mock_collection(method, os_name = nil, error = nil)
       end
     end
 
+    describe '#flush' do
+      it 'sends call to LegacyFacter' do
+        allow(LegacyFacter).to receive(:flush)
```
Filipovici-Andrei, line 408: unit test should be separated according to the AAA pattern (Arrange, Act, Assert).

Filipovici-Andrei (changes requested):

spec/facter/facter_spec.rb
sebastian-miclea: Updated. Thanks!

acceptance/tests/facter_flush.rb
sebastian-miclea: Yes, missed that. Thanks

Filipovici-Andrei (approved):

PR #2143, opened 2020-10-15
(FACT-2811) part2. Fixed all unit tests which failed with rspec --order rand

Filipovici-Andrei:
This PR fixes the remaining unit tests that fail when running them with rspec --order random

- Removed all class instance attributes from the resolvers which were caching their values. This meant that the invalidate_cache method for resolvers was not working properly.
- Instantiated the FactManager singleton for every unit test that needed it.
- Added default values for all OptionStore attributes and included them in the reset method.
- check.sh now uses rspec --order random
- Refactored some unit tests for linux partitions resolver.


Filipovici-Andrei:

Filipovici-Andrei:

sebastian-miclea (approved):

PR #2146, opened 2020-10-16
(FACT-2829) Fixed partitions and mount points facts

Filipovici-Andrei:
**Problem:** 
On ubuntu 20.04, amazon vm, the partitions paths were detected as '/dev/sys/block/<name>'
instead of '/dev/<name>'.
The root partition was detected by it's PARTUUID instead of it's path.
**Fix:**
'sys/block' is not added to the partitions path if it's data is extracted from '/sys/block/<name>/dm', or
'/sys/block/<name>/loop' folders.
If /proc/cmdline returns the root partition's PARTUUID then, if possible, we make an extra call to blkid to get it's path.


Filipovici-Andrei:

sebastian-miclea (approved):

oanatmaria (approved):

PR #2149, opened 2020-10-21
(maint) Use strings instead of symbols for os names.

IrimieBogdan:
If symbols are used for os names, the matching is not done correctly and the os hierarchy cannot be detected.


IrimieBogdan:

oanatmaria (approved):

PR #2151, opened 2020-10-22
(maint) Improve FreeBSD version detection

smortex:
`freebsd-version -k` returns the installed version of the installed kernel, but jails might have no kernel installed because they are running from the host's kernel.  In such a situation, freebsd-version fails — which is expected — but the other information we are interested in not gathered.

Rework the resolver so that freebsd-version is run twice, so that we can still gather the information of the running kernel and installed userland when no installed kernel is available.

### DIFF
--- a/lib/facter/resolvers/freebsd/freebsd_version_resolver.rb
+++ b/lib/facter/resolvers/freebsd/freebsd_version_resolver.rb
@@ -14,10 +14,13 @@ module Facter
           end
 
           def freebsd_version_system_call(fact_name)
-            output = Facter::Core::Execution.execute('/bin/freebsd-version -kru', logger: log)
-            return if output.empty?
+            output = Facter::Core::Execution.execute('/bin/freebsd-version -k', logger: log)
 
-            build_fact_list(output)
+            @fact_list[:installed_kernel] = output.strip unless output.empty?
+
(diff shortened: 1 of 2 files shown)


lib/facter/resolvers/freebsd/freebsd_version_resolver.rb
```diff
@@ -14,20 +14,22 @@ def post_resolve(fact_name)
...
+
+            build_fact_list(output) unless output.empty?
 
             @fact_list[fact_name]
           end
 
           def build_fact_list(output)
             freebsd_version_results = output.split("\n")
 
-            @fact_list[:installed_kernel]   = freebsd_version_results[0].strip
```
igalic, line 28: why is this fact gone now?

igalic (commented):
🙋🏻‍♀️

lib/facter/resolvers/freebsd/freebsd_version_resolver.rb
smortex: It's not gone, it's now collected on line 19.

This `build_fact_list` function is a bit misleading since it only does 2/3 of the job… Maybe moving the code to the calling function directly would be less confusing?

igalic: `build_some_facts_list` 😅

oanatmaria (approved):

PR #2154, opened 2020-10-26
(FACT-2813) Every facter call in the acceptance tests is run with the trace option

Filipovici-Andrei:
All acceptance tests are run with the trace=true option. 
This way, in case of an exception inside facter, the exception's backtrace 
is printed.

### DIFF
--- a/acceptance/tests/options/blocklist_no_regex.rb
+++ b/acceptance/tests/options/blocklist_no_regex.rb
@@ -1,0 +1,53 @@
+test_name 'blocking os fact does not block oss fact' do
+  tag 'risk:high'
+
+  require 'facter/acceptance/user_fact_utils'
+  extend Facter::Acceptance::UserFactUtils
+
+  custom_fact_file = 'custom_facts.rb'
+  custom_fact_name = "oss"
+  custom_fact_value = "custom_fact_value"
+
+  fact_content = <<-CUSTOM_FACT
+  Facter.add(:#{custom_fact_name}) do
+    setcode do
+      "#{custom_fact_value}"
+    end
+  end
+  CUSTOM_FACT
+
+  config_data = <<~FACTER_CONF
+    facts : {
+      blocklist : [ "os" ],
+    }
+  FACTER_CONF
+
+  agents.each do |agent|
+    fact_dir = agent.tmpdir('custom_facts')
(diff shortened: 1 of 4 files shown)


Filipovici-Andrei:

gimmyxd:
There is a simpler way to do this, without env variables and CI setup changes.
You can set options for beaker in config(eg. https://github.com/puppetlabs/facter/blob/main/acceptance/config/aio/options.rb) and this will be available in the test files in the `@options` variable

```
❯ git diff
--------------------------------------------------------------------------------------------------------------------------------------------------------------
modified: acceptance/config/aio/options.rb
--------------------------------------------------------------------------------------------------------------------------------------------------------------
@@ -1,3 +1,4 @@
 {
   :type => 'aio',
+  :cli_arguments => "--trace",
 }
--------------------------------------------------------------------------------------------------------------------------------------------------------------
modified: acceptance/tests/facter_returns_success_on_non_existent_fact.rb
--------------------------------------------------------------------------------------------------------------------------------------------------------------
@@ -3,7 +3,7 @@

   agents.each do |agent|
     step 'facter should return exit code 0 for querying non-existing-fact without --strict flag' do
-      on(agent, facter('non-existing-fact'), :acceptable_exit_codes => 0)
+      on(agent, facter("non-existing-fact #{@options[:cli_arguments]}"), :acceptable_exit_codes => 0)
     end
   end
 end
```

Filipovici-Andrei:

acceptance/lib/puppet/acceptance/common_utils.rb
```diff
@@ -2,10 +2,13 @@ module Puppet
   module Acceptance
     module CommandUtils
       def ruby_command(host)
+        extra_option = ' '
         if host['platform'] =~ /windows/ && !host.is_cygwin?
-          "cmd /V /C \"set PATH=#{host['privatebindir']};!PATH! && ruby\""
+          extra_option = " && set OPTIONS=#{@options[:facter_opt]} " if @options[:facter_opt]
+          "cmd /V /C \"set PATH=#{host['privatebindir']};!PATH!#{extra_option}&& ruby\""
```
gimmyxd, line 8: Why do you need to add the `--trace` here?

PR #2157, opened 2020-10-27
(FACT-2839) Added extra check for TTLS keys

sebastian-miclea:
This extra check helps in optimising the cache mechanism in order not to check wether a fact needs to be cached unless facts or groups are defined in facter.conf


sebastian-miclea:

IrimieBogdan (approved):

PR #2158, opened 2020-10-27
(maint) Fix puppet facts acceptance tests

sebastian-miclea:
Fixed two small issues when running puppet facts. 
1. Legacy facts were not shown by default
2. Blocked facts debug message was not printed.

### DIFF
--- a/lib/facter.rb
+++ b/lib/facter.rb
@@ -339,13 +339,15 @@ module Facter
     # @param options [Hash] parameters for the fact - attributes
     #   of {Facter::Util::Fact} and {Facter::Util::Resolution} can be
     #   supplied here
-    # @param user_queries [String] the fact names
+    # @param user_queries [Array] the fact names
     #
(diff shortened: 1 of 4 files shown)


sebastian-miclea:

lib/facter.rb
```diff
@@ -346,6 +346,8 @@ def value(user_query)
     # @api public
     def values(options, user_queries)
       init_cli_options(options, user_queries)
+      Options[:show_legacy] = true
```
IrimieBogdan, line 349: `show_legacy` should not be required. If the user queries specifically for a set of facts, Facter will automatically search in legacy fact as well.
e.g.

```ruby
require 'facter'

puts Facter.values({},["uptime"])
```

`uptime` is a legacy fact, and the result is 
```
{"uptime"=>"6 days"}
```

IrimieBogdan (approved):

PR #2161, opened 2020-10-29
(maint) Fix fact augmenter

IrimieBogdan:
When there was a searched fact with a regex in it's name, the fact augmenter tried to match it to a resolved fact using the user query, which is wrong because when no user query is provided, it will match all facts, an it will wrongfully augment them.


IrimieBogdan:

sebastian-miclea (approved):

PR #2162, opened 2020-10-30
FACT-2753-parallel-semaphore

IrimieBogdan:
Resolve facts by default in parallel (concurrently with MRI ruby). Sequential resolution of facts can be enables via the --sequential cli flag or via facter.conf (via de global.sequantial)

The semaphore is extracted in the base class compared to #2051


IrimieBogdan:

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -47,5 +55,40 @@ def resolve(searched_facts)
...
+    def start_threads(searched_facts)
+      threads = []
+      # only resolve a fact once, even if multiple search facts depend on that fact
+      searched_facts
+        .uniq { |searched_fact| searched_fact.fact_class.name }
+        .each do |searched_fact|
+        threads << Thread.new do
+          resolve_fact(searched_fact)
+        end
+      end
```
gimmyxd, line 68: you can replace `each` with `map`

```ruby
searched_facts
  .uniq { |searched_fact| searched_fact.fact_class.name }
  .map do |searched_fact|
    Thread.new do
      resolve_fact(searched_fact)
    end
  end
```

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -47,5 +55,40 @@ def resolve(searched_facts)
...
+
+      threads
+    end
+
+    def join_threads(threads, searched_facts)
+      resolved_facts = []
+
+      threads.each do |thread|
+        thread.join
```
gimmyxd, line 77: i think Thread#join is not needed as Thread#value(https://ruby-doc.org/core-2.5.1/Thread.html#method-i-value) is using join.

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
```diff
@@ -47,5 +55,40 @@ def resolve(searched_facts)
...
+        resolved_facts << thread.value unless thread.value.nil?
+      end
+
+      resolved_facts.flatten!
+
+      FactAugmenter.augment_resolved_facts(searched_facts, resolved_facts)
```
gimmyxd, line 83: Might be able to write this shorter as
```ruby
resolved_facts = threads.map(&:value)
resolved_facts.compact!

FactAugmenter.augment_resolved_facts(searched_facts, resolved_facts)
```

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: Removed join

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: Replaced each with map 👍

lib/facter/framework/core/fact/internal/internal_fact_manager.rb
IrimieBogdan: Great suggestion, refactored.

IrimieBogdan:

gimmyxd (approved):

IrimieBogdan:

PR #2165, opened 2020-11-04
(FACT-2792) Show warning and facter facts when using -p argument

IrimieBogdan:
Show facter facts even when `-p` argument is used to keep compatibility with components that have not yet migrated to `puppet facts show`.


IrimieBogdan:

sebastian-miclea (approved):

PR #2170, opened 2020-11-07
(maint) Provide virtual and is_virtual facts on FreeBSD

smortex:
This PR add the `virtual` and `is_virtual` facts on FreeBSD which seems to have disappeared at some point in Facter 4.

I was able to check the correct behavior on real systems:
* `physical` on bare metal laptop;
* `kvm` at a cloud service provider;
* `jail` in a jail on my laptop.

Overall progress:

* [x] Replace the pciconf provider with one rely on on the `kern.vm_guest` sysctl as spoted by @igalic bellow;
* [x] Ensure the FreeBSD-specific code is only run on FreeBSD.

### DIFF
--- a/lib/facter/facts/freebsd/is_virtual.rb
+++ b/lib/facter/facts/freebsd/is_virtual.rb
@@ -18,7 +18,7 @@ module Facts
       private
 
       def check_if_virtual(found_vm)
-        Facter::FactsUtils::PHYSICAL_HYPERVISORS.count(found_vm).zero?.to_s
+        Facter::FactsUtils::PHYSICAL_HYPERVISORS.count(found_vm).zero?
       end
     end
   end
--- a/lib/facter/facts/freebsd/is_virtual.rb
+++ b/lib/facter/facts/freebsd/is_virtual.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Facts
+  module Freebsd
+    class IsVirtual
+      FACT_NAME = 'is_virtual'
+
+      def initialize
+        @virtual = Facter::VirtualDetector.new
+      end
+
+      def call_the_resolver
+        fact_value = @virtual.platform
+
+        Facter::ResolvedFact.new(FACT_NAME, check_if_virtual(fact_value))
+      end
+
+      private
+
+      def check_if_virtual(found_vm)
+        Facter::FactsUtils::PHYSICAL_HYPERVISORS.count(found_vm).zero?.to_s
+      end
+    end
+  end
+end
--- a/lib/facter/facts_utils/virtual_detector.rb
+++ b/lib/facter/facts_utils/virtual_detector.rb
@@ -7,8 +7,9 @@ module Facter
     end
 
     def platform # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-      fact_value = check_docker_lxc || check_dmi || check_gce || retrieve_from_virt_what || check_vmware
-      fact_value ||= check_open_vz || check_vserver || check_xen || check_other_facts || check_lspci || 'physical'
+      fact_value = check_docker_lxc || check_dmi || check_freebsd || check_gce || retrieve_from_virt_what
+      fact_value ||= check_vmware || check_open_vz || check_vserver || check_xen || check_other_facts
+      fact_value ||= check_lspci || 'physical'
 
       fact_value
     end
@@ -56,6 +57,13 @@ module Facter
     def check_xen
       @log.debug('Checking XEN')
       Facter::Resolvers::Xen.resolve(:vm)
+    end
+
+    def check_freebsd
+      return unless Object.const_defined?('Facter::Resolvers::Freebsd::Virtual')
+
+      @log.debug('Checking if jailed')
+      Facter::Resolvers::Freebsd::Virtual.resolve(:vm)
     end
 
     def check_other_facts
--- a/lib/facter/facts/freebsd/virtual.rb
+++ b/lib/facter/facts/freebsd/virtual.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+module Facts
+  module Freebsd
+    class Virtual
+      FACT_NAME = 'virtual'
+
+      def initialize
+        @log = Facter::Log.new(self)
+        @virtual = Facter::VirtualDetector.new
+      end
+
+      def call_the_resolver
+        @log.debug('FreeBSD Virtual Resolver')
+        fact_value = @virtual.platform
+        @log.debug("Fact value is: #{fact_value}")
+
+        Facter::ResolvedFact.new(FACT_NAME, fact_value)
+      end
+    end
+  end
+end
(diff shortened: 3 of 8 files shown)


lib/facter/facts_utils/virtual_detector.rb
```diff
@@ -8,7 +8,8 @@ def initialize
 
     def platform # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       fact_value = check_docker_lxc || check_dmi || check_gce || retrieve_from_virt_what || check_vmware
-      fact_value ||= check_open_vz || check_vserver || check_xen || check_other_facts || check_lspci || 'physical'
+      fact_value ||= check_jail || check_open_vz || check_vserver || check_xen || check_other_facts
+      fact_value ||= check_lspci || check_pciconf || 'physical'
```
smortex, line 12: `check_jail` and `check_pciconf` where added to this list.  I attempted to keep the names sorted by "type" and then alphabetically, but I feel like the beginning of the list makes no sense… Please tell me how to order these if you think of a better arrangement.

lib/facter/resolvers/jail.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+
+        def check_jail(fact_name)
+          @fact_list[:vm] = is_jailed
+          @fact_list[fact_name]
+        end
+
+        def is_jailed
+          require_relative 'freebsd/ffi/ffi_helper'
+          if Facter::Freebsd::FfiHelper.sysctl_by_name(:long, 'security.jail.jailed').zero?
```
smortex, line 22: This code does not make sense outside of FreeBSD.  How would you recommend to manage this?

lib/facter/resolvers/pciconf.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module Facter
+  module Resolvers
+    class Pciconf < BaseResolver
+      init_resolver
+
+      REGEX_VALUES = { 'VirtualBox' => 'virtualbox', 'XenSource' => 'xenhvm',
+                       'Microsoft Corporation Hyper-V' => 'hyperv', 'Class 8007: Google, Inc' => 'gce',
+                       'VM[wW]are' => 'vmware', '1ab8:' => 'parallels', '[Pp]arallels' => 'parallels',
+                       '(?i)(virtio)' => 'kvm' }.freeze
```
smortex, line 11: This list is a duplicate of the one from the `Lspci` class (`pciconf` is basically the `lspci` of FreeBSD).  Maybe we can avoid duplication but I do not really know the best way to achieve this.

smortex:
Cc @igalic maybe you will be able to give this a shot? I added a bunch of in-line comments about things which worry me, if you can have a look and provide guidance, I would be really pleased!

lib/facter/resolvers/pciconf.rb
igalic: over in cloud-init, we use `sysctl -qn kern.vm_guest` and then have this translation table https://github.com/canonical/cloud-init/blob/master/tools/ds-identify#L408

igalic: also, using sysctl FFI is probably faster than shelling out

smortex: w00t!  That look like such a better solution than pciconf!

smortex: I could take some time to work on this PR.  I implemented what you suggested: thanks for pointing-out this sysctl :wink:

lib/facter/facts_utils/virtual_detector.rb
```diff
@@ -58,6 +59,13 @@ def check_xen
       Facter::Resolvers::Xen.resolve(:vm)
     end
 
+    def check_freebsd
+      return unless Object.const_defined?('Facter::Resolvers::Freebsd::Virtual')
```
smortex, line 63: I had to use that trick so that the test suite pass on Linux (CI suite on GitHub).  The file already use `Facter::Resolvers::Linux::DmiBios` which does not cause trouble running the test suite on FreeBSD, so I guess something magic can be done somewhere to improve this situation :wink:

igalic (approved):
looks a lot clearer!

lib/facter/facts/freebsd/is_virtual.rb
```diff
@@ -0,0 +1,25 @@
...
+      def call_the_resolver
+        fact_value = @virtual.platform
+
+        Facter::ResolvedFact.new(FACT_NAME, check_if_virtual(fact_value))
+      end
+
+      private
+
+      def check_if_virtual(found_vm)
+        Facter::FactsUtils::PHYSICAL_HYPERVISORS.count(found_vm).zero?.to_s
```
oanatmaria, line 21: is_virtual should be boolean, please drop the `to_s` method. (there is a bug regarding this inconsistency in fact types in the Linux `is_virtual` fact as well)

lib/facter/facts_utils/virtual_detector.rb
oanatmaria: You can try something like: `return unless Facter::Resolvers::Uname.resolve(:kernelname) == 'FreeBSD'` (not sure if the kernel name is exactly that string)

smortex: I guess it will make things worst if at some point support is added for an OS which is a FreeBSD derivative, e.g. HardenestBSD.  In such a case, this system would see and could use the FreeBSD classes, like for example the FreeBSD classes could see and use the Solaris ones when FreeBSD was (incorrectly) a Solaris derivative.

Maybe if there is no better trick for now, this can stay as is?

lib/facter/facts/freebsd/is_virtual.rb
smortex: I added a commit to fix this.  While here I rebased these changes on top of master.

lib/facter/facts_utils/virtual_detector.rb
igalic: same for Dragonfly, but i don't know how many people use that on servers

we need some kind of "family"?

igalic: can you not a HardenedBSD and show me what (the different versions of) `facter os` look like?

oanatmaria: We decide which hierarchy to load based on RbConfig::CONFIG['host_os'] (in case of FreeBSD the result is /freebsd/ https://github.com/puppetlabs/facter/blob/21c0f1a94b79c63f520f42317b6499e15b862084/lib/facter/framework/detector/os_detector.rb#L19) maybe you could use that, if there are scenarios when that is wrong also, I'm ok to leave things as they are.

smortex: @igalic, I do not have a HardenedBSD at hand, it's still on my list to things to do :yum: I similarly do not have any DragonflyBSD at hand, but I guess both systems would be detected as "bsd" and adjusting `RbConfig::CONFIG['host_os']` as suggested by @oanatmaria might be useful to improve what facter returns?

Adding support for these os is low priority for me now, so I prefer we keep it that way for now and maybe improve this if required when support of these derivative is added to facter.

If it's okay for you, we can resolve this conversation and the PR can be merged :tada:

oanatmaria (approved):

PR #2181, opened 2020-11-16
(FACT-2860) Agent release pipeline updates

GabrielNagy:
- add the packaging gem to the Gemfile
- remove '~> 2.0' constraint for bundler dependency (jenkins workers have bundler 1 and 2 can't be easily installed)
- update Rakefile to load packaging rake tasks
- create a file similar to
  https://github.com/puppetlabs/bolt/blob/main/ext/build_defaults.yaml
  so the `pl:jenkins:ship_to_artifactory` task doesn't fail horribly

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,8 @@ group(:release, optional: true) do
   gem 'octokit', '~> 4.18.0'
 end
 
+gem 'packaging', require: false
+
 local_gemfile = File.expand_path('Gemfile.local', __dir__)
 eval_gemfile(local_gemfile) if File.exist?(local_gemfile)
 
--- a/facter.gemspec
+++ b/facter.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.add_development_dependency 'bundler', '~> 2.0'
+  spec.add_development_dependency 'bundler'
   spec.add_development_dependency 'coveralls', '~> 0.8.23'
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -8,6 +8,8 @@ group(:release, optional: true) do
   gem 'octokit', '~> 4.18.0'
 end
 
+gem 'packaging', require: false
```
GabrielNagy, line 11: We can also add the gem to the `:release` group, but then we'd have to update the CJC artifact generation step to run `bundle install --with release`.

facter.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
-  spec.add_development_dependency 'bundler', '~> 2.0'
+  spec.add_development_dependency 'bundler'
```
gimmyxd, line 27: Not sure we would need to have bundler as a dependency at all. Creating a new gem using `bundle gem new_gem` does not add any dependency to bundler.
If we decide to keep it, we should lock it to avoid breaking changes, rubocop is using `'bundler', '>= 1.15.0', '< 3.0'`

Gemfile
gimmyxd: Don't really have a strong opinion from this, adding it to the release group would enable us to install it only when needed(with the change to cjc that you mentioned, since that is the only place where it is used.)

facter.gemspec
GabrielNagy: Removed the bundler dependency.

gimmyxd (approved):

florindragos (approved):

PR #2189, opened 2020-11-17
(maint) Added some new fact groups

Filipovici-Andrei:
Added some fact groups that cfacter displays dynamically, depending on the platform it's running on.

### DIFF
--- a/lib/facter/config.rb
+++ b/lib/facter/config.rb
@@ -42,12 +42,19 @@ module Facter
     end
     unless defined? FACT_GROUPS
       FACT_GROUPS = {
+        'AIX NIM type' => [
+          'nim_type'
+        ],
         'EC2' => %w[
           ec2_metadata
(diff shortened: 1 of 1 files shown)


Filipovici-Andrei:
On AIX there's also a group called AIX serial number with the fact serialnumber. That fact isn't displayed when you request all facts. It's resolved only when you make a query specifically for it. Is this a cfacter AIX easter egg?

oanatmaria:
We report the `serialnumber` fact on AIX also: https://github.com/puppetlabs/facter/blob/21c0f1a94b79c63f520f42317b6499e15b862084/lib/facter/facts/aix/serialnumber.rb#L7 , but because it's a legacy fact it's only displayed when you query specifically for it or if you run facter with `--show-legacy`.

oanatmaria (approved):

PR #2196, opened 2020-11-23
(FACT-2859) Implement load_external API method

oanatmaria:


oanatmaria:

florindragos (approved):

acceptance/tests/facter_load_external.rb
```diff
@@ -0,0 +1,65 @@
+test_name "Facter should not diaplay external facts when 'load_external' method is called with false"  do
```
Filipovici-Andrei, line 1: typo in display

acceptance/tests/facter_load_external.rb
```diff
@@ -0,0 +1,65 @@
...
+    end
+
+    step "Agent #{agent}: create facts.d directory and fact" do
+      agent.mkdir_p(factsd)
+      create_remote_file(agent, ext_fact, content)
+      agent.chmod('+x', ext_fact)
+    end
+
+    step "Agent #{agent}: ensure that external fact is loaded and resolved" do
+      on(agent, "#{Puppet::Acceptance::CommandUtils.ruby_command(agent)} #{script_name2}", :acceptable_exit_codes => 1)
```
Filipovici-Andrei, line 57: Puppet::Acceptance::CommandUtils.ruby_command does the same thing as simple ruby_command from beaker-puppet. We should prefer the latter.

lib/facter/framework/core/options/option_store.rb
```diff
@@ -178,6 +178,7 @@ def reset
         @block = true
         @cli = nil
         @custom_facts = true
+        @external_facts = true
```
Filipovici-Andrei, line 181: external_facts is being reset in the reset_config method

spec/facter/facter_spec.rb
```diff
@@ -435,6 +435,25 @@ def mock_collection(method, os_name = nil, error = nil)
       end
     end
 
+    describe '#load_external' do
+      before do
```
Filipovici-Andrei, line 439: probably we should add a test when the external facts are not loaded

Filipovici-Andrei (changes requested):

spec/facter/facter_spec.rb
oanatmaria: There already is an acceptance test added with both scenarios.

PR #2197, opened 2020-11-24
(FACT-2857) Added acceptance test for user query scenario

sebastian-miclea:


acceptance/tests/options/config_file/ttls_cached_custom_and_core_fact.rb
```diff
@@ -0,0 +1,126 @@
...
+    end
+
+    step "should create cache file once for custom_fact" do
+      on(agent, facter(custom_fact_name, environment: env))
+      output = agent.ls_ld(cache_folder + "/*").split("\n")
+
+      step "custom fact value is correct" do
+        cat_output = agent.cat("#{cache_folder}/#{custom_cache_group}")
+        cache = JSON.parse(cat_output)
+        assert_equal(custom_fact_value, cache[custom_fact_name])
```
IrimieBogdan, line 63: We should also check that the core fact did not get cached at this step

acceptance/tests/options/config_file/ttls_cached_custom_and_core_fact.rb
```diff
@@ -0,0 +1,126 @@
...
+      step "only 1 cache file is created" do
+        assert_equal(output.count, 1)
+      end
+    end
+
+    step "should update cache file once for core fact" do
+      core_value = on(agent, facter(core_fact_name, environment: env)).stdout.strip
+      ls_output = agent.ls_ld(cache_folder + "/*").split("\n")
+
+      step "custom fact value is correct" do
```
IrimieBogdan, line 79: This step checks core values, the description should be renamed to `core fact value is correct`

acceptance/tests/options/config_file/ttls_cached_custom_and_core_fact.rb
```diff
@@ -0,0 +1,126 @@
...
+        cat_output = agent.cat("#{cache_folder}/#{custom_cache_group}")
+        cache = JSON.parse(cat_output)
+
+        assert_equal(core_value, cache[core_fact_name].to_s)
```
IrimieBogdan, line 83: We should check that the custom facts is not in the cache file at this step.

acceptance/tests/options/config_file/ttls_cached_custom_and_core_fact.rb
```diff
@@ -0,0 +1,126 @@
+test_name 'missing facts should not invalidate cache' do
```
IrimieBogdan, line 1: The test does not cover all the cases from FACT-2857 description.
The test should
- check we can cache a fact
- check that when we query the second time for the same fact we get the fact value from the cache
- request to_hash
- check that the cache is no longer changed, no matter what fact we ask for

acceptance/tests/options/config_file/ttls_cached_custom_and_core_fact.rb
```diff
@@ -0,0 +1,141 @@
...
+    end
+
+    step 'request the core fact again' do
+      core_value = on(agent, facter(core_fact_name, environment: env)).stdout.strip.to_i
+
+      cat_output = agent.cat("#{cache_folder}/#{custom_cache_group}")
+      cache = JSON.parse(cat_output)
+
+      step 'check that it cached the value it printed' do
+        assert_equal(core_value, cache[core_fact_name])
```
Filipovici-Andrei, line 79: Here you should check that the cached value coincides with the one retrieved in the first request. Making the check like this, you can't be sure that facter doesn't simply solve the core fact again and rewrites the cache file.

Filipovici-Andrei (changes requested):

Filipovici-Andrei (approved):

PR #2198, opened 2020-11-25
(FACT-2870) Don't split legacy fact names

IrimieBogdan:
The name of some legacy facts gets composed at runtime. e.g. `network_<network_name>`. The network_name can contain dots (`.`) in it's name.
```
network_ens33
network_ens33.13
network_ens33.14
```

Facter interprets dots as composed facts and will try to create a fact hierarchy but will fail. Because legacy facts are never composed, we should not try to create a fact hierarchy for facts with legacy type.

This PR skips fact hierarchy creation for `legacy` facts and just uses their name as it is, no special treatment is given to dots.

### DIFF
--- a/lib/facter/models/fact_collection.rb
+++ b/lib/facter/models/fact_collection.rb
@@ -40,11 +40,16 @@ module Facter
     private
 
     def bury_fact(fact)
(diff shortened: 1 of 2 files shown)


florindragos (approved):

IrimieBogdan:
Jenkins please test this

PR #2200, opened 2020-11-26
(PUP-10793) Exclude nil custom facts from all API calls except fact()

IrimieBogdan:
Exclude custom fact with `nil` value from `to_user_output`, `values` and `to_hash` ruby API's.


IrimieBogdan:
Jenkins please test this

IrimieBogdan:
The branch has changed name to #2204

PR #2204, opened 2020-12-02
(FACT-2881) Exclude nil custom facts from Facter API

IrimieBogdan:
Exclude custom fact with nil value from to_user_output, values and to_hash ruby API's.

facter-apis


IrimieBogdan:
Jenkins please test this

oanatmaria (approved):

sebastian-miclea (approved):

PR #2207, opened 2020-12-03
Fix domain / fqdn resolution on CentOS / Debian (at least)

smortex:
After updating a CentOS node to Puppet 7, I saw that the domain name / fqdn facts have changed:

```sh-session
romain@nsXXXXXX ~ % facter hostname
nsXXXXXX
romain@nsXXXXXX ~ % facter fqdn    
nsXXXXXX.ovh.net
romain@ns565980 ~ % hostname -f
nsXXXXXX.vittoria.pro
```

CentOS seems to use a reverse IP lookup to get the FQDN from the hostname.  But facter reads /etc/resolv.conf which in my case is:

```
nameserver 127.0.0.1
search ovh.net
```

I reverted to the 6.x version of Puppet.

The same problem probably affect at least Debian nodes.

### DIFF
--- a/acceptance/tests/ticket_1238_hostname_fqdn.rb
+++ b/acceptance/tests/ticket_1238_hostname_fqdn.rb
@@ -4,8 +4,9 @@ test_name 'C93827: facter fqdn should return the hostname when its a fully quali
 
   confine :except, :platform => 'windows'
 
-  fqdn = 'foo.bar.example.org'
-  fqdn_long = 'a23456789.b23456789.c23456789.d23456789.e23456789.f23456789.wxyz'
+  fqdn_hostname = 'foo'
+  fqdn_domain = 'bar.example.org'
+  fqdn = "#{fqdn_hostname}.#{fqdn_domain}"
 
   agents.each do |agent|
     orig_hostname = on(agent, 'hostname').stdout.chomp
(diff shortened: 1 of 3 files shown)


smortex:
Just checked on a Debian node after fetching the latest facter 4 code (main branch at fd9beb8e809e3c6ef5f43af409d77e20396a5044):

```sh-session
romain@optiplex-3050 ~/Projects/facter % hostname -f
optiplex-3050.vittoria.pro
romain@optiplex-3050 ~/Projects/facter % bundle exec bin/facter fqdn domain hostname
domain => 
fqdn => optiplex-3050.
hostname => optiplex-3050
```

With this PR applied, I get the expected values:

```sh-session
romain@optiplex-3050 ~/Projects/facter % hostname -f
optiplex-3050.vittoria.pro
romain@optiplex-3050 ~/Projects/facter % bundle exec bin/facter fqdn domain hostname
domain => vittoria.pro
fqdn => optiplex-3050.vittoria.pro
hostname => optiplex-3050
```

natemccurdy:
What was the behavior in facter 3?

Not knowing the full history of this fact with older versions of Facter, I'd intuitively expect this behavior from Facter:
* `hostname` fact should match the output of running `/usr/bin/hostname`
* `fqdn` fact should match the output of running `/usr/bin/hostname -f`

How would one retrieve the short hostname from Facter with this patch?

smortex:
@natemccurdy with this patch, Facter 4 facts are backward compatible with Facter 3.

On some operating systems, `hostname` returns the FQDN (e.g. FreeBSD) and on other systems only the local part of the hostname (e.g. Linux).  So relying on `hostname` does not result in consistent results.

For this reason, `hostname -f` and `hostname -s` are preferred to get the FQDN or only the local part of the name in a portable way.

Facter's `hostname` fact is everything before the first `.` in the FQDN.  My feeling is that it's not 100% safe, but I suspect nobody set a local part with a dot in the real world.

smortex:
Close in favor for #2208

PR #2213, opened 2020-12-07
(FACT-2884) Add option to disable ec2 metadata query

florindragos:


oanatmaria (approved):

lib/facter/resolvers/ec2.rb
```diff
@@ -18,9 +18,14 @@ def post_resolve(fact_name)
         end
 
         def read_facts(fact_name)
-          @fact_list[:metadata] = {}
-          query_for_metadata(EC2_METADATA_ROOT_URL, @fact_list[:metadata])
-          @fact_list[:userdata] = get_data_from(EC2_USERDATA_ROOT_URL).strip
+          @fact_list[:metadata] = ENV['FACTER_ec2_metadata']
```
mihaibuzgau, line 21: I don't think this should be part of the resolver. Overriding the core facts with Environment facts should work out of the box

mihaibuzgau (changes requested):

PR #2218, opened 2020-12-08
(FACT-2894) Facter fails to retrieve fqdn

oanatmaria:
**Description of the problem:** Facter fails when Socket.getaddrinfo is called and it prevents fqdn information to be retrieved. 
**Description of the fix:** Rescue SocketError and continue resolving fqdn.

### DIFF
--- a/lib/facter/resolvers/hostname_resolver.rb
+++ b/lib/facter/resolvers/hostname_resolver.rb
@@ -42,7 +42,12 @@ module Facter
         end
 
         def retrieve_with_addrinfo(host)
-          name = Socket.getaddrinfo(host, 0, Socket::AF_UNSPEC, Socket::SOCK_STREAM, nil, Socket::AI_CANONNAME)[0]
+          begin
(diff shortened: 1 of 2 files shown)


lib/facter/resolvers/hostname_resolver.rb
```diff
@@ -42,7 +42,12 @@ def retrieve_from_fqdn(output)
         end
 
         def retrieve_with_addrinfo(host)
-          name = Socket.getaddrinfo(host, 0, Socket::AF_UNSPEC, Socket::SOCK_STREAM, nil, Socket::AI_CANONNAME)[0]
+          begin
+            name = Socket.getaddrinfo(host, 0, Socket::AF_UNSPEC, Socket::SOCK_STREAM, nil, Socket::AI_CANONNAME)[0]
+          rescue SocketError => e
```
joshcooper, line 47: Thanks, this solves the issue I was seeing, but one question, can `getaddrinfo` raise exceptions other than `SocketError`? Should we do `rescue => e` instead?

PR #2224, opened 2020-12-10
(maint) to_hash API method should return a hash

oanatmaria:
**Description of the problem:** dig on $facts (`'dig($facts, "os", "family")'`) fails with Facter 4. Facter's API method `to_hash` returns a different data type (Facter::FactCollection) than Facter 3.
**Description of the fix:** Ensure that `to_hash` method returns a ruby Hash instance.

### DIFF
--- a/lib/facter.rb
+++ b/lib/facter.rb
@@ -334,7 +334,8 @@ module Facter
 
       resolved_facts = Facter::FactManager.instance.resolve_facts
       resolved_facts.reject! { |fact| fact.type == :custom && fact.value.nil? }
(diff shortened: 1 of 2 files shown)


gimmyxd:
This should have a ticket as it is an issue that affects people who are using Puppet 7

sebastian-miclea (approved):

oanatmaria:
> This should have a ticket as it is an issue that affects people who are using Puppet 7

https://tickets.puppetlabs.com/browse/FACT-2897

florindragos (approved):

PR #2226, opened 2020-12-10
(FACT-2898) resolve facts sequential by default

mihaibuzgau:
Spawning one thread for each fact might cause facter to take all the file descriptors and thrown the following error:
```
Error: Could not retrieve local facts: can't create Thread: Resource
temporarily unavailable
```
This PR sets sequential to true, making fact resolution sequential.

### DIFF
--- a/lib/facter/framework/core/options/option_store.rb
+++ b/lib/facter/framework/core/options/option_store.rb
@@ -26,7 +26,7 @@ module Facter
     @config_file_external_dir = []
     @default_external_dir = []
     @fact_groups = {}
-    @sequential = false
+    @sequential = true
     @ttls = []
     @block_list = []
     @color = true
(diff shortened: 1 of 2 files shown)


joshcooper (approved):
LGTM

gimmyxd:
Defaulting to sequential menans that `--sequential` flag is not needed anymore: https://github.com/puppetlabs/facter/blob/main/lib/facter/framework/cli/cli.rb#L96-L98
Should we also remove that?

IrimieBogdan (approved):

IrimieBogdan:
@gimmyxd  I would keep `--sequential` as we can still use `--sequential false` and it will run in parallel, also in `facter.conf` we can set `sequential: false` and it will run in paralel.

PR #2232, opened 2020-12-14
(FACT-2899) Add integration tests

IrimieBogdan:
Add integration tests. The tests use rspec, and we have a special tag named `integration`.


tasks/spec.rake
```diff
@@ -0,0 +1,30 @@
...
+    until (raw_line = stderr.getc).nil?
+      print raw_line.to_s
+    end
+  end
+end
+
+desc 'Run rspec test in sequential order'
+task :spec do
+  execute_and_stream('bundle exec rspec')
+end
```
gimmyxd, line 20: Instead of relying in open3, could we rely on rspec mechanism of task def?
https://relishapp.com/rspec/rspec-core/docs/command-line/rake-task
we can decide on which files and flags are used on each rake task.

IrimieBogdan: Great suggestion, made the proposed changes.

oanatmaria (approved):

PR #2234, opened 2020-12-15
(FACT-2862) Add acceptance test for caching core and custom fact in the same group

florindragos:


acceptance/tests/options/config_file/ttls/cached_custom_group_core_custom_facts.rb
```diff
@@ -0,0 +1,79 @@
...
+        on(agent, facter("--debug --json", environment: env)) do |facter_result|
+          output_json = JSON.parse(facter_result.stdout.chomp)
+          custom_fact_value = output_json['system_uptime']['seconds']
+          assert_match(/#{custom_fact_value}/, facter_result.stdout.chomp, "#{custom_fact_name} value changed")
+          assert_match(/caching values for cached-custom-facts facts/, facter_result.stderr,
+                      'Expected debug message to state that custom facts will be cached')
+        end
+      end
+
+      step "should create a cached-custom-facts cache file that containt fact information" do
```
Filipovici-Andrei, line 61: typo in `contains`

acceptance/tests/options/config_file/ttls/cached_custom_group_core_custom_facts.rb
```diff
@@ -0,0 +1,79 @@
...
+        agent.rm_rf(fact_dir)
+        agent.rm_rf("#{cache_folder}/*")
+        agent.rm_rf(config_file)
+      end
+
+      step "should log that it creates cache file and it caches custom facts found in facter.conf" do
+        on(agent, facter("--debug --json", environment: env)) do |facter_result|
+          output_json = JSON.parse(facter_result.stdout.chomp)
+          custom_fact_value = output_json['system_uptime']['seconds']
+          assert_match(/#{custom_fact_value}/, facter_result.stdout.chomp, "#{custom_fact_name} value changed")
```
Filipovici-Andrei, line 55: I think that you should do a second request for the cached fact, and compare that value, with what facter returned in the first request

acceptance/tests/options/config_file/ttls/cached_custom_group_core_custom_facts.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      step "should create a cached-custom-facts cache file that containt fact information" do
+        result = agent.file_exist?("#{cache_folder}/cached-custom-facts")
+        assert_equal(true, result)
+        cat_output = agent.cat("#{cache_folder}/cached-custom-facts")
+        assert_match(/#{custom_fact_value}/, cat_output.strip, 'Expected cached custom fact file to contain fact information')
+        assert_match(/system_uptime/, cat_output.strip, 'Expected cached file to contain system_uptime information')
+      end
+
+      step 'should read from the cached file for a custom fact that has been cached' do
```
Filipovici-Andrei, line 69: Here you should also check that the retrieved fact value is the same with the one found in the cache file. Also you should check that the cache file is not being rewritten (it's modification date doesn't change)

florindragos: I'm not sure there is an easy way to determine the file modification date across all platforms with beaker

Filipovici-Andrei: I think you are correct. I created a ticket for this.

Filipovici-Andrei (approved):

PR #2237, opened 2020-12-16
(FACT-2901) Refactor linux networking resolver and add unit tests

Filipovici-Andrei:
Changes made to the networking linux resolver:
 - Refactored the resolver.
 - Updated the unit tests.
 - Enabled rubocop for it and updated .rubocop_todo.yaml.

PR #2241, opened 2020-12-18
(FACT-2895) add networking.primary6 fact for Linux

kenyon:
Returns the egress interface name for the IPv6 default route, analogous to the `networking.primary` fact for IPv4.

### DIFF
--- a/lib/facter/util/resolvers/networking/primary_interface.rb
+++ b/lib/facter/util/resolvers/networking/primary_interface.rb
@@ -56,6 +56,16 @@ module Facter
               primary_interface
             end
 
+            def read_from_ip6_route
+              return if Facter::Core::Execution.which('ip').nil?
+
+              ip_command_output = Facter::Core::Execution.execute('ip -json -6 route show default', logger: @log)
+              return if ip_command_output.nil? || ip_command_output.empty?
+
+              require 'json'
+              JSON.parse(ip_command_output).first['dev']
+            end
+
             def find_in_interfaces(interfaces)
               interfaces.each do |iface_name, interface|
                 interface[:bindings]&.each do |binding|
--- a/lib/facter/facts/linux/networking/primary6.rb
+++ b/lib/facter/facts/linux/networking/primary6.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Facts
+  module Linux
+    module Networking
+      class Primary6
+        FACT_NAME = 'networking.primary6'
+
+        def call_the_resolver
+          fact_value = Facter::Resolvers::Linux::Networking.resolve(:primary6_interface)
+
+          Facter::ResolvedFact.new(FACT_NAME, fact_value)
+        end
+      end
+    end
+  end
+end
--- a/lib/facter/resolvers/linux/networking.rb
+++ b/lib/facter/resolvers/linux/networking.rb
@@ -19,6 +19,7 @@ module Facter
             add_info_from_socket_reader
             add_info_from_routing_table
             retrieve_primary_interface
+            retrieve_primary6_interface
             Facter::Util::Resolvers::Networking.expand_main_bindings(@fact_list)
             add_flags
             @fact_list[fact_name]
--- a/spec/facter/facts/linux/networking/primary6_spec.rb
+++ b/spec/facter/facts/linux/networking/primary6_spec.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+describe Facts::Linux::Networking::Primary6 do
+  describe '#call_the_resolver' do
+    subject(:fact) { Facts::Linux::Networking::Primary6.new }
+
+    let(:value) { 'ens160' }
+
+    before do
+      allow(Facter::Resolvers::Linux::Networking).to receive(:resolve).with(:primary6_interface).and_return(value)
+    end
+
+    it 'calls Facter::Resolvers::Linux::Networking' do
+      fact.call_the_resolver
+      expect(Facter::Resolvers::Linux::Networking).to have_received(:resolve).with(:primary6_interface)
+    end
+
+    it 'returns networking.primary6 fact' do
+      expect(fact.call_the_resolver)
+        .to be_an_instance_of(Facter::ResolvedFact)
+        .and have_attributes(name: 'networking.primary6', value: value)
+    end
+
+    context 'when primary6 interface can not be retrieved' do
+      let(:value) { nil }
+
+      it 'returns nil' do
+        expect(fact.call_the_resolver)
+          .to be_an_instance_of(Facter::ResolvedFact).and have_attributes(name: 'networking.primary6', value: value)
+      end
+    end
+  end
+end
--- a/spec/facter/facts/linux/networking/primary_spec.rb
+++ b/spec/facter/facts/linux/networking/primary_spec.rb
@@ -10,7 +10,7 @@ describe Facts::Linux::Networking::Primary do
       allow(Facter::Resolvers::Linux::Networking).to receive(:resolve).with(:primary_interface).and_return(value)
     end
 
-    it 'calls Facter::Resolvers::NetworkingLinux' do
+    it 'calls Facter::Resolvers::Linux::Networking' do
       fact.call_the_resolver
       expect(Facter::Resolvers::Linux::Networking).to have_received(:resolve).with(:primary_interface)
     end
--- a/spec/facter/util/resolvers/networking/primary_interface_spec.rb
+++ b/spec/facter/util/resolvers/networking/primary_interface_spec.rb
@@ -70,6 +70,23 @@ describe Facter::Util::Resolvers::Networking::PrimaryInterface do
     end
   end
 
+  describe '#read_from_ip6_route' do
+    before do
+      allow(Facter::Core::Execution).to receive(:execute)
+        .and_return(load_fixture('ip_-json_-6_route_show_default').read)
+    end
+
+    it 'parses output from `ip -json -6 route show default`' do
+      allow(Facter::Core::Execution).to receive(:which).with('ip').and_return('/some/path')
+      expect(primary_interface.read_from_ip6_route).to eq('ens3')
+    end
+
+    it 'returns nil if route command does not exist' do
+      allow(Facter::Core::Execution).to receive(:which).with('ip').and_return(nil)
+      expect(primary_interface.read_from_ip6_route).to be_nil
+    end
+  end
+
   describe '#find_in_interfaces' do
     let(:interfaces) do
       { 'lo' =>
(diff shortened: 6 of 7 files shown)


lib/facter/resolvers/networking.rb
```diff
@@ -15,6 +15,7 @@ def post_resolve(fact_name)
         def read_facts(fact_name)
           interfaces_data
           primary_interface
+          primary6_interface
```
Filipovici-Andrei, line 18: This resolver is used by Osx and FreeBsd. Not sure it will work and you don't have the primary6 facts for these operating systems.

lib/facter/resolvers/networking_linux.rb
```diff
@@ -252,6 +253,11 @@ def retrieve_primary_interface
           @fact_list[:primary_interface] = primary_interface
         end
 
+        def retrieve_primary6_interface
+          primary_helper = Facter::Util::Resolvers::Networking::PrimaryInterface
+          primary6_interface = primary_helper.read_from_ip6_route
```
Filipovici-Andrei, line 258: This line seems to be redundant. I think that `@fact_list[:primary6_interface] = primary_helper.read_from_ip6_route` is sufficient.

Filipovici-Andrei (commented):
Hi Kenyon, 
Great job. This looks ok, just need to address the comments I added. You can add unit tests for the primary6 fact which should be similar to those for the linux primary interface.

Also you need unit tests for `read_from_ip6_route` util method.

As for testing the modifications inside the linux networking resolver, you should wait until https://github.com/puppetlabs/facter/pull/2237 gets merged, because it refactors the resolver and its unit tests.

lib/facter/resolvers/networking_linux.rb
kenyon: Yeah, I just wanted to follow the pattern used in `retrieve_primary_interface`.

lib/facter/resolvers/networking.rb
kenyon: I think that's OK, facter should just return empty string like for any other nonexistent or unresolvable fact.

kenyon:
> Hi Kenyon,
> Great job. This looks ok, just need to address the comments I added. You can add unit tests for the primary6 fact which should be similar to those for the linux primary interface.
> 
> Also you need unit tests for `read_from_ip6_route` util method.
> 
> As for testing the modifications inside the linux networking resolver, you should wait until #2237 gets merged, because it refactors the resolver and its unit tests.

I'll see about adding unit tests. I think I gave up on adding tests before because I couldn't get the existing test suite to pass locally, because the tests depended on the specific configuration of the machine running the tests (like network interface names), which is bad. But it's been a while since I worked on this, so I'll have to take a look again.

lib/facter/util/resolvers/networking/primary_interface.rb
```diff
@@ -59,6 +59,15 @@ def read_from_ip_route
               primary_interface
             end
 
+            def read_from_ip6_route
+              return if Facter::Core::Execution.which('ip').nil?
+
+              require 'json'
+
+              output = JSON.parse(Facter::Core::Execution.execute('ip -json -6 route show default', logger: @log))
```
Filipovici-Andrei, line 67: If the ip command is not available, the result will make JSON.parse throw an exception, which will break the resolver.

lib/facter/resolvers/networking.rb
Filipovici-Andrei: Well then at least you should add an implementation for the `primary6_interface` method. But we shouldn't add data that is not used by any fact.

kenyon: Removed this so that it's a Linux-only fact for now.

kenyon:
@puppetlabs _must_ stop treating IPv6 features like a "nice to have". It is a _requirement_.

kenyon:
I don't know why the tests fail now.

kenyon:
I added unit tests. Everything passes locally for me, except for

* random unit test failures unrelated to my changes, where the test returns facts from my system instead of from fixtures;
* `spec_integration/facter_to_hash_spec.rb` tests that return `statvfs() function failed: Permission denied`, I guess because I'm not running these tests as root.

lib/facter/util/resolvers/networking/primary_interface.rb
oanatmaria: JSON.parse fails on empty strings `JSON::ParserError (809: unexpected token at '')`, could you please verify if the output of `execute` is not empty before trying to parse it with JSON? This generates the acceptance tests failures.

kenyon: I rebased and tried to verify that the output of `execute` is not empty, but now I don't know what is going on with the failures from `spec/facter/resolvers/networking_linux_spec.rb`.

spec/facter/resolvers/networking_linux_spec.rb
```diff
@@ -36,6 +36,9 @@ def ifaddr_obj(name, ip, mac, netmask, ipv4_type)
         .with('ip route show', logger: log_spy).and_return(load_fixture('ip_route_show').read)
       allow(Facter::Core::Execution).to receive(:execute)
         .with('ip -6 route show', logger: log_spy).and_return(load_fixture('ip_-6_route_show').read)
+      allow(Facter::Core::Execution).to receive(:execute)
+        .with('ip -json -6 route show default', logger: log_spy)
+        .and_return(load_fixture('ip_-json_-6_route_show_default').read)
```
Filipovici-Andrei, line 41: You also need to stub the Facter::Core::Execution.which('ip') call.
Also you need to set the @log inside the PrimaryInterface class. If you don't, an actual logger is instantiated and the stub for Facter::Core::Execution won't work.  Use something like this: `Facter::Util::Resolvers::Networking::PrimaryInterface.instance_variable_set(:@log, log_spy)`.

spec/facter/resolvers/networking_linux_spec.rb
```diff
@@ -458,6 +465,13 @@ def ifaddr_obj(name, ip, mac, netmask, ipv4_type)
         Facter::Util::Resolvers::Networking::PrimaryInterface.instance_variable_set(:@log, log_spy)
         allow(Facter::Util::FileHelper).to receive(:safe_read).with('/proc/net/route', '').and_return('')
         allow(Facter::Core::Execution).to receive(:which).with('ip').and_return('ip available')
+        allow(Facter::Core::Execution).to receive(:execute)
+          .with('ip -json -6 route show default', logger: log_spy)
+          .and_return(load_fixture('ip_-json_-6_route_show_default').read)
+      end
+
+      it 'returns primary6 interface' do
+        expect(networking_linux.resolve(:primary6_interface)).to eq('ens3')
```
Filipovici-Andrei, line 474: This unit test fails because after the call to `Facter::Util::FileHelper` with `'/proc/net/route', ''` is made, the next one will be `Facter::Core::Execution.execute('ip route show default', logger: @log)` from `PrimaryInterface.read_from_ip_route`. So you need another stub for this call.

lib/facter/resolvers/networking_linux.rb
```diff
@@ -290,6 +291,13 @@ def retrieve_primary_interface
 
           @fact_list[:primary_interface] = primary_interface
         end
+
+        def retrieve_primary6_interface
+          primary_helper = Facter::Util::Resolvers::Networking::PrimaryInterface
+          primary6_interface = primary_helper.read_from_ip6_route
+
+          @fact_list[:primary6_interface] = primary6_interface
```
Filipovici-Andrei, line 299: I would suggest, if that's ok with you, adding some fallbacks in case the `read_from_ip6_route ` doesn't retrieve anything. Like we do inside the `retrieve_primary_interface` method. Maybe at least setting the primary6 as the first not ignored ipv6 address?

Filipovici-Andrei (commented):
Sorry for the late response. 
Thanks for pointing out the random failures. We'll try to solve them in the near future.
The linux networking resolver class grew considerably and the unit tests became a mess. PR https://github.com/puppetlabs/facter/pull/2256 tries to solve this problem. I hope that will help making the unit tests a little bit easier to write.

spec/facter/resolvers/networking_linux_spec.rb
```diff
@@ -100,6 +107,10 @@ def ifaddr_obj(name, ip, mac, netmask, ipv4_type)
       expect(networking_linux.resolve(:primary_interface)).to eq('ens160')
     end
 
+    it 'returns the primary6 interface' do
```
kenyon, line 110: It seems like the `returns the primary(6) interface` tests up here are duplicates of the ones in `context 'when asking for primary interface' do`. :question:

kenyon:
I don't see why my changes would cause the `primary` fact to become empty. https://github.com/puppetlabs/facter/pull/2241/checks?check_run_id=1715495464#step:8:7033 :man_shrugging:

spec/facter/resolvers/networking_linux_spec.rb
Filipovici-Andrei: Yes, you are correct. In `context 'when asking for primary interface'` at line 467 we test scenarios when the different ways of extracting the primary interface fail. 
So, here you should add a test for what should happen when `ip -json -6 route show default` fails to retrieve the primary(6) interface.

Filipovici-Andrei:
I think that the unit tests are failing on windows, because you are missing a mock for `Facter::Core::Execution.which('ip')`.
Now for the acceptance tests on Ubuntu 20.04 I'm not sure, because I can't see the entire log (sometimes, GitHub does that).

joshcooper:
Hi @kenyon thanks for your contribution. Could you rebase on `main`? It looks like the `networking_linux.rb` resolver was moved since you created this PR.

kenyon:
@joshcooper rebased.

PaulSD:
FYI:

For systems with IPv6 only (with no IPv4 addresses), Facter does not currently set networking.primary or networking.ip or networking.ip6 (because there is no IPv4 default route, and Facter selects networking.ip and networking.ip6 based on the IPv4 default route).

The lack of networking.ip and networking.ip6 then causes `puppet apply` to print "Warning: Could not retrieve either serverip or serverip6 fact".

It does not appear that this PR currently fixes that issue (because this PR currently only sets networking.primary6 and does not alter the selection/configuration of networking.ip6).

Either as part of this PR, or as a separate PR after this is merged, the networking.ip6 logic should be updated to use this new ipv6 default route information instead of using the IPv4 default route.

kenyon:
@PaulSD agreed. I hope @puppetlabs can work on this PR and others for better IPv6 support.

joshcooper:
Sorry it's taken so long to review. Could you investigate why specs are failing on Linux and rebase (again)? Thanks!

PR #2244, opened 2021-01-06
(FACT-2705) Facter 4 blog post

IrimieBogdan:


facter4blog.md
```diff
@@ -0,0 +1,111 @@
+# Facter 4: back to the roots
+
+Facter 4 is the default facter staring with Puppet 7. Facter 4 aims for 100% API compatible with Facter 3, but there may be some small inconsistencies. It is written in Ruby, and retains many of the same performance improvements that Facter 3 has over Facter 2. In addition to this Facter 4 also introduces granular blocking and caching for all types of facts, user defined fact groups that can be used for blocking and caching, custom and external fact hierarchies via the dot notation and profiling via `--timing` option
```
MikaelSmith, line 3: I don't think `staring` is the word you want here. Maybe `starring` or `starting`?

Also, do you have benchmarks somewhere comparing Facter 3 and FActer 4 performance?

IrimieBogdan: I only have some older benchmarks: https://docs.google.com/spreadsheets/d/1giARlXsBSGhxIWRlThV8QfmybeAfaBrNRzdr9C0pvPw/edit?usp=sharing 
and newer ones that compare parallel vs sequential vs sequential with a cache fix (with ought the fix the cache was slowing down facter, although it was not being used)

@MikaelSmith  please have a look at https://docs.google.com/document/d/1rVHQIu2NnQu8LKrYajd9pXVx3ciE3HTFD6D4nrrnKoc/edit?usp=sharing for the latest version of the blog post draft, I would appreciate a review.

igalic, line 3: Facter 4 aims for 100% API compatible with Facter 3, but there may be some small inconsistencies. ➡️ Facter 4 aims to be 100% API compatible with Facter 3, but there may be some small inconsistencies.

i highly recommend putting one sentence per line, otherwise reviews are painful

igalic (changes requested):

facter4blog.md
MikaelSmith: Have you tried "Insert a suggestion <cmd-g>"?

```suggestion
Facter 4 is the default facter staring with Puppet 7. Facter 4 aims to be 100% API compatible with Facter 3, but there may be some small inconsistencies. It is written in Ruby, and retains many of the same performance improvements that Facter 3 has over Facter 2. In addition to this Facter 4 also introduces granular blocking and caching for all types of facts, user defined fact groups that can be used for blocking and caching, custom and external fact hierarchies via the dot notation and profiling via `--timing` option
```

IrimieBogdan:
Moved the blog post content to a google.

PR #2245, opened 2021-01-06
(FACT-2882) Fix domain fact on Windows

oanatmaria:
**Description of the problem:** Facter fails to retrieve `domain` fact when Windows does not expose the host's primary DNS suffix via WMI and it's only present in the registry.
**Description of the fix:** Try to retrieve domain from registry if the domain wasn't retrieved with GetAdaptersAddresses call. Also fixed the trailing dot problem with `fqdn` fact when the domain is missing.

### DIFF
--- a/lib/facter/resolvers/windows/networking.rb
+++ b/lib/facter/resolvers/windows/networking.rb
@@ -28,6 +28,8 @@ module Facter
 
             Facter::Util::Resolvers::Networking.expand_main_bindings(@fact_list)
 
+            retrieve_domain_from_registry if !@fact_list[:domain] || @fact_list[:domain].empty?
+
             @fact_list[fact_name]
           end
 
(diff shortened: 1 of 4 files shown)


lib/facter/resolvers/windows/networking.rb
```diff
@@ -130,6 +132,14 @@ def find_primary_interface(sock_addr, name, addr)
               @fact_list[:primary_interface] = name.to_s
             end
           end
+
+          def retrieve_domain_from_registry
+            reg = ::Win32::Registry::HKEY_LOCAL_MACHINE.open('SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters')
+            reg.each do |name, _value|
+              @fact_list[:domain] = reg[name] if name == 'Domain'
+            end
+            reg.close
```
gimmyxd, line 141: You can use a block to avoid explicitly closing the register:
```ruby
::Win32::Registry::HKEY_LOCAL_MACHINE.open('SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters') do |reg|
  ...
end
```

eg: ::Win32::Registry::HKEY_LOCAL_MACHINE.open('SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters')

Should this handle registry read errors, like `Win32::Registry::Error` ?

oanatmaria: Done

IrimieBogdan (approved):

gimmyxd (approved):

PR #2248, opened 2021-01-10
(maint) spec/fixtures/ip_route_show_default: correct output

kenyon:
`ip route show default` would only output one line, the default route. So remove the other line.

### DIFF
--- a/spec/fixtures/ip_route_show_default
+++ b/spec/fixtures/ip_route_show_default
@@ -1,2 +1,1 @@
-10.16.112.0/20 dev ens160 proto kernel scope link src 10.16.124.1
 default via 10.16.112.1 dev ens160


oanatmaria:
@kenyon thank you for your contribution! As what you stated is right for some operating systems, there are some OSes for which this is not applicable (Ex: for Centos 7 the output is not a single line):
```
~ip route show default
default via 10.16.x.y dev ens160  proto static  metric 100
10.16.x.y/20 dev ens160  proto kernel  scope link  src 10.16.x.y  metric 100
10.32.x.y via 10.16.112.1 dev ens160  proto dhcp  metric 100
```

kenyon:
Yes, that's weird, I guess because EL7 is quite old and it was different back then. EL8 gives only the single line. Oh well.

PR #2249, opened 2021-01-11
(PE-30895) Improces regex for detecting legacy processor facts

IrimieBogdan:
Problem: The regex for blocking processor legacy facts was blocking the core "processors" fact.
Fix: The regex takes into account that processor legacy facts contain a number and only excludes processor facts that have a number appended.


gimmyxd:
This should be a Facter ticket, otherwise, release notes won't get it.

lib/facter/config.rb
```diff
@@ -312,8 +312,7 @@ module Config
           'operatingsystemrelease',
           'osfamily',
           'physicalprocessorcount',
-          'processor.*',
-          'processor*',
+          'processor[0-9]+.*',
```
gimmyxd, line 315: Can we have another ticket to have an integration/acceptance tests that blocks legacy facts and matches the resulted fact keys?

gimmyxd:
@IrimieBogdan the check_commits task is broken, it should have failed for this PR, but if you check the output, it runs on another commit: https://github.com/puppetlabs/facter/pull/2249/checks?check_run_id=1679991795

PR #2251, opened 2021-01-11
(maint)Enable puppet facts to disable legacy facts

oanatmaria:
Enable `puppet facts show` application to decide if legacy facts should be displayed.

This PR depends on [this one](https://github.com/puppetlabs/puppet/pull/8475) and fixes: [PUP-10850](https://tickets.puppetlabs.com/browse/PUP-10850)

### DIFF
--- a/spec/facter/facter_spec.rb
+++ b/spec/facter/facter_spec.rb
@@ -365,10 +365,10 @@ describe Facter do
         expect(Facter.values({}, ['os.name'])).to eq(result)
       end
 
-      it 'sets show_legacy to true' do
+      it 'show_legacy is false by default' do
         Facter.values({}, [])
 
-        expect(Facter::Options[:show_legacy]).to be true
+        expect(Facter::Options[:show_legacy]).to be false
       end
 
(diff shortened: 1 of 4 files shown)


spec/facter/facter_spec.rb
```diff
@@ -368,7 +368,7 @@ def mock_collection(method, os_name = nil, error = nil)
       it 'sets show_legacy to true' do
         Facter.values({}, [])
 
-        expect(Facter::Options[:show_legacy]).to be true
+        expect(Facter::Options[:show_legacy]).to be false
```
joshcooper, line 371: nit, the description still says `sets show_legacy to true`.

It might be good to add a test that shows how the behavior changes when `show_legacy` is `true` vs `false`. For example, when true `Facter.values` returns an `uptime` fact. When `false` it doesn't.

oanatmaria: Good catch! I've added integration tests to show the behaviour in the two cases you described.

Filipovici-Andrei (approved):

PR #2254, opened 2021-01-13
(maint) Add debug message

IrimieBogdan:
Add debug message with path to external and custom fact file sources

### DIFF
--- a/lib/facter/custom_facts/util/fact.rb
+++ b/lib/facter/custom_facts/util/fact.rb
@@ -43,6 +43,8 @@ module Facter
         @used_resolution_weight = 0
 
         @value = nil
+
+        @log = Facter::Log.new(self)
       end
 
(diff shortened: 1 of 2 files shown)


gimmyxd (approved):

Filipovici-Andrei (approved):

lib/facter/custom_facts/util/fact.rb
```diff
@@ -201,11 +203,27 @@ def find_first_real_value(resolutions)
...
+          log_fact_path(resolve)
+
+          return value
         end
         nil
       end
 
+      def log_fact_path(resolve)
+        fact_type, resolved_from = if resolve.instance_of?(Facter::Core::Aggregate)
+                                     ['aggregate', resolve.instance_variable_get(:@aggregate).source_location[0]]
```
joshcooper, line 217: This is causing CI failures as `source_location`  can be nil. It would be good to add tests for this change also.

PR #2259, opened 2021-01-18
(FACT-2917) Blocking `legacy` facts slows down facter

IrimieBogdan:
In order to exclude legacy facts we are using a mechanism that compares the name of all facts to all facts from [legacy group](https://github.com/puppetlabs/facter/blob/5f00e96aaff19bfa5af85d195624bb873463233f/lib/facter/config.rb#L244). The comparison is done using [regex](https://github.com/puppetlabs/facter/blob/5f00e96aaff19bfa5af85d195624bb873463233f/lib/facter/framework/core/fact_loaders/fact_loader.rb#L103) and for large numbers of facts it is slow. 

The fix: instead of using the regex comparison, we can use the fact type to block legacy facts.

### DIFF
--- a/lib/facter/framework/config/fact_groups.rb
+++ b/lib/facter/framework/config/fact_groups.rb
@@ -24,6 +24,9 @@ module Facter
       fact_list = []
 
       @block_list.each do |group_name|
+        # legacy is a special group and does not need to be broken into facts
+        next if group_name == 'legacy'
+
         facts_for_block = @groups[group_name]
 
         fact_list += facts_for_block || [group_name]
--- a/lib/facter/framework/core/fact_loaders/fact_loader.rb
+++ b/lib/facter/framework/core/fact_loaders/fact_loader.rb
@@ -83,6 +83,8 @@ module Facter
     def block_facts(facts, options)
       blocked_facts = options[:blocked_facts] || []
 
+      facts.reject! { |fact| fact.type == :legacy } if options[:block_list]&.include?('legacy')
+
       reject_list_core, reject_list_legacy = construct_reject_lists(blocked_facts, facts)
 
       facts = facts.reject do |fact|
--- a/spec/framework/config/fact_groups_spec.rb
+++ b/spec/framework/config/fact_groups_spec.rb
@@ -54,6 +54,20 @@ describe Facter::FactGroups do
       end
     end
 
+    context 'with blocked `legacy` group' do
+      before do
+        stub_const('Facter::Config::FACT_GROUPS', 'legacy' => %w[legacy_fact1 legacy_fact2])
+        allow(config_reader).to receive(:block_list).and_return(%w[legacy])
+      end
+
+      # legacy facts are excluded because they are blocked by another mechanism that takes into account the fact's type
+      it 'excludes legacy facts' do
+        blk_list = fact_groups.new
+
+        expect(blk_list.blocked_facts).to be_empty
+      end
+    end
+
     context 'without block_list' do
       before do
         allow(config_reader).to receive(:block_list).and_return([])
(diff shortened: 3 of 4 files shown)


spec/framework/config/fact_groups_spec.rb
```diff
@@ -54,6 +54,20 @@
...
+        stub_const('Facter::Config::FACT_GROUPS', 'legacy' => %w[legacy_fact1 legacy_fact2])
+        allow(config_reader).to receive(:block_list).and_return(%w[legacy])
+      end
+
+      # legacy facts are excluded because they are blocked by another mechanism that takes into account the fact's type
+      it 'excludes legacy facts' do
+        blk_list = fact_groups.new
+
+        expect(blk_list.blocked_facts).to be_empty
+      end
```
gimmyxd, line 68: is this test working properly? I removed the stubs in the `before` block and it is still passing.

lib/facter/framework/core/fact_loaders/fact_loader.rb
```diff
@@ -83,6 +83,12 @@ def load_external_facts(options)
     def block_facts(facts, options)
       blocked_facts = options[:blocked_facts] || []
 
+      if options[:block_list]&.include?('legacy')
+        facts = facts.reject do |fact|
+          fact.type == :legacy
+        end
```
gimmyxd, line 89:
```ruby
facts.reject! { |fact| fact.type == :legacy } if options[:block_list]&.include?('legacy')
```

spec/framework/config/fact_groups_spec.rb
IrimieBogdan: The stubs mock facter's default groups and the config from `facter.conf`. `blocked_facts` method takes the facts and groups from `facter.conf` and tries to break the groups in facts (using FACT_GROUPS from [config.rb](https://github.com/puppetlabs/facter/blob/5f00e96aaff19bfa5af85d195624bb873463233f/lib/facter/config.rb#L44), which is also mocked) 

The new condition just tells that `blocked_facts` should not break the `legacy` group in facts, it should just ignore it.

The stubs tell what group to transform into facts (in this case `legacy` group) and how to do this (FACT_GROUPS defines what facts belong to a group). When you removed the mocks you basically removed `facter.conf` and the `blocklist: ['legacy']` condition. In this case, the new logic does nothing.

lib/facter/framework/core/fact_loaders/fact_loader.rb
IrimieBogdan: done

lib/facter/framework/config/fact_groups.rb
```diff
@@ -24,6 +24,9 @@ def blocked_facts
       fact_list = []
 
       @block_list.each do |group_name|
+        # legacy is a special group and does not need to be broken in facts
```
Filipovici-Andrei, line 27: `broken into facts`

Filipovici-Andrei (approved):

lib/facter/framework/config/fact_groups.rb
IrimieBogdan: Done

gimmyxd (approved):

gimmyxd:
@IrimieBogdan maybe we should document that if you block the keyword "legacy", all the legacy facts will be blocked. 
If i create a block group `"legacy": ["fact1", "fact2"]`, all the legacy facts will be blocked? 
If i want to block the legacy facts, fact1 and fact2, my group name should have another name?

IrimieBogdan:
> If i want to block only the legacy facts fact1 and fact2, my group name should have another name?

@gimmyxd  I totally agree, Facter behaves exactly as you described. Will add documentation on the JIRA ticket

PR #2261, opened 2021-01-20
(FACT-2924) Facter fails with invalid config file

oanatmaria:
**Description of the problem:** Facter fails when an invalid config file is provided.
**Description of the fix:** Log a warning message stating that the parsing of config file failed and continue retrieving facts with the default options.

### DIFF
--- a/lib/facter/framework/config/config_reader.rb
+++ b/lib/facter/framework/config/config_reader.rb
@@ -33,9 +33,16 @@ module Facter
 
       def refresh_config(config_path)
         @conf = File.readable?(config_path) ? Hocon.load(config_path) : {}
+      rescue StandardError => e
+        log.warn("Facter failed to read config file #{config_path} with the following error: #{e.message}")
+        @conf = {}
       end
 
       private
+
+      def log
+        @log ||= Log.new(self)
+      end
 
       def default_path
         os = OsDetector.instance.identifier
--- a/spec/framework/config/config_reader_spec.rb
+++ b/spec/framework/config/config_reader_spec.rb
@@ -120,6 +120,23 @@ describe Facter::ConfigReader do
       end
     end
 
+    context 'with invalid config file' do
+      let(:config) { 'some corrupt information' }
+      let(:log) { instance_spy(Facter::Log) }
+
+      before do
+        allow(Facter::Log).to receive(:new).and_return(log)
+        allow(Hocon).to receive(:load).and_raise(StandardError)
+        allow(log).to receive(:warn)
+      end
+
+      it 'loggs a warning' do
+        config_reader.init
+
+        expect(log).to have_received(:warn).with(/Facter failed to read config file/)
+      end
+    end
+
     context 'with global section in config file' do
       let(:config) { { 'global' => 'global_config' } }
 


lib/facter/framework/config/config_reader.rb
```diff
@@ -33,6 +35,9 @@ def fact_groups
 
       def refresh_config(config_path)
         @conf = File.readable?(config_path) ? Hocon.load(config_path) : {}
+      rescue StandardError => e
+        @log.warn("Facter failed to read config file #{config_path} with the following error: #{e}")
```
sebastian-miclea, line 39: Is the exception printed out properly? Wouldn't it be easier to read **e.message**?

acceptance/tests/options/config_file/invalid_file.rb
```diff
@@ -0,0 +1,30 @@
+test_name "FACT-2923: providing an invalid config file prints a warning" do
```
gimmyxd, line 1: Do we need an acceptance test for this? We can just force garbage into `Hocon.load` and test how facter behaves there(like it calls the log.warn)

lib/facter/framework/config/config_reader.rb
```diff
@@ -2,6 +2,8 @@
 
 module Facter
   class ConfigReader
+    @log = Log.new(self)
```
gimmyxd, line 5: Is the common practice in Facter to initialize the logger this early? why not add it in the `init` method, to be initialized only when it is used?

oanatmaria: `refresh_config` method is called 4 times, that means a new logger every time that method is called
```
[2021-01-20 17:08:33.766582 ] WARN Facter::ConfigReader - Facter failed to read config file /Users/oana.tanasoiu/projects/bad_file.conf with the following error: /Users/oana.tanasoiu/projects/bad_file.conf: 2: Key 'some invalid info' may not be followed by token: end of file
[2021-01-20 17:08:33.766905 ] WARN Facter::ConfigReader - Facter failed to read config file /Users/oana.tanasoiu/projects/bad_file.conf with the following error: /Users/oana.tanasoiu/projects/bad_file.conf: 2: Key 'some invalid info' may not be followed by token: end of file
[2021-01-20 17:08:33.772585 ] WARN Facter::ConfigReader - Facter failed to read config file /Users/oana.tanasoiu/projects/bad_file.conf with the following error: /Users/oana.tanasoiu/projects/bad_file.conf: 2: Key 'some invalid info' may not be followed by token: end of file
[2021-01-20 17:08:33.772945 ] WARN Facter::ConfigReader - Facter failed to read config file /Users/oana.tanasoiu/projects/bad_file.conf with the following error: /Users/oana.tanasoiu/projects/bad_file.conf: 2: Key 'some invalid info' may not be followed by token: end of file
```

gimmyxd: You  can memoize the logger
```ruby
@log ||= Log.new(self)
```

acceptance/tests/options/config_file/invalid_file.rb
oanatmaria: Done

sebastian-miclea (approved):

spec/framework/config/config_reader_spec.rb
```diff
@@ -120,6 +120,23 @@
...
+      before do
+        allow(Facter::Log).to receive(:new).and_return(log)
+        allow(Hocon).to receive(:load).and_raise(StandardError)
+        allow(log).to receive(:warn)
+      end
+
+      it 'returns loggs a warning' do
+        config_reader.init
+
+        expect(log).to have_received(:warn)
```
gimmyxd, line 136: It would be good to also match the message, to make sure that is the codepath that you are testing.
```ruby
expect(log).to have_received(:warn).with(/Facter failed to read config file/)
```

gimmyxd (approved):

PR #2264, opened 2021-01-27
(FACT-2927) Fix computation of metrics prefix

usev6:
For speed values whose log10 is not in (3, 6, 9, 12), the old code lead
to an error when trying to stringify prefix[power] (which was nil).

The error surfaced when a linux system reported a cpu speed of less
than 1 GHz:

```
puts Math.log10(999_999_999).floor
8
```

Not really relevant for processor speed, but the hash 'prefix' now
also handles the case of no metrics prefix at all (just ```Hz```).

### DIFF
--- a/lib/facter/util/facts/unit_converter.rb
+++ b/lib/facter/util/facts/unit_converter.rb
@@ -17,10 +17,11 @@ module Facter
             speed = speed.to_i
             return if !speed || speed.zero?
 
-            prefix = { 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
+            prefix = { 0 => '', 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
             power = Math.log10(speed).floor
-            validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**power)
-            format('%<displayed_speed>.2f', displayed_speed: validated_speed).to_s + ' ' + prefix[power] + 'Hz'
+            display_power = power - power % 3
+            validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**display_power)
+            format('%<displayed_speed>.2f', displayed_speed: validated_speed).to_s + ' ' + prefix[display_power] + 'Hz'
           end
 
           def bytes_to_human_readable(bytes)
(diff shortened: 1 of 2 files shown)


usev6:
It looks like my change made rubocop unhappy. I aimed at keeping the old logic, but I don't see an obvious way to improve the situation wrt ```Metrics/AbcSize``` with the current logic. A suggestion would be much appreciated.

Apart from that I wasn't able to sign the CLA. I found https://cla.puppet.com/, but trying to 'sign for myself' and then 'authenticate on github' just brought me back to the first page (with an redirect_uri_mismatch error in the URL). So this seems to be a bit more complicated than I hoped.

lib/facter/util/facts/unit_converter.rb
```diff
@@ -17,10 +17,11 @@ def hertz_to_human_readable(speed)
             speed = speed.to_i
             return if !speed || speed.zero?
 
-            prefix = { 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
+            prefix = { 0 => '', 3 => 'k', 6 => 'M', 9 => 'G', 12 => 'T' }
             power = Math.log10(speed).floor
-            validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**power)
-            format('%<displayed_speed>.2f', displayed_speed: validated_speed).to_s + ' ' + prefix[power] + 'Hz'
+            display_power = power - power % 3
+            validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**display_power)
```
gimmyxd, line 23: You could extract this into a private method:
```ruby
def compose_speed_data(speed)
  power = Math.log10(speed).floor
  display_power = power - power % 3
  validated_speed = power.zero? ? speed.to_f : speed.fdiv(10**display_power)
  [display_power, validated_speed]
end
```

and it will be used here as:
```ruby
display_power, validated_speed = compose_speed_data(speed)
format('%<displayed_speed>.2f', displayed_speed: validated_speed).to_s + ' ' + prefix[display_power] + 'Hz'
```

This way, the changes will be minimal and rubocop will be happy

usev6: I'll try something in that direction.

Since I noticed that I also messed up the commit message (wrong issue is referenced), I'll open a second PR. (Alternatively I could force-push to this PR.)

usev6: I amended the wrong commit message and force-pushed. @gimmyxd I didn't take your suggestion literally, but decided to move the computation of the metric prefix to the new private method, too. I hope that makes sense.

gimmyxd (approved):

PR #2265, opened 2021-01-28
(FACT-2921) Fix os.release fact on Linux

oanatmaria:
**Description of the problem:** Os.release fact is retrieved from the `/etc/os-release` file but Facter 3 reads other release files based on OS.
**Description of the fix** Retrieve os.release from the specific release file for every os.

### DIFF
--- a/lib/facter/facts/devuan/os/distro/release.rb
+++ b/lib/facter/facts/devuan/os/distro/release.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Facts
+  module Devuan
+    module Os
+      module Distro
+        class Release
+          FACT_NAME = 'os.distro.release'
+          ALIASES = %w[lsbdistrelease lsbmajdistrelease lsbminordistrelease].freeze
+
+          def call_the_resolver
+            fact_value = Facter::Resolvers::LsbRelease.resolve(:release)
+            return Facter::ResolvedFact.new(FACT_NAME, nil) unless fact_value
+
+            release = construct_release(fact_value)
+
+            [Facter::ResolvedFact.new(FACT_NAME, release),
+             Facter::ResolvedFact.new(ALIASES[0], fact_value, :legacy),
+             Facter::ResolvedFact.new(ALIASES[1], release['major'], :legacy),
+             Facter::ResolvedFact.new(ALIASES[2], release['minor'], :legacy)]
+          end
+
+          def construct_release(version)
+            versions = version.split('.')
+
+            {}.tap do |release|
+              release['full'] = version
+              release['major'] = versions[0]
+              release['minor'] = versions[1] if versions[1]
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/facter/facts/photon/os/release.rb
+++ b/lib/facter/facts/photon/os/release.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+module Facts
+  module Photon
+    module Os
+      class Release
+        FACT_NAME = 'os.release'
+        ALIASES = %w[operatingsystemmajrelease operatingsystemrelease].freeze
+
+        def call_the_resolver
+          version = from_specific_file || from_os_release
+
+          return Facter::ResolvedFact.new(FACT_NAME, nil) unless version
+
+          [Facter::ResolvedFact.new(FACT_NAME, version),
+           Facter::ResolvedFact.new(ALIASES.first, version['major'], :legacy),
+           Facter::ResolvedFact.new(ALIASES.last, version['full'], :legacy)]
+        end
+
+        def from_specific_file
+          version = Facter::Resolvers::SpecificReleaseFile.resolve(:release,
+                                                                   { release_file: '/etc/lsb-release',
+                                                                     regex: /DISTRIB_RELEASE="(\d+)\.(\d+)/ })
+          return if version.nil? || version[1].nil? || version[2].nil?
+
+          major = version[1].to_s
+          minor = version[2].to_s
+          version_data = major + '.' + minor
+
+          {
+            'full' => version_data,
+            'major' => major,
+            'minor' => minor
+          }
+        end
+
+        def from_os_release
+          version = Facter::Resolvers::OsRelease.resolve(:version_id)
+
+          Facter::Util::Facts.release_hash_from_string(version)
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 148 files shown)


lib/facter/resolvers/release_from_first_line.rb
```diff
@@ -0,0 +1,43 @@
...
+
+          output = Facter::Util::FileHelper.safe_read(release_file, nil)
+          return @fact_list[fact_name] = nil if output.nil?
+
+          @fact_list[:release] = retreive_version(output)
+
+          @fact_list[fact_name]
+        end
+
+        def retreive_version(output)
```
GabrielNagy, line 31: nit: retreive -> retrieve

lib/facter/resolvers/release_from_first_line.rb
```diff
@@ -0,0 +1,43 @@
...
+          if output[/(Rawhide)$/]
+            'Rewhide'
```
GabrielNagy, line 33: nit: Rawhide?

GabrielNagy (commented):
just some minor comments

lib/facter/util/facts/facts_utils.rb
```diff
@@ -24,6 +24,30 @@ def discover_family(os)
...
+          versions = output.split('.')
+          fact_value = {}
+          fact_value['full'] = output
+          fact_value['major'] = versions[0]
+          fact_value['minor'] = versions[1] if versions[1]
+          fact_value
+        end
+
+        def release_hash_from_matchdata(data)
+          return unless data[1]
```
Filipovici-Andrei, line 40: you should also add a guard clause for when data is nil

lib/facter/facts/photon/os/release.rb
```diff
@@ -0,0 +1,45 @@
...
+          [Facter::ResolvedFact.new(FACT_NAME, version),
+           Facter::ResolvedFact.new(ALIASES.first, version['major'], :legacy),
+           Facter::ResolvedFact.new(ALIASES.last, version['full'], :legacy)]
+        end
+
+        def from_specific_file
+          version = Facter::Resolvers::ReleaseWithRegex.resolve(:release,
+                                                                { release_file: '/etc/lsb-release',
+                                                                  regex: /DISTRIB_RELEASE="(\d+)\.(\d+)/ })
+          return if version[1].nil? || version[2].nil?
```
Filipovici-Andrei, line 24: I think you also need a guard clause for when version is nil.

Filipovici-Andrei (changes requested):

lib/facter/resolvers/aix/networking.rb
```diff
@@ -9,7 +9,7 @@ class Networking < BaseResolver
         class << self
           private
 
-          def post_resolve(fact_name)
+          def post_resolve(fact_name, _options)
```
Filipovici-Andrei, line 12: why is this necessary?

lib/facter/framework/detector/os_detector.rb
```diff
@@ -64,7 +64,14 @@ def detect_family
     Facter::Resolvers::OsRelease.resolve(:id_like)
   end
 
+  def detect_based_on_release_file
+    @identifier = :devuan if File.readable?('/etc/devuan_version')
+  end
+
   def detect_distro
+    detect_based_on_release_file
```
Filipovici-Andrei, line 72: I think this os is not as popular as the other ones, so I think that we should make this search probably before or after the SuseRelease. I noticed that we also extract the version, which is not used anywhere in the code (I could be wrong though).

Filipovici-Andrei (approved):

lib/facter/resolvers/release_with_regex.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class ReleaseWithRegex < BaseResolver
```
gimmyxd, line 5: This is the same as `SpecificReleseFile` + a regex. Since we exposed the `options`, why not have only a resolver which does the following:
 - reads version from a options[:release_file] or a default location of options[:release_file] is not sent
 - apply a regex if options[:regex] is sent

Maybe we can also compact `ReleaseFromFirstLine` into that.

lib/facter/facts/devuan/os/distro/release.rb
```diff
@@ -0,0 +1,35 @@
...
+             Facter::ResolvedFact.new(ALIASES[2], release['minor'], :legacy)]
+          end
+
+          def construct_release(version)
+            versions = version.split('.')
+            release = {}
+            release['full'] = version
+            release['major'] = versions[0]
+            release['minor'] = versions[1] if versions[1]
+            release
```
gimmyxd, line 29: You can use tap to avoid creating an empty has
```ruby
{}.tap do |release|
  release['full'] = version
  release['major'] = versions[0]
  release['minor'] = versions[1] if versions[1]
end
```

spec/fixtures/os_release_mint
```diff
@@ -0,0 +1,2 @@
+DISTR_NAME=“Linuxmint”
```
GabrielNagy, line 1: not sure if it matters but these are weird quotes `“` and `”`, should probably be `"`

lib/facter/facts/photon/os/release.rb
```diff
@@ -0,0 +1,45 @@
...
+
+          major = version[1].to_s
+          minor = version[2].to_s
+          version_data = major + '.' + minor
+
+          fact_value = {}
+          fact_value['full'] = version_data
+          fact_value['major'] = major
+          fact_value['minor'] = minor
+          fact_value
```
gimmyxd, line 34: Return the hash directly
```ruby
{
  'full' => version_data,
  'major' => major,
  'minor' => minor
}
```

lib/facter/util/facts/facts_utils.rb
```diff
@@ -24,6 +24,23 @@ def discover_family(os)
...
+
+        def release_hash_from_string(output)
+          return unless output
+
+          versions = output.split('.')
+          fact_value = {}
+          fact_value['full'] = output
+          fact_value['major'] = versions[0]
+          fact_value['minor'] = versions[1] if versions[1]
+          fact_value
```
gimmyxd, line 36: You can use `tap` to make it shorter

GabrielNagy (approved):
looks good :+1:, just a minor fixture comment

lib/facter/resolvers/release_with_regex.rb
oanatmaria: `SpecificReleseFile` and `ReleaseWithRegex` are unified in `SpecificReleseFile`. The change is under the last commit for more visibility.

gimmyxd (approved):

gimmyxd:
Looks good, @oanatmaria can you squash the commits before merging this?

PR #2270, opened 2021-02-02
(FACT-2932) fix acceptance tests leakage 

Filipovici-Andrei:
On windows platforms, the directory containing a custom fact is not properly deleted during teardown.


Filipovici-Andrei:

gimmyxd (approved):

PR #2271, opened 2021-02-03
(FACT-2934)Set last exit code after Execution.exec

oanatmaria:
**Description of the problem:** The exec and execute methods of the Facter::Core::Execution module are commonly used to execute external processes in custom facts. In Facter 4, these methods no longer set the global Ruby variables $CHILD_STATUS or $? because Open3 is used and it consumes the exit status of the child process.
**Description of the fix:** Reimplement Open3.popen3 to not use Process.detach but Process.wait instead.


lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -70,8 +72,7 @@ def execute_command(command, on_fail = nil, logger = nil, time_limit = nil)
             opts = { 'LC_ALL' => 'C', 'LANG' => 'C' }
             require 'timeout'
             @log.debug("Executing command: #{command}")
-            out, stderr = Open3.popen3(opts, command.to_s) do |_, stdout, stderr, wait_thr|
-              pid = wait_thr.pid
+            out, stderr = Popen3.popen3e(opts, command.to_s) do |_, stdout, stderr, pid|
```
ciprianbadescu, line 75: It would better be something like `Open3.popen3_e` to suggest that is patch of Open3 module with `_e` as it provides the child exit status

PR #2279, opened 2021-02-05
(FACT-2936) Facter::Core::Execution tests fix

Dorin-Pleava:
On Solaris platforms behaves different that on other linux platforms:

root@keen-taste:~# false
root@keen-taste:~# echo $?
1
root@keen-taste:~# which false
/usr/bin/false
root@keen-taste:~# /usr/bin/false
root@keen-taste:~# echo $?
255

One way to force an exit code is to run `/usr/bin/bash -c 'exit 1'`


Dorin-Pleava:
Jenkins, please test this on all

Dorin-Pleava:

gimmyxd (approved):

PR #2280, opened 2021-02-08
(FACT-2931) resolve os.distro without lsb

gimmyxd:
`os.distro` facts will now work even if lsb release is not installed.
For the legacy lsb facts, lsb_release needs to be installed in order for them to be displayed
This is changing the following platforms
- [x] rhel(el, centos, fedora)
- [x] amazon
- [x] sles

The functionality is extracted from the `lsb_release` packages

### DIFF
--- a/lib/facter/facts/amzn/os/distro/codename.rb
+++ b/lib/facter/facts/amzn/os/distro/codename.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+module Facts
+  module Amzn
+    module Os
+      module Distro
+        class Codename
+          FACT_NAME = 'os.distro.codename'
+
+          def call_the_resolver
+            fact_value = Facter::Resolvers::SpecificReleaseFile.resolve(
+              :release, release_file: '/etc/system-release'
+            ).match(/.*release.*(\(.*\)).*/)
+
+            fact_value = fact_value[1].gsub(/\(|\)/, '') if fact_value
+            fact_value ||= 'n/a'
+
+            Facter::ResolvedFact.new(FACT_NAME, fact_value)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/facter/facts/sles/os/distro/codename.rb
+++ b/lib/facter/facts/sles/os/distro/codename.rb
@@ -6,13 +6,12 @@ module Facts
       module Distro
         class Codename
           FACT_NAME = 'os.distro.codename'
-          ALIASES = 'lsbdistcodename'
 
           def call_the_resolver
-            fact_value = Facter::Resolvers::LsbRelease.resolve(:codename)
+            fact_value = Facter::Resolvers::OsRelease.resolve(:version_codename)
+            fact_value = 'n/a' if !fact_value || fact_value.empty?
 
-            [Facter::ResolvedFact.new(FACT_NAME, fact_value),
-             Facter::ResolvedFact.new(ALIASES, fact_value, :legacy)]
+            Facter::ResolvedFact.new(FACT_NAME, fact_value)
           end
         end
       end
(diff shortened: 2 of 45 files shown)


gimmyxd:
Jenkins please test this on centos6-64,redhat7-64,sles12-64,sles15-64,fedora32-64

gimmyxd:

gimmyxd:

lib/facter/facts/amzn/os/distro/codename.rb
```diff
@@ -0,0 +1,24 @@
...
+  module Amzn
+    module Os
+      module Distro
+        class Codename
+          FACT_NAME = 'os.distro.codename'
+
+          def call_the_resolver
+            fact_value = Facter::Resolvers::SpecificReleaseFile.resolve(
+              :release, release_file: '/etc/system-release'
+            ).match(/.*release.*(\(.*\)).*/)
```
gimmyxd, line 13: Use `Facter::Resolvers::OsRelease.resolve(:pretty_name)`

lib/facter/facts/sles/os/distro/codename.rb
```diff
@@ -6,13 +6,16 @@ module Os
       module Distro
         class Codename
           FACT_NAME = 'os.distro.codename'
-          ALIASES = 'lsbdistcodename'
 
           def call_the_resolver
-            fact_value = Facter::Resolvers::LsbRelease.resolve(:codename)
+            fact_value = Facter::Resolvers::SpecificReleaseFile.resolve(
+              :release, release_file: '/etc/os-release'
+            ).match(/.*release.*(\(.*\)).*/)
```
gimmyxd, line 13: Use `Facter::Resolvers::OsRelease.resolve(:pretty_name)`

lib/facter/facts/sles/os/distro/id.rb
```diff
@@ -6,13 +6,24 @@ module Os
...
+            description = Facter::Resolvers::OsRelease.resolve(:pretty_name)
 
-            [Facter::ResolvedFact.new(FACT_NAME, fact_value),
-             Facter::ResolvedFact.new(ALIASES, fact_value, :legacy)]
+            fact_value = if description.include?('release')
+                           description.split('release')[0].split.reject { |el| el.casecmp('linux').zero? }.join
+                         else
+                           default_fact_value
+                         end
+            Facter::ResolvedFact.new(FACT_NAME, fact_value)
```
gimmyxd, line 18: this is hardcoded in lsb_release, return value per version

lib/facter/facts/sles/os/distro/codename.rb
oanatmaria: fact_value can be nil if VERSION_CODENAME is not resent in `/etc/os-release`

gimmyxd:

oanatmaria (approved):

ciprianbadescu:
comparing cu facter 3 when lsb packages are installed the `specification => ":core-4.1-aarch64:core-4.1-noarch"` fact is missing

gimmyxd:
> comparing cu facter 3 when lsb packages are installed the `specification => ":core-4.1-aarch64:core-4.1-noarch"` fact is missing

Hmm, that should be inherited from linux: https://github.com/puppetlabs/facter/blob/main/lib/facter/facts/linux/os/distro/specification.rb, i'll check if i have to add that fact under each distribution

Dorin-Pleava (approved):
Tested on centos and amazon, and it works as expected.
Specification fact is missing when redhat-lsb-core is not installed, but this is normal as specification fact is basically the lsb version from redhat-lsb-core.

ciprianbadescu:
It looks like `specification` fact is practically `LSB Version`. It does not make sense that is available without having lsb release package installed. 
Documenting that the `specification` fact is missing when lsb packages are not installed should be enough

PR #2290, opened 2021-02-19
(FACT-2926) Ensure all methods have rubydocs

oanatmaria:
Added (and checked) rubydocs to all public methods for the following classes:
- module Facter
- Facter::Core::Execution
- Facter::Util::Fact
- Facter::Util::Resolution
- LegacyFacter::Core::Resolvable
- LegacyFacter::Core::Suitable
- Facter::Core:Aggregate

### DIFF
--- a/lib/facter.rb
+++ b/lib/facter.rb
@@ -44,6 +44,12 @@ module Facter
       end
     end
 
+    # Method used by cli to set puppet paths
+    # in order to retrieve puppet custom and external facts
+    #
+    # @return nil
+    #
+    # @api private
     def puppet_facts
       require 'puppet'
 
@@ -110,10 +116,10 @@ module Facter
       nil
     end
 
-    # Gets the value for a core fact, external or custom facts are
-    #   not returned with this call. Returns `nil` if no such fact exists.
+    # Retrieves the value of a core fact. External or custom facts are
+    # not returned with this call. Returns `nil` if no such fact exists.
     #
-    # @return [FactCollection] hash with fact names and values
+    # @return [FactCollection] A hash with fact names and values
     #
     # @api private
     def core_value(user_query)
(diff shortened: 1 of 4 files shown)


lib/facter.rb
```diff
@@ -346,7 +368,7 @@ def search_path
     # Gets a hash mapping fact names to their values
     # The hash contains core facts, legacy facts, custom facts and external facts (all facts that can be resolved).
     #
-    # @return [FactCollection] hash with fact names and values
+    # @return a hash with fact names and values
```
gimmyxd, line 371: Maybe include the type?
```ruby
# @return [Hash] A hash with fact names and values
```

Filipovici-Andrei, line 371: returns

lib/facter/custom_facts/core/resolvable.rb
```diff
@@ -59,6 +59,17 @@ def flush
         @on_flush_block&.call
       end
 
+      # Resolves the value or loggs an error if the value
```
Filipovici-Andrei, line 62: Does `Resolves the custom fact value` sound a little better than `Resolves the value`?

IrimieBogdan: I would use `Resolve the facts value` because this might be a custom or an external fact

lib/facter.rb
oanatmaria: the "@return" form is everywhere, I didn't wanted to break the pattern.

IrimieBogdan (approved):

IrimieBogdan:
in https://github.com/puppetlabs/facter/blob/392b8022c8869010d5d854f2d50f98691e9fcf3c/lib/facter.rb#L119 we could update the description to 
```
    # Retrieves the value of a core fact, external or custom facts are
    #  not returned with this call. Returns `nil` if no such fact exists.
    #
    # @return [FactCollection] hash with fact names and values
    #
    # @api private
```

IrimieBogdan:
https://github.com/puppetlabs/facter/blob/392b8022c8869010d5d854f2d50f98691e9fcf3c/lib/facter.rb#L367 we could update to 
```
    # Retirves a fact's value. Returns `nil` if no such fact exists.
    #
    # @param user_query [String] the fact name
    # @return [String] the value of the fact, or nil if no fact is found
    #
    # @api public
```

IrimieBogdan:
https://github.com/puppetlabs/facter/blob/392b8022c8869010d5d854f2d50f98691e9fcf3c/lib/facter.rb#L478
we could update to 

```
    # Returns a list with the names of all resolved facts
    # @return [Array] the list with all the fact names
    #
    # @api public
```

Filipovici-Andrei (approved):

PR #2291, opened 2021-02-19
(FACT-2134) Distinguish between CentOS and CentOS Stream

nbarrientos:
```
[root@c8stream ~]# cat /etc/os-release
NAME="CentOS Stream"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Stream 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

Before:

```
[root@c8stream ~]# /opt/puppetlabs/puppet/bin/facter-ng os.name os.distro.id os.release.major os.release.minor
os.distro.id => CentOSStream
os.name => CentOS
os.release.major => 8
```

After:

```
[root@c8stream ~]# RUBYLIB=/root/facter/lib ./facter/bin/facter os.name os.distro.id os.release.major os.release.minor
os.distro.id => CentOSStream
os.name => CentOSStream
os.release.major => 8
```

https://en.wikipedia.org/wiki/CentOS#CentOS_Stream


nbarrientos:
Will add a test.

oanatmaria (approved):

nbarrientos:
CERN has signed the CLA as organisation so individual signatures are not necessary.

lucywyman:
Hey y'all - I think it's fine to merge this against the CLA check until we get the bot sorted. Apologies for the inconvenience!

mihaibuzgau:
thanks @nbarrientos for your contribution!

duritong:
Hi all, please consider this quite a major breaking change!

Switching from CentOS Linux 8 to CentOS Stream 8 is 2 commands with the same content (just 2 different point in time).

BUT for puppet this now means that any detection have to be changed!

And there are many modules around that do have `if $facts['os']['name'] == 'CentOS'` and they will all break!

This change should be considered only for a new major factor release.

duritong:
Just look at all those modules breaking: https://github.com/search?q=if+%24facts%5B%27os%27%5D%5B%27name%27%5D+%3D%3D+%27CentOS%27&type=code

And there are likely many folks that already switched to Stream.

Don't make this harder than it already is.

roidelapluie:
The ID= is still centos. It does not make sense to distinguish centos and centos stream as both will not exist in parallel in the future.

duritong:
On my Stream 8 & puppet 7.4.1 / facter 4.0.51 I do not have `os.distro` - but from @nbarrientos output it looks like one was already able to differentiate over that.

nbarrientos:
Hi,

Unfortunately some (like us *grin*) have to maintain both CentOS 8 and CentOS Stream 8 until the former is dead so it's necessary to have a way to distinguish reliably between them early in the bootstrapping process. Think of a Puppet module that configures the system DNF repositories. It needs a source of information to know if the OS being configured is Stream or not to correctly setup the URLs of the repositories.

I understand though that changing `os.name` implies patching modules to explicitly declare support for this "new" distribution. I can think of some alternatives that might suit our use case and be less disruptive:

* Rely only on `os.distro.id` to make such distinction. This is what we're currently doing in Puppet 6 which Facter 3.x, however this requires some RPMs to be installed very early in the boostrapping process. In our case this would be doable as we control the images and the magic to bootstrap them so we can add whatever is necessary. https://github.com/puppetlabs/facter/pull/2280 is interesting indeed. I haven't read the whole patch but how does it [handle ](https://github.com/puppetlabs/facter/pull/2280/files#diff-fb5eee5f3456dbe9fb6e2e66549ee0dd9edb58973c98d7e4cec1cda10fc49a8fR132)the CentOS Stream case?

* If `os.distro.id` must be `centos` (which is not currently the case), then a new top-level key in the `os` fact called `flavor` or similar could be created so we can consistently rely on it to detect Stream.

* We internally develop a custom made fact nobody knows about :)

gimmyxd:
Hi @nbarrientos @duritong
I've looked over the changes on #2280 and with that `os.distro.id` can be resolved without the need of additional packages installed.

```
# on CentOS Linux
❯ facter os.distro.id
CentOS


# on CentOS Stream 
❯ facter os.distro.id
CentOSStream
```

@nbarrientos would that fact be enough to differentiate CentOS Linux and CentOS Stream? Then would be ok to revert this changes?

<details>
<summary>Full facter.os output</summary>

```
# on CentOS Linux
❯ facter os
{
  architecture => "x86_64",
  distro => {
    codename => "Core",
    description => "CentOS Linux release 8.0.1905 (Core)",
    id => "CentOS",
    release => {
      full => "8.0.1905",
      major => "8",
      minor => "0"
    }
  },
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "8.0.1905",
    major => "8",
    minor => "0"
  },
  selinux => {
    enabled => false
  }
}


# on CentOS Stream 
❯ facter os
{
  architecture => "x86_64",
  distro => {
    description => "CentOS Stream release 8",
    id => "CentOSStream",
    release => {
      full => "8",
      major => "8"
    }
  },
  family => "RedHat",
  hardware => "x86_64",
  name => "CentOS",
  release => {
    full => "8",
    major => "8"
  },
  selinux => {
    enabled => false
  }
}
```

</details>

nbarrientos:
That would work for us (we're already using that approach with P6 and Facter 3.x) but I believe that it's not aligned to @roidelapluie's comments.

PR #2292, opened 2021-02-20
(FACT-2943) Add AlmaLinux support

cschug:
This change adds support for [AlmaLinux](https://almalinux.org/),
which is 1:1 clone of Red Hat Enterprise Linux.


GabrielNagy (approved):
Tested this on an AlmaLinux docker container and it works as expected:

```sh-session
> facter os.name
AlmaLinux
```

Thanks for the contribution! :+1:

PR #2294, opened 2021-02-25
(FACT-1383) Add Azure instance metadata fact

luchihoratiu:
This commit adds the `az_metadata` fact that contains all available metadata information of the Microsoft Azure instance on which facter is being run. Support added for Linux and Windows.

### DIFF
--- a/acceptance/tests/options/no_cache_should_not_cache_facts.rb
+++ b/acceptance/tests/options/no_cache_should_not_cache_facts.rb
@@ -36,7 +36,7 @@ test_name "C99968: --no-cache command-line option causes facter to not cache fac
 
     step "facter should not cache facts when --no-cache is specified" do
       on(agent, facter("--no-cache")) do |facter_output|
-        assert_no_match(/caching/, facter_output.stderr, "facter should not have tried to cache any facts")
+        assert_no_match(/caching values for/, facter_output.stderr, "facter should not have tried to cache any facts")
       end
     end
   end
--- a/acceptance/tests/options/no_cache_should_not_refresh_cached_facts.rb
+++ b/acceptance/tests/options/no_cache_should_not_refresh_cached_facts.rb
@@ -52,7 +52,7 @@ FILE
       agent.modified_at(cached_fact_file, '198001010000')
 
       on(agent, facter("--no-cache")) do |facter_output|
-        assert_no_match(/caching/, facter_output.stderr, "facter should not have tried to refresh the cache")
+        assert_no_match(/caching values for/, facter_output.stderr, "facter should not have tried to refresh the cache")
       end
 
       cat_output = agent.cat(cached_fact_file)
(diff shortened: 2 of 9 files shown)


acceptance/tests/options/no_cache_should_not_cache_facts.rb
```diff
@@ -11,7 +11,12 @@
   cached_fact_name = "uptime"
   config = <<-FILE
   cli : { debug : true }
-  facts : { ttls : [ { "#{cached_fact_name}" : 30 minutes } ] }
+  facts : {
+    blocklist : [ "az_metadata" ],
```
GabrielNagy, line 15: I would prefer to update the test assertion rather than changing the preconditions here

acceptance/tests/options/no_cache_should_not_refresh_cached_facts.rb
```diff
@@ -18,7 +18,12 @@
 
   config = <<-FILE
 cli : { debug : true }
-facts : { ttls : [ { "#{cached_fact_name}" : 30 minutes } ] }
+facts : {
+  blocklist : [ "az_metadata" ],
```
GabrielNagy, line 22: I would prefer to update the test assertion rather than changing the preconditions here

GabrielNagy (commented):
code looks good to me, had some minor comments related to the tests

GabrielNagy (approved):
Looks good to me :+1:! Tested both this and the Ruby version and these are my findings:
1. the fact correctly resolves if on hyperv and the endpoint is available ✔️ 
2. the fact silently fails to resolve if on hyperv and the endpoint is not available ✔️ 
3. the fact does not resolve at all if not on hyperv ✔️ 

The only difference I was able to spot was how empty arrays are rendered, but this is something that appears to be different in the Facter 4 code that pretty-prints the json, which is unrelated to this work:

```
$ diff -u <(bundle exec facter az_metadata) <(../3.x_facter/bin/facter az_metadata)
--- /dev/fd/63	2021-03-02 17:22:36.085844995 +0000
+++ /dev/fd/62	2021-03-02 17:22:36.085844995 +0000
@@ -21,9 +21,7 @@
     platformFaultDomain => "0",
     platformUpdateDomain => "0",
     provider => "Microsoft.Compute",
-    publicKeys => [
-
-    ],
+    publicKeys => [],
     publisher => "canonical",
     resourceGroupName => "test-facter-azure_group",
     resourceId => "/subscriptions/de8aa4b1-87d7-41be-ada7-5e2e4b4768e5/resourceGroups/test-facter-azure_group/providers/Microsoft.Compute/virtualMachines/test-facter-azure",
@@ -87,9 +85,7 @@
     },
     subscriptionId => "de8aa4b1-87d7-41be-ada7-5e2e4b4768e5",
     tags => "",
-    tagsList => [
-
-    ],
+    tagsList => [],
     version => "20.04.202102020",
     vmId => "9e5e7e81-8db0-4ff1-adc3-2dcfcd7f2727",
     vmScaleSetName => "",
@@ -114,9 +110,7 @@
           ]
         },
         ipv6 => {
-          ipAddress => [
-
-          ]
+          ipAddress => []
         },
         macAddress => "000D3A38BFF9"
       }
```

PR #2297, opened 2021-02-27
feature: Add serial number and wwid to disks fact

KervyN:
This PR adds serial numbers and wwid to the disks facts.

@IrimieBogdan said I should open this: https://puppetcommunity.slack.com/archives/C0W298S9G/p1614364073352100?thread_ts=1614252410.207700&cid=C0W298S9G


valia0906:
It`s really helpfull and good feature.
Dear facter team, please look at this issue.

valia0906:
Can one of the admins look at this issue?

lib/facter/resolvers/disk.rb
```diff
@@ -9,7 +9,13 @@ class Disk < BaseResolver
         init_resolver
 
         DIR = '/sys/block'
-        FILE_PATHS = { model: 'device/model', size: 'size', vendor: 'device/vendor', type: 'queue/rotational' }.freeze
+        FILE_PATHS = { model: 'device/model',
+                       size: 'size',
+                       vendor: 'device/vendor',
+                       type: 'queue/rotational',
+                       sn: 'false',
+                       wwid: 'false'
```
GabrielNagy, line 17: Is there any reason you're not just reading the values from their respective files?

WWID can also be read from `/sys/block/<disk>/wwid`, so you'd just have to put `wwid` here instead of `false`.
SN on my system can be read from `/sys/block/<disk>/device/serial` so you'd have to put `device/serial` instead of `false`.

KervyN: I‘ve tested it with rotational disks and SSDs on centos7 and there were no files that give me the values reliable. 

So I‘ve put false in there to circle around the problem.

valia0906:
What is next?

GabrielNagy:
Hey, sorry for letting the PR languish for so long.

The PR looks good, the following should be done in order to get it merged:
- create a ticket for the feature at [tickets.puppetlabs.com/projects/FACT](https://tickets.puppetlabs.com/projects/FACT), and edit the commit message to include the ticket ID (you can check other PRs for examples); this is so we can internally track the ticket, and also to make it possible to give you credit on our release notes page
- have some tests for this behavior; tests for this resolver live in https://github.com/puppetlabs/facter/blob/main/spec/facter/resolvers/disk_spec.rb - at a first glance you need to add the new `false` keys to https://github.com/puppetlabs/facter/blob/main/spec/facter/resolvers/disk_spec.rb#L8 and then stub the Facter::Core::Execution calls as needed in the before blocks (example: https://github.com/puppetlabs/facter/blob/main/spec/facter/resolvers/disk_spec.rb#L34)
- ideally we should have tests for both when the command works and returns something, and also for when it does not (and returns an empty string)

Let me know if you need any help with anything!

GabrielNagy:
@KervyN I opened https://github.com/puppetlabs/facter/pull/2461 which includes your work. In the PR I changed the fact names from `sn` to `serial`, and `wwid` to `wwn` (for simplicity reasons, to match the lsblk option name). It would be great if you can give that PR a test.

KervyN:
> @KervyN I opened #2461 which includes your work. In the PR I changed the fact names from `sn` to `serial`, and `wwid` to `wwn` (for simplicity reasons, to match the lsblk option name). It would be great if you can give that PR a test.

@GabrielNagy absolutly. This looks very awesome. I really appreciate the work you put into this. Thank you.
Renaming things to match the lsblk output is a really great idea.

Is there an easy way to test it on production hardware? The last times I just rewrote the disk.rb file and ran `puppet facts`.
What I tried now on ubuntu focal: install `ruby-dev`, `bundler`, `make` and run `bundle install` as a local user. This builds but I can not run `./bin/facter` as I get an exceptions: 
```
~/facter$ ./bin/facter
	2: from ./bin/facter:5:in `<main>'
	1: from /usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
/usr/lib/ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- facter/framework/cli/cli_launcher.rb (LoadError)
```

On a centos7 system I can't get this build because ruby is just too old. 

Next thing I would try is to clone it into `/opt/puppetlabs` and use the packaged ruby, but I am not sure if this would change anything, so I ask you.

Please excuse my lack of knowledge when it comes to building things.

GabrielNagy:
@KervyN a simple way would be to use the packaged ruby to test it (on a machine with puppet-agent installed):

```
/opt/puppetlabs/puppet/bin/gem install bundler
git clone https://github.com/GabrielNagy/facter && cd facter
git checkout add_sn_wwid_to_disks_fact
/opt/puppetlabs/puppet/bin/bundle install
/opt/puppetlabs/puppet/bin/bundle exec facter disks
```

Sample output for me:
```sh-session
> /opt/puppetlabs/puppet/bin/bundle exec facter disks
{
  sda => {
    model => "Virtual disk",
    size => "70.00 GiB",
    size_bytes => 75161927680,
    type => "hdd",
    vendor => "VMware"
  },
  sr0 => {
    model => "VMware IDE CDR00",
    serial => "00000000000000000001",
    size => "1.00 GiB",
    size_bytes => 1073741312,
    type => "hdd",
    vendor => "NECVMWar"
  },
  sr1 => {
    model => "VMware IDE CDR01",
    serial => "01000000000000000001",
    size => "1.00 GiB",
    size_bytes => 1073741312,
    type => "hdd",
    vendor => "NECVMWar"
  }
}
```

KervyN:
@GabrielNagy it works like I would expect it.
I just tested on centos7 and ubuntu 20.04 and tested some systems to know if it reports correctly on different kind of disk vendors, which it does.

GabrielNagy:
Sounds great, thanks for testing! I'll close this PR then since it's replaced by https://github.com/puppetlabs/facter/pull/2461

PR #2298, opened 2021-03-02
(FACT-1383) Add Azure instance metadata fact

luchihoratiu:
This commit adds the `az_metadata` fact that contains all available metadata information of the Microsoft Azure instance on which facter is being run. Support added for Linux and Windows.

### DIFF
--- a/lib/src/facts/resolvers/az_resolver.cc
+++ b/lib/src/facts/resolvers/az_resolver.cc
@@ -1,0 +1,261 @@
+#include <internal/facts/resolvers/az_resolver.hpp>
+#include <facter/facts/collection.hpp>
+#include <facter/facts/fact.hpp>
+#include <facter/facts/scalar_value.hpp>
+#include <facter/facts/array_value.hpp>
+#include <facter/facts/map_value.hpp>
+#include <facter/facts/vm.hpp>
+#include <leatherman/logging/logging.hpp>
+#include <leatherman/locale/locale.hpp>
+#include <leatherman/util/environment.hpp>
+#include <boost/algorithm/string.hpp>
+#include <rapidjson/reader.h>
+#include <rapidjson/error/en.h>
+#include <stack>
+#include <tuple>
+#include <stdexcept>
+
+#ifdef USE_CURL
+#include <leatherman/curl/client.hpp>
+#include <leatherman/curl/request.hpp>
+#include <leatherman/curl/response.hpp>
+namespace lth_curl = leatherman::curl;
+#endif
+
+using namespace std;
+using namespace rapidjson;
+using namespace leatherman::util;
+
+namespace facter { namespace facts { namespace resolvers {
+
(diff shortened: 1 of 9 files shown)


lib/src/facts/resolvers/az_resolver.cc
```diff
@@ -0,0 +1,261 @@
...
+                LOG_ERROR("failed to parse Azure metadata: {1}.", GetParseError_En(result.Code()));
+                return;
+            }
+
+            if (!data->empty()) {
+                facts.add(fact::az_metadata, move(data));
+            }
+        } catch (lth_curl::http_request_exception& ex) {
+            LOG_DEBUG("Azure instance metadata fact is unavailable: not running under an Azure instance or Azure is not responding in a timely manner.");
+            LOG_TRACE("EC2 metadata request failed: {1}", ex.what());
```
GabrielNagy, line 249: nit: should be AZ/Azure

GabrielNagy (approved):
Looks good to me :+1:, just a minor comment about a leftover `EC2`. Tested both this and the Ruby version and these are my findings:
1. the fact correctly resolves if on hyperv and the endpoint is available ✔️ 
2. the fact silently fails to resolve if on hyperv and the endpoint is not available ✔️ 
3. the fact does not resolve at all if not on hyperv ✔️ 

The only difference I was able to spot was how empty arrays are rendered, but this is something that appears to be different in the Facter 4 code that pretty-prints the json, which is unrelated to this work:

```
$ diff -u <(bundle exec facter az_metadata) <(../3.x_facter/bin/facter az_metadata)
--- /dev/fd/63	2021-03-02 17:22:36.085844995 +0000
+++ /dev/fd/62	2021-03-02 17:22:36.085844995 +0000
@@ -21,9 +21,7 @@
     platformFaultDomain => "0",
     platformUpdateDomain => "0",
     provider => "Microsoft.Compute",
-    publicKeys => [
-
-    ],
+    publicKeys => [],
     publisher => "canonical",
     resourceGroupName => "test-facter-azure_group",
     resourceId => "/subscriptions/de8aa4b1-87d7-41be-ada7-5e2e4b4768e5/resourceGroups/test-facter-azure_group/providers/Microsoft.Compute/virtualMachines/test-facter-azure",
@@ -87,9 +85,7 @@
     },
     subscriptionId => "de8aa4b1-87d7-41be-ada7-5e2e4b4768e5",
     tags => "",
-    tagsList => [
-
-    ],
+    tagsList => [],
     version => "20.04.202102020",
     vmId => "9e5e7e81-8db0-4ff1-adc3-2dcfcd7f2727",
     vmScaleSetName => "",
@@ -114,9 +110,7 @@
           ]
         },
         ipv6 => {
-          ipAddress => [
-
-          ]
+          ipAddress => []
         },
         macAddress => "000D3A38BFF9"
       }
```

PR #2301, opened 2021-03-04
(FACT-2950) return root fact if defined

gimmyxd:
Because the toplevel fact does not exists, root of
structured core facts cannot be overriden by a
custom fact. This PR updates the QueryParser logic
to return the root fact, if that is found in the
loaded facts list.

eg: the list contains both `os.name` and `os` facts,
the result will be the `os` fact.

This new functionality allows completely redefinition
of core facts. Eg:

```
❯ cat custom_facts/my_custom_fact.rb

Facter.add(:os) do
  has_weight(999)
  setcode do
    {"architecture"=>"x64"}
  end
end

❯ bx facter os
{
  architecture => "x64"
}

❯ bx facter os.name

❯ bx facter os.architecture
x64
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -68,6 +68,7 @@ Metrics/AbcSize:
     - 'lib/facter/resolvers/aix/ffi/ffi_helper.rb'
     - 'lib/facter/custom_facts/core/execution/popen3.rb'
     - 'lib/facter.rb'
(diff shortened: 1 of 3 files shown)


gimmyxd:

GabrielNagy (approved):

PR #2302, opened 2021-03-04
(FACT-1847) Fix cloud fact implementation for Azure

GabrielNagy:
Since we now have an az_metadata fact, we can reimplement the cloud fact using this. The previous implementation was Linux-only and worked by searching for a specific string in the DHCP lease file, which does not work anymore (at least not on Ubuntu 20.04 Azure VMs).

For the new implemetation, extract the cloud fact into its own resolver, as it indirectly depends on the virtual fact which is also part of the virtualization resolver. The new implementation should work regardless of OS, as it just checks for the existence of the az_metadata fact.


GabrielNagy:
jenkins please test this on redhat7-64a,ubuntu2004-64a,windows2012r2-64a

lib/src/facts/resolvers/cloud_resolver.cc
```diff
@@ -0,0 +1,58 @@
...
+            result.provider = cloud_provider;
+        }
+
+        return result;
+    }
+
+    string cloud_resolver::get_azure(collection& facts)
+    {
+        string provider;
+#ifdef USE_CURL
```
ciprianbadescu, line 48: is this check needed here (dependency inversion)?

GabrielNagy: nope, should be good to remove it since the az resolver also checks for curl

ciprianbadescu (approved):

luchihoratiu (approved):

PR #2305, opened 2021-03-04
(FACT-2952) Align the lsbmajdistrelease fact with Facter 3

luchihoratiu:
The `lsbmajdistrelease` fact in Facter 4 was not showing the correct value on Ubuntu due to their unique approach on versioning.

Example:
In version `20.04.2`, they consider as follows:
 - `20.04` = major version
 - `2`     = minor version

Most of the other Linux operating systems respect the generic semantic versioning rules (`major.minor.patch`). This commit aligns the fact's output with Facter 3 accordingly.

Also removed and cleaned up some tests (Debian will use the Linux resolver so the case shouldn't be covered under Ubuntu).

### DIFF
--- a/lib/facter/facts/ubuntu/lsbdistrelease.rb
+++ b/lib/facter/facts/ubuntu/lsbdistrelease.rb
@@ -15,8 +15,8 @@ module Facts
         version = fact_value.split('.')
 
         [Facter::ResolvedFact.new(FACT_NAME, fact_value, :legacy),
-         Facter::ResolvedFact.new(ALIASES[0], version[0], :legacy),
-         Facter::ResolvedFact.new(ALIASES[1], version[1], :legacy)]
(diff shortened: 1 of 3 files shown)


gimmyxd (approved):
Looks good now 👍 
```
# before
facter --show-legacy | grep lsb
lsbdistcodename => bionic
lsbdistdescription => Ubuntu 18.04.4 LTS
lsbdistid => Ubuntu
lsbdistrelease => 18.04
lsbmajdistrelease => 18
lsbminordistrelease => 04

# with the fix
facter --show-legacy | grep lsb
lsbdistcodename => bionic
lsbdistdescription => Ubuntu 18.04.4 LTS
lsbdistid => Ubuntu
lsbdistrelease => 18.04
lsbmajdistrelease => 18.04
```

PR #2306, opened 2021-03-05
(FACT-2956) rework Facter.value

gimmyxd:
Updated Facter.value to resolve the facts in a
similar way to Facter 3:
  - ~~search in the internal collection(not ported to Facter 4)~~
  - load the `fact_name.rb` from the configured custom directories
  - load all the core facts, external facts and env facts
  - load all custom facts

Side note: `Facter.value` becomes considerably faster for core and external facts on systems that have many custom facts because it will not call `Kernel.load` to evaluate all the custom facts files.

Run unit tests:

- [x] puppetlabs-puppet_agent
- [x] puppetlabs-stdlib
- [x] puppet-nginx
- [x] puppetlabs-apt
- [x] puppetlabs-java
- [x] puppetlabs-acl
- [x] puppet-prometheus

```
puppetlabs-stdlib unit tests
  Facter 4(4.0.51):     Took 86 seconds (1:26) 
  Facter 4(FACT-2956):  Took 51 seconds

puppetlabs-apt unit tests
  Facter 4(4.0.51):    Took 27 seconds
  Facter 4(FACT-2956): Took 21 seconds


puppetlabs-acl unit tests
  Facter 4(4.0.51):    Took 26 seconds
  Facter 4(FACT-2956): Took 26 seconds

puppet-prometheus unit tests
   Facter 4(4.0.51):    Took 201 seconds (3:21)
   Facter 4(FACT-2956): Took 155 seconds (2:35)
```

TODO:
- [x] squash commits after review

### DIFF
--- a/acceptance/tests/api/value/core_and_custom.rb
+++ b/acceptance/tests/api/value/core_and_custom.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+test_name 'Facter.value(core_fact) when custom fact is defined' do
+  confine :to, platform: 'ubuntu'
+  tag 'audit:high'
+
+  require 'facter/acceptance/base_fact_utils'
+  require 'facter/acceptance/api_utils'
+  extend Facter::Acceptance::BaseFactUtils
+  extend Facter::Acceptance::ApiUtils
+
+  agents.each do |agent|
+    fact_name = 'os.name'
+    core_fact_value = os_processors_and_kernel_expected_facts(agent)[fact_name]
+
+    step 'in different file than fact name' do
+      facts_dir = agent.tmpdir('facts')
+
+      teardown do
+        agent.rm_rf(facts_dir)
+      end
+
+      fact_file = File.join(facts_dir, 'test_fact.rb')
+      fact_content = <<-RUBY
+        Facter.add('#{fact_name}') do
+          has_weight(100)
+          setcode { 'custom_fact' }
+        end
+      RUBY
+
+      create_remote_file(agent, fact_file, fact_content)
+
+      step 'returns core fact value' do
+        facter_rb = facter_value_rb(agent, fact_name, custom_dir: facts_dir)
+        fact_value = on(agent, "#{ruby_command(agent)} #{facter_rb}").stdout&.strip
+
+        assert_match(fact_value, core_fact_value, 'Incorrect fact value for os.name')
+      end
+    end
+  end
+end
--- a/lib/facter/custom_facts/util/collection.rb
+++ b/lib/facter/custom_facts/util/collection.rb
@@ -106,6 +106,11 @@ module LegacyFacter
         @loaded = false
       end
 
+      def custom_fact(fact_name)
+        internal_loader.load(fact_name)
+        @custom_facts = @facts.select { |_k, v| v.options[:fact_type] == :custom }
+      end
+
       # Builds a hash of custom facts
       def custom_facts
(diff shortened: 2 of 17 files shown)


gimmyxd:

lib/facter/custom_facts/util/collection.rb
```diff
@@ -107,13 +107,16 @@ def reload_custom_facts
...
-      def custom_facts
+      def custom_facts(fact_name = nil)
         return @custom_facts if @valid_custom_facts
 
-        @valid_custom_facts = true
-
-        internal_loader.load_all unless @loaded
-        @loaded = true
+        if fact_name
+          internal_loader.load(fact_name)
```
gimmyxd, line 114: this should be a separate method, and drop the f`act_name` argument

spec/framework/core/fact_manager_spec.rb
```diff
@@ -4,90 +4,349 @@
...
+          expect(logger).to have_received(:debug)
+            .with("Searching fact: #{user_query} in core facts and external facts")
+        end
+
+        it 'loads all custom facts' do
+          Facter::FactManager.instance.resolve_fact(user_query)
+
+          expect(logger).to have_received(:debug)
+            .with("Searching fact: #{user_query} in all custom facts")
+        end
```
GabrielNagy, line 178: Do we want to maybe assert that this behavior happens in a specific order?

spec/framework/core/fact_manager_spec.rb
```diff
@@ -4,90 +4,349 @@
...
+        expect(logger).to have_received(:debug)
+          .with("Searching fact: #{user_query} in file: #{user_query}.rb")
+      end
+
+      it 'loads core and external facts' do
+        Facter::FactManager.instance.resolve_fact(user_query)
+
+        expect(logger).to have_received(:debug)
+          .with("Searching fact: #{user_query} in core facts and external facts")
+      end
```
GabrielNagy, line 251: same comment about order

spec/framework/core/fact_manager_spec.rb
```diff
@@ -4,90 +4,349 @@
...
+
+      it 'loads all custom facts' do
+        Facter::FactManager.instance.resolve_fact(user_query)
+
+        expect(logger).to have_received(:debug)
+          .with("Searching fact: #{user_query} in all custom facts")
```
GabrielNagy, line 314: same comment about order

acceptance/tests/api/value/core_and_custom.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+test_name 'Facter.value(core_fact) when custom fact is defined' do
```
GabrielNagy, line 3: should probably differentiate between this test's title and the following one from `core_and_custom_overwrite.rb` as they have the same name

GabrielNagy (commented):
The code looks good to me, had some comments on the tests.

Tested this manually with a variety of core/custom/external fact combinations and everything seems to happen in the correct order.

spec/framework/core/fact_manager_spec.rb
gimmyxd: Tried, but it will break rubocop multiple expectations cops. On the other hand, given the way tests are executed, if the order is not expected, and something else gets on the logger, the expectation will fail.

GabrielNagy (approved):

PR #2311, opened 2021-03-10
(FACT-2930) Facter 4 Date and Time no support

Dorin-Pleava:
As facter 3 did not support Date and Time types, facter 4 should not
support them.

Using the following custom fact,
```
Facter.add(:now) do
  setcode { Time.now }
end
```

Facter 3 outputs empty fact:
...
now =>
...

Facter 4 outputs an error and no empty fact
ERROR Facter - Fact resolution fact='now', resolution='<anonymous>' resolved to an invalid value: Expected 2021-03-10 12:12:45 +0200 to be one of [Integer, Float, TrueClass, FalseClass, NilClass, Symbol, String, Array, Hash], but was Time

### DIFF
--- a/lib/facter/custom_facts/util/normalization.rb
+++ b/lib/facter/custom_facts/util/normalization.rb
@@ -16,8 +16,6 @@ module LegacyFacter
       # @return [void]
       def normalize(value)
         case value
-        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol, Date
-          value
         when String
           normalize_string(value)
         when Array
@@ -25,7 +23,7 @@ module LegacyFacter
         when Hash
           normalize_hash(value)
         else
-          raise NormalizationError, "Expected #{value} to be one of #{VALID_TYPES.inspect}, but was #{value.class}"
+          value
         end
       end
 
--- a/spec/custom_facts/util/fact_spec.rb
+++ b/spec/custom_facts/util/fact_spec.rb
@@ -196,6 +196,22 @@ describe Facter::Util::Fact do
         having_attributes(value: '1', used_resolution_weight: 1)
       )
     end
+
+    [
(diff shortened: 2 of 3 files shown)


Dorin-Pleava:
Note: When adding --json flag to facter 3, the `now` fact will show as `"now": null`.
Also if a custom fact returns nil, facter 3 will not display it, but will show a debug message: `fact "now" resolved to null and will not be added.`.

lib/facter/custom_facts/util/normalization.rb
```diff
@@ -16,16 +16,14 @@ class NormalizationError < StandardError; end
       # @return [void]
       def normalize(value)
         case value
-        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol, Date
```
gimmyxd, line 19: I'm not sure we should be this permissive. I think now facter 4 cli will behave differently than it does in facter 3.

Dorin-Pleava: Yep, I think the same.
I tried to make it simmilar to how facter 3 returns objects in irb:
```
2.7.2 :008 > Facter.add(:dd) do
2.7.2 :009 >   setcode { StandardError.new("test") }
2.7.2 :010 > end
 => #<Facter::Util::Fact:0x00007fe07f137ec0>
2.7.2 :011 > Facter.value(:dd)
 => #<StandardError: test>
2.7.2 :012 > Facter.value(:dd).class
 => StandardError
2.7.2 :013 >
```
 Also created https://github.com/puppetlabs/facter/pull/2314 just to add Date and Time in case anyone relied on only these types.

gimmyxd:
closing in favour of https://github.com/puppetlabs/facter/pull/2314

PR #2314, opened 2021-03-10
(FACT-2930) Add Date and Time to supported valid types

Dorin-Pleava:
Added supported valid types just for Date and Time

### DIFF
--- a/lib/facter/custom_facts/util/normalization.rb
+++ b/lib/facter/custom_facts/util/normalization.rb
@@ -5,7 +5,10 @@ module LegacyFacter
     module Normalization
       class NormalizationError < StandardError; end
 
-      VALID_TYPES = [Integer, Float, TrueClass, FalseClass, NilClass, Symbol, String, Array, Hash].freeze
+      # load Date and Time classes
+      require 'time'
+
+      VALID_TYPES = [Integer, Float, TrueClass, FalseClass, NilClass, Symbol, String, Array, Hash, Date, Time].freeze
 
       module_function
 
@@ -16,8 +19,10 @@ module LegacyFacter
       # @return [void]
       def normalize(value)
         case value
-        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol, Date
+        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol
           value
+        when Date, Time
+          value.iso8601
         when String
           normalize_string(value)
         when Array
--- a/spec/custom_facts/util/normalization_spec.rb
+++ b/spec/custom_facts/util/normalization_spec.rb
@@ -95,17 +95,37 @@ describe LegacyFacter::Util::Normalization do
     end
   end
 
-  [1, 1.0, true, false, nil].each do |val|
-    it "accepts #{val.inspect}:#{val.class}" do
-      expect(normalization.normalize(val)).to eq(val)
-    end
-  end
+  describe '.normalize' do
+    context 'when Time object' do
+      it 'returns a string in an ISO 8601 format' do
+        value = Time.utc(2020)
+        expected = '2020-01-01T00:00:00Z'
 
-  [Object.new, Set.new].each do |val|
-    it "rejects #{val.inspect}:#{val.class}" do
-      expect do
-        normalization.normalize(val)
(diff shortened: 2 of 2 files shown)


joshcooper (approved):
Given some custom facts on the puppetserver:

```
# mkdir -p /etc/puppetlabs/code/environments/production/modules/customfacts/lib/facter
# cat <<END > /etc/puppetlabs/code/environments/production/modules/customfacts/lib/facter/types.rb
require 'date'

Facter.add(:custom_date) do
  setcode { Date.new(2020) }
end

Facter.add(:custom_time) do
  setcode { Time.now }
end

Facter.add(:custom_object) do
  setcode { Object.new }
end

Facter.add(:custom_err) do
  setcode { StandardError.new('something') }
end
END
# chown -R puppet:puppet /etc/puppetlabs/code/environments/production/modules
```

If I run a puppet6 and puppet7 agents and compare the facts they submitted, I see:

**puppet6**
```
# cat /opt/puppetlabs/server/data/puppetserver/server_data/facts/fugual-protocol.delivery.puppetlabs.net.json | jq . | grep -E '(custom_(date|time|object|err))|puppetversion'
    "custom_date": "2020-01-01",
    "custom_err": "something",
    "custom_object": "#<Object:0x0000000002aec868>",
    "custom_time": "2021-03-11 17:58:55 +0000",
    "puppetversion": "6.22.0",
```

**puppet7**
```
# cat /opt/puppetlabs/server/data/puppetserver/server_data/facts/pliant-leather.delivery.puppetlabs.net.json | jq . | grep -E '(custom_(date|time|object|err))|puppetversion'
    "puppetversion": "7.5.0",
    "custom_date": "2020-01-01",
```

So allowing Date and Time in Facter 4 (to match Facter 3), but continuing to exclude arbitrary objects and exceptions makes sense to me.

gimmyxd:
Should we normalize the Date format, or just pass back whatever the custom fact creates:
```ruby

 require 'facter'

 Facter.add(:jisx0301) do
   setcode { Date.today.jisx0301 }
 end

 Facter.add(:iso8601) do
   setcode { Date.today.iso8601 }
 end

 Facter.add(:today) do
   setcode { Date.today }
 end

 Facter.value(:jisx0301)
 => "R03.03.12"

 Facter.value(:iso8601)
 => "2021-03-12"

 Facter.value(:today)
 => #<Date: 2021-03-12 ((2459286j,0s,0n),+0s,2299161j)>
```

lib/facter/custom_facts/util/normalization.rb
```diff
@@ -16,8 +19,10 @@ class NormalizationError < StandardError; end
       # @return [void]
       def normalize(value)
         case value
-        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol, Date
+        when Integer, Float, TrueClass, FalseClass, NilClass, Symbol
           value
+        when Date, Time
+          value.iso8601
```
GabrielNagy, line 25: In facter 3, `Time.now` seems to more closely resemble `#to_s` rather than `#iso8601`. For `Date` it seems to be the same... so maybe just call `to_s` on them?

```
2.6.3 :001 > Facter.add(:custom_time) { setcode { Time.now } }
 => #<Facter::Util::Fact:0x000055d8ee5de608> 
2.6.3 :002 > Facter.value(:custom_time)
 => 2021-03-18 15:25:16 +0200 
```

But since we didn't fully "support" these type of facts before, I'm ok with converting them to ISO 8601 if the user doesn't do the normalization.

gimmyxd: Seems like `.to_s` calls `iso8601` on Date https://apidock.com/ruby/Date/to_s

GabrielNagy: hm... I would call .to_s on both then, just to be consistent

gimmyxd: I would keep iso8601 as is the standard for working with datetimes. Calling `to_s` we rely on Ruby to delegate to the right methods, which can change/be different depending on the ruby version. Also having a standardised format would make it easier for someone that wants to parse the resulted time string(back to time objects), in another programming language.

GabrielNagy: Makes sense to me 👍

GabrielNagy (approved):

PR #2315, opened 2021-03-10
(FACT-2944) Retrieve fqdn and hostname facts with ffi

oanatmaria:


mihaibuzgau (approved):

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,19 +36,23 @@ def self.getffihostname
...
-            return if res != 0 || res != EAI_NONAME
+            res = Hostname.getaddrinfo(hostname, '', hints.to_ptr, ret)
+            log = Log.new(self)
+            log.debug("FFI Getaddrinfo finished with exit status: #{res}")
+            return if res != 0 && res != EAI_NONAME
 
-            name_ptr = hints[:ai_canonname]
-            return if hints.to_ptr != FFI::Pointer::NULL || !name_ptr || hostname == name_ptr.read_string
+            addr = Facter::Util::Resolvers::Ffi::AddrInfo.new(ret.read_pointer)
+            name_ptr = addr[:ai_canonname]
```
joshcooper, line 53: `getaddrinfo` defines this field as `char *` but our struct above is using `:pointer`. I think we want to update our struct to be `:string` as well.

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -19,12 +19,12 @@ class AddrInfo < ::FFI::Struct
 
         module Hostname
           HOST_NAME_MAX = 64
-          EAI_NONAME = 8
+          EAI_NONAME = -2
```
joshcooper, line 22: I see `/usr/include/netdb.h` defines this as:

```
# define EAI_NONAME       -2    /* NAME or SERVICE is unknown.  */
```

But when I call the method with a bogus hostname I get `8`:

```
irb(main):052:0> Facter::Util::Resolvers::Ffi::Hostname.getaddrinfo('nada', '', hints.to_ptr, ret)
=> 8
```

We defined the `getaddrinfo` function as returning a signed `:int`, so there isn't a signed/unsigned mismatch... Also if I attach `gai_strerror`, the message looks correct for `8`, but not any negative number:

```
irb(main):053:0>  Facter::Util::Resolvers::Ffi::Hostname.gai_strerror(8)
=> "nodename nor servname provided, or not known"
irb(main):054:0>  Facter::Util::Resolvers::Ffi::Hostname.gai_strerror(-2)
=> "Unknown error"
```

So I'm inclined to leave it as `EAI_NONAME = 8` and maybe a note about the discrepancy?

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,19 +36,23 @@ def self.getffihostname
...
             hints[:ai_family] = Socket::AF_UNSPEC
             hints[:ai_socketype] = Socket::SOCK_STREAM
             hints[:ai_flags] = Socket::AI_CANONNAME
 
-            res = Hostname.getaddrinfo(hostname_ptr, FFI::Pointer::NULL, hints.to_ptr, FFI::Pointer::NULL)
-            return if res != 0 || res != EAI_NONAME
+            res = Hostname.getaddrinfo(hostname, '', hints.to_ptr, ret)
+            log = Log.new(self)
+            log.debug("FFI Getaddrinfo finished with exit status: #{res}")
+            return if res != 0 && res != EAI_NONAME
```
joshcooper, line 50: if `res == 0`, then from that point forward we own the native memory returned by `getaddrinfo` and we need to free it by calling `freeaddrinfo`, preferably in an ensure block:

> The freeaddrinfo() function frees the memory that was allocated for the dynamically allocated linked list res.

lib/facter/util/resolvers/ffi/hostname.rb
joshcooper: Very strange... I also see -2 when using jruby in puppetserver packages. Given this monkey patch to facter 4:

```ruby
require 'facter'
require 'facter/framework/core/file_loader'
require 'ffi'

module Facter
  module Util
    module Resolvers
      module Ffi
        class AddrInfo < ::FFI::Struct
          layout  :ai_flags, :int,
                  :ai_family, :int,
                  :ai_socketype, :int,
                  :ai_protocol, :int,
                  :ai_addrlen, :uint,
                  :ai_addr, :pointer,
                  :ai_canonname, :pointer,
                  :ai_next, :pointer
        end

        module Hostname
          extend ::FFI::Library
          ffi_lib ::FFI::Library::LIBC

          attach_function :getaddrinfo, %i[string string pointer pointer], :int
          attach_function :gai_strerror, %i[int], :string
        end
      end
    end
  end
end

ret = FFI::MemoryPointer.new(Facter::Util::Resolvers::Ffi::AddrInfo)

hints = Facter::Util::Resolvers::Ffi::AddrInfo.new
hints[:ai_family] = Socket::AF_UNSPEC
hints[:ai_socketype] = Socket::SOCK_STREAM
hints[:ai_flags] = Socket::AI_CANONNAME

res = Facter::Util::Resolvers::Ffi::Hostname.getaddrinfo('nada', '', hints.to_ptr, ret)
puts res
puts Facter::Util::Resolvers::Ffi::Hostname.gai_strerror(res)
```

On Redhat 8, I get:

```
[root@seismic-subtype ~]# java -version
openjdk version "1.8.0_282"
OpenJDK Runtime Environment (build 1.8.0_282-b08)
OpenJDK 64-Bit Server VM (build 25.282-b08, mixed mode)
[root@seismic-subtype ~]# /opt/puppetlabs/server/bin/puppetserver ruby --version
jruby 9.2.14.0 (2.5.7) 2020-12-08 ebe64bafb9 OpenJDK 64-Bit Server VM 25.282-b08 on 1.8.0_282-b08 +jit [linux-x86_64]
[root@seismic-subtype ~]# /opt/puppetlabs/server/bin/puppetserver ruby getaddrinfo.rb
-2
Name or service not known
```

But using jruby 9.2.13.0 on macOS (I can't get 9.2.14.0 to install):

```
$ java -version
java version "1.8.0_162"
Java(TM) SE Runtime Environment (build 1.8.0_162-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.162-b12, mixed mode)
$ bx ruby --version
jruby 9.2.13.0 (2.5.7) 2020-08-03 9a89c94bcc Java HotSpot(TM) 64-Bit Server VM 25.162-b12 on 1.8.0_162-b12 +jit $ bx ruby getaddrinfo.rb
8
nodename nor servname provided, or not known
```

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,19 +36,23 @@ def self.getffihostname
             raw_hostname.read_string
           end
 
-          def self.getffiaddrinfo(hostname)
-            hostname_ptr = FFI::MemoryPointer.new(hostname)
+          def self.getffiaddrinfo(hostname) # rubocop:disable  Metrics/AbcSize
+            ret = FFI::MemoryPointer.new(Facter::Util::Resolvers::Ffi::AddrInfo)
```
ciprianbadescu, line 40:
```       int getaddrinfo(const char *node, const char *service,
                       const struct addrinfo *hints,
                       struct addrinfo **res);
```
based on this, ret, should be pointer to pointer, I assume that allocating a `:pointer` should be enough

lib/facter/util/resolvers/ffi/hostname.rb
ciprianbadescu: I assume this value is `networking resolver` dependent, -2 on Linux using glibc/resolv and 8 on *BSD (including macOS) using ISC libbind

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,19 +36,23 @@ def self.getffihostname
...
 
             hints = Facter::Util::Resolvers::Ffi::AddrInfo.new
             hints[:ai_family] = Socket::AF_UNSPEC
             hints[:ai_socketype] = Socket::SOCK_STREAM
             hints[:ai_flags] = Socket::AI_CANONNAME
 
-            res = Hostname.getaddrinfo(hostname_ptr, FFI::Pointer::NULL, hints.to_ptr, FFI::Pointer::NULL)
-            return if res != 0 || res != EAI_NONAME
```
ciprianbadescu, line 48: We could log here a debug/warning message, similar to facter 3

lib/facter/util/resolvers/ffi/hostname.rb
ciprianbadescu: We could log here a debug/warning message, similar to facter 3

lib/facter/util/resolvers/ffi/hostname.rb
oanatmaria: :ai_canonname was changed to :string

lib/facter/util/resolvers/ffi/hostname.rb
oanatmaria: I've had dropped the check for EAI_NONAME because CFacter only uses it to log a different message: https://github.com/puppetlabs/facter/blob/ce1f2bb4a91a1ac4ae5852091c96ae6ee3712e23/lib/src/facts/posix/networking_resolver.cc#L96

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,21 +36,32 @@ def self.getffihostname
             raw_hostname.read_string
           end
 
-          def self.getffiaddrinfo(hostname)
-            hostname_ptr = FFI::MemoryPointer.new(hostname)
+          def self.getffiaddrinfo(hostname) # rubocop:disable  Metrics/AbcSize
+            ret = FFI::MemoryPointer.new(:pointer)
```
ciprianbadescu, line 40: could we make sure that the new pointer is NULL (or check, it may be default to NULL from ruby)

lib/facter/util/resolvers/ffi/hostname.rb
```diff
@@ -36,21 +36,32 @@ def self.getffihostname
...
-            res = Hostname.getaddrinfo(hostname_ptr, FFI::Pointer::NULL, hints.to_ptr, FFI::Pointer::NULL)
-            return if res != 0 || res != EAI_NONAME
+            res = Hostname.getaddrinfo(hostname, '', hints.to_ptr, ret)
+            log = Log.new(self)
+            log.debug("FFI Getaddrinfo finished with exit status: #{res}")
+            log.debug("FFI Getaddrinfo returned an AddrInfo struct at address: #{ret}")
+            return unless res.zero?
+
+            begin
+              return if ret == FFI::Pointer::NULL
```
ciprianbadescu, line 54: it may be enough to check `ret.read_pointer` only?

ciprianbadescu (approved):

h-haaks:
Hi, 
I'm having som issues with my unit tests in a module with custom types that I think could be related to this PR.
Since 16.03.2021 at 00:01:40 my rspec unit tests with puppet6 started failing like this
```
An error occurred while loading ./spec/unit/puppet/provider/am_agent_j2ee/ruby_spec.rb.
Failure/Error: defaultto "amAgent_#{Facter.value('fqdn').gsub('.', '_')}_80.log"
Puppet::Error:
  Could not autoload puppet/type/am_agent_j2ee: undefined method `gsub' for nil:NilClass
```
It seems like Facter 4.0.52 does not report the fqdn fact ....

ekohl:
Same, Facter 4.0.52 is broken for me. I wonder if https://github.com/puppetlabs/facter/pull/2318 is the correct fix for this.

gimmyxd:
@ekohl would it be possible to set a custom ref for Facter to `main` or `b33d60d86d6edd6747f2237a054df4cb956e8604` and see if it fixes your issue?

ekohl:
I've pushed a commit to https://github.com/theforeman/foreman-installer/pull/662, CI is running now.

ekohl:
Yes, that also makes it green. I'd appreciate a quick 4.0.53 release to fix it.

Magisus:
This is also causing issues for our Travis tests for the `puppetserver-ca-cli` gem. I rolled forward to https://github.com/puppetlabs/facter/commit/c5b9592e25611315f1c3e9321cad1fe6fb42c955 (head of main) and I'm still seeing the issues. Failing build: https://travis-ci.com/github/puppetlabs/puppetserver-ca-cli/jobs/493577152

PR #2317, opened 2021-03-18
(FACT-2988) Only query cloud provider on HyperV

GabrielNagy:
Due to some bug in Ruby/Windows, the Azure metadata fact takes 20+ seconds before timing out, regardless of the set timeout.

To fix this, only attempt to resolve Azure metadata if we're on HyperV.

Additionally, add a guard to the HTTP resolver to ensure long timeouts don't happen on Windows (the Linux implementation respects timeouts).

### DIFF
--- a/lib/facter/facts/linux/cloud/provider.rb
+++ b/lib/facter/facts/linux/cloud/provider.rb
@@ -6,10 +6,17 @@ module Facts
       class Provider
         FACT_NAME = 'cloud.provider'
 
-        def call_the_resolver
-          az_metadata = Facter::Resolvers::Az.resolve(:metadata)
+        def initialize
+          @virtual = Facter::Util::Facts::VirtualDetector.new
+        end
 
-          Facter::ResolvedFact.new(FACT_NAME, az_metadata&.empty? ? nil : 'azure')
+        def call_the_resolver
+          provider = case @virtual.platform
+                     when 'hyperv'
+                       'azure' unless Facter::Resolvers::Az.resolve(:metadata).empty?
+                     end
+
+          Facter::ResolvedFact.new(FACT_NAME, provider)
         end
       end
     end
--- a/lib/facter/util/resolvers/http.rb
+++ b/lib/facter/util/resolvers/http.rb
@@ -10,7 +10,7 @@ module Facter
           CONNECTION_TIMEOUT = 0.6
           SESSION_TIMEOUT = 5
 
-          # Makes a GET http request and returns it's response.
+          # Makes a GET http request and returns its response.
           #
           # Params:
           # url: String which contains the address to which the request will be made
@@ -39,6 +39,12 @@ module Facter
             uri = URI.parse(url)
             http = http_obj(uri, timeouts)
             request = request_obj(headers, uri, request_type)
+
+            # The Windows implementation of sockets does not respect net/http
+            # timeouts, so check if the target is reachable in a different way
+            if Gem.win_platform?
+              Socket.tcp(uri.host, uri.port, connect_timeout: timeouts[:connection] || CONNECTION_TIMEOUT)
+            end
 
             # Make the request
             response = http.request(request)
--- a/lib/facter/facts/windows/az_metadata.rb
+++ b/lib/facter/facts/windows/az_metadata.rb
@@ -5,10 +5,6 @@ module Facts
     class AzMetadata
       FACT_NAME = 'az_metadata'
 
-      def initialize
-        @virtual = Facter::Util::Facts::VirtualDetector.new
-      end
-
       def call_the_resolver
         return Facter::ResolvedFact.new(FACT_NAME, nil) unless azure_hypervisor?
 
@@ -20,7 +16,7 @@ module Facts
       private
 
       def azure_hypervisor?
-        @virtual.platform == 'hyperv'
+        Facter::Resolvers::Virtualization.resolve(:virtual) == 'hyperv'
       end
     end
   end
--- a/lib/facter/facts/windows/cloud/provider.rb
+++ b/lib/facter/facts/windows/cloud/provider.rb
@@ -6,10 +6,17 @@ module Facts
       class Provider
         FACT_NAME = 'cloud.provider'
 
-        def call_the_resolver
-          az_metadata = Facter::Resolvers::Az.resolve(:metadata)
+        def initialize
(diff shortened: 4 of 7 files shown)


lib/facter/facts/linux/cloud/provider.rb
```diff
@@ -6,10 +6,17 @@ module Cloud
...
+        def initialize
+          @virtual = Facter::Util::Facts::VirtualDetector.new
+        end
+
         def call_the_resolver
-          az_metadata = Facter::Resolvers::Az.resolve(:metadata)
+          provider = case @virtual.platform
+                     when 'hyperv'
+                       'azure' unless Facter::Resolvers::Az.resolve(:metadata).empty?
+                     end
```
gimmyxd, line 17: [optional] Don't really need the `case when` for a single case, could be a simple `if`

lib/facter/util/resolvers/http.rb
```diff
@@ -40,6 +40,12 @@ def make_request(url, headers, timeouts, request_type)
             http = http_obj(uri, timeouts)
             request = request_obj(headers, uri, request_type)
 
+            # The Windows implementation of sockets does not respect net/http
+            # timeouts, so check if the target is reachable in a different way
+            if Gem.win_platform?
+              Socket.tcp(uri.host, uri.port, connect_timeout: timeouts[:connection] || CONNECTION_TIMEOUT)
+            end
```
gimmyxd, line 47: Might be good to use https://github.com/bblimke/webmock to refactor and add more tests to the Http client. We can create another ticket for that, since it is a technical debt.

gimmyxd: This will work for the `open_timeout` but i think there is still an issue with `read_timeout` that might not be respected on windows.

gimmyxd (commented):
The CONNECTION_TIMEOUT is now respected on Windows, we should also check that SESSION_TIMEOUT works.

lib/facter/util/resolvers/http.rb
GabrielNagy: I did some more digging about `open_timeout` and `read_timeout`. For `open_timeout`, Ruby [wraps it](https://github.com/ruby/ruby/blob/9e71f1aa4cea42c895a2857c0f9c252865468d53/lib/net/http.rb#L984) in a `Timeout.timeout` block which is known to be [problematic](https://vaneyckt.io/posts/the_disaster_that_is_rubys_timeout_method/) and [dangerous](http://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/): 

However, `read_timeout` is only passed when creating a BufferedIO object, and is used [here](https://github.com/ruby/ruby/blob/9e71f1aa4cea42c895a2857c0f9c252865468d53/lib/net/protocol.rb#L219
).

I think the underlying issue is within the `Timeout` module, so `read_timeout` should not be affected. To confirm this, I created a simple reproduction case which creates a server, accepts a connection then waits for 5 seconds before sending a response:

`server.rb`
```ruby
require 'socket'
server = TCPServer.new("127.0.0.1", 8888)

while session = server.accept do
  session.gets
  sleep(5)
  response = <<-HEREDOC
HTTP/1.1 200 OK

hello world
  HEREDOC
  session.puts response
  session.close
end
```

Then, in an irb session we can try different timeouts which are respected:

```
irb(main):001:0> require 'net/http'
=> true
irb(main):002:0> http = Net::HTTP.new('127.0.0.1', 8888)
=> #<Net::HTTP 127.0.0.1:8888 open=false>
irb(main):003:0> http.read_timeout = 1
=> 1
irb(main):004:0> http.get('/')
       16: from C:/Program Files/Puppet Labs/Puppet/puppet/bin/irb.bat:31:in `load'
       15: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
       14: from (irb):4
       13: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1241:in `get'
       12: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1483:in `request'
       11: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:933:in `start'
       10: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1485:in `block in request'
        9: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1492:in `request'
        8: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1519:in `transport_request'
        7: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1519:in `catch'
        6: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http.rb:1528:in `block in transport_request'
        5: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http/response.rb:31:in `read_new'
        4: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/http/response.rb:42:in `read_status_line'
        3: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/protocol.rb:201:in `readline'
        2: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/protocol.rb:191:in `readuntil'
        1: from C:/Program Files/Puppet Labs/Puppet/puppet/lib/ruby/2.7.0/net/protocol.rb:217:in `rbuf_fill'
Net::ReadTimeout (Net::ReadTimeout)
irb(main):005:0> http.read_timeout = 7
=> 7
irb(main):006:0> http.get('/')
=> #<Net::HTTPOK 200 OK readbody=true>
```

lib/facter/facts/linux/cloud/provider.rb
GabrielNagy: Was thinking this could be a bit future-proof for when we add more cloud providers

gimmyxd (approved):

luchihoratiu (approved):

PR #2319, opened 2021-03-19
(FACT-2956) fix loading of env facts

gimmyxd:
https://github.com/puppetlabs/facter/pull/2306 changed
the order of fact loading when all facts are loaded,
which made environment facts not to be loaded anymore
if there is no other fact already loaded with the
same name.

This commit changes back to the same order of loading
facts.


GabrielNagy (approved):
Before:
```sh-session
> FACTER_test=a bundle exec facter test
 
```
After:
```sh-session
> FACTER_test=a bundle exec facter test
a
```

PR #2328, opened 2021-03-25
(FACT-2994) Check for config file when resolving selinux

GabrielNagy:
Previously we considered `selinux.enabled` to be true just by checking for a `selinuxfs` mountpoint. Facter 3 also checks for the existence of the `/etc/selinux/config` file.

This commit aligns the behavior of Facter 4 with Facter 3, and rearranges the tests based on possible use cases.

Facter 3 commit: https://github.com/puppetlabs/facter/commit/125a79e4da408bb4d4a86ebb7dd71c0ca27e288f

### DIFF
--- a/lib/facter/resolvers/selinux.rb
+++ b/lib/facter/resolvers/selinux.rb
@@ -13,21 +13,21 @@ module Facter
         end
 
         def retrieve_facts(fact_name)
-          mountpoint = read_mounts_file if @fact_list[:enabled].nil?
+          mountpoint = selinux_mountpoint
+
+          @fact_list[:enabled] = !mountpoint.empty? && read_selinux_config
           read_other_selinux_facts(mountpoint) if @fact_list[:enabled]
 
           @fact_list[fact_name]
         end
 
-        def read_mounts_file
+        def selinux_mountpoint
           output = Facter::Core::Execution.execute('cat /proc/self/mounts', logger: log)
-          @fact_list[:enabled] = false
           mountpoint = ''
 
           output.each_line do |line|
             next unless line =~ /selinuxfs/
 
-            @fact_list[:enabled] = true
             mountpoint = line.split("\s")[1]
             break
           end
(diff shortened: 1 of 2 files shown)


lib/facter/resolvers/selinux.rb
```diff
@@ -60,7 +58,7 @@ def read_selinux_config
             @fact_list[:config_policy] = line.split('=').last.strip if line =~ /^SELINUXTYPE=/
           end
 
-          true unless file_lines.empty?
+          file_lines.empty? ? false : true
```
ciprianbadescu, line 61: `!file_lines.empty?` would be a synonym

lib/facter/resolvers/selinux.rb
```diff
@@ -13,30 +13,28 @@ def post_resolve(fact_name, _options)
         end
 
         def retrieve_facts(fact_name)
-          mountpoint = read_mounts_file if @fact_list[:enabled].nil?
+          mountpoint = selinux_mountpoint
+
+          @fact_list[:enabled] = !mountpoint.empty? && read_selinux_config
```
ciprianbadescu, line 18: if `mountpoint.empty?` is true, it will never execute `read_selinux_config`, is this intended?

GabrielNagy: Yes, `mountpoint.empty?` translates to selinuxfs not mounted, so there's no point to check for the config file (from what I understand). This is similar to the facter 3 implementation: https://github.com/puppetlabs/facter/commit/125a79e4da408bb4d4a86ebb7dd71c0ca27e288f#diff-c69186a8ef3e4d86f44def025a4e1bf8f132faa49d31b0e0b90adf55333fe325R70

lib/facter/resolvers/selinux.rb
GabrielNagy: will update, sounds more readable to me

lib/facter/resolvers/selinux.rb
ciprianbadescu: You are right. Filling @fact_list in various places is not easy reading for me.

ciprianbadescu (approved):

PR #2331, opened 2021-03-26
(FACT-2984) Align mount options fact on OSX with Facter 3

luchihoratiu:
This commit adds partial remapping of mountpoints options to match with Facter 3 output.
Before:
```ruby
OPT_NAMES[MNT_RDONLY]      = 'read-only'
OPT_NAMES[MNT_ASYNC]       = 'asynchronous'
OPT_NAMES[MNT_SYNCHRONOUS] = 'synchronous'
OPT_NAMES[MNT_QUOTA]       = 'quotas'
OPT_NAMES[MNT_ROOTFS]      = 'rootfs'
OPT_NAMES[MNT_DEFWRITE]    = 'defwrite'
```
After:
```ruby
OPT_NAMES[MNT_RDONLY]      = 'readonly'
OPT_NAMES[MNT_ASYNC]       = 'async'
OPT_NAMES[MNT_SYNCHRONOUS] = 'noasync'
OPT_NAMES[MNT_QUOTA]       = 'quota'
OPT_NAMES[MNT_ROOTFS]      = 'root'
OPT_NAMES[MNT_DEFWRITE]    = 'deferwrites'
```

See https://github.com/puppetlabs/facter/blob/3.x/lib/src/facts/bsd/filesystem_resolver.cc#L55-L76 and https://github.com/djberg96/sys-filesystem/blob/9880f127465e680534263ec0de078179d73ef18b/lib/sys/unix/sys/filesystem.rb#L16-L35 for comparisson.

### DIFF
--- a/lib/facter/resolvers/macosx/mountpoints.rb
+++ b/lib/facter/resolvers/macosx/mountpoints.rb
@@ -21,7 +21,7 @@ module Facter
               device = fs.name
               filesystem = fs.mount_type
               path = fs.mount_point
-              options = fs.options.split(',').map(&:strip).map { |o| o == 'rootfs' ? 'root' : o }
+              options = read_options(fs.options)
 
               mounts[path] = read_stats(path).tap do |hash|
                 hash[:device] = device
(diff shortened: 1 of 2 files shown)


lib/facter/patches/sysfilesystem/sys/statvfs.rb
```diff
@@ -4,6 +4,16 @@
 
 module Sys
   class Filesystem
+
+    OPT_NAMES = OPT_NAMES.dup if OPT_NAMES.frozen?
```
gimmyxd, line 8: Won't this issue a warning because `OPT_NAMES` is redefined?
I think you'll have to remove the constant first, then set it to the new value

GabrielNagy (approved):

PR #2332, opened 2021-03-30
(FACT-2999) Handle unexpected JRuby behavior

GabrielNagy:
In some cases, JRuby can fail the first time it executes an external command. In our case, this happens through the `Facter::Core::Execution` module, where `Process.wait` fails in the underlying call to popen3.

The exception raised is `Errno::ENOENT` which is unexpected for waitpid, this has been traced to a JRuby issue that is still unresolved[1].

We found that even if ENOENT is raised, the output still gets filled in, so as a workaround it's enough to continue the execution if this specific exception is raised and we're running under JRuby.

Other exception types are re-raised, and the logging was improved to print a debug message in this case.

[1] https://github.com/jruby/jruby/issues/5971

### DIFF
--- a/lib/facter/custom_facts/core/execution/popen3.rb
+++ b/lib/facter/custom_facts/core/execution/popen3.rb
@@ -11,6 +11,8 @@ module Facter
   module Core
     module Execution
       class Popen3
+        @log ||= Log.new(self)
+
         def self.popen_rune(cmd, opts, child_io, parent_io) # :nodoc:
           pid = spawn(*cmd, opts)
           child_io.each(&:close)
@@ -20,7 +22,17 @@ module Facter
               return yield(*result)
             ensure
               parent_io.each(&:close)
-              Process.wait(pid)
+              begin
+                Process.wait(pid)
+              rescue Errno::ENOENT
+                # For some reason, the first Process.wait executed in JRuby
+                # always fails with ENOENT. However, the command output is
+                # filled in so we just need to silently continue.
+                # https://github.com/jruby/jruby/issues/5971
+                raise unless RUBY_PLATFORM == 'java'
+
+                @log.debug('Caught ENOENT during Process.wait on JRuby, continuing...')
+              end
             end
           end
           result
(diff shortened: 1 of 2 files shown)


luchihoratiu (approved):
LGTM! 👍 
Manually tested these changes locally on macOSX as described in the ticket and using https://puppet.com/docs/puppet/7.5/server/dev_debugging.html. Commands are executed as expected (first one included).

lib/facter/custom_facts/core/execution/popen3.rb
```diff
@@ -20,7 +20,15 @@ def self.popen_rune(cmd, opts, child_io, parent_io) # :nodoc:
...
               parent_io.each(&:close)
-              Process.wait(pid)
+              begin
+                Process.wait(pid)
+              rescue Errno::ENOENT
+                # For some reason, the first Process.wait executed in JRuby
+                # always fails with ENOENT. However, the command output is
+                # filled in so we just need to silently continue.
+                # https://github.com/jruby/jruby/issues/5971
+                raise unless RUBY_PLATFORM == 'java'
```
ciprianbadescu, line 30: could we call micro-sleep and call Process.wait(pid) again instead on silently ignoring the error?

GabrielNagy: Subsequent calls to `Process.wait(pid)` with the same pid fail with `ECHILD` so I assume the PID doesn't exist anymore.

luchihoratiu: Can confirm without any sleep:
> Errno::ECHILD: No child processes - No child processes

GabrielNagy:
Added a log message:

```
2021-03-31 10:52:32,394 DEBUG [clojure-agent-send-pool-0] [puppetserver] Facter Resolving facts sequentially
2021-03-31 10:52:32,479 DEBUG [clojure-agent-send-pool-0] [puppetserver] Facter Executing command: ip link show tun0
2021-03-31 10:52:32,522 DEBUG [clojure-agent-send-pool-0] [puppetserver] Facter Caught ENOENT during Process.wait on JRuby, continuing...
```

luchihoratiu (approved):
Tested latest changes. Still working as expected and the new debug log is looking good:

> 2021-03-31 15:07:01,530 DEBUG [clojure-agent-send-pool-0] [puppetserver] Facter Caught ENOENT during Process.wait on JRuby, continuing...

PR #2333, opened 2021-03-30
(FACT-3000) allow segments wrapped in quotes

gimmyxd:
```

Facter.add('"f.g.h"') do
  setcode { "no_space" }
end

# Facter 4 -> "f.g.h" -> { "f.g.h": "no_space" }  -> $facts["f.g.h"]

Facter.value('"f.g.h"')
 => "no_space"

❯ bx facter \"f.g.h\"
no_space								

-------------------------------------

Facter.add('m."n.p"') do
  setcode { "structured" }
end

# Facter 4 -> "m.\"n.p\"" -> { "m" => { "n.p": "structured" } }  -> $facts["m"]["n.p"]

❯ bx facter m
{
  n.p => "structured"
}

❯ bx facter m.n


❯ bx facter m.n.p


❯ bx facter m."n.p"


❯ bx facter 'm."n.p"'
structured
```

### DIFF
--- a/lib/facter/framework/lookup.rb
+++ b/lib/facter/framework/lookup.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+module Facter
+  module Framework
+    module Lookup
+      class << self
+        SPECIAL = /['"\.]/.freeze
+        SEGMENTS_REGEX = /(\s*"[^"]+"\s*|\s*'[^']+'\s*|[^'".]+)/.freeze
+
+        # Split key into segments. A segment may be a quoted string (both single and double quotes can
+        # be used) and the segment separator is the '.' character. Whitespace will be trimmed off on
+        # both sides of each segment. Whitespace within quotes are not trimmed.
+        #
+        # If the key cannot be parsed, the inital key is returned
+        #
+        # Example:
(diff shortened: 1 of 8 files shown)


lib/facter/framework/lookup.rb
```diff
@@ -0,0 +1,81 @@
...
+        # Example:
+        # ["a", "b", "c"] => "a.b.c"
+        # ["a", "b.c"]    => "a.\"b.c\""
+        # ["a.\"b.c"]     => "a.\"b.c"
+        # ["a.b.c"]       => "\"a.b.c\""
+        #
+        # @param arr [Array<String>] the Array of Strings
+        # @return [String] the joined String
+        def join_keys(arr)
+          return arr[0] if arr.size == 1 && arr[0] !~ /^((\w+\.\w+)+(\.\w+)?)$/
```
gimmyxd, line 53: This might not cover all the cases.

Generally if the `arr` size is 1, means that `split_key` did not manage to split the original string into segments, so it is safe to return it, or the original string was a segment(like `"\"a.b.c\""`) which was converted to ["a.b.c"] and now we need to make it back to `"\"a.b.c\""`

PR #2334, opened 2021-03-30
(FACT-3002) add option to sanitize fact name 

ciprianbadescu:
To enable fact name sanitize option, add `sanitize-fact-name` to
`global` section and set it to `true`:
```
global : {
    sanitize-fact-name : true
}
```

The sanitize option will sanitize facts names containing `"` or ```
using the following algorithm:
 - name will be split in segments that either:
   - are quoted or ticked strings
   - are not containing `.`
 - segments are joined using `.` separator

This will happen just before fact is saved in the collection

### DIFF
--- a/lib/inc/facter/util/string.hpp
+++ b/lib/inc/facter/util/string.hpp
@@ -70,11 +70,9 @@ namespace facter { namespace util {
     boost::optional<int> maybe_stoi(std::string const& str);
 
     /**
-     * Converts the given string to a sanitized string using the same algorithm
-     * as in puppet:
-https://github.com/puppetlabs/puppet/blob/85bd9ed07e3e6c956f1972624a50be914292e1ab/lib/puppet/pops/lookup/sub_lookup.rb#L6-L44
+     * If the string is fully surrounded in quotes/ticks, remove them
      * @param str The string to sanitize.
-     * @return Returns the sanitize string
+     * @return Returns the sanitized string
      */
     std::string sanitize_fact_name(std::string str);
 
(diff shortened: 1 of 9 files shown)


lib/inc/facter/util/string.hpp
```diff
@@ -69,4 +69,14 @@ namespace facter { namespace util {
...
     boost::optional<int> maybe_stoi(std::string const& str);
 
+    /**
+     * Converts the given string to a sanityzed string using the same algorithm
+     * as in puppet:
+https://github.com/puppetlabs/puppet/blob/85bd9ed07e3e6c956f1972624a50be914292e1ab/lib/puppet/pops/lookup/sub_lookup.rb#L6-L44
+     * @param str The string to sanityze.
+     * @return Returns the sanityze string
+     */
+    std::string sanityze_name(std::string str);
```
GabrielNagy, line 79: nit: spelling should be `sanitize`

lib/tests/util/string.cc
```diff
@@ -287,3 +287,27 @@ SCENARIO("converting strings to integers") {
...
+    GIVEN("quotes protected facts in string") {
+        THEN("should remove quotes") {
+          REQUIRE(sanityze_name("\"a.b\".c") == "a.b.c");
+        }
+    }
+    GIVEN("ticks protected facts in string") {
+        THEN("should remove ticks") {
+          REQUIRE(sanityze_name("a.'b.c'") == "a.b.c");
+        }
+    }
```
joshcooper, line 312: might be good to add cases for `""`, `"a.'b.c"`, `"a.\"b.c"`

joshcooper:
I think we can close this?

PR #2338, opened 2021-04-01
(FACT-2998) Fix CLI fact matching

GabrielNagy:
See the individual commit messages for details.

### DIFF
--- a/lib/facter/models/fact_collection.rb
+++ b/lib/facter/models/fact_collection.rb
@@ -21,13 +21,20 @@ module Facter
       self
     end
 
+    def dig_fact(user_query)
+      split_user_query = Facter::Utils.split_user_query(user_query)
+      fact = dig(*split_user_query) || dig(user_query)
+    rescue TypeError
+      # An incorrect user query (e.g. mountpoints./.available.asd) can cause
+      # Facter to call dig on a string, which raises a type error.
+      # If this happens, we assume the query is wrong and silently continue.
+    ensure
+      @log.debug("Fact \"#{user_query}\" does not exist") unless fact
+      fact
+    end
+
     def value(user_query)
-      fetch(user_query) do
-        split_user_query = Facter::Utils.split_user_query(user_query)
-        split_user_query.reduce(self) do |memo, key|
-          memo.fetch(key.to_s)
-        end
-      end
+      dig_fact(user_query)
     end
 
     def bury(*args)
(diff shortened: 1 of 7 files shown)


GabrielNagy:
me when PR tests pass

lib/facter/models/fact_collection.rb
```diff
@@ -17,13 +21,20 @@ def build_fact_collection!(facts)
...
     def value(user_query)
-      fetch(user_query) do
-        split_user_query = Facter::Utils.split_user_query(user_query)
-        split_user_query.reduce(self) do |memo, key|
-          memo.fetch(key.to_s)
-        end
-      end
+      dig_fact(user_query)
+    rescue NoMethodError, TypeError
+      nil
```
GabrielNagy, line 37: this is not needed anymore

lib/facter/models/fact_collection.rb
```diff
@@ -17,13 +23,18 @@ def build_fact_collection!(facts)
...
     end
 
+    def dig_fact(user_query)
+      split_user_query = Facter::Utils.split_user_query(user_query)
+      dig(*split_user_query) || dig(user_query)
+    rescue NoMethodError, TypeError
+      # An incorrect user query (e.g. mountpoints./.available.asd) can cause
+      # Facter to call dig on a string, which raises a type error.
+      # If this happens, we assume the query is wrong and silently continue.
+      nil
```
ciprianbadescu, line 33: if this is not happening often, we could log a debug message with the exception message instead of silently ignoring exception

lib/facter/framework/formatters/yaml_fact_formatter.rb
```diff
@@ -15,7 +15,7 @@ def format(resolved_facts)
                            FormatterHelper.retrieve_facts_to_display_for_user_query(user_queries, resolved_facts)
                          end
 
-      facts_to_display = Psych.parse_stream(facts_to_display.to_yaml)
+      facts_to_display = Psych.parse_stream(facts_to_display.deep_to_h.to_yaml)
```
ciprianbadescu, line 18: does it helps to keep the inherited `to_yaml` or would be better to rewrite it to call `deep_to_h.to_yaml` and make `deep_to_h` private?

lib/facter/models/fact_collection.rb
GabrielNagy: updated to always log if dig result is nil, or if we get an exception

PR #2340, opened 2021-04-05
(FACT-2907): add ifa_flags to networking.bindings6

b4ldr:
(FACT-2907): add ifa_flags to networking.bindings6

This PR adds the [IFA flags][1] to the networking.bindings6 fact
This is usefull information for deciding which ip6 address would make a
more logical candidate for networking.ip6, instead of just picking
 networking.interfaces[networking.primary]['bindings6'].first

I have commented out [the other flags][2] as [only the lower 8bits are 
available in /proc/net/if_inet6][3].  Other then adding netlink to ruby
I'm not sure how to access the higher bit flags :(

[1]: https://github.com/torvalds/linux/blob/fcadab740480e0e0e9fa9bd272acd409884d431a/include/uapi/linux/if_addr.h#L44-L53
[2]: https://github.com/torvalds/linux/blob/fcadab740480e0e0e9fa9bd272acd409884d431a/include/uapi/linux/if_addr.h#L54-L57
[3]: https://unix.stackexchange.com/questions/643692/calculating-ifa-flags-larger-then-0xff-from-proc-net-if-inet6


GabrielNagy:
Hi @b4ldr, thanks for the contribution! This looks like an useful addition to the networking fact.

We're a bit busy right now with the upcoming Puppet & Facter releases, but we'll try to get this reviewed when we can.

GabrielNagy:

GabrielNagy:
jenkins please test this on linux

GabrielNagy:
jenkins please test this on linux

GabrielNagy:
jenkins please test this on linux

GabrielNagy (approved):

b4ldr:
@GabrielNagy thanks for merging however there was an old commit i had locally which never made it here.  Basicaly this batch only exposes the 8 bit flags and not the 32bit flags.  There inclusion here shouldn't cause any issue other then confusion however i they shouldn't be there so i have created a patch to comment them out #2370

I still think this PR is a useful addition but understand if you want to revert this and wait for a complete solution which works with the 32bit flags (as noted in the other pr i have not found a way forward for such a solution)

GabrielNagy:
Thanks @b4ldr, I noticed this in the commit message. This is definitely useful even without the higher bit flags, and we can circle back to this if/when a saner solution to calculate the flags comes up in Linux.

I'll make sure we document this caveat when publishing release notes for this ticket.

b4ldr:
@GabrielNagy  SGTM thanks

PR #2341, opened 2021-04-05
(FACT-2992) Add Socket and Thread details for processors fact

BobosilaVictor:
This pull request aims to add Socket (Cores/Socket) and Threads(Threads/Core) details for the processors fact on all supported platforms. 
For example, here is the old processors fact for AIX : 
```
{
  "processors": {
    "count": 32,
    "isa": "powerpc",
    "models": [
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8"
    ],
    "speed": "3.43 GHz"
  }
}
```
And the new one with cores per socket and threads per core details :
```
{
  "processors": {
    "cores": 4,
    "count": 32,
    "isa": "powerpc",
    "models": [
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8",
      "PowerPC_POWER8"
    ],
    "speed": "3.43 GHz",
    "threads": 8
  }
}

```

Notes:

- Created a new resolver for linux to get the threads per core and cores per socket, by using ` lscpu | grep -e "Core(s)" -e "Thread(s)"`, because /proc/cpuinfo does not output relevant data to calculate those facts for PowerPC and arm64 architectures


lib/facter/resolvers/processors_lscpu.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    module Linux
+      class ProcessorsLscpu < BaseResolver
+        init_resolver
+
+        ITEMS = { threads_per_core: "| grep -e 'Thread(s)'",
+                  scores_per_socket: "-e 'Core(s)'" }.freeze
```
b4ldr, line 10: i think you probably wanted `| grep` here as well

lib/facter/resolvers/processors_lscpu.rb
```diff
@@ -0,0 +1,44 @@
...
+          #:threads_per_core
+
+          private
+
+          def post_resolve(fact_name, _options)
+            @fact_list.fetch(fact_name) { read_cpuinfo(fact_name) }
+          end
+
+          def read_cpuinfo(fact_name)
+            lscpu_output = Facter::Core::Execution.execute("lscpu #{ITEMS.values.join(' ')}", logger: log)
```
ciprianbadescu, line 23: I think the initial comment suggested to have uniform values in ITEMS, something like
```
        ITEMS = { threads_per_core: "-e 'Thread(s)'",
                         cores_per_socket: "-e 'Core(s)'" }.freeze
```
and you could have here `lscpu | grep #{ITEMS.values.join(' ')}` instead `lscpu #{ITEMS.values.join(' ')}`

GabrielNagy:

lib/facter/resolvers/solaris/processors.rb
```diff
@@ -28,6 +28,17 @@ def parse_output(output)
             @fact_list[:physical_count] = output.scan(/chip_id .*/).uniq.size
             @fact_list[:speed] = output.scan(/current_clock_Hz .*/).first.gsub(/[a-zA-z\s]+/, '').to_i
             @fact_list[:models] = output.scan(/brand .*/).map { |elem| elem.gsub(/brand(\s+)/, '') }
+            calculate_threads(output)
+            calculate_cores(output)
+          end
+
+          def calculate_threads(output)
+            @fact_list[:threads_per_core] = output.scan(/module/).size / (output.scan(/core_id .*/).uniq.size -
```
GabrielNagy, line 36: Won't `output.scan(/core_id .*/)` also match `pkg_core_id`?

Also, you can use `@fact_list[:logical_count]` instead of calling `output.scan(/module/).size` again. Same for `output.scan(/chip_id .*/).uniq.size`.

Wondering if it might be better to also assign `output.scan(/core_id .*/).uniq.size` to a `@fact_list[:core_count]` variable and use it in these methods.

lib/facter/resolvers/windows/processors.rb
```diff
@@ -19,14 +19,20 @@ def post_resolve(fact_name, _options)
 
         def read_fact_from_win32_processor(fact_name)
           win = Facter::Util::Windows::Win32Ole.new
-          proc = win.exec_query('SELECT Name,Architecture,NumberOfLogicalProcessors FROM Win32_Processor')
+          query_string = 'SELECT Name,'\
+          'Architecture,'\
+          'NumberOfLogicalProcessors,'\
+          'ThreadCount,'\
+          'NumberOfCores FROM Win32_Processor'
+          proc = win.exec_query(query_string)
```
GabrielNagy, line 27: I think this query does not work anymore, I'm getting this on Windows 2012R2:

```
C:\ProgramData\PuppetLabs>facter processors
{
  cores => 0,
  count => 0,
  models => [

  ],
  physicalcount => 0,
  threads => 0
}
```

Jenkins seems to fail for the same reason

BobosilaVictor: The issue was that `ThreadCount` wasn't a valid wmic cpu query for windows2012r2, so I removed it and used `NumberOfLogicalProcessors` as the total number of threads instead, which also solves a previous issue where ThreadCount queries would return 0 or nil in windows virtual machines.

lib/facter/resolvers/solaris/processors.rb
BobosilaVictor: `output.scan(/core_id .*/)` does also match` pkg_core_id`, so i have subtracted from the amount of `core_id` matches the amount of `pkg_core_id` matches, in hopes that this solves the issue

BobosilaVictor:

lib/facter/resolvers/solaris/processors.rb
```diff
@@ -28,6 +28,13 @@ def parse_output(output)
             @fact_list[:physical_count] = output.scan(/chip_id .*/).uniq.size
             @fact_list[:speed] = output.scan(/current_clock_Hz .*/).first.gsub(/[a-zA-z\s]+/, '').to_i
             @fact_list[:models] = output.scan(/brand .*/).map { |elem| elem.gsub(/brand(\s+)/, '') }
+            calculate_threads_cores(output)
+          end
+
+          def calculate_threads_cores(output)
+            @fact_list[:threads_per_core] = @fact_list[:logical_count] / (output.scan(/core_id .*/).uniq.size -
```
GabrielNagy, line 35: @BobosilaVictor I think it would be better to be more specific with the regex expression, such as `/\Wcore_id .*/` or `/\score_id .*/` which tries to match either a non-alphanumeric character or a whitespace before `core_id`.

Also, if we need to scan for `core_id` twice, my preference would be to add it to a `@fact_list` variable since regex calls are expensive.

lib/facter/resolvers/windows/processors.rb
```diff
@@ -40,7 +45,29 @@ def iterate_proc(result)
             isa ||= find_isa(proc.Architecture)
           end
 
-          { models: models, isa: isa, logical_count: logical_count.zero? ? models.count : logical_count }
+          { models: models,
+            isa: isa,
+            logical_count: logical_processors_count(logical_count, models.count) }
+        end
+
+        def calculate_cores_threads(result, result2)
```
GabrielNagy, line 53: what do `result` and `result2` stand for? 😄 can we find more suggestive names?

lib/facter/resolvers/windows/processors.rb
```diff
@@ -40,7 +45,29 @@ def iterate_proc(result)
...
+          cores = 0
+          threads_per_core = 0
+          result.each do |proc|
+            cores = proc.NumberOfCores
+            threads_per_core = if check_hyperthreading(result2[:logical_count], cores) == true ||
```
GabrielNagy, line 58:
```suggestion
            threads_per_core = if check_hyperthreading(result2[:logical_count], cores) ||
```

spec/facter/resolvers/processors_spec.rb
```diff
@@ -6,18 +6,31 @@
...
       ['Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz', 'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz',
        'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz', 'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz']
     end
     let(:physical_processors) { 1 }
+    let(:cores_per_socket) { 1 }
+    let(:threads_per_core) { 2 }
+    let(:lscpu_output) do
+      ["| grep -e 'Thread(s) per core': 2",
+       "-e 'Cores(s) per socket': 1"].join("\n")
+    end
```
GabrielNagy, line 21: I still fail to understand how this is something that `lscpu` outputs. Output should be similar to this:

```
2.6.3 :001 > require 'facter'
 => true 
2.6.3 :002 > Facter::Core::Execution.execute("lscpu | grep -e 'Thread(s)' -e 'Core(s)'")
 => "Thread(s) per core:              2\nCore(s) per socket:              4" 
```

BobosilaVictor:

spec/facter/resolvers/processors_spec.rb
```diff
@@ -6,18 +6,31 @@
...
       ['Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz', 'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz',
        'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz', 'Intel(R) Xeon(R) CPU E5-2697 v4 @ 2.30GHz']
     end
     let(:physical_processors) { 1 }
+    let(:cores_per_socket) { 1 }
+    let(:threads_per_core) { 2 }
+    let(:lscpu_output) do
+      ["'Thread(s) per core': 2",
+       "'Cores(s) per socket': 1"].join("\n")
+    end
```
GabrielNagy, line 21: Sorry, just saw these tests. Since you're testing a different resolver here (`ProcessorsLscpu`) it would be better to have the tests in a separate file.
Also, should we change the name of the resolver to just `lscpu`, since that's what the resolver executes?

spec/facter/facts/solaris/processors/threads_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+describe Facts::Solaris::Processors::Threads do
+  describe '#call_the_resolver' do
+    subject(:fact) { Facts::Solaris::Processors::Threads.new }
+
+    let(:threads_per_core) { '4' }
```
GabrielNagy, line 7: About string vs int here, the Solaris processors resolver (from what I understand) returns int values. We should confirm that this is the case, and mock the fact tests with the same types. We should always aim to mock with "real life" values, so we don't get confused in the future if/when we return to this code.

Can you do a quick pass through all the resolvers and cross-check if the facts are tested with the correct types for `threads_per_core` and `cores_per_socket`?

spec/facter/facts/windows/processors/threads_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+describe Facts::Windows::Processors::Threads do
+  describe '#call_the_resolver' do
+    it 'returns a fact' do
+      expected_fact = instance_double(Facter::ResolvedFact, name: 'processors.threads', value: 'value')
+      allow(Facter::Resolvers::Processors).to receive(:resolve).with(:threads_per_core).and_return('value')
```
GabrielNagy, line 7: Same as my other comment about keeping mock values close to reality. I know we test the fact/resolver classes in isolation but it would be good to ensure the mocks are matching.

spec/facter/resolvers/linux/processors_lscpu_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+describe Facter::Resolvers::Linux::Processors do
```
GabrielNagy, line 3: This should be `Facter::Resolvers::Linux::Lscpu`. The filename should also be changed to `lscpu_spec.rb`.

spec/facter/resolvers/linux/processors_lscpu_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  before do
+    Facter::Resolvers::Linux::Lscpu.instance_variable_set(:@log, log_spy)
+    allow(Facter::Core::Execution)
+      .to receive(:execute)
+      .with("lscpu | grep -e 'Thread(s)' -e 'Core(s)'", logger: log_spy)
+      .and_return(lscpu_output)
+  end
+
+  after do
+    Facter::Resolvers::Linux::Processors.invalidate_cache
```
GabrielNagy, line 13: This should be `Facter::Resolvers::Linux::Lscpu.invalidate_cache`

BobosilaVictor:

BobosilaVictor:

BobosilaVictor:

GabrielNagy (approved):

PR #2350, opened 2021-04-09
(FACT-2998) Fix filter tokens and fact matching

GabrielNagy:
This reverts the behavior introduced in FACT-2950, which allowed core facts to be fully ovewritten, which broke filter_tokens for top-level facts. Due to the addition of the `force-dot-resolution` setting, we need to figure out a better way to implement this.

In addition, to avoid issues with custom facts that can have names with repeating tokens (such as: `a.b.c.b`), change how filter tokens are constructed. Previously this was done by subtracting two arrays which could remove more elements than intended.

Do the following checks when matching searched facts to resolved facts:
- if the queried fact has filter tokens, and the resolved fact is not a array or hash, do not match the facts
- if the value of the resolved fact is a hash or array, attempt to dig through it using the filter tokens of the searched fact; if not found,   do not match the facts
- if the names of the searched fact and the resolved fact are not identical, do not match the facts

### DIFF
--- a/lib/facter/framework/core/fact_augmenter.rb
+++ b/lib/facter/framework/core/fact_augmenter.rb
@@ -13,7 +13,6 @@ module Facter
       augumented_resolved_facts
     end
 
-    # rubocop:todo Metrics/AbcSize
     private_class_method def self.get_resolved_facts_for_searched_fact(searched_fact, resolved_facts)
       if searched_fact.name.include?('.*')
         resolved_facts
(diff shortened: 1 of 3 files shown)


ciprianbadescu (approved):

GabrielNagy:
Note: in the future we should consolidate this inside https://github.com/puppetlabs/facter/blob/main/lib/facter/framework/core/fact_filter.rb#L12 which does part of the filtering

luchihoratiu (approved):

PR #2354, opened 2021-04-15
(FACT-3017) Update product check for Big Sur

BobosilaVictor:


acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -355,7 +355,7 @@ def osx_expected_facts(agent)
             'os.hardware'              => 'x86_64',
             'os.name'                  => 'Darwin',
             'os.macosx.build'          => /\d+[A-Z]\d{2,4}\w?/,
-            'os.macosx.product'        => 'Mac OS X',
+            'os.macosx.product'        => agent['platform'].match(/osx-11-/) ? 'macOS' : 'Mac OS X',
```
gimmyxd, line 358: Usually, `match` is used when you need the matched data. Since you just want to check the platform, `agent['platform'] =~ /osx-11-/ is the better option.

gimmyxd:
@BobosilaVictor this will also be needed for facter#main as we do not mergeup facter#3.x in main

gimmyxd (approved):

PR #2355, opened 2021-04-15
(FACT-3021) Support 20-octet InfiniBand MAC addresses

ananace:
I'm co-opting the old issue from Facter 3 for this, as the functionality seems to have been lost in the move to Facter 4

### DIFF
--- a/lib/facter/resolvers/networking.rb
+++ b/lib/facter/resolvers/networking.rb
@@ -63,7 +63,8 @@ module Facter
         end
 
         def extract_mac(raw_data, parsed_interface_data)
-          mac = raw_data.match(/(?:ether|lladdr)\s+(\w?\w:\w?\w:\w?\w:\w?\w:\w?\w:\w?\w)/)&.captures&.first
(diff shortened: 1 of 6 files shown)


GabrielNagy:
Hi @ananace, thanks for the contribution! And for improving the existing regex! 😄 

I took the liberty to clone the old ticket, as we'd want to have your contribution recognized in our release notes. The new ticket is [FACT-3021](https://tickets.puppetlabs.com/browse/FACT-3021) and specifically targets Facter 4.

All we need from you is to amend your commit message with the new ticket ID, so our automation includes it for the upcoming release.

Thanks again for taking the time to contribute!

GabrielNagy (commented):
Looks good! I had a question before merging: I'm curious about the OS you encountered this on. Linux OSes should use the networking resolver from `lib/facter/resolvers/linux/networking.rb` which has a different way of obtaining the mac address (by calling `Socket.getifaddrs`).

This resolver is only used on macOS and FreeBSD.

ananace:
That is a good point actually, I should patch that part as well. I'll throw up a new version of the patch momentarily.

GabrielNagy (approved):

PR #2357, opened 2021-04-16
(FACT-3023) xen command: avoid xen-toolstack warning

lollipopman:
xen-toolstack is deprecated and so on Puppet runs in Debian buster you
receive this nettlesome message:

  warning: something called deprecated script /usr/lib/xen-common/bin/xen-toolstack

Prior to this change we used xen-toolstack if it is present, but that is
only necessary if more than one stack is installed. Instead check if we
have multiple tool stacks and if we do use xen-toolstack if it is
present.

### DIFF
--- a/lib/facter/resolvers/xen.rb
+++ b/lib/facter/resolvers/xen.rb
@@ -56,7 +56,12 @@ module Facter
         end
 
         def find_command
-          return XEN_TOOLSTACK if File.exist?(XEN_TOOLSTACK)
+          num_stacks = 0
+          XEN_COMMANDS.each do |command|
+            num_stacks += 1 if File.exist?(command)
+          end
+
+          return XEN_TOOLSTACK if num_stacks > 1 && File.exist?(XEN_TOOLSTACK)
 
           XEN_COMMANDS.each { |command| return command if File.exist?(command) }
         end
--- a/lib/schema/facter.yaml
+++ b/lib/schema/facter.yaml
@@ -1894,7 +1894,7 @@ xen:
     type: map
     description: Return metadata for the Xen hypervisor.
     resolution: |
-        POSIX platforms: use `/usr/lib/xen-common/bin/xen-toolstack` to locate xen admin commands if available, otherwise fallback to `/usr/sbin/xl` or `/usr/sbin/xm`. Use the found command to execute the `list` query.
+        POSIX platforms: if `/usr/sbin/xl` and `/usr/sbin/xm` is installed use `/usr/lib/xen-common/bin/xen-toolstack` to choose, otherwise try `/usr/sbin/xl` then `/usr/sbin/xm` in order. Use the found command to execute the `list` query.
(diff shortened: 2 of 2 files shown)


GabrielNagy:
Hi @lollipopman, thanks for the contribution! :tada:

I have a couple of questions, and I don't have much experience with Xen (and no hypervisor VM on hand), so please bear with me:
- Is this warning shown on stdout/stderr when executing Facter without `--debug`? Normally, the execution API should redirect all external command output to debug.
- I was able to find some context on this -- a [thread](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=912381#39) on bugs.debian.org which mentions that:
> The xm toolstack is long gone now, and xl is the only choice.

Initially I wanted to ask if there was an accepted way of getting the toolstack in case of multiple stacks, but from what I understand there's only `xl` now?

It would be good to update https://github.com/puppetlabs/facter/blob/main/lib/schema/facter.yaml#L1897 to reflect the new behavior.

Also, I opened [FACT-3023](https://tickets.puppetlabs.com/browse/FACT-3023) to track the work on this. Can you amend your commit title to include the ticket ID instead of `maint`?

Thanks again for your contribution!

lollipopman:
> * Is this warning shown on stdout/stderr when executing Facter without `--debug`? Normally, the execution API should redirect all external command output to debug.

The output goes to stderr, the full text of the script is:

```
#!/bin/sh
set -e
echo >&2 "warning: something called deprecated script $0"
exec xl "$@"
```

> * I was able to find some context on this -- a [thread](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=912381#39) on bugs.debian.org which mentions that:
> 
> > The xm toolstack is long gone now, and xl is the only choice.

This is mostly true, but it was still available in Debian Jessie, which has long term support until June of 2022, [[1]].

> Initially I wanted to ask if there was an accepted way of getting the toolstack in case of multiple stacks, but from what I understand there's only `xl` now?

That is true now, but I assumed we wanted to be backward compatible, what are the compatibility guarantees for facter?

> It would be good to update https://github.com/puppetlabs/facter/blob/main/lib/schema/facter.yaml#L1897 to reflect the new behavior.

Will do, once we agree upon the behavior

> Also, I opened [FACT-3023](https://tickets.puppetlabs.com/browse/FACT-3023) to track the work on this. Can you amend your commit title to include the ticket ID instead of `maint`?

done!

> Thanks again for your contribution!

Thank you!

[1]: https://en.wikipedia.org/wiki/Debian_version_history#Release_table

GabrielNagy:
Yep @lollipopman I agree we should keep this backwards-compatible. The current approach of executing xen-toolstack only if multiple stacks are available seems fine to me.

lollipopman:
thanks @GabrielNagy updated with your feedback, let me know if you want any other changes!

GabrielNagy (approved):

PR #2358, opened 2021-04-19
(FACT-2914) Fix output for major version fact on Big Sur

BobosilaVictor:
Starting with macOS 11, Apple is incrementing the second component of the release number for point releases (e.g., 11.0, 11.1, 11.2). It's assumed that going forward a new major version would be 12, 13, etc.
We fix this by changing the output of the major version fact depending on the macOS version. 

Example output for macOS 11 :
```
{
  full => "11.2.1",
  major => "11",
  minor => "2.1"
}
```

### DIFF
--- a/lib/facter/facts/macosx/os/macosx/version.rb
+++ b/lib/facter/facts/macosx/os/macosx/version.rb
@@ -10,13 +10,21 @@ module Facts
 
           def call_the_resolver
             fact_value = Facter::Resolvers::SwVers.resolve(:productversion)
-            versions = fact_value.split('.')
-            ver = { 'full' => fact_value, 'major' => "#{versions[0]}.#{versions[1]}", 'minor' => versions[-1] }
+            ver = version_hash(fact_value)
 
             [Facter::ResolvedFact.new(FACT_NAME, ver),
              Facter::ResolvedFact.new(ALIASES[0], fact_value, :legacy),
              Facter::ResolvedFact.new(ALIASES[1], ver['major'], :legacy),
              Facter::ResolvedFact.new(ALIASES[2], ver['minor'], :legacy)]
+          end
+
+          def version_hash(fact_value)
+            versions = fact_value.split('.')
+            if versions[0] == '10'
+              { 'full' => fact_value, 'major' => "#{versions[0]}.#{versions[1]}", 'minor' => versions[-1] }
+            else
+              { 'full' => fact_value, 'major' => versions[0], 'minor' => "#{versions[1]}.#{versions[-1]}" }
+            end
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/facter/facts/macosx/os/macosx/version.rb
```diff
@@ -11,13 +11,21 @@ class Version
           def call_the_resolver
             fact_value = Facter::Resolvers::SwVers.resolve(:productversion)
             versions = fact_value.split('.')
```
ciprianbadescu, line 13: you could also move this line into `check_os_version` method

lib/facter/facts/macosx/os/macosx/version.rb
```diff
@@ -10,14 +10,22 @@ class Version
...
 
             [Facter::ResolvedFact.new(FACT_NAME, ver),
              Facter::ResolvedFact.new(ALIASES[0], fact_value, :legacy),
              Facter::ResolvedFact.new(ALIASES[1], ver['major'], :legacy),
              Facter::ResolvedFact.new(ALIASES[2], ver['minor'], :legacy)]
           end
+
+          def check_os_version(fact_value)
+            versions = fact_value.split('.')
+            if versions[0] < '11'
```
gimmyxd, line 23: might be better to use  `versions[0] == '10'`  instead of  `if strings with <`

lib/facter/facts/macosx/os/macosx/version.rb
```diff
@@ -10,14 +10,22 @@ class Version
...
+              { 'full' => fact_value, 'major' => "#{versions[0]}.#{versions[1]}", 'minor' => versions[-1] }
+            else
+              { 'full' => fact_value, 'major' => versions[0], 'minor' => "#{versions[-2]}.#{versions[-1]}" }
```
gimmyxd, line 26: To keep similarity with the first
`
{ 'full' => fact_value, 'major' => versions[0], 'minor' => "#{versions[1]}.#{versions[-1]}" }
`

lib/facter/facts/macosx/os/macosx/version.rb
```diff
@@ -10,14 +10,22 @@ class Version
...
-            ver = { 'full' => fact_value, 'major' => "#{versions[0]}.#{versions[1]}", 'minor' => versions[-1] }
+            ver = check_os_version(fact_value)
 
             [Facter::ResolvedFact.new(FACT_NAME, ver),
              Facter::ResolvedFact.new(ALIASES[0], fact_value, :legacy),
              Facter::ResolvedFact.new(ALIASES[1], ver['major'], :legacy),
              Facter::ResolvedFact.new(ALIASES[2], ver['minor'], :legacy)]
           end
+
+          def check_os_version(fact_value)
```
gimmyxd, line 21: `versions_hash` would be a better name for this method

gimmyxd (commented):
Would be also nice to add a test for this fact that checks it works with both 10 and 11 versions of mac OS

gimmyxd (approved):

GabrielNagy (approved):

PR #2359, opened 2021-04-19
(FACT-2914) Fix OS version matching on Big Sur

GabrielNagy:
We need to also take into account that the major version is now 11 instead of 11.2. Update the tests to reflect this behavior.

Also make the tests more future-proof by reversing the checks so (presumably) we won't have to update them on every macOS release.

Follow up to https://github.com/puppetlabs/facter/pull/2358

### DIFF
--- a/acceptance/lib/facter/acceptance/base_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/base_fact_utils.rb
@@ -219,12 +219,8 @@ module Facter
 
       # osx
       def osx_expected_facts(agent)
-        version = agent['platform'].match(/osx-(10\.\d+)/)
-        os_version = if version.nil?
-                       /\d+\.\d+/
-                     else
-                       /#{version[1]}/
-                     end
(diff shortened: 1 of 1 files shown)


GabrielNagy:
Tested this with:

```
export SHA=973611680f7a77641f32faf28c27bc98dc80d10d                                                                                                           
export BP_ROOT=/home/gabi/repo/beaker-puppet                                                                                                                  
bundle exec beaker init --hosts osx11-64a{hypervisor=abs}-osx1015-64a{hypervisor=abs} --options-file config/aio/options.rb                                    
bundle exec beaker provision                                                                                                                                  
bundle exec beaker exec --pre-suite $BP_ROOT/setup/common/000-delete-puppet-when-none.rb,$BP_ROOT/setup/aio/010_Install_Puppet_Agent.rb
bundle exec beaker exec tests/facts/os_processors_and_kernel.rb                                                                                               
```

mihaibuzgau (approved):

PR #2371, opened 2021-04-26
(FACT-2928) Better error for mountpoints stats

Dorin-Pleava:
When Sys::Filesystem.stat(mountopint_path) fails, it will throw
`statvfs() function failed: No such file or directory`, but will not
show what file is missing.

Steps I used to reproduce a similar error as in FACT-2928:

`$ yum install bind-chroot -y`
...
`$ systemctl start named-chroot`

Place a binding to stop execution above https://github.com/puppetlabs/facter/blob/main/lib/facter/util/resolvers/filesystem_helper.rb#L20

`$ puppet facts show`

  `require 'sys/filesystem'
=> binding.pry
   Sys::Filesystem.stat(path)`

When binding.pry is hit, run `$ systemctl stop named-chroot`, then
continue execution.

### DIFF
--- a/lib/facter/resolvers/mountpoints.rb
+++ b/lib/facter/resolvers/mountpoints.rb
@@ -67,22 +67,30 @@ module Facter
         end
 
         def get_mount_sizes(mount)
-          stats = Facter::Util::Resolvers::FilesystemHelper.read_mountpoint_stats(mount[:path])
+          begin
+            stats = Facter::Util::Resolvers::FilesystemHelper.read_mountpoint_stats(mount[:path])
+            get_bytes_data(mount, stats)
+          rescue Sys::Filesystem::Error => e
+            @log.debug("Could not get stats for mountpoint #{mount[:path]}, got #{e}")
+            mount[:size_bytes] = mount[:available_bytes] = mount[:used_bytes] = 0
+          end
 
-          get_bytes_data(mount, stats)
-
-          total_bytes = mount[:used_bytes] + mount[:available_bytes]
-          mount[:capacity] = Facter::Util::Resolvers::FilesystemHelper.compute_capacity(mount[:used_bytes], total_bytes)
-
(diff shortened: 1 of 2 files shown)


lib/facter/resolvers/mountpoints.rb
```diff
@@ -67,23 +67,31 @@ def get_mount_data(file_system, mount)
         end
 
         def get_mount_sizes(mount)
-          stats = Facter::Util::Resolvers::FilesystemHelper.read_mountpoint_stats(mount[:path])
+          begin
+            stats = Facter::Util::Resolvers::FilesystemHelper.read_mountpoint_stats(mount[:path])
+            get_bytes_data(mount, stats)
+          rescue Sys::Filesystem::Error => e
+            @log.debug("Could not get stats for mountpoint #{mount[:path]}, got #{e}")
+            mount[:size_bytes] = mount[:available_bytes] = mount[:used_bytes] = 0
```
gimmyxd, line 75: Does it help if we add this facts with value `0`? since this is not the real information.
What happens if we do not set a value for this fields?

ciprianbadescu: I think it is for consistency, we have similar behavior in osx resolver: https://github.com/puppetlabs/facter/blob/64502d3561da0db674ccfdb9f0d9a3f3d70cde07/lib/facter/resolvers/macosx/mountpoints.rb#L43

Dorin-Pleava: It will fail when trying to calculate total_bytes, https://github.com/puppetlabs/facter/blob/main/lib/facter/resolvers/mountpoints.rb#L74 and the fact will return nil.
I added the 0 values as the mountpoint fact already had some real data in it, like mount[:device], mount[:filesystem], mount[:path] and mount[:options]

gimmyxd (approved):

PR #2372, opened 2021-04-26
(FACT-2295) Replace check.sh script and fix failing tests

luchihoratiu:
### Remove env dependant integration tests
Tests checking the Facter CLI and API output were wrongfully assuming that IPv6 was always set and its related facts always existed. This was causing unjustified test failures when IPv6 was not set up.

This commit removes said tests since they were not vital for testing Facter output in the given scenarios.

------

### Fix order dependent test
Before this commit, running rspec with random order could give unexpected failures with certain seeds in os release tests. This was due to the cached resolver being invalidated only after tests instead of before also.

This commit adds the necessary cache invalidation and allows tests to pass without dependency to order execution.

------

### Update workflow to run rspec tests in random order
This commit allows the rspec tests to run in random order to better avoid future order dependency over tests. 

------

### Add rubocop as rake task
This commit adds the `rubocop` as rake task in preparation for the `check` rake task.

------

### Replace `check.sh` with `check` rake task
This commit replaces the `check.sh` script with equivalent rake task.


spec_integration/facter_to_hash_spec.rb
```diff
@@ -48,12 +48,6 @@
         expect(out).to match(/[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}/)
       end
 
-      it 'returns ipaddress6' do
```
mihaibuzgau, line 51: would it be possible to run this integration test only on OSes that have ipv6 addresses?

luchihoratiu: Removed the commit and created https://tickets.puppetlabs.com/browse/FACT-3032.

ciprianbadescu (approved):

BobosilaVictor (approved):

PR #2378, opened 2021-05-06
(FACT-3034) Use require_relative

joshcooper:
Updates facter to use `require_relative` instead of `require` to reduce the amount of file IO needed to load facter. See also https://github.com/puppetlabs/puppet/pull/8595 

Note this doesn't compute the shortest relative path to the file-to-be-required. So if the relative path includes `..` then we traverse to the top and back down. This produces longer paths, but I find it to be clearer, since we don't omit directories. I may be the only one that thinks that?

Note the  `..` doesn't result in extra file IO, because ruby realpath's the file to be included. For example, once we find `facter/a.rb` then it uses `openat` passing the absolute path to `b.rb`:

```
stat("/home/josh/work/facter/lib/facter/a.rb", {st_mode=S_IFREG|0664, st_size=48, ...}) = 0
openat(AT_FDCWD, "/home/josh/work/facter/lib/facter/b.rb", O_RDONLY|O_NONBLOCK|O_CLOEXEC) = 7
```


joshcooper:
Hmm this doesn't seem to affect facter in any significant way on redhat or windows.

Before

```
$ strace -fc -e trace=file bundle exec facter > /dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 72.74    0.103014           2     35577     28142 openat
 13.95    0.019760           1     11503      4089 stat
  7.94    0.011241           1      9741         5 lstat
  2.62    0.003704           2      1310       351 access
  1.51    0.002132           7       298       116 execve
  0.39    0.000551           2       235       116 readlink
  0.33    0.000474           2       167        50 faccessat
  0.24    0.000340           4        72           readlinkat
  0.14    0.000198           2        87         8 statfs
  0.11    0.000149           4        36        10 newfstatat
  0.02    0.000025           2        11        11 mkdir
  0.02    0.000024           2        10           chdir
  0.01    0.000017           1        13           getcwd
  0.00    0.000000           0         2           unlink
------ ----------- ----------- --------- --------- ----------------
100.00    0.141629                 59062     32898 total
```


After

```
$ strace -fc -e trace=file bundle exec facter > /dev/null
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 70.64    0.095940           2     35576     28142 openat
 13.65    0.018534           1     11503      4089 stat
 10.78    0.014641           1      9741         5 lstat
  2.16    0.002929           9       298       116 execve
  1.90    0.002583           1      1310       351 access
  0.24    0.000323           4        72           readlinkat
  0.22    0.000300           3        87         8 statfs
  0.19    0.000263           1       167        50 faccessat
  0.10    0.000140           0       235       116 readlink
  0.07    0.000101           2        36        10 newfstatat
  0.01    0.000020           1        11        11 mkdir
  0.01    0.000018           1        10           chdir
  0.01    0.000014           1        13           getcwd
  0.01    0.000010           5         2           unlink
------ ----------- ----------- --------- --------- ----------------
100.00    0.135816                 59061     32898 total
```

I think it's because facter predominately calls `require` with an absolute path based on the location of the `file_loader` already, eg

```
require '/home/josh/work/facter/lib/facter/framework/core/../../resolvers/linux/docker_uptime.rb'
```

I'm going to close this for now.

PR #2387, opened 2021-05-20
(FACT-3042) retrieve IMDSv2 by default

gimmyxd:
This commit changes the way Facter retrieves
ec2_metadata by favoring IMDSv2 over IMDSv1.

This is achieved by trying to retrieve an AWS token
and add it to the `X-aws-ec2-metadata-token`.
If the token cannot be retrieved, IMDSv1 is used.

### DIFF
--- a/spec/facter/util/resolvers/aws_token_spec.rb
+++ b/spec/facter/util/resolvers/aws_token_spec.rb
@@ -7,6 +7,10 @@ describe 'Facter::Util::Resolvers::Http' do
     Facter::Util::Resolvers::AwsToken.reset
   end
 
+  after do
+    Facter::Util::Resolvers::AwsToken.reset
+  end
+
   describe '#get' do
     it 'calls Facter::Util::Resolvers::Http.put_request' do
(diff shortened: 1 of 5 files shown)


spec/facter/util/resolvers/aws_token_spec.rb
```diff
@@ -24,7 +24,7 @@
     it 'makes a second request if token is expired' do
       allow(Facter::Util::Resolvers::Http).to receive(:put_request).and_return('token')
       aws_token.get(1)
-      sleep 1
+      Timecop.travel(Time.now + 2)
```
gimmyxd, line 27: makes unit tests faster with 1 second 🚀

GabrielNagy (approved):
Tested this on an EC2 VM and it seems to work as expected. I wasn't able to modify the instance metadata and force IMDSv1/IMDSv2 (don't think I have enough rights) but the logic works.

Maybe it would be helpful to have a `--http_debug` option like Puppet has, and show all HTTP requests/responses. In a separate ticket, obviously.

PR #2392, opened 2021-06-04
(FACT-3046) Concatenate facter 4 flags

Dorin-Pleava:
This PR moves `--puppet` and `-p` subcommand to Thor class_option so it
can be concatenated with other class_option flags.

Now -p flag can be concatenated with -j(json), -y(yaml), -d(debug) and
-t(timing) flags.

### DIFF
--- a/spec_integration/facter_to_hash_spec.rb
+++ b/spec_integration/facter_to_hash_spec.rb
@@ -35,6 +35,46 @@ describe Facter do
       end
     end
 
+    context 'when concatenating short flags' do
+      it 'returns no error' do
+        _, err = IntegrationHelper.exec_facter('-pjdt')
+
+        expect(err).not_to match(/ERROR Facter::OptionsValidator - unrecognised option/)
+      end
+
+      it 'returns error' do
+        _, err = IntegrationHelper.exec_facter('-pjdtz')
+
+        expect(err).to match(/ERROR Facter::OptionsValidator - .*unrecognised option '-z'/)
+      end
+
+      context 'when concatenating JSON and DEBUG flags' do
+        out, err = IntegrationHelper.exec_facter('-jd')
+
+        it 'outputs in valid JSON format' do
+          expect do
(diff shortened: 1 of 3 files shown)


acceptance/tests/concatenated_option_flags.rb
```diff
@@ -0,0 +1,80 @@
+# This test is intended to ensure that command-line short options work
+# properly when concatenated.
+test_name "Facter short flags can be concatenated" do
```
gimmyxd, line 3: I would see this as an integration test in `spec_integration` where you could use `IntegrationHelper.exec_facter('-jp')` and expect not to output `/ERROR Facter::OptionsValidator - unrecognised option/`

gimmyxd: And something like `IntegrationHelper.exec_facter('-jpv')` to output  `/ERROR Facter::OptionsValidator - unrecognised option/`

luchihoratiu (approved):

spec_integration/facter_to_hash_spec.rb
```diff
@@ -35,6 +35,46 @@
...
+        _, err = IntegrationHelper.exec_facter('-pjdt')
+
+        expect(err).not_to match(/ERROR Facter::OptionsValidator - unrecognised option/)
+      end
+
+      it 'returns error' do
+        _, err = IntegrationHelper.exec_facter('-pjdtz')
+
+        expect(err).to match(/ERROR Facter::OptionsValidator - .*unrecognised option '-z'/)
+      end
```
gimmyxd, line 49: might be good to split this into context, to be clear, eg. 
```ruby
context 'when concatenating short flags' do
  context 'when using valid flags' do
    it 'returns no error' do
  end

  context 'when using flags and subcommands' do
    it 'returns validation error' do
  end
end
```

gimmyxd (approved):

PR #2394, opened 2021-06-08
(FACT-3050) Use empty header in Ec2 resolver when token is nil

Dorin-Pleava:
Before this commit, rspec tests for the Ec2 resolver were failing on
some GitHub Actions Ubuntu machines with ruby 2.3.8p459.

This fix checks if the token is nil before sending the HTTP request that
gathers the needed data. Based on this token the header will be filled
accordingly.

### DIFF
--- a/spec/facter/resolvers/ec2_spec.rb
+++ b/spec/facter/resolvers/ec2_spec.rb
@@ -118,7 +118,8 @@ describe Facter::Resolvers::Ec2 do
     end
 
     let(:token) { 'v2_token' }
-    let(:headers) { { 'X-Aws-Ec2-Metadata-Token' => token } }
+    let(:headers) { { 'X-aws-ec2-metadata-token' => token } }
+    let(:expected_header) { headers }
 
     it_behaves_like 'ec2'
   end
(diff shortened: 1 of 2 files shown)


spec/facter/resolvers/ec2_spec.rb
```diff
@@ -94,6 +94,12 @@
           }
         )
       end
+
+      it "uses the expected header" do
+        allow(Facter::Resolvers::Ec2).to receive(:v2_token).and_return(token)
+        expect(Facter::Util::Resolvers::Http).to receive(:get_request).with(metadata_uri, expected_header, any_args)
```
gimmyxd, line 100: There should be a spy on `Facter::Util::Resolvers::Http` and expect that spy to `have_received(:get_request)`

luchihoratiu (approved):

spec/facter/resolvers/ec2_spec.rb
```diff
@@ -130,7 +131,17 @@
...
+    let(:expected_header) { {} }
 
     it_behaves_like 'ec2'
   end
+
+  it 'does not add headers if token is nil' do
+    allow(Facter::Resolvers::Ec2).to receive(:v2_token).and_return(nil)
+
+    stub_request(:any, metadata_uri).with { |request| !request.headers.key?('X-aws-ec2-metadata-token') }
+    stub_request(:any, userdata_uri).with { |request| !request.headers.key?('X-aws-ec2-metadata-token') }
```
gimmyxd, line 143: stub the specific request type(`:get`) to reflect the use case of this scenarion

spec/facter/resolvers/ec2_spec.rb
```diff
@@ -118,7 +118,8 @@
     end
 
     let(:token) { 'v2_token' }
-    let(:headers) { { 'X-Aws-Ec2-Metadata-Token' => token } }
+    let(:headers) { { 'X-aws-ec2-metadata-token' => token } }
+    let(:expected_header) { headers }
```
gimmyxd, line 122: I think this is not used anymore

PR #2401, opened 2021-06-18
(FACT-3045) performance improvement on many network interfaces

ciprianbadescu:
O a system with ~1500 network interfaces:
```
for i in $(seq 1 380); do ip tuntap add name dummy_tun$i mode tun; done
for i in $(seq 1 1274); do ip link add name dummy_br$i type bridge; done
```
we got the following times before and after the changes:
```
[root@dreary-boogie facter]# time bundle exec facter > initial

real	0m9.853s
user	0m3.282s
sys	0m7.078s
[root@dreary-boogie facter]# time bundle exec facter > dhcp_command_check

real	0m5.521s
user	0m2.138s
sys	0m3.802s
[root@dreary-boogie facter]# time bundle exec facter > ip_link_show_update

real	0m5.236s
user	0m1.823s
sys	0m3.407s
[root@dreary-boogie facter]# time bundle exec facter > both_changes

real	0m1.007s
user	0m0.658s
sys	0m0.276s
```


lib/facter/util/linux/socket_parser.rb
```diff
@@ -59,8 +59,23 @@ def mac_from_bonded_interface(interface_name)
...
+          def get_ip_link_show_data(interface_name)
+            @ip_link_show_data ||= read_ip_link_show_data
+            @ip_link_show_data[interface_name]
+          end
+
+          def read_ip_link_show_data
+            ip_link_show_data = {}
+            output = Facter::Core::Execution.execute('ip -o link show', logger: @log)
+            output.each_line do |line|
+              interface_name = line&.split(':')&.[](1)&.strip
```
GabrielNagy, line 75: Could be more readable like: `line.split(':')[1].strip if line`

lib/facter/util/linux/dhcp.rb
```diff
@@ -75,6 +75,11 @@ def search_internal_leases(interface_name)
           def search_with_dhcpcd_command(interface_name)
             @log.debug("Attempt to get DHCP for interface #{interface_name}, from dhcpcd command")
 
+            if @dhcpcd_command_exists.nil?
+              @dhcpcd_command_exists = !Facter::Core::Execution.execute('which dhcpcd', logger: @log).empty?
+            end
+            return unless @dhcpcd_command_exists
+
             output = Facter::Core::Execution.execute("dhcpcd -U #{interface_name}", logger: @log)
```
GabrielNagy, line 83: I know it's out of scope for this ticket (dhcpcd is not functional/present), but maybe we can optimize calls to dhcpcd as well? (In a different ticket)

```
     -U, --dumplease [interface]
             Dumps the current lease for the interface to stdout.  If no interface is given then all interfaces are dumped.  Use the -4 or -6 flags to specify an address family.  If a lease is piped in via standard input
             then that is dumped.  In this case, specifying an address family is mandatory.
```

lib/facter/util/linux/socket_parser.rb
ciprianbadescu: We need to make sure the code is not raising, maybe `line.split(':')[1]&.strip if line`?

lib/facter/util/linux/dhcp.rb
ciprianbadescu: It might not really work as documented:
```
user@machine:~ $ sudo dhcpcd -U eth0
broadcast_address='192.168.200.255'
dhcp_lease_time='259200'
dhcp_message_type='5'
dhcp_server_identifier='192.168.200.1'
domain_name_servers='192.168.200.1'
ip_address='192.168.200.3'
network_number='192.168.200.0'
routers='192.168.200.1'
subnet_cidr='24'
subnet_mask='255.255.255.0'
dhcp6_client_id='0001000125d83160dca33344bac4'
dhcp6_name_servers='fe80::1%eth0'
dhcp6_server_id='00010001dc43e795443335916115'
user@machine:~ $ sudo dhcpcd -U 
: dhcp_dump: Invalid argument

```

lib/facter/util/linux/socket_parser.rb
GabrielNagy: Good point, didn't consider that, your suggestion sounds good then 👍🏼

GabrielNagy (approved):

lib/facter/util/linux/dhcp.rb
```diff
@@ -75,6 +75,11 @@ def search_internal_leases(interface_name)
           def search_with_dhcpcd_command(interface_name)
             @log.debug("Attempt to get DHCP for interface #{interface_name}, from dhcpcd command")
 
+            if @dhcpcd_command_exists.nil?
+              @dhcpcd_command_exists = !Facter::Core::Execution.execute('which dhcpcd', logger: @log).empty?
+            end
+            return unless @dhcpcd_command_exists
```
GabrielNagy, line 81: We can also use the which method: `Facter::Core::Execution.which('dhcpcd')`

And the logic can be simplified to: 

```ruby
@dhcpcd_command ||= Facter::Core::Execution.which('dhcpcd')
return unless @dhcpcd_command
```

PR #2405, opened 2021-06-30
(FACT-3054) convert ip6 ranges into single ip as it does for ip4

rvstaveren:
Filed as https://tickets.puppetlabs.com/browse/FACT-3054

There might be some discussion on the regexp for mixed ipv4/ipv6 use, but then again this regexp would also work with invalid IPv4 addresses if presented to Facter somehow

### DIFF
--- a/lib/facter/resolvers/networking.rb
+++ b/lib/facter/resolvers/networking.rb
@@ -37,7 +37,9 @@ module Facter
 
         def clean_up_interfaces_response(response)
           # convert ip ranges into single ip. eg. 10.16.132.213 -->  10.16.132.213 is converted to 10.16.132.213
-          response.gsub!(/(\d+(\.\d+)*)\s+-->\s+\d+(\.\d+)*/, '\\1')
+          # convert ip6 ranges into single ip. eg. 2001:db8:cafe::132:213 -->
+          # 2001:db8:cafe::132:213 is converted to 2001:db8:cafe::132:213
+          response.gsub!(/([\da-fA-F]+([\.:]+[\da-fA-F]+)*)\s+-->\s+[\da-fA-F]+([\.:]+[\da-fA-F]+)*/, '\\1')
         end
 
         def parse_interfaces_response(response)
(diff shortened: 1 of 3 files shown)


GabrielNagy:
Thanks for the contribution @rvstaveren,

A few things...

Can you prefix your commit message with the ticket ID so it looks similar to `(FACT-3054) commit title`? Also, it's a good practice to keep a short commit title and put any details in the commit body. Something like:

```
(FACT-3054) Convert IPv6 ranges into single IP

Convert IPv6 ranges into single IP. eg.
2001:db8:cafe::132:213 -->  200cd1d9b91:db8:cafe::132:213
is converted to 2001:db8:cafe::132:213
```
 
In addition, it would be good to have a test for this behavior. You could add a sample network interface to https://github.com/puppetlabs/facter/blob/main/spec/fixtures/ifconfig_mac and assert the interface is created correctly here https://github.com/puppetlabs/facter/blob/main/spec/facter/resolvers/networking_spec.rb.

Lastly, in order for us to merge your contribution you have to sign the Contributor License Agreement as outlined in the comment above.

lib/facter/resolvers/networking.rb
```diff
@@ -37,7 +37,8 @@ def interfaces_data
 
         def clean_up_interfaces_response(response)
           # convert ip ranges into single ip. eg. 10.16.132.213 -->  10.16.132.213 is converted to 10.16.132.213
-          response.gsub!(/(\d+(\.\d+)*)\s+-->\s+\d+(\.\d+)*/, '\\1')
+          # convert ip6 ranges into single ip. eg. 2001:db8:cafe::132:213 -->  2001:db8:cafe::132:213 is converted to 2001:db8:cafe::132:213
```
GabrielNagy, line 40: rubocop fails on this comment line due to its length, you could wrap it:

```
# convert ip6 ranges into single ip. eg.
# 2001:db8:cafe::132:213 -->  2001:db8:cafe::132:213
# is converted to 2001:db8:cafe::132:213
```

rvstaveren:
Thanks for the feedback, I've also added the interface output to the fixture

GabrielNagy:
Thanks @rvstaveren, I updated the spec tests to assert the data from the new fixture.

GabrielNagy (approved):

PR #2406, opened 2021-07-02
(maint) fix gemspec stack level too deep

gimmyxd:
Because of an r10k issue(https://github.com/puppetlabs/r10k/issues/760#issuecomment-325825267)
we cannot use any require statemens inside `.gemspec`
files.

This commit updates `facter.gemspec` and `facter-ng.gemspec`
to build the file paths without using `pathname`

### DIFF
--- a/agent/facter-ng.gemspec
+++ b/agent/facter-ng.gemspec
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require 'pathname'
-
 lib = File.expand_path('../lib', __dir__)
 $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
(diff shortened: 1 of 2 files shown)


luchihoratiu (approved):

joshcooper:
Thanks for fixing this! FWIW, puppet and hiera set `files` to a single entry, eg https://github.com/puppetlabs/puppet/blob/99a52d47373f02a17bbbad31de66a8d7f795655e/.gemspec#L27. It kind of doesn't matter what we set it to since we use `packaging` to build the gem, not the usual `gem build` command.

gimmyxd:
> Thanks for fixing this! FWIW, puppet and hiera set `files` to a single entry, eg [puppetlabs/puppet@`99a52d4`/.gemspec#L27](https://github.com/puppetlabs/puppet/blob/99a52d47373f02a17bbbad31de66a8d7f795655e/.gemspec#L27). It kind of doesn't matter what we set it to since we use `packaging` to build the gem, not the usual `gem build` command.

Makes sense, in puppet-agent#main we build the facter gem using packaging, but for 6.x we use the gem command for facter-ng: https://github.com/puppetlabs/puppet-agent/blob/6.x/configs/components/facter-ng.rb#L35

PR #2408, opened 2021-07-05
(FACT-3053) Change license from MIT to Apache 2.0

mihaibuzgau:
Puppet is licensed as Apache 2.0. 
This PR updates the license of Facter to be on par with Puppet

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,202 @@
-MIT License
 
-Copyright (c) 2019 Puppet
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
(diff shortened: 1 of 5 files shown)


GabrielNagy:
Should also change https://github.com/puppetlabs/facter/blob/main/facter.gemspec#L15 and https://github.com/puppetlabs/facter/blob/main/agent/facter-ng.gemspec#L15

gimmyxd:
might be good to also add it to project_data: https://github.com/puppetlabs/facter/blob/main/ext/project_data.yaml

gimmyxd (approved):

PR #2410, opened 2021-07-14
(FACT-3032) skip ipv6 tests outside of CI

gimmyxd:
Because ipv6 tests usually fail when running locally,
updated the RSpec configuration to skip them by
default. These tests will be run only when setting
the environment variable `CI`.

The `skip_outside_ci` tag can be used for future tests that
need to be excluded when running locally.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,7 +48,11 @@ top of things.
 
 * Make sure you have added the necessary tests for your changes.
 * Run _all_ the tests to assure nothing else was accidentally broken.
-  * We recommend running `./check.sh` to have the same checks run as on github
+* We recommend running the follwing rake tasks:
+  * `bundle exec rake rubocop` - runs rubocop cheks
+  * `bundle exec rake spec_random` - runs unit tests
+  * `bundle exec rake spec_integration` - runs integration tests. Note that tests with the `skip_outside_ci` tag are excluded outside of CI, you can run them with `CI=true bundle exec rake spec_integration`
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -48,7 +48,9 @@ top of things.
 
 * Make sure you have added the necessary tests for your changes.
 * Run _all_ the tests to assure nothing else was accidentally broken.
-  * We recommend running `./check.sh` to have the same checks run as on github
+* We recommend running the follwing rake tasks:
+  * `bundle exec rake spec_random` - runs unit tests
+  * `bundle exe rake spec_integration` runs integration tests. Note that tests with the `skip_outside_ci` tag are excluded outside of CI, you can run them with `CI=true bundle exe rake spec_integration`
```
luchihoratiu, line 53: Maybe we should also recommend running `bundle exec rake rubocop` and `bundle exec rake check` as a 3 in 1 solution?

luchihoratiu, line 53: nit: dash before `runs` to be consistent with above.

luchihoratiu, line 53: nit: `bundle exec` instead of `bundle exe` (at beginning and end of line)

luchihoratiu (approved):
Besides small doc comments, LGTM! 👍 
Tested locally without IPv6 set and works as expected (`bundle exec rake spec_integration`, `CI=true bundle exec rake spec_integration`, `export CI=true` + `bundle exec rake spec_integration` + `bundle exec rake check`).

PR #2413, opened 2021-07-21
Remove "master" in generated docs

clairecadman:


lib/schema/facter.yaml
```diff
@@ -2057,6 +2057,6 @@ zpool_version:
 
 nim_type:
     type: string
-    description: Tells if the node is master or standalone inside an AIX Nim environment.
+    description: Tells if the node is a primary server or standalone inside an AIX Nim environment.
```
gimmyxd, line 2060: I think this is based on the upstream naming and we might not retain the same understanding if we replace `nim master` with a `primary master` https://www.ibm.com/docs/en/aix/7.2?topic=nim-configuring-master-creating-basic-installation-resources

clairecadman: @gimmyxd ah okay, can I delete this PR then?

gimmyxd: Yes, we can close this, as the values returned by the fact can be `standalone` or `master`

PR #2419, opened 2021-08-13
(FACT-3059) Fix AIX reporting odd number of arguments for Hash 

BobosilaVictor:
Previously, the aix `info_extractor` would make the assumption that keys are
separated from values with multiple spaces, but in some instances there would be
only one space.

This commit fixes this issue by searching for each known key
(from the `PROPERTIES` variable), and finds the value, then ignores the rest of
the line. The algorithm repeats for each key.

### DIFF
--- a/lib/facter/util/aix/info_extractor.rb
+++ b/lib/facter/util/aix/info_extractor.rb
@@ -6,17 +6,68 @@ module Facter
       module InfoExtractor
         MEGABYTES_EXPONENT = 1024**2
         GIGABYTES_EXPONENT = 1024**3
+        PROPERTIES = {
+          lslv: [
+            'LOGICAL VOLUME:',
+            'VOLUME GROUP:',
+            'LV IDENTIFIER:',
+            'PERMISSION:',
+            'VG STATE:',
+            'LV STATE:',
+            'TYPE:',
+            'WRITE VERIFY:',
+            'MAX LPs:',
+            'PP SIZE:',
+            'COPIES:',
+            'SCHED POLICY:',
+            'LPs:',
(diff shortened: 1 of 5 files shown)


lib/facter/util/aix/info_extractor.rb
```diff
@@ -6,17 +6,71 @@ module Aix
...
-
-            line.split(/:\s*|\s{2,}/)
+        def self.extract(content, cmd)
+          property_hash = {}
+          properties = case cmd
+                       when :lspv
+                         PROPERTIES_LSPV
+                       when :lslv
+                         PROPERTIES_LSLV
+                       end
```
GabrielNagy, line 64: can also have a single `PROPERTIES` constant that is a hash of type `PROPERTIES = { :lspv => <lspv_properties>, :lslv => <lslv_properties> }` then you can rewrite this line as `properties = PROPERTIES[cmd]`

GabrielNagy (approved):

lib/facter/util/aix/info_extractor.rb
```diff
@@ -6,17 +6,67 @@ module Aix
...
+            'FREE PPs:',
+            'HOT SPARE:',
+            'USED PPs:',
+            'MAX REQUEST:',
+            'FREE DISTRIBUTION:',
+            'USED DISTRIBUTION:',
+            'MIRROR POOL:'
+          ]
+        }.freeze
+        def self.extract(content, cmd)
```
gimmyxd, line 58: nit: needs a newline here.

gimmyxd (approved):

PR #2420, opened 2021-08-16
(FACT-3057) Downcase environment facts

GabrielNagy:
Previously, environment facts were not downcased prior to being added to the fact collection. This contradicts the Facter 3 [behavior](https://github.com/puppetlabs/facter/blob/f123fc0c371114e4c6bc61ab245cdb3a32cf5ebb/lib/src/facts/collection.cc#L257), and breaks some use cases where the fact name is upcased when added to the collection and downcased when resolved via the CLI/Ruby API.

Downcase environment facts before adding them to the collection, and downcase the user queries passed to `Facter.fact` and `Facter.value`, which matches the [behavior of Facter 3](https://github.com/puppetlabs/facter/blob/f123fc0c371114e4c6bc61ab245cdb3a32cf5ebb/lib/src/ruby/module.cc#L483).

Link to the Facter documentation in the README, where this behavior will also be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,15 @@
 [![Maintainability](https://api.codeclimate.com/v1/badges/3bd4be86f4b0b49bc0ca/maintainability)](https://codeclimate.com/github/puppetlabs/facter/maintainability)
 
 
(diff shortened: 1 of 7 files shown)


GabrielNagy:
Suggested documentation update: https://gist.github.com/GabrielNagy/20c0929e601c88915e74652ae1b44665 /cc @clairecadman

gimmyxd (approved):

Dorin-Pleava (approved):

PR #2428, opened 2021-08-29
Fix documentation for Facter::Core::Execution.execute

johanfleury:
The `time_limit` parameter [is actually called `limit`](https://github.com/puppetlabs/facter/blob/main/lib/facter/custom_facts/core/execution/base.rb#L117) and the documentation for `options` was badly formated

### DIFF
--- a/lib/facter/custom_facts/core/execution.rb
+++ b/lib/facter/custom_facts/core/execution.rb
@@ -96,22 +96,25 @@ module Facter
       end
 
       # Execute a command and return the output of that program.
+      #
       # @param command [String] Command to run
       #
       # @param options [Hash] Hash with options for the command
-      #
-      # Options accepted values :on_fail How to behave when the command could
+      # @option options [Symbol] :on_fail How to behave when the command could
       #   not be run. Specifying :raise will raise an error, anything else will
-      #   return that object on failure. Default is :raise.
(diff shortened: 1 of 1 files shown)


lib/facter/custom_facts/core/execution.rb
```diff
@@ -96,22 +96,25 @@ def exec(command)
       end
 
       # Execute a command and return the output of that program.
+      #
       # @param command [String] Command to run
       #
       # @param options [Hash] Hash with options for the command
-      #
-      # Options accepted values :on_fail How to behave when the command could
+      # @option options [Symbol] :on_fail How to behave when the command could
```
johanfleury, line 103: Can someone who knows ruby confirm that `Symbol` is actually the correct type here?

gimmyxd: https://github.com/puppetlabs/facter/blob/main/lib/facter/custom_facts/core/execution/base.rb#L114

gimmyxd:
Looks good, @johanfleury can you prefix your commit message with `(docs)`?
Something like: `(docs) Fix documentation for Facter::Core::Execution.execute`

GabrielNagy:
Thanks @johanfleury! I'll close this in favor of https://github.com/puppetlabs/facter/pull/2429 and I'll make sure you are also credited for the contribution.

PR #2429, opened 2021-09-08
(FACT-3073) Fix timeout option processing for Facter::Core::Execution.execute

smortex:
The documentation of Facter::Core::Execution.execute indicate that the
options hash accept 3 values (`expand` is not documented):
  - `on_fail`
  - `logger`
  - `time_limit`

However, when options are extracted by the extract_options method, the
key `limit` is checked instead of `time_limit` as documented.

Finally, previous versions of facter used a `timeout` parameter.

Fix these inconsistencies by using a single name `timeout` consistently.

Adjust the logic so that the documented name that was previously not
working (:time_limit) and the actual name which was working (:time) also
work the same way.

### DIFF
--- a/lib/facter/custom_facts/core/execution/base.rb
+++ b/lib/facter/custom_facts/core/execution/base.rb
@@ -117,6 +117,11 @@ module Facter
           timeout = (options[:timeout] || options[:time_limit] || options[:limit]).to_i
           timeout = timeout.positive? ? timeout : nil
 
+          extra_keys = options.keys - %i[on_fail expand logger timeout]
+          unless extra_keys.empty?
+            @log.warn("Unexpected key passed to Facter::Core::Execution.execute option: #{options.keys.join(',')}")
+          end
+
           [on_fail, expand, logger, timeout]
         end
 
--- a/lib/facter/custom_facts/core/execution/base.rb
+++ b/lib/facter/custom_facts/core/execution/base.rb
@@ -42,7 +42,7 @@ module Facter
         end
 
         def execute(command, options = {})
-          on_fail, expand, logger, time_limit = extract_options(options)
+          on_fail, expand, logger, timeout = extract_options(options)
 
           expanded_command = if !expand && builtin_command?(command) || logger
                                command
@@ -59,12 +59,12 @@ module Facter
             return on_fail
           end
 
-          out, = execute_command(expanded_command, on_fail, logger, time_limit)
+          out, = execute_command(expanded_command, on_fail, logger, timeout)
           out
         end
 
-        def execute_command(command, on_fail = nil, logger = nil, time_limit = nil)
-          time_limit ||= 300
+        def execute_command(command, on_fail = nil, logger = nil, timeout = nil)
+          timeout ||= 300
           begin
             # Set LC_ALL and LANG to force i18n to C for the duration of this exec;
             # this ensures that any code that parses the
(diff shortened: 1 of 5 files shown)


smortex:
The issue was [spotted and reported](https://github.com/puppetlabs/puppetlabs-docker/pull/770#discussion_r697941218) by @johanfleury.

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -114,7 +114,7 @@ def extract_options(options)
           on_fail = options.fetch(:on_fail, :raise)
           expand = options.fetch(:expand, true)
           logger = options[:logger]
-          time_limit = options[:limit].to_i
+          time_limit = options[:time_limit].to_i
```
smortex, line 117: Other things we can do:

1. See if `:limit` is set and if so use if as `:time_limit` and emit a warning;
2. Emit a warning for each unexpected key in the hash.

GabrielNagy:
This is a bit embarrassing... it looks like on Facter 3 the option was called `timeout`: https://github.com/puppetlabs/facter/blob/3.x/lib/src/ruby/module.cc#L959

I'm not sure how `limit` and `time_limit` came to be. I think it would be best if we updated the documentation and code to refer to `timeout` as the correct parameter, to make it easier for folks who upgrade from Facter 3.

lib/facter/custom_facts/core/execution/base.rb
GabrielNagy: 1 makes sense, with the mention that it would be better to use `timeout` (as per my comment below).

2 also makes sense, we can have a list of deprecated options (which would be only `limit` as of now) and a list of accepted options: `timeout`, `expand`, `on_fail`, `logger` (not 100% sure what this does). If the user sets a deprecated option we should accept it and warn, and if the user sets something else we should warn.

It would be good to have a ticket created for this change. Let me know if you can work on this @smortex

johanfleury:
See also #2428

smortex:
@johanfleury Ah, I missed your PR!  I am not sure about the syntax you used to alter the docstrings to document each value… and I can't find a properly generated doc… rdoc might not be the right tool?

As requested @GabrielNagy I used `timeout` everywhere, added backward compatibility with the documented key and the actual key, and added a warning in case of unexpected key.

GabrielNagy:
Thanks a lot @smortex and @johanfleury! A couple more things:
- we should file a ticket for this change; otherwise it won't get in the release notes: https://tickets.puppetlabs.com/ under the FACT project
- commit(s) should be updated with the ticket ID instead of `(maint)`
- it would be ideal to have some unit tests for the warn on unexpected key behavior (as the behavior of `timeout` is already tested in acceptance) - unfortunately it seems that there are no unit tests at all for this class; I can help with this if you want, just let me know

spec/custom_facts/core/execution_spec.rb
```diff
@@ -44,4 +44,50 @@
...
+    end
+
+    context 'with a deprecated time_limit' do
+      it 'execute the found command with a timeout' do
+        expect(impl).to receive(:execute_command).with('/under/the/honey/are/the/waffles', :raise, nil, 90)
+        execution.execute('waffles', time_limit: 90)
+      end
+
+      it 'emits a warning' do
+        expect_any_instance_of(Facter::Log).to receive(:warn)
```
GabrielNagy, line 74: I think you can do something similar to https://github.com/puppetlabs/facter/blob/main/spec/facter/resolvers/windows/memory_spec.rb to avoid `expect_any_instance_of`.

smortex: I tried, unfortunately it does not work.

I feel like the logger is initialized when the module `Facter::Core::Execution` is evaluated (`lib/facter/custom_facts/core/execution.rb` is loaded, `@@impl` is initialized with a `Facter::Core::Execution::Base` derived class, the Logger is instanciated and the module loading continue).

I was pushing this code hopping someone could help :wink:

johanfleury:
> @johanfleury Ah, I missed your PR! I am not sure about the syntax you used to alter the docstrings to document each value… and I can't find a properly generated doc… rdoc might not be the right tool?

My understanding is that they use yard.

spec/custom_facts/core/execution_spec.rb
smortex: @GabrielNagy I could workaround this issue by getting the logger differently:

```
let(:logger) { impl.instance_variable_get(:@log) }
```

Not very clean, but cleaner I guess since we have "weird" thing in only one place.

GabrielNagy (approved):
Looks good! Thanks @smortex for also fixing the documentation rendering. I've confirmed this by locally generating the docs using `yard`.

PR #2431, opened 2021-09-09
(FACT-3067) Fix resolution of custom facts that partial match legacy fact names

gimmyxd:
Before this commit, custom facts that had names which partially matched core legacy facts were being resolved as expected in CLI but not with the API. Due to the nature of how legacy facts are defined in Facter 4, with the regular expression used in `FACT_NAME`, `Facter::FactManager` was not resolving custom facts if core or external facts had any results.

This fix adds a new method to check if the user query matches the name of already resolved facts to decide whether or not we should also resolve custom facts.

### DIFF
--- a/lib/facter/framework/core/fact_manager.rb
+++ b/lib/facter/framework/core/fact_manager.rb
@@ -51,7 +51,9 @@ module Facter
       core_and_external_facts = core_or_external_fact(user_query) || []
       resolved_facts = core_and_external_facts + custom_facts
 
-      resolved_facts = all_custom_facts(user_query) if resolved_facts.empty?
+      if resolved_facts.empty? || resolved_facts.none? { |rf| rf.resolves?(user_query) }
+        resolved_facts.concat(all_custom_facts(user_query))
+      end
 
       @cache_manager.cache_facts(resolved_facts)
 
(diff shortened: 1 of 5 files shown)


Dorin-Pleava (approved):
LGTM, but not sure if there are some cases that I missed
```
# cat /Users/dorin.pleava/Documents/Work/facter/custom_fact.rb
Facter.add('network_nexthop_ip') do
  setcode do
    'This is my custom fact, not network_...'
  end
end
```
before
```
± % bx irb
2.7.3 :001 > require 'facter'
 => true
2.7.3 :002 > Facter::OptionStore.custom_dir << "/Users/dorin.pleava/Documents/Work/facter"
 => ["/Users/dorin.pleava/Documents/Work/facter"]
2.7.3 :003 > Facter.value(:network_nexthop_ip)
 => nil
2.7.3 :004 > exit
```
after

```
± % bx irb
2.7.3 :001 > require 'facter'
 => true
2.7.3 :002 > Facter::OptionStore.custom_dir << "/Users/dorin.pleava/Documents/Work/facter"
 => ["/Users/dorin.pleava/Documents/Work/facter"]
2.7.3 :003 > Facter.value(:network_nexthop_ip)
...
 => "This is my custom fact, not network_..."
...
2.7.3 :004 > exit
```

BobosilaVictor (approved):

PR #2432, opened 2021-09-14
(FACT-2955) Only use the available processors

BobosilaVictor:
Previously, after querying for CuDv, facter added all the procs found in the
said query, without checking the proc's status.  This commit adds a check for
the each proc's status, and if the status is different from available, we skip
the processor.

Tested this PR manually by changing the logic in the CuDv logic to check for `status == 2`. By doing so, if all the processors were available, `facter processors` should only output the isa, without any processors details.
```
> ./facter processors
{
  isa => "powerpc"
}
```

### DIFF
--- a/lib/src/facts/aix/processor_resolver.cc
+++ b/lib/src/facts/aix/processor_resolver.cc
@@ -56,8 +56,12 @@ namespace facter { namespace facts { namespace aix {
             string query = (boost::format("PdDvLn=%1%") % type).str();
             auto cu_dv_query = odm_class<CuDv>::open("CuDv").query(query);
             for (auto& cu_dv : cu_dv_query) {
-                LOG_DEBUG("got a processor: {1}", cu_dv.name);
-                processor_names.push_back(cu_dv.name);
+                if (cu_dv.status == 1) {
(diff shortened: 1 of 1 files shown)


luchihoratiu:
jenkins please test this on aix71-POWERfa

luchihoratiu (approved):
LGTM! 👍  Manually tested these changes on AIX 7.2 and seems to be working as expected.

lib/src/facts/aix/processor_resolver.cc
```diff
@@ -56,8 +56,12 @@ namespace facter { namespace facts { namespace aix {
             string query = (boost::format("PdDvLn=%1%") % type).str();
             auto cu_dv_query = odm_class<CuDv>::open("CuDv").query(query);
             for (auto& cu_dv : cu_dv_query) {
-                LOG_DEBUG("got a processor: {1}", cu_dv.name);
-                processor_names.push_back(cu_dv.name);
+                if (cu_dv.status == 1){
```
GabrielNagy, line 59:
```suggestion
                if (cu_dv.status == 1) {
```

This is just for my OCD, otherwise the PR looks great

GabrielNagy (approved):

PR #2433, opened 2021-09-15
(FACT-2955) Only use the available processors (4.x)

BobosilaVictor:
Previously, after querying for CuDv, facter added all the procs found in the
said query, without checking the proc's status. This commit adds a check for
the each proc's status, and if the status is different from available, we skip
the processor.


luchihoratiu (approved):
LGTM! 👍  Manually tested these changes on AIX 7.2 and seems to be working as expected.

GabrielNagy (approved):

PR #2434, opened 2021-09-18
(FACT-1644) Facter 4 does not resolve `physical_count` on AIX

BobosilaVictor:
Previously, in Facter 3, e73ea8d removed the
physical_count fact for AIX systems.  This commit adds the said fact back, and
sets the `physical_count` fact to the number of `CuDv` query results, that have
the `status` equal to 1 (see cf0eb1bc6db428d8e4cf5ba754659e9b30d56f38). This makes sense since each proc is a different
processor.
 To verify this, one can check the number of procs with the output from
`prtconf`

```
[AIX] # prtconf
System Model: IBM,8284-22A
Machine Serial Number: 21684EW
Processor Type: PowerPC_POWER8
Processor Implementation Mode: POWER 8
Processor Version: PV_8_Compat
Number Of Processors: 4 <--
```
And the `CuDv` query result:
```
[AIX] # odmget -q 'PdDvLn=processor/sys/proc_rspc' CuDv

CuDv:
        name = "proc0"
	...
CuDv:
        name = "proc8"
	...
CuDv:
        name = "proc16"
	...
CuDv:
        name = "proc24"
	...
```

Output before fix:
```
[0] [AIX] # facter processors
{
  cores => 4,
  count => 32,
  isa => "powerpc",
  models => [
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8"
  ],
  speed => "3.43 GHz",
  threads => 8
}
```
Output after fix:
```
{
  cores => 4,
  count => 32,
  isa => "powerpc",
  models => [
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8",
    "PowerPC_POWER8"
  ],
  physicalcount => 4,
  speed => "3.43 GHz",
  threads => 8
}
```

### DIFF
--- a/lib/facter/facts/aix/processors/physicalcount.rb
+++ b/lib/facter/facts/aix/processors/physicalcount.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Facts
+  module Aix
+    module Processors
+      class Physicalcount
+        FACT_NAME = 'processors.physicalcount'
+
+        def call_the_resolver
+          fact_value = Facter::Resolvers::Aix::Processors.resolve(:physical_count)
+          Facter::ResolvedFact.new(FACT_NAME, fact_value)
+        end
+      end
+    end
+  end
+end
--- a/lib/facter/resolvers/aix/processors.rb
+++ b/lib/facter/resolvers/aix/processors.rb
@@ -18,6 +18,7 @@ module Facter
             @fact_list[:logical_count] = 0
             @fact_list[:cores_per_socket] = 0
             @fact_list[:threads_per_core] = 0
+            @fact_list[:physical_count] = 0
 
             odmquery = Facter::Util::Aix::ODMQuery.new
             odmquery.equals('class', 'processor')
@@ -42,9 +43,11 @@ module Facter
             return unless result
 
             names = retrieve_from_array(result.scan(/name\s=\s.*/), 1)
+
             status = retrieve_from_array(result.scan(/\s+status\s=\s.*/), 1)
 
             names.each_with_index { |elem, idx| query_cuat(elem) if status[idx] == '1' }
+            @fact_list[:physical_count] = status.count('1')
           end
 
           def query_cuat(name)
--- a/spec/facter/facts/aix/processors/physicalcount_spec.rb
+++ b/spec/facter/facts/aix/processors/physicalcount_spec.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+describe Facts::Aix::Processors::Physicalcount do
+  describe '#call_the_resolver' do
+    subject(:fact) { Facts::Aix::Processors::Physicalcount.new }
+
+    let(:physical_count) { '2' }
+
+    before do
+      allow(Facter::Resolvers::Aix::Processors).to \
+        receive(:resolve).with(:physical_count).and_return(physical_count)
+    end
+
+    it 'calls Facter::Resolvers::Aix::Processors' do
+      fact.call_the_resolver
+      expect(Facter::Resolvers::Aix::Processors).to have_received(:resolve).with(:physical_count)
+    end
+
+    it 'returns processors physical count fact' do
+      expect(fact.call_the_resolver).to be_an_instance_of(Facter::ResolvedFact).and \
+        have_attributes(name: 'processors.physicalcount', value: physical_count)
+    end
+  end
+end
--- a/spec/facter/resolvers/aix/processors_spec.rb
+++ b/spec/facter/resolvers/aix/processors_spec.rb
@@ -97,5 +97,9 @@ describe Facter::Resolvers::Aix::Processors do
     it 'returns threads fact' do
       expect(resolver.resolve(:threads_per_core)).to eq(8)
     end
+
+    it 'returns physical_count fact' do
+      expect(resolver.resolve(:physical_count)).to eq(1)
+    end
   end
 end


GabrielNagy:
I'm not seeing a consensus on how this should be treated on AIX. I found this [thread](https://community.ibm.com/community/user/power/communities/community-home/digestviewer/viewthread?GroupId=6049&MID=138050&CommunityKey=daa942cb-b783-4fd3-ba27-a2d7462f9530&tab=digestviewer) (see post 10) which makes the case that based on the `lsdev -Cc processor` output the physicalcount should be 1. An important thing to note from the thread is that AIX's concept of physical processors does not necessarily map to actual physical chips on the motherboard (or whatever it is they have), especially in virtualized environments like we have.

Also AIX runs on POWER architecture which from what I understand is quite different in concepts compared to x86_64. So I think the argument from https://github.com/puppetlabs/facter/commit/e73ea8d3d0a01e9000190e851b27397fb997ccfd makes sense.

Right now your implementation suggests that there are 4 physical cores with 4 cores per each, with 8 threads per each. Totaling this we get to 128 threads which doesn't make sense. The number of threads/CPUs in our facter output would make sense if there were a single physical CPU, not 4.

BobosilaVictor:
> I'm not seeing a consensus on how this should be treated on AIX. I found this [thread](https://community.ibm.com/community/user/power/communities/community-home/digestviewer/viewthread?GroupId=6049&MID=138050&CommunityKey=daa942cb-b783-4fd3-ba27-a2d7462f9530&tab=digestviewer) (see post 10) which makes the case that based on the `lsdev -Cc processor` output the physicalcount should be 1. An important thing to note from the thread is that AIX's concept of physical processors does not necessarily map to actual physical chips on the motherboard (or whatever it is they have), especially in virtualized environments like we have.
> 
> Also AIX runs on POWER architecture which from what I understand is quite different in concepts compared to x86_64. So I think the argument from [e73ea8d](https://github.com/puppetlabs/facter/commit/e73ea8d3d0a01e9000190e851b27397fb997ccfd) makes sense.
> 
> Right now your implementation suggests that there are 4 physical cores with 4 cores per each, with 8 threads per each. Totaling this we get to 128 threads which doesn't make sense. The number of threads/CPUs in our facter output would make sense if there were a single physical CPU, not 4.

This makes sense. It seems that I was confused by the terminology used when describing processors, both by the AIX docs and other forums that I have searched in. 

So I searched again and found a different possible solution for this problem. From [this](https://www.linuxquestions.org/questions/aix-43/how-to-find-the-number-of-physical-processors-in-a-aix-machine-738595/#post3610125)(11) answer and also [this](https://www.linuxquestions.org/questions/aix-43/how-to-find-the-number-of-physical-processors-in-a-aix-machine-738595/#post3611405)(14) one, I gathered that one can check the number of actual chips inserted in a socket on the motherboard by checking the `lscfg -vp|grep WAY` command. On one AIX instance, I get this result: 
```
[0] [AIX] # lscfg -vp|grep WAY
      10-WAY PROC CUOD:
      10-WAY PROC CUOD:
```
From my understanding, this means that the server has 2 10 core chips inside it, which can be verified with: 
```
[0] [AIX] # lparstat -i|grep ^Active\ Phys
Active Physical CPUs in system             : 20
```
Also, by checking the model name of the system, I have found the actual model sold by IBM ([here](https://www.ibm.com/common/ssi/ShowDoc.wss?docURL=/common/ssi/rep_sm/1/760/ENUS8284-_h01/index.html#toc)), and indeed it uses a 2 socket system. So, I think that this system has a physical count of 2, and the virtual machine is allocated 4 cores from one of the chips.

Now I'm not sure what the `physicalcount` fact should output here. The fact that there are 2 chips on the whole system, or the fact that the virtual machine only uses 1? I think the second option makes the most sense, but I'm not sure how this can be achieved...

GabrielNagy:
I did a quick check with @luchihoratiu on how the `physicalcount` fact behaves across OSes and Facter versions:

docs ([schema.yaml](https://github.com/puppetlabs/facter/blob/3.x/lib/schema/facter.yaml#L1362)): The count of physical processors.

What should a physical processor map to? A core on the CPU or the chip from the motherboard?

### macOS
**Facter 2:** `physicalcount` - does not exist
**Facter 3:** `physicalcount` - shows number of physical cores
**Facter 4:** `physicalcount` - shows number of physical cores

### Linux
**Facter 2:** `physicalcount` - shows number of motherboard chips
**Facter 3:** `physicalcount` - shows number of motherboard chips
**Facter 4:** `physicalcount` - shows number of motherboard chips

### Windows
**Facter 2:** `physicalcount` - shows number of  physical cores
**Facter 3:** `physicalcount` - shows number of physical cores
**Facter 4:** `physicalcount` - shows number of physical cores

### AIX
**Facter 2:** `physicalcount` - shows number of physical cores
**Facter 3:** `physicalcount` - does not exist
**Facter 4:** `physicalcount` - does not exist - handled in this PR - implementation would show number of physical cores

### Solaris
**Facter 2:** `physicalcount` - shows number of physical cores
**Facter 3:** `physicalcount` - shows number of physical cores
**Facter 4:** `physicalcount` - shows number of physical cores

This seems like a larger issue that we should address at some point (in a major release?). I think physicalcount should always map to a physical processor on a motherboard, especially now since we have the `cores` fact which kind of duplicates the information of `physicalcount` on platforms like Solaris, AIX, Windows and macOS. Linux seems to be the only platform that has reported the number of actual physical processors for all Facter versions.

PR #2435, opened 2021-09-23
(maint) Disable DHCP test on EL 8 platforms

GabrielNagy:
AlmaLinux and Rocky come with newer versions of NetworkManager which
don't provide the necessary DHCP options for us to resolve the dhcp
fact, see https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426.

It looks like some of the options were added, but they are not available
in the NetworkManager version found on Alma and Rocky.

As Alma/Rocky/RHEL/CentOS are the same platform, the simplest way is to
disable the test on all el-8 platforms. We will probably encounter this
again when we migrate to GCP and the RedHat templates there have newer
NetworkManager versions anyway.


GabrielNagy:
Manually ran acceptance on this and it passed.

PR #2437, opened 2021-09-27
(FACT-3075) Fix the `os.release` fact on Windows 2022

luchihoratiu:
This commit allows the `os.release.full` and `os.release.major` to correctly detect and output Windows 2022 instead of '2019' as previously.

### DIFF
--- a/lib/facter/util/facts/windows_release_finder.rb
+++ b/lib/facter/util/facts/windows_release_finder.rb
@@ -23,9 +23,11 @@ module Facter
           private
 
           def check_version_10(consumerrel, kernel_version)
+            return '10' if consumerrel
+
             build_number = kernel_version[/([^.]*)$/].to_i
-            if consumerrel
-              '10'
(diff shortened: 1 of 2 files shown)


lib/facter/util/facts/windows_release_finder.rb
```diff
@@ -26,6 +26,8 @@ def check_version_10(consumerrel, kernel_version)
             build_number = kernel_version[/([^.]*)$/].to_i
             if consumerrel
```
ciprianbadescu, line 27: I think we should decompose this `if` sequence to specific `if` by variable and with easy to read limits/values

BobosilaVictor (approved):
LGTM! Tested this on a Windows Server 2022 (Standard Edition) and got the expected results:
image
(Note: I also had the changes from https://github.com/puppetlabs/facter/pull/2439)

GabrielNagy (approved):

PR #2438, opened 2021-09-27
(maint) Testing jruby tests

luchihoratiu:


GabrielNagy:
JRuby 9.3.10 was [released](https://www.jruby.org/2021/10/13/jruby-9-3-1-0.html). We could pin to that if we want (or remove the pin to be on the bleeding edge™).

PR #2439, opened 2021-09-27
(FACT-3058) Fix `os.windows.release_id` fact and add `os.windows.display_version` fact

luchihoratiu:
This commit fixes `os.windows.release_id` fact on Windows 2022 by retrieving the information from the correct registry key. The `ReleaseId` registry key from `SOFTWARE\Microsoft\Windows NT\CurrentVersion` seems to be abandoned since build `2009` (is still kept on newer builds and versions of Windows but not updated anymore) and replaced by the `DisplayVersion` registry key from the same location.

This commit also adds a new fact, `os.windows.release_id`, with above information. For backwards compatibility reasons, when `DisplayVersion` is available, both facts will be populated from this registry key. When `DisplayVersion` is not available, only `os.windows.release_id` will be populated but from `ReleaseId` registry key instead (same as before this fix). 

TODO in Facter 5: remove `os.windows.release_id` fact to avoid duplicated information


BobosilaVictor (approved):
LGTM! Tested this on a Windows Server 2022 (Standard Edition) machine and got the expected results:
image
The `display_version` and `release_id`  are shown correctly. 
image

GabrielNagy (approved):

PR #2443, opened 2021-09-28
(FACT-3058) Fix `os.windows.release_id` fact and add `os.windows.display_version` fact

luchihoratiu:
This commit fixes `os.windows.release_id` fact on Windows 2022 by retrieving the information from the correct registry key. The `ReleaseId` registry key from `SOFTWARE\Microsoft\Windows NT\CurrentVersion` seems to be abandoned since build `2009` (is still kept on newer builds and versions of Windows but not updated anymore) and replaced by the `DisplayVersion` registry key from the same location.

This commit also adds a new fact, `os.windows.release_id`, with above information. For backwards compatibility reasons, when `DisplayVersion` is available, both facts will be populated from this registry key. When `DisplayVersion` is not available, only `os.windows.release_id` will be populated but from `ReleaseId` registry key instead (same as before this fix). 

TODO in Facter 5: remove `os.windows.release_id` fact to avoid duplicated information

See https://github.com/puppetlabs/facter/pull/2439 for same implementation in Facter 4

### DIFF
--- a/lib/inc/facter/facts/fact.hpp
+++ b/lib/inc/facter/facts/fact.hpp
@@ -451,6 +451,11 @@ namespace facter { namespace facts {
         constexpr static char const* windows_release_id = "windows_release_id";
 
         /**
+         * The fact for Windows Display Version.
+         */
+        constexpr static char const* windows_display_version = "windows_display_version";
+
+        /**
          * The fact for Windows native system32 directory.
          */
         constexpr static char const* windows_system32 = "system32";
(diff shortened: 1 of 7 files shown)


luchihoratiu:
jenkins please test this on windows2019-64a

luchihoratiu:
jenkins please test this on windows2019-64a

luchihoratiu:
jenkins please test this on windows2019-64a

luchihoratiu:
Output on Windows 2022 with these changes (winver and facter):
image

Output on Windows 2022 without these changes (winver and facter):
image

Build for further testing: `a44b5c630d52d530bd18945d617df491ac25fe9d`

TODO: We should also manually test these on non-server Windows and older servers than 2022 to make sure of the behaviour.

BobosilaVictor (approved):
LGTM! Tested this on a Windows 2019 machine and got the following output with the fix:
image
The `release_id` can be confirmed with running `winver`:
image

Also tested this on a Windows 10 Enterprise machine:
image
image

PR #2444, opened 2021-10-05
(FACT-3077) Fix inconsistencies in Facter output

BobosilaVictor:
Previously, facter 4 wouldn't resolve custom facts that had the value equal
to `false`. For example the custom fact `custom.yaml` with the content
of:
```
---
key1: true
key2: false
```
Would output:
```
key1 => true
key2 =>
```

This commit fixes this issue by checking for `value.nil?` instead of
`value`, to avoid having boolean values skipped when they are set to
`false`

### DIFF
--- a/lib/facter/custom_facts/util/fact.rb
+++ b/lib/facter/custom_facts/util/fact.rb
@@ -117,7 +117,7 @@ module Facter
       #
       # @api public
       def value
-        return @value if @value
+        return @value unless @value.nil?
 
         if @resolves.empty?
           log.debug format('No resolves for %<name>s', name: @name)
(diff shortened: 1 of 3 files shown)


Dorin-Pleava (approved):

luchihoratiu (approved):
LGTM! 👍 
Manually tested these changes on Windows with Puppet 7 and using:
`C:\ProgramData\PuppetLabs\facter\facts.d\yaml_facts.yaml`:
```yaml
---
key1_yaml: true
key2_yaml: false
key3_yaml: 42
key4_yaml: "string"
key5_yaml: nil
key6_yaml: null
key7_yaml:
```
`C:\ProgramData\PuppetLabs\facter\facts.d\json_facts.json`:
```json
{
"key1_json": true,
"key2_json": false,
"key3_json": 42,
"key4_json": "string"
}
```
Got the following output without these changes:
image

And got the following output with these changes:
image

PR #2446, opened 2021-10-08
Revert "Merge pull request #2445 from luchihoratiu/FACT-3068"

ciprianbadescu:
This reverts commit 142ad0203f2dcabe1b8f703741b2813f004ca69f, reversing
changes made to f3404502fa4907e31bd269f53b893dc54731aea0.

PR #2447, opened 2021-10-11
(FACT-3068) Force encoding of timezone fact output on Windows

luchihoratiu:
This commit adds back commit 5822fc32b61ea144c44ab109d93269d441f174f5 and encodes the new resolver's output to 'UTF-8' to avoid code assumptions that end up using an invalid 'UTF-8' string. More information available:
- [ticket discussion](https://tickets.puppetlabs.com/browse/FACT-3068?focusedCommentId=829527&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-829527)
- [revert pr](https://github.com/puppetlabs/facter/pull/2446)
- [original pr](https://github.com/puppetlabs/facter/pull/2445)

The original commit adds a new `timezone` resolver specific to Windows which checks the system codepage and uses it for encoding the `timezone` fact to avoid unwanted characters on non-English OS. 

### Windows FR:
image

### Windows JA:
image
image

And `puppet agent -t --preferred_serialization_format=pson --http_debug --facterng` exited with 0 on a machine with these changes.

TODO: Run Puppet Acceptance tests on Windows when possible.


GabrielNagy:
jenkins please test this,puppet on windows

lib/facter/resolvers/windows/timezone.rb
```diff
@@ -0,0 +1,41 @@
...
+        class << self
+          private
+
+          def post_resolve(fact_name, _options)
+            @fact_list.fetch(fact_name) { determine_timezone }
+          end
+
+          def determine_timezone
+            timezone = Time.now.zone
+            @fact_list[:timezone] = timezone.force_encoding("CP#{codepage}").encode('UTF-8')
```
GabrielNagy, line 18:
```suggestion
            @fact_list[:timezone] = timezone.force_encoding("CP#{codepage}").encode('UTF-8', invalid: :replace)
```

maybe we can do this so we don't end up (even by accident) in a case where the bytes aren't valid UTF-8, and we end up breaking all Facter

PR #2448, opened 2021-10-13
(FACT-3081) Replace invalid UTF characters in dmi.rb

Heidistein:
One server has some invalid (looks like binary data/corruption) in the DMI-data. Productname and -manufacturer are garbled. On one hand this should not happen, on the other hand puppet(facter) should not break over it.
This change should just replace the invalid characters with a questionmark-gliph-thingy, and continue not burning down.


Heidistein:
Right, I was messing with manufacturer_spec, trying to call the DmiBios with the data, your approach is simpler and even got it working. Thanks.

PR #2465, opened 2021-12-06
(FACT-3099) ec2_userdata fails if it contains ASCII-8BIT chars

Dorin-Pleava:
When using facter to get ec2_userdata that contains some ASCII-8Bit values, a
conversion error would be thrown.

The error is caused by amazon ec2 instances return base64-encoded userdata strings,
so it can contain non UTF-8 data.

Forcing a UTF-8 encoding seems to fix the problem, but not sure if there are
cases when we would want to have non UTF-8 data.

### DIFF
--- a/spec/facter/resolvers/ec2_spec.rb
+++ b/spec/facter/resolvers/ec2_spec.rb
@@ -142,4 +142,20 @@ describe Facter::Resolvers::Ec2 do
 
     ec2.resolve(:userdata)
   end
+
+  context 'when data received are ASCII-8BIT encoded' do
+    let(:headers) { { 'Accept' => '*/*' } }
+
+    it 'converts data received to UTF-8 encoding' do
+      allow(Facter::Resolvers::Ec2).to receive(:v2_token).and_return(nil)
+
+      expected_str = "nofail\xCC\xA6\"".strip
+      bin_str = "nofail\xCC\xA6\"".strip.force_encoding(Encoding::ASCII_8BIT)
+      stub_request(:get, userdata_uri).with(headers: headers).to_return(status: 200, body: bin_str, headers:
+        { 'Content-Type' => 'application/octet-stream' })
+      stub_request(:get, metadata_uri).with { |request| !request.headers.key?('X-aws-ec2-metadata-token') }
+
+      expect(ec2.resolve(:userdata)).to eql(expected_str)
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


Dorin-Pleava:
Tried to create some specs to test the `force_encoding` but I think `stub_request` cannot return ASCII-8BIT and always returns UTF-8
```
ascii_str = "nofail\xCC\xA6\""
stub_request(:get, userdata_uri).with(headers: headers).to_return(status: 200, body: ascii_str)
...
response =  "nofail\u0326\""
```

joshcooper:
By default, literal strings in ruby source files are UTF-8 encoded (unless an explicit encoding is specified as a pragma at the top of the file or if `Encoding.default_internal` is set to an explicit value). I think you can simulate `stub_request` returning a binary string by forcing the encoding of the string for the response **body**:

```ruby
bin_str = "nofail\xCC\xA6\"".force_encoding(Encoding::ASCII_8BIT)
stub_request(:get, userdata_uri).with(headers: headers).to_return(status: 200, body: bin_str)
...
```

Note in ruby `Encoding::BINARY` and `ASCII_8BIT` are the same (one is aliased to the other).

Also to correctly simulate the request/response, you may want to return `Content-Type: application/octet-stream` in the response header, I think it's something like:

```ruby
to_return(status: 200, body: bin_str, headers: { 'Content-Type' => 'application/octet-stream'})

**Comment by traylenator** [comment]
Hi,
Can I just try reproducing this with another fact - sure it should be possible...

File: spec/facter/resolvers/ec2_spec.rb [hunk 1]
```diff
@@ -142,4 +142,19 @@
...
     ec2.resolve(:userdata)
   end
+
+  context 'when data received are ASCII-8BIT encoded' do
+    let(:headers) { { 'Accept' => '*/*' } }
+
+    it 'converts data received to UTF-8 encoding' do
+      allow(Facter::Resolvers::Ec2).to receive(:v2_token).and_return(nil)
+
+      bin_str = "nofail\xCC\xA6\"".strip.force_encoding(Encoding::ASCII_8BIT)
```
GabrielNagy, line 152: You can also have a variable like `expected_str = "nofail\xCC\xA6\"".strip` which should be UTF-8, and then change the expectation to `expect(ec2.resolve(:userdata).encoding).to eql(expected_str)`. This way the string contents will also be checked.

Also, I think you can remove the `strip` calls since they don't change the string contents. It looks like the string is frozen so you can use `dup` instead for better readability.

traylenator:
> Hi, Can I just try reproducing this with another fact - sure it should be possible...

Seems fine in fact, caching a trivial unicode fact, editing it on disk then the fact comes back correctly from cache.

```
 facter unicode
Stévé
[root@bag8 facts.d]# cat /etc/facter/facts.d/unicode.txt 
unicode=Stévé123
```
Ignore my comment.

GabrielNagy (approved):

PR #2467, opened 2021-12-14
FACT-3100) fix disks fact

valia0906:
https://tickets.puppetlabs.com/browse/FACT-3100
Fix disks fact could not get serial bacause of using absolute path for lsblk.


GabrielNagy (commented):
Thanks for the PR @valia0906,

Looks like there are some failing tests that need updating in `spec/facter/resolvers/disks_spec.rb`. It should be as simple as changing `/usr/bin/lsblk` to `lsblk` there as well.

If you update that and sign the CLA (see the comment above) this should be good to merge!

PR #2471, opened 2022-01-11
(PUP-11370) Github workflow now uses windows 2019

Dorin-Pleava:
Upgraded github actions to use windows 2019 instead of windows 2016 as
it will be removed on March 15, 2022.


GabrielNagy (approved):
Also updated the required status check from windows-2016 to windows-2019.

PR #2475, opened 2022-02-03
(FACT-1556) Add KVM detection for cloud fact

cthorn42:
The cloud fact will now detect if the KVM hypervisor is being used and if so look to see which cloud provider is used.

### DIFF
--- a/lib/facter/facts/linux/cloud/provider.rb
+++ b/lib/facter/facts/linux/cloud/provider.rb
@@ -10,6 +10,8 @@ module Facts
           provider = case Facter::Util::Facts::Posix::VirtualDetector.platform
                      when 'hyperv'
                        'azure' unless Facter::Resolvers::Az.resolve(:metadata).empty?
+                     when 'kvm', 'xen'
+                       'aws' unless Facter::Resolvers::Ec2.resolve(:metadata).empty?
                      end
 
           Facter::ResolvedFact.new(FACT_NAME, provider)
--- a/lib/facter/facts/windows/cloud/provider.rb
+++ b/lib/facter/facts/windows/cloud/provider.rb
@@ -11,6 +11,8 @@ module Facts
           provider = case virtual
                      when 'hyperv'
                        'azure' unless Facter::Resolvers::Az.resolve(:metadata).empty?
+                     when 'kvm', 'xen'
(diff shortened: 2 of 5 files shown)


lib/facter/facts/linux/cloud/provider.rb
```diff
@@ -10,6 +10,9 @@ def call_the_resolver
           provider = case Facter::Util::Facts::Posix::VirtualDetector.platform
                      when 'hyperv'
                        'azure' unless Facter::Resolvers::Az.resolve(:metadata).empty?
+                     when 'kvm'
+                       # hypervisor.kvm fact is what we reuse
+                       Facts::Linux::Hypervisors::Kvm.new.call_the_resolver.value.first[0]
```
joshcooper, line 15: Is `call_the_resolver.value` or `call_the_resolver.value.first` ever `nil`? It seems like it may be possible if the hypervisor is `virtualbox` or `parallels`:

https://github.com/puppetlabs/facter/blob/636c4955fcd5449d625f8390a99dad6564857e12/lib/facter/facts/linux/hypervisors/kvm.rb#L17

Or if `kvm?` is false, so `fact_value` is `nil` in https://github.com/puppetlabs/facter/blob/636c4955fcd5449d625f8390a99dad6564857e12/lib/facter/facts/linux/hypervisors/kvm.rb#L19-L21

Could we follow the Azure example and rely on the EC2 metadata to detect AWS, something like:

```ruby
when 'kvm'
  'aws' unless Facter::Resolvers::Ec2.resolve(:metadata).empty?
end
```

Also side note, we'll need a similar change for the [Windows cloud provider](https://github.com/puppetlabs/facter/blob/main/lib/facter/facts/windows/cloud/provider.rb).

cthorn42: Good call, that looks much more straight forward.

joshcooper (approved):
LGTM, might be good to install a dev build on linux and windows and confirm `facter cloud.provider` is what we expect?

cthorn42:
Okay with the latest push, I tested both an EL-7 and a Windows Server 2019 on AWS. Ran into an issue where the Windows server had a 'xen' hypervisor. So updated the logic to look for both hypervisors, so I think that should cover us. 
With the most recent change can confirm `cloud.provider` fact is show up as `aws` on both OSes I tested.

PR #2476, opened 2022-02-14
(FACT-3104) Update the Snyk monitor action

cthorn42:
* Removed cron timer, only need to run on push
* Update the key


cthorn42:
Example of the action running: https://github.com/puppetlabs/facter/actions/runs/1843628121

joshcooper (approved):

PR #2477, opened 2022-03-11
(FACT-3090) Fix windows 11 detection

smortex:
Windows 11 has version 10.0.22000 (yeah, versioning is **hard** at microsoft):

https://docs.microsoft.com/en-us/windows/release-health/windows11-release-information

Adjust the logic so that os.release.full is set to 11, not 10.

### DIFF
--- a/lib/facter/util/facts/windows_release_finder.rb
+++ b/lib/facter/util/facts/windows_release_finder.rb
@@ -14,7 +14,7 @@ module Facter
             kernel_version = input[:kernel_version]
 
             if version =~ /10.0/
-              check_version_10(consumerrel, kernel_version)
+              check_version_10_11(consumerrel, kernel_version)
             else
               check_version_6(version, consumerrel) || check_version_5(version, consumerrel, description) || version
             end
@@ -22,10 +22,12 @@ module Facter
 
           private
 
-          def check_version_10(consumerrel, kernel_version)
+          def check_version_10_11(consumerrel, kernel_version)
+            build_number = kernel_version[/([^.]*)$/].to_i
+
+            return '11' if build_number >= 22_000
             return '10' if consumerrel
 
-            build_number = kernel_version[/([^.]*)$/].to_i
             if build_number >= 20_348
               '2022'
             elsif build_number >= 17_623
--- a/spec/facter/util/facts/windows_release_finder_spec.rb
+++ b/spec/facter/util/facts/windows_release_finder_spec.rb
@@ -25,6 +25,17 @@ describe Facter::Util::Facts::WindowsReleaseFinder do
     end
   end
 
+  describe '#find windows release when version is 11' do
+    let(:cons) { true }
+    let(:desc) {}
+    let(:k_version) { '10.0.22000' }
+    let(:version) { '10.0' }
+
+    it 'returns 11' do
+      expect(Facter::Util::Facts::WindowsReleaseFinder.find_release(input)).to eql('11')
+    end
+  end
+
   describe '#find windows release when version is 2022' do
     let(:cons) { false }
     let(:desc) {}


joshcooper:
Thank you for your contribution @smortex! Does this resolve all of the issues in https://tickets.puppetlabs.com/browse/FACT-3090? If so, could you add that ticket to your commit summary?

smortex:
@joshcooper it might… I'll try to give a look at $WORK next week :wink:

smortex:
Got windows running in VirtualBox, I think it does the right thing: only the version change:
Screenshot 2022-03-11 212425

joshcooper:
Thanks @smortex! It looks like `os.windows.product_name => 'Windows 10 Enterprise Evaluation'` needs updating. Also, is this really the display version `os.windows.display_version = '21H2'`? (I haven't been following MSFT versioning schemes lately)

smortex:
> It looks like `os.windows.product_name => 'Windows 10 Enterprise Evaluation'` needs updating.

Well, that's the raw value of the registry key `HKEY_LOCAL_MACHINE/SOFTWARE/Microsoft/Windows NT/CurrentVersion/ProductName` on my virtual machine (what Facter get it from)… :smile: 

See also:
* https://answers.microsoft.com/en-us/windows/forum/all/product-name-not-updated-in-registry-on-windows-11/ab21721a-9a80-4ab5-ae27-2b08c97976a2
* https://docs.microsoft.com/en-us/answers/questions/555857/windows-11-product-name-in-registry.html

Do you think we should add quirks if facter to workaround Microsoft inconsistencies (`s/10/11` in ProductName if version is 11), or should we rather just hope that Microsoft will fix it at some point in the future?

> Also, is this really the display version `os.windows.display_version = '21H2'`? (I haven't been following MSFT versioning schemes lately)

I *think* so: I see this version on this development VM, coworkers nodes, issue FACT-3090 in jira and on https://docs.microsoft.com/en-us/windows/release-health/windows11-release-information

There also seems to be `Windows 10 21H2`.  I feel like the "right" way to check a version is by looking at the build number, 22000+ for Windows 11, what is doing this PR… but it does not seem to be exposed by facter ATM.

Windows 11 is really is Windows 10 with lipstick on it :clown_face: .

canihavethisone:
I feel that querying build greater than may, or may not prove reliable (knowing Microsoft).  I have created a custom fact in my environment that uses WMI for a more guaranteed result.  As I still use Puppet 5 on some nodes, this is required for me as new features in facter will not be received.  Happy to get any feedback on this approach:
```Facter.add('windows_version') do
  confine :osfamily => :windows
  setcode do
    begin
      require 'win32ole'
      wmi = WIN32OLE.connect('winmgmts:\\\\.\\root\\cimv2')
      # Retrieve operatingsystem hash from WMI as regsitry values were not incremented by Microsoft
      windows_version = wmi.ExecQuery('SELECT * FROM Win32_operatingsystem').each.first

      # Create hash for values
      value = {}
      value['product_name'] = windows_version.Caption.split(' ')[1..-1].join(' ')
      value['release'] = windows_version.Caption.split(' ')[2]

      # Return hash
      value
    end
  end
end
```

canihavethisone:
@joshcooper any thoughts on using wmi lookup as above instead of relying on build number being greater than?  I feel that the wmi method is more reliable and appropriate.

joshcooper:
WMI can be useful sometimes, but there are a few caveats:

1. It's really slow. This was the main reason Facter 2.x was so slow on Windows, combined with Facter 2 retrieving network information multiple times (each nic attribute times each nic)
2. If you have to use WMI, select only the properties you care about, e.g. `SELECT product_name, release FROM Win32_operatingsystem`. Otherwise, `SELECT *` will return all of the data in the `Win32_OperatingSystem` class, only to discard most of it.
3. Never use `Win32_Product` as it can cause packages to be reinstalled.

canihavethisone:
@joshcooper thanks for that info.  I have updated my code suggestion above to use PowerShell's WMI query (instead of Ruby) and removed the retrieval of *.  There was no perceivable time benefit however.

Running the fact within measure command, it takes ~420 milliseconds so I don't find it slow.

joshcooper:
> Running the fact within measure-command, it takes ~1400 milliseconds so I don't find it too slow.

That seems really slow to me. Most facts resolve in tens of msec or less:

```
C:\> cmd /c facter --timing
fact 'processor.*', took: (0.023844) seconds
fact 'ssh.*key', took: (0.000491) seconds
fact 'sshfp_.*', took: (0.000010) seconds
fact 'az_metadata', took: (0.007484) seconds
fact 'mtu_.*', took: (0.012659) seconds
fact 'netmask6_.*', took: (0.000011) seconds
fact 'ssh', took: (0.000011) seconds
fact 'cloud.provider', took: (0.000007) seconds
fact 'netmask_.*', took: (0.000008) seconds
fact 'network6_.*', took: (0.000007) seconds
fact 'dhcp_servers', took: (0.000022) seconds
fact 'processors.isa', took: (0.000009) seconds
fact 'hardwareisa', took: (0.000004) seconds
fact 'processors.models', took: (0.000008) seconds
fact 'processors.physicalcount', took: (0.000008) seconds
fact 'physicalprocessorcount', took: (0.000007) seconds
fact 'processors.threads', took: (0.000008) seconds
fact 'processors.cores', took: (0.000008) seconds
fact 'processors.count', took: (0.000007) seconds
fact 'processorcount', took: (0.000004) seconds
fact 'dmi.product.name', took: (0.003807) seconds
fact 'productname', took: (0.000015) seconds
fact 'dmi.product.uuid', took: (0.000016) seconds
```

joshcooper:
> Do you think we should add quirks if facter to workaround Microsoft inconsistencies

I'd say no. I think facter should report on the actual state of the system. Presumably MS will fix this as the questions keep coming in https://docs.microsoft.com/en-us/answers/questions/711686/win11-machine-is-still-showing-windows-10-as-produ.html

joshcooper:
jenkins please test this on windows2019-64a,windows10ent-64a

joshcooper (approved):

PR #2480, opened 2022-04-12
(FACT-1592) A key/value pair for Interface Speed/Duplex

jcpunk:
I'll confess, I'll need help with the tests....

This partially resolves the issue for Linux systems.

Facter will have a `speed` fact equal to the link speed determined by the kernel for each supported interface.
Facter will have a `duplex` fact equal to the duplex determined by the kernel for each supported interface.

For wired networks this should report the speed/duplex even if the interface is not assigned an address.

### DIFF
--- a/lib/facter/resolvers/linux/networking.rb
+++ b/lib/facter/resolvers/linux/networking.rb
@@ -33,6 +33,8 @@ module Facter
               dhcp(interface_name, mtu_and_indexes)
               operstate(interface_name)
               physical(interface_name)
+              linkspeed(interface_name)
+              duplex(interface_name)
 
               @log.debug("Found interface #{interface_name} with #{@fact_list[:interfaces][interface_name]}")
             end
@@ -60,6 +62,30 @@ module Facter
                                                          else
                                                            false
                                                          end
+          end
+
+          def duplex(interface_name)
+            return unless @fact_list[:interfaces][interface_name][:physical]
+
+            # not all interfaces support this, wifi for example causes an EINVAL (Invalid argument)
+            begin
+              plex = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/duplex", nil)
+              @fact_list[:interfaces][interface_name][:duplex] = plex.strip if plex
+            rescue StandardError => e
+              @log.debug("Failed to read '/sys/class/net/#{interface_name}/duplex': #{e.message}")
+            end
+          end
+
+          def linkspeed(interface_name)
+            return unless @fact_list[:interfaces][interface_name][:physical]
+
+            # not all interfaces support this, wifi for example causes an EINVAL (Invalid argument)
+            begin
+              speed = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/speed", nil)
+              @fact_list[:interfaces][interface_name][:speed] = speed.strip if speed
+            rescue StandardError => e
+              @log.debug("Failed to read '/sys/class/net/#{interface_name}/speed': #{e.message}")
+            end
           end
 
           def parse_ip_command_line(line, mtu_and_indexes)
--- a/spec/facter/resolvers/linux/networking_spec.rb
+++ b/spec/facter/resolvers/linux/networking_spec.rb
@@ -27,6 +27,10 @@ describe Facter::Resolvers::Linux::Networking do
         .with('/sys/class/net/lo/operstate', nil).and_return('unknown')
       allow(Facter::Util::FileHelper).to receive(:safe_read)
         .with('/sys/class/net/ens160/operstate', nil).and_return('up')
+      allow(Facter::Util::FileHelper).to receive(:safe_read)
+        .with('/sys/class/net/ens160/speed', nil).and_return(1000)
+      allow(Facter::Util::FileHelper).to receive(:safe_read)
+        .with('/sys/class/net/ens160/duplex', nil).and_return('full')
     end
 
     after do
@@ -293,6 +297,7 @@ describe Facter::Resolvers::Linux::Networking do
               { address: 'fe80::250:56ff:fe9a:8481', netmask: 'ffff:ffff:ffff:ffff::',
                 network: 'fe80::', scope6: 'link', flags: ['permanent'] }
             ],
+            duplex: 'full',
             ip6: 'fe80::250:56ff:fe9a:8481',
             mac: '00:50:56:9a:61:46',
             mtu: 1500,
@@ -300,7 +305,8 @@ describe Facter::Resolvers::Linux::Networking do
             network6: 'fe80::',
             operational_state: 'up',
             physical: true,
-            scope6: 'link'
+            scope6: 'link',
+            speed: 1000
           }
         }
       end


tskirvin:
For the record, I'd like to see this functionality.  It looks useful for day-to-day work.

lib/facter/resolvers/linux/networking.rb
```diff
@@ -56,6 +58,24 @@ def parse_ip_command_line(line, mtu_and_indexes)
             mtu_and_indexes[name] = { index: index, mtu: mtu }
           end
 
+          def duplex(interface_name)
+            # not all interfaces support this, wifi for example does not
+            unless interface_name == 'lo'
+              @fact_list[:interfaces][interface_name][:duplex] = File.read("/sys/class/net/#{interface_name}/duplex").strip
```
joshcooper, line 64: Is it safe to hardcode `lo` here? What about `lo:1`? Or should we be using a more general way of determining if the interface is loopback?

Can `@fact_list[:interfaces][interface_name]` be nil?

Can `"/sys/class/net/#{interface_name}/duplex"` not exist, if so `File.read` will raise.

jcpunk: Alas, I can only answer one of these:

> Can `"/sys/class/net/#{interface_name}/duplex"` not exist, if so `File.read` will raise.

This file will always exist for each interface.  When it isn't relevant to the type of interface it is set to `unknown`.

lib/facter/resolvers/linux/networking.rb
```diff
@@ -56,6 +58,24 @@ def parse_ip_command_line(line, mtu_and_indexes)
...
+
+          def linkspeed(interface_name)
+            # not all interfaces support this, wifi for example does not
+            unless interface_name == 'lo'
+              @fact_list[:interfaces][interface_name][:speed] = File.read("/sys/class/net/#{interface_name}/speed").strip
+            end
+          rescue
+            nil
+          end
+
```
joshcooper, line 78:
```suggestion
            if interface_name != 'lo'
              speed = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/speed", nil)
              @fact_list[:interfaces][interface_name][:speed] = speed.strip if speed
```

lib/facter/resolvers/linux/networking.rb
jcpunk: In theory I've got this sorted and have a check in place to avoid listing `lo:1` etc.

lib/facter/resolvers/linux/networking.rb
```diff
@@ -56,6 +58,26 @@ def parse_ip_command_line(line, mtu_and_indexes)
...
           end
 
+          def duplex(interface_name)
+            # not all interfaces support this, wifi for example does not
+            unless interface_name.start_with?('lo')
+              plex = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/duplex", nil)
+              @fact_list[:interfaces][interface_name][:duplex] = plex.strip if plex
+            end
+          rescue StandardError
+            nil
```
joshcooper, line 68: The `safe_read` should eliminate the need for the rescuing `StandardError` here and below (since it checks if the file is readable before trying to read it).

jcpunk: I've dropped the offending bits.

lib/facter/resolvers/linux/networking.rb
```diff
@@ -56,6 +58,22 @@ def parse_ip_command_line(line, mtu_and_indexes)
...
+            unless interface_name.start_with?('lo')
+              plex = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/duplex", nil)
+              @fact_list[:interfaces][interface_name][:duplex] = plex.strip if plex
+            end
+          end
+
+          def linkspeed(interface_name)
+            # not all interfaces support this, wifi for example does not
+            unless interface_name.start_with?('lo')
+              speed = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/speed", nil)
```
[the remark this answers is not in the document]
jcpunk: Interesting, I didn't look into how wireless devices would work...  Those seem to need to parse out `cat /proc/net/wireless` perhaps?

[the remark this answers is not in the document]
joshcooper: The error is coming from `read` system call. Not sure why it's readable but can't be read. Would be good to understand what's special about this file before blinding rescuing `EINVAL`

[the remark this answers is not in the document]
jcpunk: It appears that this is kicking out `EINVAL` as the code to produce speed measurements for WiFi devices does not use this backing store.

lib/facter/resolvers/linux/networking.rb
```diff
@@ -62,6 +64,30 @@ def physical(ifname)
...
           end
 
+          def duplex(interface_name)
+            # not all interfaces support this, wifi for example causes an EINVAL (Invalid argument)
+            unless interface_name.start_with?('lo')
+              begin
+                plex = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/duplex", nil)
+                @fact_list[:interfaces][interface_name][:duplex] = plex.strip if plex
+              rescue
+                @fact_list[:interfaces][interface_name][:duplex] = nil
```
joshcooper, line 74: Could you log the error at debug level:

```suggestion
              rescue => e
                @log.debug("Failed to read '/sys/class/net/#{interface_name}/duplex': #{e.message}")
                @fact_list[:interfaces][interface_name][:duplex] = nil
              end
```

lib/facter/resolvers/linux/networking.rb
```diff
@@ -62,6 +64,30 @@ def physical(ifname)
...
+            end
+          end
+
+          def linkspeed(interface_name)
+            # not all interfaces support this, wifi for example causes an EINVAL (Invalid argument)
+            unless interface_name.start_with?('lo')
+              begin
+                speed = Facter::Util::FileHelper.safe_read("/sys/class/net/#{interface_name}/speed", nil)
+                @fact_list[:interfaces][interface_name][:speed] = speed.strip if speed
+              rescue
```
joshcooper, line 85: similar comment here

joshcooper:
I disabled the `Metrics/ClassLength` rubocop check so if you rebase you'll get that change. Not sure about the other cops.

jcpunk:
rebased

lib/facter/resolvers/linux/networking.rb
```diff
@@ -62,6 +64,32 @@ def physical(ifname)
                                                          end
           end
 
+          def duplex(interface_name)
+            # not all interfaces support this, wifi for example causes an EINVAL (Invalid argument)
+            unless interface_name.start_with?('lo')
```
joshcooper, line 69: I think rubocop is suggesting to return early like:

```suggestion
            return unless interface_name.start_with?('lo')
```

and then shift everything left

jcpunk: I think I've cleaned up what it was upset about and made sure there is a default case for `lo`.

spec/facter/resolvers/linux/networking_spec.rb
```diff
@@ -27,6 +27,10 @@
         .with('/sys/class/net/lo/operstate', nil).and_return('unknown')
       allow(Facter::Util::FileHelper).to receive(:safe_read)
         .with('/sys/class/net/ens160/operstate', nil).and_return('up')
+      allow(Facter::Util::FileHelper).to receive(:safe_read)
+        .with('/sys/class/net/ens160/speed', nil).and_return(1000)
+      allow(Facter::Util::FileHelper).to receive(:safe_read)
+        .with('/sys/class/net/ens160/duplex').and_return('full')
```
joshcooper, line 33: This fixes some of the issues:

```suggestion
        .with('/sys/class/net/ens160/duplex', nil).and_return('full')
```

jcpunk: Fixed.

joshcooper:
I made a slight modification to a5267b3dac3af96b6d5c1c799d306087d6d81ba9, added a refactor commit and opened a new PR to supersede this one. Thank you for your contribution (and patience!)

PR #2481, opened 2022-04-21
resolvers/networking.rb: avoid calling ipconfig unless Windows

spchamp:
The following message was observed when running **`puppet catalog compile`** under `RUBYOPT=--debug` with Puppet 7 and Facter 4.2.9 on FreeBSD 12.3

~~~~
Exception `Errno::ENOENT' at /opt/puppet/puppet_wk/puppet_ctl/tools/bundle/ruby/3.0/gems/facter-4.2.9/lib/facter/custom_facts/core/execution/popen3.rb:17 - No such file or directory - ipconfig
~~~~

In order to prevent the call to the Microsoft Windows **ipconfig** tool, this changeset adds an additional test under the method **Facter::Resolvers::Networking.extract_dhcp**

The patched networking.rb has been tested with Puppet 7 from a local installation of Puppet, using Ruby 3.0 from FreeBSD ports, on a FreeBSD 12.3 platform

(cherry picked from commit 614213a104971ffb89c1728d58081f60d7f65af2)

### DIFF
--- a/lib/facter/resolvers/networking.rb
+++ b/lib/facter/resolvers/networking.rb
@@ -71,7 +71,8 @@ module Facter
         end
 
         def extract_dhcp(interface_name, raw_data, parsed_interface_data)
-          return unless raw_data =~ /status:\s+active/
+          return unless ((raw_data =~ /status:\s+active/) &&
+                         (OsDetector.instance.identifier == :windows))
 
           result = Facter::Core::Execution.execute("ipconfig getoption #{interface_name} " \
                                                      'server_identifier', logger: log)


lib/facter/resolvers/networking.rb
```diff
@@ -71,7 +71,8 @@ def extract_mac(raw_data, parsed_interface_data)
         end
 
         def extract_dhcp(interface_name, raw_data, parsed_interface_data)
-          return unless raw_data =~ /status:\s+active/
+          return unless ((raw_data =~ /status:\s+active/) &&
+                         (OsDetector.instance.identifier == :windows))
```
joshcooper, line 75: `ipconfig` is valid on macOS too:

```
# uname -a
Darwin taxable-stump 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_X86_64 x86_64
# ipconfig getoption en0 server_identifier
10.32.22.9
```

spchamp: Candidly, I'm not familiar with the OS X environment, or so much of Windows networking. I was aware that there is an ipfconfig tool on Windows, did not take a look at the cmd syntax though.

It looks like :macosx is actually where the matching ipconfig would be found. I'll update the patch

spchamp:
I wasn't sure of adding any `require` calls to the change - assuming it all gets linked together properly once facter is all loaded

joshcooper:
@spchamp  Thanks for your contribution! Could you squash your commit and update the commit summary (since we're only calling ipconfig on macOS)? It's a little bit of a code smell to check of macos in the "base" resolver, but I think it's ok for now. If we need more macOS specific networking behavior then we can move that to a subclass later.

joshcooper (changes requested):
Hi @spchamp could you squash your commits?

PR #2482, opened 2022-04-21
resolvers/freebsd/virtual.rb: Adding kenv check for bhyve

spchamp:
This changeset adds a test for bhyve virtualization. 

Presently, bhyve VMs would be reported as 'generic' under  the 'virtual' key in facter. This would match the value of the `kern.vm_guest` sysctl under bhyve.

This changeset detects the generic case, then providing a string-downcase translation of the `smbios.bios.vendor` field under FreeBSD kenv. This might typically be the string, 'bhyve'. 

This more specialized representation for a bhyve vm might be easier to interpret for its relevance, in Puppet manifest code, contrated to the string, 'generic'.

Candidly, this may have an impact on any existing code that would assume that the value 'generic' will be produced for that key. If this patch was adopted, perhaps the updated behavior could be annotated in the release notes or in documentation?

(cherry picked from commit 98ca8d7d229f2df8e22f6bf32808448ca5aec330)

### DIFF
--- a/lib/facter/resolvers/freebsd/virtual.rb
+++ b/lib/facter/resolvers/freebsd/virtual.rb
@@ -31,6 +31,11 @@ module Facter
 
               vm = VM_GUEST_SYSCTL_NAMES[vm] if VM_GUEST_SYSCTL_NAMES.key?(vm)
 
+              if (vm == 'generic')
+                ## may be bhyve
+                vm = Facter::Freebsd::FfiHelper.kenv(:get, 'smbios.bios.vendor').downcase
+              end
+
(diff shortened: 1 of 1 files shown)


spchamp:
This patch might be limited in its effects to a FreeBSD VM under bhyve. 

It seems that Linux under bhyve may already report `virtual => bhyve` in facter. This might be commensurate with the definition of **Facter::Util::Facts::HYPERVISORS_HASH**.

In effect, the changeset proposed here might serve to bring the behavior for facter installed in a VM running FreeBSD into alignment with that for a VM running Linux, with either under bhyve.

PR #2484, opened 2022-04-22
(maint) Add Ruby 3.1 to CI

bastelfreak:
This is used by modern distributions so I think it's a good idea to test
against it.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -21,6 +21,7 @@ jobs:
           - '2.3'
           - '2.7'
           - '3.0'
+          - '3.1'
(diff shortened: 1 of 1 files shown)


joshcooper:
Thanks @bastelfreak! We have a ticket filed to investigate why specs fail on Ruby 3.1 https://tickets.puppetlabs.com/browse/FACT-3102 There's one for PUP too

joshcooper:
Thank you for your contribution @bastelfreak! This has been superseded by https://github.com/puppetlabs/facter/pull/2525 which adds both 3.1 (windows) and 3.2 preview 2 (ubuntu) to the rspec matrix

PR #2486, opened 2022-04-27
(FACT-3116) Ignore EROFS when deleting fact cache

kajinamit:
When fact cache is placed in read only file system, deleting a file
fails with EROFS instead of EACCES .


joshcooper:
Thank you @kajinamit for your contribution!

kajinamit:
The error in CI looks irrelevant. I intended to rerun CI but it seems it again requires approval ...

joshcooper:
jenkins please test this on redhat7-64a,windows2019-64a

joshcooper (approved):

PR #2489, opened 2022-05-03
(FACT-3113) Locale Bug

BartoszBlizniak:
Prior to this commit, if the encoding wasn't set to default to UTF_8 facter would crash if the system locale didn't match with facter code causing a runtime error. This only occurs on version 4 of Facter due to code migration.

### DIFF
--- a/lib/facter/util/file_helper.rb
+++ b/lib/facter/util/file_helper.rb
@@ -9,14 +9,14 @@ module Facter
         DEBUG_MESSAGE = 'File at: %s is not accessible.'
 
         def safe_read(path, default_return = '')
-          return File.read(path, :encoding => 'UTF_8') if File.readable?(path)
+          return File.read(path, :encoding => Encoding::UTF_8) if File.readable?(path)
 
           log_failed_to_read(path)
           default_return
         end
 
         def safe_readlines(path, default_return = [])
-          return File.readlines(path, :encoding => 'UTF_8') if File.readable?(path)
+          return File.readlines(path, :encoding => Encoding::UTF_8) if File.readable?(path)
 
           log_failed_to_read(path)
           default_return
--- a/lib/facter/util/file_helper.rb
+++ b/lib/facter/util/file_helper.rb
@@ -9,14 +9,14 @@ module Facter
         DEBUG_MESSAGE = 'File at: %s is not accessible.'
 
         def safe_read(path, default_return = '')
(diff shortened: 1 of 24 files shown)


joshcooper:
Hi @BartoszBlizniak thanks for you contribution. Just a few PR clean up requests. Could you create a feature branch in your fork:

```
git checkout -b locale_3113
git reset --hard fd3e25db1c2911e8b6e14324f88d658b012ff8f4
```

And squash the commits into one commit:

```
git reset 356fcc265516f5d9504d745847c887956dff8453^
git add lib
git commit
```

Then push your feature branch to your fork and create a new PR:

```
git push -f origin HEAD
```

lib/facter/util/file_helper.rb
```diff
@@ -9,14 +9,14 @@ class << self
         DEBUG_MESSAGE = 'File at: %s is not accessible.'
 
         def safe_read(path, default_return = '')
-          return File.read(path) if File.readable?(path)
+          return File.read(path, :encoding => 'UTF_8') if File.readable?(path)
```
joshcooper, line 12: Nit better to use the encoding constant instead of a temporary string:

```suggestion
          return File.read(path, :encoding => Encoding::UTF_8) if File.readable?(path)
```

BartoszBlizniak:
Hi @joshcooper - I'll keep this PR open for the spec tests for time being, any ideas as to why this is failing? 

```
rspec ./spec/facter/util/file_helper_spec.rb:41 # Facter::Util::FileHelper#safe_read when successfully read the file content returns the file content
rspec ./spec/facter/util/file_helper_spec.rb:45 # Facter::Util::FileHelper#safe_read when successfully read the file content File.readable? is called with the correct path
```

joshcooper:
> any ideas as to why this is failing?

There's more detail in the logs earlier on like:

```
       #<File (class)> received :read with unexpected arguments
         expected: ("/Users/admin/file.txt")
              got: ("/Users/admin/file.txt", {:encoding=>#<Encoding:UTF-8>})
```

This means existing tests are doing something like:

```ruby
expect(File).to receive(:read).with("/Users/admin/file.txt")
```

But they are now receiving an options hash containing the `:encoding` key.

I would change the expectation for existing tests to be:

```ruby
expect(File).to receive(:read).with("/Users/admin/file.txt", anything)
```

And add a test to verify it can load content as UTF-8

PR #2490, opened 2022-05-04
(FACT-3109) Add handling for Windows NIC names

mhashizume:
In Windows, a network interface's (NIC) "friendly name" is set by
the user and encoded in UTF-16LE. We currently convert these names
to UTF-8 and invalid characters can cause an error.

This commit replaces any invalid characters during the conversion to
UTF-8.

### DIFF
--- a/lib/facter/resolvers/windows/ffi/ffi.rb
+++ b/lib/facter/resolvers/windows/ffi/ffi.rb
@@ -24,26 +24,33 @@ MAX_PATH = 32_767
 
 module FFI
   WIN32FALSE = 0
-  END_OF_WCHAR_STRING = "\0\0".encode('UTF-16LE')
+  END_OF_WCHAR_STRING = (+"\0\0").force_encoding(Encoding::UTF_16LE).freeze
 
   class Pointer
     def read_wide_string_with_length(char_length)
       # char_length is number of wide chars (typically excluding NULLs), *not* bytes
-      str = get_bytes(0, char_length * 2).force_encoding('UTF-16LE')
-      str.encode('UTF-8', str.encoding)
+      str = get_bytes(0, char_length * 2).force_encoding(Encoding::UTF_16LE)
+      str.encode(Encoding::UTF_8, str.encoding)
     end
 
-    def read_wide_string_without_length
+    def read_wide_string_without_length(replace_invalid_chars: false)
       wide_character = get_bytes(0, 2)
+      wide_character.force_encoding(Encoding::UTF_16LE)
       i = 2
       str = []
 
-      while wide_character.encode('UTF-16LE') != END_OF_WCHAR_STRING
+      while wide_character != END_OF_WCHAR_STRING
         str << wide_character
         wide_character = get_bytes(i, 2)
+        wide_character.force_encoding(Encoding::UTF_16LE)
         i += 2
       end
-      str.join.force_encoding('UTF-16LE').encode('UTF-8')
+
+      if replace_invalid_chars
+        str.join.force_encoding(Encoding::UTF_16LE).encode(Encoding::UTF_8, Encoding::UTF_16LE, invalid: :replace)
+      else
+        str.join.force_encoding(Encoding::UTF_16LE).encode(Encoding::UTF_8)
+      end
     end
 
     def read_win32_bool
--- a/spec/facter/resolvers/windows/networking_spec.rb
+++ b/spec/facter/resolvers/windows/networking_spec.rb
@@ -201,9 +201,9 @@ describe Facter::Resolvers::Windows::Networking do
                        DnsSuffix: dns_ptr, FriendlyName: friendly_name_ptr, Flags: 0, Mtu: 1500,
                        FirstUnicastAddress: ptr, Next: ptr, to_ptr: FFI::Pointer::NULL)
       end
-      let(:ptr) { instance_spy(FFI::Pointer) }
-      let(:dns_ptr) { instance_spy(FFI::Pointer) }
-      let(:friendly_name_ptr) { instance_spy(FFI::Pointer) }
+      let(:ptr) { FFI::Pointer.new }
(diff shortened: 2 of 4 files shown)


spec/facter/resolvers/windows/networking_spec.rb
```diff
@@ -250,5 +250,52 @@
...
+    context 'when it successfully retrieved networking information but the interface has non-UTF-8 friendly name' do
+      let(:error_code) { NetworkingFFI::ERROR_SUCCES }
+      let(:adapter) do
+        OpenStruct.new(OperStatus: NetworkingFFI::IF_OPER_STATUS_UP, IfType: NetworkingFFI::IF_TYPE_ETHERNET_CSMACD,
+                       DnsSuffix: dns_ptr, FriendlyName: friendly_name_ptr, Flags: 0, Mtu: 1500,
+                       FirstUnicastAddress: ptr, Next: ptr, to_ptr: FFI::Pointer::NULL)
+      end
+      let(:ptr) { instance_spy(FFI::Pointer) }
+      let(:dns_ptr) { instance_spy(FFI::Pointer) }
+      let(:friendly_name_ptr) { instance_spy(FFI::Pointer) }
```
joshcooper, line 263: Based on https://stackoverflow.com/questions/57541170/utf16-big-endian-to-utf8-conversion-for-failed-for-0xdcf0, the two byte sequence `0xDC`, `0xF0` is not a valid unicode code point. You could set the Friendly Name to that and verify facter correctly replaces it with the replacement character `\uFFFD`. And rather then stubbing out the FFI pointer, just create one that points to the desired byte array. Note since we're using UTF-16LE (little endian) the byte array needs to be reversed below:

```ruby
let(:friendly_name_ptr) do
  bytes = "\xF0\xDC".force_encoding(Encoding::UTF_16LE).bytes
  ptr = FFI::MemoryPointer.new(:uchar, bytes.length)
  ptr.write_array_of_uchar(bytes)
  ptr
```
Then in the test when you call `resolver.resolve(:interfaces)`, the network resolver will call `adapter_address[:FriendlyName]`, which will return the `friendly_name_ptr` above and call `read_wide_string_without_length(replace_invalid_chars: true)` on it. Then in your test you can verify that:

```ruby
expect(resolved_interface.keys.first).to eql("\uFFFD")
```

joshcooper:
Hey @mhashizume I think the failing test is because the tests you added in commit 6e25035b29a81a83f5e377f11d6be2cbac8f671c are using `FFI::MemoryPointer` but need to use `FFI::Pointer` instead and call `stub_utf16le_bytes` to stub out the `get_bytes` calls. Might be easier to move your commit after my two commits and resolve conflicts. I can help you with that part?

spec/facter/resolvers/windows/networking_spec.rb
```diff
@@ -217,6 +218,15 @@
       end
       let(:domain) { 'my_domain' }
 
+      def stub_utf16le_bytes(ptr, str)
+        i = 0
+        str.encode(Encoding::UTF_16LE).each_char do |ch|
+          allow(ptr).to receive(:get_bytes).with(i, 2).and_return(ch)
+          i += 2
+        end
+        allow(ptr).to receive(:get_bytes).with(i, 2).and_return("\0\0")
```
joshcooper, line 227: The line I added isn't quite right. `"\0\0"` is a UTF-8 encoded string representing two null **characters**. We really should be returning a UTF-16LE string for a single null wide character like:

```suggestion
        allow(ptr).to receive(:get_bytes).with(i, 2).and_return("\0".encode(Encoding::UTF_16LE))
```

It just so happens that they have the same byte sequence:

```
irb(main):001:0> "\0".encode(Encoding::UTF_16LE).bytes == "\0\0".bytes
=> true
```

Or even better, pull the null wide character into a constant or let, like we do in puppet:
https://github.com/puppetlabs/puppet/blob/2f14ddd269c26cba36771d9b3464c2c29422e46a/lib/puppet/ffi/windows/api_types.rb#L23

spec/facter/resolvers/windows/networking_spec.rb
```diff
@@ -249,6 +260,14 @@
       it 'returns domain' do
         expect(resolver.resolve(:domain)).to eql(domain)
       end
+
+      it 'replaces invalid characters in the friendly name' do
+        allow(friendly_name_ptr).to receive(:get_bytes).with(0, 2).and_return(+"\xf0\xdc")
+        allow(friendly_name_ptr).to receive(:get_bytes).with(2, 2).and_return(+"\0\0")
```
joshcooper, line 266: Should be able to replace both lines above with a single call to:

```suggestion
        stub_utf16le_bytes(friendly_name_ptr, "\xf0\xdc")
```

You won't have to worry about the frozen string since `stub_utf16le_bytes` calls `str.encode(Encoding::UTF_16LE)` which makes a copy. Also it handles adding the null terminator.

lib/facter/resolvers/windows/ffi/ffi.rb
```diff
@@ -24,26 +24,33 @@
 
 module FFI
   WIN32FALSE = 0
-  END_OF_WCHAR_STRING = "\0\0".encode('UTF-16LE')
+  END_OF_WCHAR_STRING = (+"\0\0").force_encoding(Encoding::UTF_16LE)
```
joshcooper, line 27:
```suggestion
  END_OF_WCHAR_STRING = "\0".encode(Encoding::UTF_16LE).freeze
```

mhashizume: I'm not sure I understand why this is necessary, isn't that already implicitly frozen from `# frozen_string_literal: true` on line 1?

mhashizume:
Accidentally created some commit wonkiness through bad rebasing, trying to fix before merge

joshcooper (approved):

PR #2491, opened 2022-05-06
(PA-4486) Update Networking Facts test

cthorn42:
In GCE the VMs we use can have a `networking.netmask` of 255.255.255.255.
This causes failure in the networking_facts.rb test, this PR updates the regex
to allow for such an IP.


joshcooper (approved):

joshcooper:
@cthorn42 I think you'll need to cherry-pick to `main` since we don't auto mergeup (as the 3.x and main code bases are so different)

PR #2492, opened 2022-05-10
(FACT-3121) Detect Illumos (SmartOS/OmniOS) LX virtualization

smokris:
See https://tickets.puppetlabs.com/browse/FACT-3121.

Facter uses `virt-what`'s output which is `lxc` (on `virt-what` 1.20) or null (on `virt-what` 1.21+), so I added `check_illumos_lx` prior to `retrieve_from_virt_what`.

### DIFF
--- a/spec/facter/util/facts/posix/virtual_detector_spec.rb
+++ b/spec/facter/util/facts/posix/virtual_detector_spec.rb
@@ -74,6 +74,7 @@ describe Facter::Util::Facts::Posix::VirtualDetector do
         allow(Facter::Resolvers::Containers).to receive(:resolve).with(:vm).and_return(nil)
         allow(Facter::Resolvers::Freebsd::Virtual).to receive(:resolve).with(:vm).and_return(nil)
         allow(Facter::Resolvers::Linux::DmiBios).to receive(:resolve).with(:bios_vendor).and_return(nil)
+        allow(Facter::Resolvers::Uname).to receive(:resolve).with(:kernelversion).and_return(nil)
         allow(Facter::Resolvers::VirtWhat).to receive(:resolve).with(:vm).and_return(value)
       end
 
@@ -96,6 +97,7 @@ describe Facter::Util::Facts::Posix::VirtualDetector do
         allow(Facter::Resolvers::Freebsd::Virtual).to receive(:resolve).with(:vm).and_return(nil)
         allow(Facter::Resolvers::Linux::DmiBios).to receive(:resolve).with(:bios_vendor).and_return(nil)
         allow(Facter::Resolvers::VirtWhat).to receive(:resolve).with(:vm).and_return(nil)
+        allow(Facter::Resolvers::Uname).to receive(:resolve).with(:kernelversion).and_return(nil)
         allow(Facter::Resolvers::Vmware).to receive(:resolve).with(:vm).and_return(value)
       end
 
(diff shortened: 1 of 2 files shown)


spec/facter/util/facts/posix/virtual_detector_spec.rb
```diff
@@ -241,6 +241,19 @@
       end
     end
 
+    context 'when smartos-lx' do
+      let(:value) { 'smartos-lx' }
+
+      before do
+        allow(Facter::Resolvers::VirtWhat).to receive(:resolve).with(:vm).and_return('lxc')
```
bahamat, line 248: Hi, I’m one of the maintainers for SmartOS. In general, I support this change to facter.

But, I’m not sure returning `lxc` here is the best option. SmartOS LX brand zones are not the same as LXC, and we are not attempting to emulate LXC. This may cause some programs to attempt to use LXC facilities that won’t be present on SmartOS and/or not use facilities present (e.g., `mdata-*` commands). If possible, I would suggest returning `lx` here.

smokris: Thanks, @bahamat.  This MR intends to do (almost) that (please see [this change](https://github.com/puppetlabs/facter/pull/2492/files#diff-6a457370eec506d706714e7494e9f6debeb0e914837f76224b2eef0ad66a2e69R29).  Currently, `facter` returns `lxc` (passing through the value from `virt-what`), whereas this MR changes it to return `smartos-lx`.  (I figured the `smartos-` prefix would make it clearer than simply `lx`.)

The `spec` (automated test) line you commented on is just mocking the return value of `virt-what`.

bahamat: Ok, that makes sense. Sorry for the confusion.

bahamat (approved):
Hi, I’m one of the maintainers for SmartOS. It looks to me like the behavior presented here is desirable from our perspective. (Though, I have not evaluated this code functionally, I’ll leave that to those who are more familiar).

spec/facter/util/facts/posix/virtual_detector_spec.rb
smokris: In a6107cb, I renamed the identifier from `smartos-lx` to `illumos-lx` based on [this discussion](https://github.com/TritonDataCenter/illumos-joyent/pull/415#issuecomment-1123304950).

smokris:
Hi, @joshcooper.  I saw that you merged some PRs in this repo over the last few months.  Are you available to review this PR sometime soon?

mhashizume:
Closing and reopening to retrigger tests

mhashizume:
Hi @smokris , thank you for your contribution and sorry for the delay.

Would you be able to rebase you PR and resolve the conflict?

Thank yoU!

smokris:
@mhashizume, yes, I pushed d959b40 a few minutes ago, in which I've rebased onto current `main`, and squashed the 2 commits into 1 (since there's no need to preserve the name-change commit).  Investigating test failures now…

smokris:
GitHub CI is running the tests in a different random order each time, and the tests were failing intermittently depending on the order they ran.  For example, `bundle exec rspec --order random:47312` consistently failed, while `bundle exec rspec --order random:36802` consistently passed.

If I understand correctly, the problem was that, if the test calls a resolver that isn't mocked, then the resolver will execute on the current system and cache its results (rather than merely returning the mocked results without caching), and the cached results break subsequent tests (even if mocked).

I pushed 430d549, which I think adds all the necessary mocks to enable the tests to pass regardless of order.

mhashizume (approved):
Thank you for your contribution @smokris !

mhashizume:
Closing and reopening to pick up new check

PR #2495, opened 2022-06-08
(maint) Update Networking Facts test

cthorn42:
In GCE the VMs we use can have a `networking.netmask` of 255.255.255.255.
This causes failure in the networking_facts.rb, and networking_facts_with_secondary_ips.rb test, this PR updates the regex
to allow for such an IP.
Also we need to have some small changes to the dmi test because if we use a Google Cloud VM
some default DMI settings are changed.


acceptance/tests/facts/dmi.rb
```diff
@@ -20,12 +20,12 @@
...
       expected_facts.merge!({'dmi.bios.release_date' => /\d+\/\d+\/\d+/,
                              'dmi.bios.vendor'       => /\w+/,
-                             'dmi.bios.version'      => /\d+/,
+                             'dmi.bios.version'      => /(\d+|Google)/,
                              'dmi.chassis.type'      => /\w+/,
                             })
     end
     unless agent['platform'] =~ /windows|cisco|aarch64|el-/
-      expected_facts.merge!({'dmi.chassis.asset_tag' => /\w+/})
+      expected_facts.merge!({'dmi.chassis.asset_tag' => /(\w+|)/})
```
cthorn42, line 28: This is looking to have either a word, or nothing at all. I'm kind of worried that this might mask a true failure at some point in the future.

joshcooper: Ah I didn't understand this part? It looks like `asset_tag` is a UUID for GCE VMs:

```
  * verify that dmi structured fact contains facts
    
    sullen-eyeball.vmpooler-prod.puppet.net (sullen-eyeball.vmpooler-prod.puppet.net) 14:19:57$ facter --json dmi
    ...
      {
        "dmi": {
          "bios": {
            "release_date": "01/01/2011",
            "vendor": "Google",
            "version": "Google"
          },
          "board": {
            "asset_tag": "A5AF127A-C3A5-EB4D-FF93-7A761A07E1E1",
            "manufacturer": "Google",
            "product": "Google Compute Engine",
            "serial_number": "Board-GoogleCloud-A5AF127AC3A5EB4DFF937A761A07E1E1"
          },
```

And the UUID should match `\w+`. However, one nit, the `-` character doesn't match, so our regexp only matches the first component of the UUID:

```ruby
irb(main):003:0> "A5AF127A-C3A5-EB4D-FF93-7A761A07E1E1".match(/(\w+)/)
=> #<MatchData "A5AF127A" 1:"A5AF127A">
```

So maybe specify the regexp as `/([\w-]+)/` to match the entire UUID (word or dash characters one or more times):

```ruby
irb(main):004:0> "A5AF127A-C3A5-EB4D-FF93-7A761A07E1E1".match(/([\w-]+)/)
=> #<MatchData "A5AF127A-C3A5-EB4D-FF93-7A761A07E1E1" 1:"A5AF127A-C3A5-EB4D-FF93-7A761A07E1E1">
```

joshcooper:
nit the commit check is failing because of the `PA` JIRA ticket. Maybe just add `(maint) ...` and then reference the PA ticket in the body?

acceptance/tests/facts/dmi.rb
cthorn42: I'm a bit confused on this, you're looking at the `dmi.board.asset_tag`, but what the test looks at is `dmi.chassis.asset_tag`. Or are those the same thing?

joshcooper: Ohhh, I completely missed that. So I'm seeing the `dmi.chassis.asset_tag` fact is missing: 

```
        "dmi": {
          ...
          },
          "chassis": {
            "type": "Other"
          },
```

And it appears facter is doing the "right thing" since the asset tag for the chassis is not specified:

```
root@sullen-eyeball:~# dmidecode --type chassis
# dmidecode 3.3
Getting SMBIOS data from sysfs.
SMBIOS 2.4 present.

Handle 0x0099, DMI type 3, 20 bytes
Chassis Information
...
	Asset Tag: Not Specified
```

Checking that the fact matches nothing is a little surprising to me. It might be clearer to explicitly exclude GCE VMs (as I'm assuming this will be an issue as we migrate VMs to GCE). Maybe exclude the `dmi.chassis.asset_tag` check based on the `virtual` fact:

```
root@sullen-eyeball:~# facter virtual
gce
```

You can do something like this to check the value on the host:

```ruby
virtual = on(agent, facter('virtual')).stdout.chomp
```

joshcooper (approved):
Verified this works using:

```
$ export BEAKER_VERSION=https://github.com/voxpupuli/beaker#master
$ bundle update
...
$ SHA=20a2de087552b2b6201aa922042eafa39f28497b setup.sh ubuntu2204-64a 
...
$ bundle exec beaker exec tests/facts/dmi.rb,tests/facts/networking_facts.rb,tests/facts/networking_facts_with_secondary_ips.rb
```

PR #2496, opened 2022-06-23
(FACT-3102) Use non-deprecated YAML safe loading

ekohl:
This PR suppresses the following `Psych.safe_load` args warn when using Psych 3.1.0 (Ruby 2.6+).

    Passing permitted_classes with the 2nd argument of Psych.safe_load is deprecated. Use keyword argument like Psych.safe_load(yaml, permitted_classes: ...) instead.

This is needed because in Psych 4 (included in Ruby 3.1) it is no longer allowed to use the non-keyword style.

An alternative to https://github.com/puppetlabs/facter/pull/2484

### DIFF
--- a/lib/facter/custom_facts/util/parser.rb
+++ b/lib/facter/custom_facts/util/parser.rb
@@ -70,7 +70,11 @@ module LegacyFacter
         def parse_executable_output(output)
           res = nil
           begin
-            res = YAML.safe_load(output, [Symbol, Time])
+            res = if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0') # Ruby 2.6+
+                    YAML.safe_load(output, permitted_classes: [Symbol, Time])
+                  else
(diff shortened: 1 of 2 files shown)


ekohl:
Perhaps I should have used FACT-3123 instead.

lib/facter/custom_facts/util/parser.rb
```diff
@@ -70,7 +70,11 @@ def parse_results
         def parse_executable_output(output)
           res = nil
           begin
-            res = YAML.safe_load(output, [Symbol, Time])
+            res = if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0') # Ruby 2.6+
```
bastelfreak, line 73: We used the exact fix in other places, for example bolt. :+1: from my side.

bastelfreak (approved):

joshcooper (approved):

PR #2497, opened 2022-06-23
(FACT-3130) Add ipaddress_* ipaddress6_* facts on FreeBSD

smortex:


smortex:
FYI, this patch is included in the FreeBSD package of facter:
https://github.com/freebsd/freebsd-ports/blob/3d5c9e55165378b7554b5d4c920a2a64bcb86b72/sysutils/rubygem-facter/Makefile#L8

joshcooper (approved):
At first I was wondering why these are copy/pasted from `lib/facter/facts/linux/*` and then remembered `Linux` and `Bsd` are in different hierarchies https://github.com/puppetlabs/facter/blob/7fda3bb695d70354a2d1654daa391c63ee885d39/lib/facter/config.rb#L8-L49

PR #2501, opened 2022-07-07
Resolve DMI ChassisType for Windows too

yakatz:
We use the `dmi.chassis.type` fact (as a poor way) to determine whether a system in compatible with Dell server management tools. No reason it shouldn't be available on Windows.

### DIFF
--- a/lib/facter/resolvers/windows/dmi_systemenclosure.rb
+++ b/lib/facter/resolvers/windows/dmi_systemenclosure.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class DMISystemEnclosure < BaseResolver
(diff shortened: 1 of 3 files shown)


lib/facter/resolvers/windows/dmi_systemenclosure.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class DMISystemEnclosure < BaseResolver
+      @log = Facter::Log.new(self)
```
joshcooper, line 6: It's no longer recommended/needed to create a log instance, see 016811658a536f5c9c90d53e56fd7260f470bb8d Could you rebase on main and drop this line?

joshcooper:
Going to close and reopen to rekick tests

PR #2507, opened 2022-07-14
(maint) Update os.macosx.version.full fact

cthorn42:
MacOS M1 full version is only major and minor version, the patch version is
dropped. This PR fixes the acceptance test for that.

### DIFF
--- a/acceptance/lib/facter/acceptance/base_fact_utils.rb
+++ b/acceptance/lib/facter/acceptance/base_fact_utils.rb
@@ -276,7 +276,11 @@ module Facter
           expected_facts['os.macosx.version.full'] = /#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}/
         else
           expected_facts['os.macosx.version.patch'] = /\d+/
-          expected_facts['os.macosx.version.full'] = /^#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}\.#{expected_facts['os.macosx.version.patch']}$/
+          if agent['platform'] =~ /arm64/
+            expected_facts['os.macosx.version.full'] = /^#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}$/
+          else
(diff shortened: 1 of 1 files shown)


acceptance/lib/facter/acceptance/base_fact_utils.rb
```diff
@@ -276,7 +276,11 @@ def osx_expected_facts(agent)
           expected_facts['os.macosx.version.full'] = /#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}/
         else
           expected_facts['os.macosx.version.patch'] = /\d+/
-          expected_facts['os.macosx.version.full'] = /^#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}\.#{expected_facts['os.macosx.version.patch']}$/
+          if agent['platform'] =~ /arm64/
+            expected_facts['os.macosx.version.full'] = /#{expected_facts['os.macosx.version.major']}\.#{expected_facts['os.macosx.version.minor']}/
```
tvpartytonight, line 280: Does this regex need a `^` and `$`?

cthorn42: It doesn't require it, test passed with it as it is.

tvpartytonight: The test passes but the regex is very lenient, so `10.16` and also `234.10.16.234` would also pass`/10.16/`. It seems like there should be parity with this check with the previous one that is here.

Why is the `patch` version omitted as well? I just looked at the versions for Monterey, and there are both `12.0` major-minor versions along with `12.2.1` major-minor-patch versions.

cthorn42: Good point, updated the regex.
For the patch not being there, that is because I believe the full version fact is based off of the product version from the sw_vers command:
```
pix-arm64-macos12-4:bin root# sw_vers
ProductName:	macOS
ProductVersion:	12.4
BuildVersion:	21F79
```

On x86_64 mac that command:
```
lazy-icebox:~ root# sw_vers
ProductName:	macOS
ProductVersion:	12.2.1
BuildVersion:	21D62
```

PR #2513, opened 2022-08-11
(FACT-3135) Add support for yaml anchors in Facter 4

AriaXLi:


lib/facter/custom_facts/util/parser.rb
```diff
@@ -119,9 +119,9 @@ def parse_results
           cont = content.gsub(TIME, '"\1"')
 
           if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0') # Ruby 2.6+
-            YAML.safe_load(cont, permitted_classes: [Date])
+            YAML.safe_load(cont, permitted_classes: [Date], aliases: true)
           else
-            YAML.safe_load(cont, [Date])
+            YAML.safe_load(cont, [Date], aliases: true)
```
joshcooper, line 124: On older ruby versions, we'll need to pass this parameter as a positional argument. Here's the older API signature https://ruby-doc.org/stdlib-2.5.0/libdoc/psych/rdoc/Psych.html#method-c-safe_load

AriaXLi: Ahh okay, thank you for catching that, I had no idea

joshcooper:
Ah sorry forgot to mention it'd be good to add a unit test for this. For example, there's a fixture for dates in external yaml facts https://github.com/puppetlabs/facter/blob/main/spec/fixtures/external_fact_yaml_date and there's a unit test to verify we can load the external fact.

AriaXLi:
> Ah sorry forgot to mention it'd be good to add a unit test for this. For example, there's a fixture for dates in external yaml facts https://github.com/puppetlabs/facter/blob/main/spec/fixtures/external_fact_yaml_date and there's a unit test to verify we can load the external fact.

Ah okay, I'll add that test too to this PR!

mhashizume (approved):

PR #2517, opened 2022-08-25
(FACT-3133) allow `echo` for Windows custom facts

tvpartytonight:


lib/facter/custom_facts/core/execution/windows.rb
```diff
@@ -8,7 +8,14 @@ def search_paths
...
         DEFAULT_COMMAND_EXTENSIONS = %w[.COM .EXE .BAT .CMD].freeze
 
+        def aliased_command?(command)
+          _output, status = Open3.capture2("powershell.exe Get-Alias #{command}")
+          status.success?
+        end
+
         def which(bin)
+          return bin if aliased_command?(bin)
+
```
joshcooper, line 18: My understanding is that facter 3.x special cases `echo` to be a builtin on Windows. Originally the change was made in facter 3.x, see https://github.com/puppetlabs/facter/pull/1010 and then later the code was moved to [leatherman](https://github.com/puppetlabs/leatherman/commit/281a95ed5f2c034dd390b9585237b1353bb68c40). I'd be inclined to do the same here instead of invoking `powershell`.

Also note the default shell on Windows (as defined in leatherman) is [`cmd.exe`](https://github.com/puppetlabs/leatherman/blob/2703f1196680ae6cb586e4d77af3ce01c6d88b97/execution/src/windows/execution.cc#L37), so I think we want to avoid any powershell-isms.

tvpartytonight: ah, thanks for giving some context for this; I sort of naively assumed that it was ok to use powershell since the [docs](https://puppet.com/docs/puppet/7/install_agents.html#install_windows_agents) state that powershell 5.0 is required. I'm happy to update the PR to special case `echo`, but I wrote it this was because I thought it might become a game of whack-a-mole with anything else that might be aliased this way.

joshcooper (approved):

PR #2518, opened 2022-08-29
(maint) only check `builtin_command?` for posix

tvpartytonight:
The `builtin_command?` only is useful in posix, where the `type`
function is defined. Windows Powershell does not have `type` defined.


tvpartytonight:
This is related to #2517 ; I didn't want to change some of the [logic](https://github.com/puppetlabs/facter/blob/fd9e6ca0df5fe4a25b9c43372811233078a941ff/lib/facter/custom_facts/core/execution/base.rb#L48) here or change the `expand` default set, so I moved the logic to the `which` in that PR and just submitted this patch separately to keep it cleaner.

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -138,9 +138,9 @@ def log_stderr(msg, command, logger)
           logger.debug(format(STDERR_MESSAGE, command, msg.strip))
         end
 
-        def builtin_command?(command)
-          output, _status = Open3.capture2("type #{command}")
-          output.chomp =~ /builtin/ ? true : false
+        # optional function to be defined in subclass if necessary
+        def builtin_command?(_command_)
+          nil
```
smortex, line 143: Maybe keep `false` for consistency?  Also the "Ruby way" for unused variable is to name them `_variable`.

Overall LGTM.

joshcooper:
@tvpartytonight is this still needed? If so, I agree with @smortex's comments.

joshcooper:
Coming back to this, is this still needed? It looks like PowerShell can query if a command is a function like https://superuser.com/questions/1648302/powershell-equivalent-of-unix-shell-buitin-type But I'm not sure that'd be useful here?

PR #2521, opened 2022-09-09
(FACT-3078) Accept relative load path for --custom-dir

smortex:


joshcooper (approved):
This behavior was introduced in https://tickets.puppetlabs.com/browse/FACT-2043 in an attempt to maintain compatibility with Facter 2.x, but the behavior in 3.x had changed. This restores the behavior that existed in 3.x:

```
❯ cat custom/custom.rb 
Facter.add(:custom) do
  setcode { true }
end

❯ bx facter --custom-dir custom custom
true

❯ gco main
Switched to branch 'main'
Your branch is up to date with 'upstream/main'.

❯ bx facter --custom-dir custom custom

❯
```

joshcooper:
jenkins please test this with redhat9-64a,windows2022-64a

PR #2523, opened 2022-09-16
(FACT-3149) Fix memory usage reporting on FreeBSD

smortex:
* In vmstat(8), "free memory" correspond to unused memory that does not
  contain any data, and does not include cache / inactive memory which
  has some data but is immediatly available to the system if memory is
  needed.
  Prefer to get active and wired memory page count form sysctl(3) which
  correpond to the used memory that can (active) and cannot (wired) be
  swapped out, and multiply these by the size of a memory page.  This
  give a better overview of the current memory usage.
* Prefer `vm.stats.vm.v_page_count` over `hw.physmem`:  The available
  memory can be lower that what is physically present in the system and
  the free / used memory is a portion of the usable memory more than a
  portion of the physical memory.

Description of the sysctl entries:
* `vm.stats.vm.v_page_count`: Total number of pages in system
* `vm.stats.vm.v_active_count`: Active pages
* `vm.stats.vm.v_wire_count`: Wired pages
* `vm.stats.vm.v_page_size`: Page size in bytes

### DIFF
--- a/lib/facter/resolvers/freebsd/system_memory.rb
+++ b/lib/facter/resolvers/freebsd/system_memory.rb
@@ -15,25 +15,33 @@ module Facter
 
           def calculate_system_memory(fact_name)
             read_total_memory_in_bytes
-            read_available_memory_in_bytes
+            read_used_memory_in_bytes
 
-            @fact_list[:used_bytes] = @fact_list[:total_bytes] - @fact_list[:available_bytes]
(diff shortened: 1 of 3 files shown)


smortex:
Rebased to fix conflicts

joshcooper:
@smortex sorry for the delay, could you rebase on `main`?

smortex:
@joshcooper rebased without conflict and CI is green!

joshcooper:
Thank you @smortex!

joshcooper (approved):

PR #2525, opened 2022-10-04
(FACT-3154) Adds Ruby 3.2 to rspec tests

mhashizume:
In preparation for Puppet 8, which will likely use Ruby 3.2, this commit adds Ruby 3.2.0-preview1 to the rspec test matrix.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -21,7 +21,7 @@ jobs:
           - '2.3'
           - '2.7'
           - '3.0'
-          - '3.1'
+          - '3.2.0-preview2'
           - 'jruby-9.3.7.0'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


mhashizume:
This PR does not add Ruby 3.2.0 preview builds to Windows tests as they are not yet available through the `setup-ruby` action.

.github/workflows/unit_tests.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
           - '2.3'
           - '2.7'
           - '3.0'
-          - '3.1'
+          - '3.2.0-preview2'
```
joshcooper, line 24: I think the branch protection rules are still requiring 3.1, so those will need to be updated.

Also wondering if we can add 3.1 for Windows, until the `3.2.0-preview2` version becomes available? And update branch protection rules.

mhashizume: I've added Ruby 3.1 to the Windows testing matrix and updated the branch protection rules accordingly

joshcooper (approved):

PR #2526, opened 2022-10-08
(FACT-3159) Fix install.rb for Ruby 3.0

jcharaoui:


install.rb
```diff
@@ -54,7 +54,7 @@ def do_configs(configs, target, strip = 'ext/')
     Dir.mkdir(target) unless File.directory? target
     configs.each do |cf|
       ocf = File.join(InstallOptions.config_dir, cf.gsub(/#{strip}/, ''))
-      install(cf, ocf, { mode: 0o644, preserve: true, verbose: true })
+      install(cf, ocf, mode: 0o644, preserve: true, verbose: true)
```
joshcooper, line 57: Do older ruby versions accept the keyword argument form of `FileUtils.install`?

jcharaoui: From what I can tell, this works for Ruby 2.7 and up. The old form (using the last argument as keyword parameter) has been deprecated since Ruby 2.7. The details around that change are explained here: [https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/)

joshcooper: yep yep, makes sense. Looks like it works in 2.5.8 too: 

```
irb(main):001:0> require 'fileutils'
=> true
irb(main):002:0> FileUtils.install('Gemfile', '/tmp/Gemfile', mode: 0o644, preserve: true, verbose: true)
install -c -p -m 644 Gemfile /tmp/Gemfile
=> nil
irb(main):003:0> RUBY_VERSION
=> "2.5.8"
```

joshcooper:
Hi @jcharaoui thank you for your contribution! We recently updated our rspec checks to test on newer ruby versions, but your PR was branched before then. Could you rebase your changes onto upstream/main? Also nit, could you amend your commit summary to include why this change is necessary?

jcharaoui:
All done @joshcooper HTH

joshcooper (approved):

PR #2527, opened 2022-10-11
Fix FACT-3151

fxpires:
Fix for https://tickets.puppetlabs.com/browse/FACT-3151. Allows comments on /etc/os-release file.

### DIFF
--- a/lib/facter/resolvers/os_release.rb
+++ b/lib/facter/resolvers/os_release.rb
@@ -49,7 +49,7 @@ module Facter
 
           pairs = []
           content.each do |line|
-            pairs << line.strip.delete('"').split('=', 2)
+            pairs << line.strip.delete('"').split('=', 2) unless line.start_with?('#')
           end
 
           pairs


joshcooper:
Hi @fxpires thank you for your contribution! Could you add a test to `spec/facter/resolvers/os_release_spec.rb`? Also please amend your commit summary with `(FACT-3151)` and force push to your feature branch?

smortex:
@fxpires your tests should go into `spec/facter/resolvers/os_release_spec.rb`. a good start might be to add a fixture file `spec/fixtures/os_release_*` with some problematic content you copy from a system where the code does not work as expected. Then add tests for your use-case.

Regarding rewording your commit, from your working directory run:

```
git rebase -i origin/main # Interactively rewrite history
```

This will bring your editor listing all commits in your branch.  Replace `pick` with `reword` for all commits you want to fix, then save and quit, your editor will pop again for each commit and allow you to rewrite them.

Then push your updated branch:

```
git push -f            # Send the changes (-f is required because we re-wrote history)
```

AriaXLi:
Hi @fxpires! I added a spec test for your commit in #2563 and merged both your commit and the test and made sure to credit you as a co-author so I'm going to close out this PR now. Thank you again for your contribution!

PR #2528, opened 2022-10-16
(FACT-3160) Fix size calculations for AIX blockdevice and partitions

loopway:
* disks.rb: TOTAL PPs = USED PPs + FREE PPs. Therefor adding FREE PPs to TOTAL PPs produces incorrect disk sizes.

* partitions.rb: Using physical_partitions (PPs) to calculate LV sizes are not suitable if there is more than one mirror (can be two or three). LPs * nr_of_mirrors = PPs.

* info_extractor.rb: Adjusted regexp because using property LPs also matched MAX LPs.

### DIFF
--- a/lib/facter/resolvers/aix/disks.rb
+++ b/lib/facter/resolvers/aix/disks.rb
@@ -47,7 +47,7 @@ module Facter
           end
 
           def compute_size(size_hash)
-            physical_partitions = size_hash['TOTAL PPs'].to_i + size_hash['FREE PPs'].to_i
+            physical_partitions = size_hash['TOTAL PPs'].to_i
             size_physical_partition = size_hash['PP SIZE']
             exp = if size_physical_partition[/mega/]
                     Facter::Util::Aix::InfoExtractor::MEGABYTES_EXPONENT
--- a/lib/facter/resolvers/aix/partitions.rb
+++ b/lib/facter/resolvers/aix/partitions.rb
@@ -56,14 +56,14 @@ module Facter
           end
 
           def compute_size(info_hash)
-            physical_partitions = info_hash['PPs'].to_i
+            logical_partitions = info_hash['LPs'].to_i
             size_physical_partition = info_hash['PP SIZE']
             exp = if size_physical_partition[/mega/]
                     Facter::Util::Aix::InfoExtractor::MEGABYTES_EXPONENT
                   else
                     Facter::Util::Aix::InfoExtractor::GIGABYTES_EXPONENT
                   end
-            size_physical_partition.to_i * physical_partitions * exp
+            size_physical_partition.to_i * logical_partitions * exp
           end
         end
       end
--- a/lib/facter/util/aix/info_extractor.rb
+++ b/lib/facter/util/aix/info_extractor.rb
@@ -61,7 +61,7 @@ module Facter
           properties = PROPERTIES[cmd]
           properties.each do |property|
             str = (properties - [property]).join('|')
-            matcher = content.match(/#{Regexp.escape(property)}([^\n]*?)(#{str}|\n|$)/s)
+            matcher = content.match(/(?:^|^[^:]+:[^:]+)#{Regexp.escape(property)}([^\n]*?)(#{str}|\n|$)/s)
             if matcher
               value = matcher[1].strip
               property_hash[property.split(':').first] = value


smortex:
I guess you'll have to update the tests accordingly in `spec/facter/facts/aix/disks_spec.rb` and `spec/facter/facts/aix/partitions_spec.rb`?

loopway:
@smortex I don't think it's necessary, since the sizes in the specs are specified without using nr. of LP's and PP's. Tests would have only failed in previous versions if mirrored LV's have been used and/or there was free space (PPs) in the VGs.

joshcooper:
@loopway thank you for your pull request. Could you rebase your changes on top of `main`?

lib/facter/resolvers/aix/disks.rb
```diff
@@ -47,7 +47,7 @@ def find_size(name)
           end
 
           def compute_size(size_hash)
-            physical_partitions = size_hash['TOTAL PPs'].to_i + size_hash['FREE PPs'].to_i
+            physical_partitions = size_hash['TOTAL PPs'].to_i
```
joshcooper, line 50: Is this covered by an existing test? If not, can you add one?

lib/facter/resolvers/aix/partitions.rb
```diff
@@ -56,14 +56,14 @@ def populate_from_lslv(name)
...
-            physical_partitions = info_hash['PPs'].to_i
+            logical_partitions = info_hash['LPs'].to_i
             size_physical_partition = info_hash['PP SIZE']
             exp = if size_physical_partition[/mega/]
                     Facter::Util::Aix::InfoExtractor::MEGABYTES_EXPONENT
                   else
                     Facter::Util::Aix::InfoExtractor::GIGABYTES_EXPONENT
                   end
-            size_physical_partition.to_i * physical_partitions * exp
+            size_physical_partition.to_i * logical_partitions * exp
```
joshcooper, line 66: I'm not sure I understand this part, don't we want to report on the physical partition size?

loopway: If we want to report the size of the partition (file system) correctly, we need to only consider the number of logical partitions. For example a 2GB `/usr` file system resides on the logical volume `/dev/hd2` which has:
- a `PP Size` of 512MB
- 4 `LP`s (4x 512MB = 2048MB)
- 2 `copies` (1 mirror), therefor 2x 4 `LP`s = 8 `PP`s
- 8 `PP`s (2x 4x 512MB = 4096MB)
The partition/file system size should be reported as 2GB (4 `LP`s) and not 4GB (8 `PP`s). As long as logical volume has no additional copies (>1), we could obviously use either `PP`s or `LP`s. But for mirrored volumes we must use the number of `LP`s. Hope this makes  sense and explains it. For a different phrasing see also the [official IBM AIX Documentation](https://www.ibm.com/docs/en/aix/7.2?topic=lvsc-logical-partitions)

lib/facter/resolvers/aix/disks.rb
loopway: I unfortunately do not have any experience with Rspec, but looking at [aix/partitions_spec.rb](https://github.com/puppetlabs/facter/blob/main/spec/facter/facts/aix/partitions_spec.rb) and [aix/mountpoints_spec.rb](https://github.com/puppetlabs/facter/blob/main/spec/facter/facts/aix/mountpoints_spec.rb), it seems to me that the expected size values are checked against MiB/GiB. So as long as the test environment did not use additional logical volume copies (mirrors), these tests would have succeed in the past. As far as I can tell the tests will also work with the proposed code changes in this pull request for mirrored logical volumes.

joshcooper:
retriggering actions

lib/facter/resolvers/aix/disks.rb
joshcooper: This makes sense. I have no idea why were adding free physical partitions to the total

lib/facter/resolvers/aix/partitions.rb
joshcooper: Ah thanks for the example! So if we're mirroring, then from a logical perspective we have 2GB, but from physical perspective we have 4GB, and you're saying facter should report the former. That makes sense to me.

joshcooper:
@loopway did facter 3.x (in puppet-agent 5.x) correctly report `blockdevice_hdisk*_size` and `partitions./dev/*.size` facts?

The reason I ask is because facter 3.x on AIX called odm to collect this information: https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/src/facts/aix/disk_resolver.cc#L18-L69 But in my testing facter 3.x incorrectly reported the amount of used disk space, and therefore capacity. For example, given

```
# df -Im /    
Filesystem    MB blocks      Used      Free %Used Mounted on
/dev/hd4        2048.00    288.61   1759.39   15% /
```

Facter 3 incorrectly reports used as `2.00 GiB`:

```
mountpoints => {
  / => {
    available => "1.77 GiB",
    available_bytes => 1898340352,
    capacity => "53.08%",
    device => "/dev/hd4",
    filesystem => "jfs2",
    options => [
      "rw",
      "log=/dev/hd8"
    ],
    size => "2.00 GiB",
    size_bytes => 2147483648,
    used => "2.00 GiB",
    used_bytes => 2147483648
  }
```

While Facter 4 correctly reports `288.59 MiB`:

```
mountpoints => {
  / => {
    available => "1.72 GiB",
    available_bytes => 1844871168,
    capacity => "14.09%",
    device => "/dev/hd4",
    filesystem => "jfs2",
    options => [
      "rw",
      "log=/dev/hd8"
    ],
    size => "2.00 GiB",
    size_bytes => 2147483648,
    used => "288.59 MiB",
    used_bytes => 302612480
  },
```

joshcooper:
@loopway I submitted a new PR with your original commit along with a test fix, so I'm going to close this PR. Thank you for your contribution!

PR #2529, opened 2022-10-21

buzzdeee:
Most inspiration taken from FreeBSD, rest from other OS, and some own creations. Not all core facts supported, but makes puppet facts works well, as well as puppet agent --test

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -54,7 +54,7 @@ class Installer
     Dir.mkdir(target) unless File.directory? target
     configs.each do |cf|
       ocf = File.join(InstallOptions.config_dir, cf.gsub(/#{strip}/, ''))
-      install(cf, ocf, { mode: 0o644, preserve: true, verbose: true })
+      install(cf, ocf, mode: 0o644, preserve: true, verbose: true)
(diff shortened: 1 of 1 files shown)


buzzdeee:
just this one RuboCop left, where I don't know, how to fix that complexity requirement.

joshcooper:
Hi @buzzdeee thank you for your contribution!

The failing cops can be configured in https://github.com/puppetlabs/facter/blob/d84fea9ad320a85b405f8dfe6f8a981b7e29cb89/.rubocop.yml#L74-L95 I don't see much value in those cops, so I'd be fine if you disabled them like:

```yaml
Metrics/PerceivedComplexity:
  Enabled: false

Metrics/CyclomaticComplexity:
  Enabled: false
```

and then removed the specific `Exclude`s

Also it would be good to have unit test coverage for this files. I think you could copy most of what's in spec/unit/facts/freebsd

buzzdeee:
GRR, messed with the branch... and it auto-closed, new: https://github.com/puppetlabs/facter/pull/2531

PR #2530, opened 2022-10-25
/etc/os-release parsing: strip comments and blank lines

sandynomad:
In the function "read_and_parse_os_release_file" the "pairs" variable is constructed with a very simple parse of the file. This patch will cleanup the content of os-release by: stripping white space from either side of each line string; discarding lines that are either empty or start with a '#'.

This is required as Arista EOS is based on CentOS 7 with customisations. In /etc/os-release there are comments that cause facter to throw an error.

### DIFF
--- a/lib/facter/resolvers/os_release.rb
+++ b/lib/facter/resolvers/os_release.rb
@@ -48,7 +48,11 @@ module Facter
           return nil if content.empty?
 
           pairs = []
-          content.each do |line|
+          content.map {
+                | line | line.strip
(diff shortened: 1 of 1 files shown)


joshcooper:
Thank you for your contribution @sandynomad I believe this has already been fixed in 30325d0d2754056a5c2963075931d7506c9d3566 (ignoring comments and striping white space), so I'm going to close this. Please reopen and rebase against `main` if I'm misunderstanding.

PR #2531, opened 2022-10-26

buzzdeee:
Most inspiration taken from FreeBSD, rest from other OS, and some own creations. Not all core facts supported, but makes puppet facts works well, as well as puppet agent --test

also disable RuboCop Metrics/PerceivedComplexity
and Metrics/CyclomaticComplexity
as suggested by @joshcooper


buzzdeee:
new after messing up: https://github.com/puppetlabs/facter/pull/2529 #2529

joshcooper:
I think the CI failures are due to a change in behavior with rspec-mocks. We had a similar problem in puppet, so I'll take a look.

buzzdeee:
thanks @joshcooper 
 yay, all checks passed.

give me a bit, and I'll look into the unit tests

buzzdeee:
yeah, going to squash them, once I'm done with it, I'll let you know when I think I'm good (:

buzzdeee:
@joshcooper I think I'm done. 

All OpenBSD facts have tests, and all but the dhcp resolver. I didn't implement a full Module/Class there, only subclassed Facter::Resolvers::Networking and override the extract_dhcp method. There I don't know how to properly test it, I made a number of different attempts, but all failed in various ways. 

other changes from initial PR:

- add system_uptime facts I forgot
- fix networking.dhcp fact
- case insensitive match in facter/framework/detector/os_detector.rb

the last one I needed because when facter is installed with puppetserver gem install facter, it returned OpenBSD and it just didn't matched it. My final goal is to get puppetserver to run on OpenBSD to replace my dated puppet master based on puppet 5.

klemensn:
It would be great to get this landed so we can finish up Puppet 7 (and eventually retire 5) on OpenBSD.

bigio:
I am currently running Puppetserver 7.9.2 on OpenBSD with this diff in production and it's working fine.
Having this committed will ease the move to Puppet7 on OpenBSD.

buzzdeee:
with the file loading change to 4.4, this was quite a hurdle to find. 
updated to the latest, seems to all working.
It would be really great to get this landed.

Also wondering why the FACTER queue in Jira is archived?

joshcooper (approved):

joshcooper:
Thanks for you work on this @buzzdeee!

joshcooper:
Sorry but this introduced order dependent spec failures and had to be reverted. You can reproduce the failure using:

```
$ rspec -fd './spec/facter/resolvers/openbsd/virtual_spec.rb[1:2:1]' './spec/facter/util/facts/posix/virtual_detector_spec.rb[1:1:12:1]' --order random:17181
```

buzzdeee:
I'm looking

buzzdeee:
Hi @joshcooper 

thanks for the rspec line to reproduce the issue.

with a certainty of about 80% I understand what went wrong, and tried to address it in #2642

PR #2533, opened 2022-11-02
(maint) Expect `with` hash arguments instead of kwargs

joshcooper:
Facter verifies partial doubles so arguments are checked against the original method and to ensure methods that don't exist can't be stubbed. RSpec 3.11.2 changes how keyword arguments are verified[1].

So previously you could write `expect(:m).with(a: true)` and it would pass if the method was called with a hash or keyword arguments. But now the expectation must expect a hash, not keyword arguments.

[1] https://github.com/rspec/rspec-mocks/commit/bc1a68725d57e167c885763e7f3edc236f92b7b5

### DIFF
--- a/spec/custom_facts/core/execution_spec.rb
+++ b/spec/custom_facts/core/execution_spec.rb
@@ -37,7 +37,7 @@ describe Facter::Core::Execution do
   end
 
   it 'delegates #exec to #execute' do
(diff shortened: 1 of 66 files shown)


joshcooper:
See also https://github.com/puppetlabs/puppet/pull/8942

tvpartytonight (approved):

PR #2536, opened 2022-12-08
(FACT-3147) Updates minimum Ruby version to 2.5

mhashizume:
This PR both updates the minimum Ruby version for Facter to Ruby 2.5 (removing all instances of Ruby 2.3 and 2.4 in the codebase) and increments the next Facter version to 4.3.0.

### DIFF
--- a/spec/custom_facts/util/parser_spec.rb
+++ b/spec/custom_facts/util/parser_spec.rb
@@ -354,7 +354,9 @@ describe LegacyFacter::Util::Parser do
         let(:yaml_content) { load_fixture('external_fact_yaml_date').read }
 
         it 'loads date' do
-          pending 'There is a bug in newer versions of Psych and Timecop' if RUBY_VERSION =~ /^3\.2/
+          if RUBY_VERSION.to_f == 3.2 && Gem::Version.new(Timecop::VERSION) < Gem::Version.new('0.9.6')
+            pending 'There is a bug in newer versions of Psych and timecop'
+          end
           expected_result = { 'testsfact' => { 'date' => Date.parse('2020-04-28') } }
 
           expect(yaml_parser.parse_results).to eq(expected_result)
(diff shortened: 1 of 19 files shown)


mhashizume:
I've also updated the branch protection rules to use the Ruby 2.5 version of unit tests instead of Ruby 2.3.

spec/custom_facts/util/parser_spec.rb
```diff
@@ -354,7 +354,7 @@ def expects_to_parse_powershell(cmd, result)
         let(:yaml_content) { load_fixture('external_fact_yaml_date').read }
 
         it 'loads date' do
-          pending 'There is a bug in newer versions of Psych and Timecop' if RUBY_VERSION =~ /^3\.2/
+          pending 'There is a bug in newer versions of Psych and Timecop' if RUBY_VERSION.match?(/3.2/)
```
joshcooper, line 357:
```suggestion
          pending 'There is a bug in newer versions of Psych and Timecop' if RUBY_VERSION.to_f == 3.2 && Gem::Version.new(Timecop::VERSION) < Gem::Version.new("0.9.6")
```

mhashizume:
> Not sure what's up with the acceptance test? Maybe we're expecting 20.04 but `ubuntu-latest` is 22.04?
> 
> ```
> Expected /Ubuntu (?-mix:20.04)/ to match "Ubuntu 22.04.1 LTS".>
>     Test line: tests/facts/os_processors_and_kernel.rb:15:in `block (6 levels) in run_test'
> ```

Yeah, looks like you're right: https://github.com/actions/runner-images/issues/6399

I'll put up another commit that just hard codes 20.04 for now

mhashizume:
I updated the branch protection rules to reflect the change in the Ubuntu runner.

joshcooper (approved):

PR #2539, opened 2022-12-31
(FACT-3171) Fix IPv6 link-local unicast address filtering

smortex:
RFC4291 define the Link-local unicast range as `fe80::/10`, not `fe80::/16` (what we compared against because of the String comparison). As a consequence, some Link-local unicast addresses where not correctly filtered, e.g. `fe87::`.

While here, add a "working" example with an actual global IPv6 address from the `2001:0DB8::/32` range reserved for documentation (RFC3849).

### DIFF
--- a/lib/facter/util/resolvers/networking/networking.rb
+++ b/lib/facter/util/resolvers/networking/networking.rb
@@ -65,7 +65,7 @@ module Facter
           end
 
           IPV4_LINK_LOCAL_ADDR = IPAddr.new('169.254.0.0/16').freeze # RFC5735
-          IPV6_LINK_LOCAL_ADDR = IPAddr.new('fe80::/16').freeze # RFC4291
+          IPV6_LINK_LOCAL_ADDR = IPAddr.new('fe80::/10').freeze # RFC4291
 
           def ignored_ip_address(addr)
             return true if addr.empty?
--- a/lib/facter/util/resolvers/networking/networking.rb
+++ b/lib/facter/util/resolvers/networking/networking.rb
@@ -64,8 +64,23 @@ module Facter
             bindings.empty? ? nil : bindings.first
           end
 
+          IPV4_LINK_LOCAL_ADDR = IPAddr.new('169.254.0.0/16').freeze # RFC5735
+          IPV6_LINK_LOCAL_ADDR = IPAddr.new('fe80::/16').freeze # RFC4291
+
           def ignored_ip_address(addr)
-            addr.empty? || addr.start_with?('127.', '169.254.') || addr.start_with?('fe80') || addr.eql?('::1')
+            return true if addr.empty?
+
+            ip = IPAddr.new(addr)
+            return true if ip.loopback?
+
+            [
+              IPV4_LINK_LOCAL_ADDR,
+              IPV6_LINK_LOCAL_ADDR
+            ].each do |range|
+              return true if range.include?(ip)
+            end
+
+            false
           end
 
           def calculate_mask_length(netmask)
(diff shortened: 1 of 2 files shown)


lib/facter/util/resolvers/networking/networking.rb
```diff
@@ -65,7 +65,19 @@ def find_valid_binding(bindings)
           end
 
           def ignored_ip_address(addr)
-            addr.empty? || addr.start_with?('127.', '169.254.') || addr.start_with?('fe80') || addr.eql?('::1')
+            return true if addr.empty?
+
+            ip = IPAddr.new(addr)
```
joshcooper, line 70: Can `addr` be invalid? If so, you'll want to rescue:

```ruby
ruby -r ipaddr -e "puts IPAddr.new('42')"
	3: from -e:1:in `<main>'
	2: from -e:1:in `new'
	1: from /home/josh/.rbenv/versions/2.7.6/lib/ruby/2.7.0/ipaddr.rb:586:in `initialize'
/home/josh/.rbenv/versions/2.7.6/lib/ruby/2.7.0/ipaddr.rb:651:in `in6_addr': invalid address (IPAddr::InvalidAddressError)
	3: from -e:1:in `<main>'
	2: from -e:1:in `new'
	1: from /home/josh/.rbenv/versions/2.7.6/lib/ruby/2.7.0/ipaddr.rb:557:in `initialize'
/home/josh/.rbenv/versions/2.7.6/lib/ruby/2.7.0/ipaddr.rb:598:in `rescue in initialize': invalid address: 42 (IPAddr::InvalidAddressError)
```

lib/facter/util/resolvers/networking/networking.rb
```diff
@@ -65,7 +65,19 @@ def find_valid_binding(bindings)
...
+            return true if ip.loopback?
+
+            [
+              IPAddr.new('169.254.0.0/16'), # Link-local unicast RFC5735
+              IPAddr.new('fe80::/10')       # Link-local unicast RFC4291
+            ].each do |range|
+              return true if range.include?(ip)
+            end
```
joshcooper, line 78: Suggest creating frozen constants with the parsed IPAddr at the top:

```ruby
LINK_LOCAL_ADDR = IPAddr.new('169.254.0.0/16').freeze
```

And then check the ip against it:

```ruby
return true if LINK_LOCAL_ADDR.include?(ip)
```

lib/facter/util/resolvers/networking/networking.rb
smortex: > Can `addr` be invalid?

I don't think so: on line 18 when we build the binding, we already succeeded in parsing this address. Maybe it would make sense to only have `IPAddr` in `bindind` and build the `address` / `netmask` / `network` from the IPAddr later, but I didn't want to rework too much the code so decided to parse the addr a second time.

lib/facter/util/resolvers/networking/networking.rb
smortex: Done!

joshcooper:
@smortex we just merged some changes to address the test failures. If you rebase onto d3447d7a08703231c68cfef8c06c41133b38dff9 I think your PR will pass

smortex:
Awesome!  Thanks for the head-up!

joshcooper (approved):
Looks correct to me based on https://www.rfc-editor.org/rfc/rfc4291.html#section-2.4

```
      Link-Local unicast   1111111010           FE80::/10       2.5.6
```

 and https://www.rfc-editor.org/rfc/rfc5735.html#section-3

> 169.254.0.0/16 - This is the "link local" block.  As described in
   [[RFC3927](https://www.rfc-editor.org/rfc/rfc3927)], it is allocated for communication between hosts on a
   single link.  Hosts obtain these addresses by auto-configuration,
   such as when a DHCP server cannot be found.

PR #2540, opened 2022-12-31
(FACT-3172) Fix IPv6 unique-local unicast address filtering

smortex:
We do not want these addresses to be reported as the default IPv6 address in `networking.ip6`.  Quoting RFC4193:

> This document defines an IPv6 unicast address format that is globally
> unique and is intended for local communications, usually inside of a
> site.  These addresses are not expected to be routable on the global
> Internet.

This PR also include:
* #2539

### DIFF
--- a/lib/facter/util/resolvers/networking/networking.rb
+++ b/lib/facter/util/resolvers/networking/networking.rb
@@ -65,7 +65,7 @@ module Facter
           end
 
           IPV4_LINK_LOCAL_ADDR = IPAddr.new('169.254.0.0/16').freeze # RFC5735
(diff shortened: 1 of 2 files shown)


joshcooper:
@smortex we just merged some changes to address the test failures. If you rebase onto https://github.com/puppetlabs/facter/commit/d3447d7a08703231c68cfef8c06c41133b38dff9 I think your PR will pass

joshcooper (approved):
Matches https://www.rfc-editor.org/rfc/rfc4193#section-3.2.1:

```
      Prefix            FC00::/7 prefix to identify Local IPv6 unicast
                        addresses.
```

PR #2545, opened 2023-01-24
(FACT-3170) Add logic to Facter.value and Facter.fact to use cache

cthorn42:
Looks like there was something overlooked when it comes to caching with the value and fact methods. This small fix will make it so we use the cache instead of looking up the value again.


spec/facter/facter_spec.rb
```diff
@@ -212,6 +212,15 @@ def mock_collection(method, os_name = nil, error = nil)
       expect(Facter.value('os.name')).to eq('Ubuntu')
     end
 
+    it 'call value twice and and resolved once' do
+      mock_fact_manager(:resolve_fact, [os_fact])
+
+      allow(fact_collection_spy).to receive(:value).with('os.name').and_return('Ubuntu')
+
+      expect(Facter.value('os.name')).to eq('Ubuntu')
+      expect(Facter.value('os.name')).to eq('Ubuntu')
```
joshcooper, line 221: suggest adding the expectation on:

```
expect(fact_collection_spy).to receive(:value).with('os.name').and_return('Ubuntu')
```

and then calling `Facter.value('os.name')` twice, so it's clearer what's being tested. Also won't need to change rubocop.

lib/facter.rb
```diff
@@ -410,7 +410,7 @@ def trace(bool)
     # @api public
     def value(user_query)
       user_query = user_query.to_s.downcase
-      resolve_fact(user_query)
+      resolve_fact(user_query) unless @already_searched[user_query]&.value
```
joshcooper, line 413: ah one gotcha, if a fact value is `false`, then thjis would resolve it multiple times. Recommend checking if the key is in the hash:

```ruby
... unless @already_searched.include?(user_query)
```

spec/facter/facter_spec.rb
cthorn42: Yup that did the track! 👍 No rubocop exception required.

lib/facter.rb
```diff
@@ -427,7 +427,7 @@ def value(user_query)
     # @api public
     def fact(user_query)
       user_query = user_query.to_s.downcase
-      resolve_fact(user_query)
+      resolve_fact(user_query) unless @already_searched[user_query]
```
joshcooper, line 430:
```suggestion
      resolve_fact(user_query) unless @already_searched.include?(user_query)
```

joshcooper (approved):
I confined the posix file provider to `:operatingsystem => :Ubuntu` similar to what the `windows` provider does and it took puppet 42 seconds to manage 10k files that were already insync:

```
$ cat files
Integer[0, 10000].each |$i| {
  file { "/tmp/file$i":
    ensure => present
  }
}
$ time bundle exec puppet apply files.pp
Notice: Compiled catalog for localhost in environment production in 1.44 seconds
Notice: Applied catalog in 42.31 seconds
bundle exec puppet apply files.pp  46.21s user 2.52s system 98% cpu 49.588 total
```

With this change it only takes:

```
$ time bundle exec puppet apply files.pp                            
Notice: Compiled catalog for localhost in environment production in 1.56 seconds
Notice: Applied catalog in 3.81 seconds
bundle exec puppet apply files.pp  10.02s user 0.25s system 92% cpu 11.141 total
```

PR #2549, opened 2023-02-22
(maint) improve and regenerate manpage

jcharaoui:
This improves the facter manpage to fix parsing errors due to the missing short description in the NAME section, makes the formatting consistent with most other manpages, and updates the DESCRIPTION section using the one in the project's README, which is more useful this context.

### DIFF
--- a/lib/facter/templates/man.erb
+++ b/lib/facter/templates/man.erb
@@ -1,11 +1,14 @@
 <%# encoding: UTF-8%>
+facter - collect and display facts about the current system
+===========================================================
+
 SYNOPSIS
 --------
-  facter [options] [query] [query] [...]
+<b>facter</b> [options] [query] [query] [...]
 
 DESCRIPTION
 -----------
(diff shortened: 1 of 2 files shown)


jcharaoui:
Oh I see there's also an ERB template involved, so this needs some extra work.

joshcooper:
Thanks @jcharaoui Could you update this template https://github.com/puppetlabs/facter/blob/07a7a8c4d8076ba19f4564cfb2d3896bedf515b9/lib/facter/templates/man.erb instead? There's also a rake task to regenerate the man page. Thanks!

joshcooper:
Hi @jcharaoui could you rebase on `main` and update the erb template? Also I think we do want to keep the `\fBfacter\fR` in the name so that it's bold in the man output. Thanks!

jcharaoui:
> Hi @jcharaoui could you rebase on `main` and update the erb template? Also I think we do want to keep the `\fBfacter\fR` in the name so that it's bold in the man output. Thanks!

Done and done!

joshcooper (approved):

joshcooper:
Thank you @jcharaoui!

PR #2555, opened 2023-04-01
(FACT-3195) Fix inconsistency in disks serial number reporting

smortex:
When support for serial numbers of disks was added in Facter in a676bbd727ed1ba40a70df27147a0f37bdf55e61, the key was named `serial_number` and documented in `lib/schema/facter.yaml`.

When a similar feature was added for Linux, the key was initially named `sn` (57dbe34e0e9682b7ae98d05abf1630c5a532eac1) then renamed to `serial` (006f4fbd1f4cb6bc32e20328e27c322c52f148c8).

Align disks serial number reporting to match the schema on Linux.

### DIFF
--- a/lib/facter/resolvers/disks.rb
+++ b/lib/facter/resolvers/disks.rb
@@ -52,18 +52,26 @@ module Facter
 
                 next unless result
 
-                value[key] = case key
-                             when :size
-                               # Linux always considers sectors to be 512 bytes long
-                               # independently of the devices real block size.
-                               construct_size(value, result)
-                             when :type
-                               result == '0' ? 'ssd' : 'hdd'
-                             else
-                               result
-                             end
+                result = case key
+                         when :size
+                           # Linux always considers sectors to be 512 bytes long
+                           # independently of the devices real block size.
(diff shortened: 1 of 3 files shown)


joshcooper:
Renaming fact names is more than the usual major break because those fact names are specified in puppetdb queries, hiera lookups, PE classification. In order to change them we would need to:

1. collect both the old and new fact names and send them to puppetserver for awhile (thru 8.x)
2.update the puppet-lint plugin (like we did for legacy facts)
3. Ship the plugin in the pdk
4. Migrate PE classification rules during upgrades.
5. Docs changes
6. Ask everyone else to update pdb queries, hiera lookups, puppet manifests
...

Given all of this, I don't see it happening.

AriaXLi:
Hi @smortex, thank you for reporting this issue. While we agree this is likely an improvement, we do not anticipate addressing this any time soon due to this change requiring a substantial architectural change to support it. As such, this PR and the ticket FACT-3195 will be closed. We may revisit this at a later time, and if so, will re-open this ticket.

PR #2556, opened 2023-04-04
(FACT-3194) Difference in Facter 3 & 4 ordering

AriaXLi:
In Facter 4.x, Facter.to_hash() and Facter.resolve() return hashes whose keys are sorted differently depending on the parameters or which of the two  methods is called. In Facter 3.x, the hashes are always returned in a stable, consistent order when using either method.

This commit ensures the returned hashes from Facter.resolve() and Facter.to_hash() have a stable order by using Facter::Utils.sort_hash_by_key().

### DIFF
--- a/spec/facter/facter_spec.rb
+++ b/spec/facter/facter_spec.rb
@@ -8,9 +8,17 @@ describe Facter do
     # note the type is the :nil symbol
     Facter::ResolvedFact.new('missing_fact', nil, :nil, 'missing_fact')
   end
+  let(:hash_fact) do
+    Facter::ResolvedFact.new('mountpoints', { '/boot' => { 'filesystem' => 'ext4', 'device' => 'dev1' },
+                                              '/' => { 'filesystem' => 'ext4', 'device' => 'dev2' } },
+                             :core, 'mountpoints')
+  end
 
   let(:logger) { instance_spy(Facter::Log) }
   let(:config_reader_double) { class_spy(Facter::ConfigReader) }
+
+  sorted_fact_hash = { 'mountpoints' => { '/' => { 'device' => 'dev2', 'filesystem' => 'ext4' },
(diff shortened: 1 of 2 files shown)


joshcooper (approved):
For posterity, I think facter 3 had a stable order, because it used `std::map` for the [fact collection's elements](https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/inc/facter/facts/collection.hpp#L348) and for the elements of a [`map_value`](https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/inc/facter/facts/map_value.hpp#L124). And [`std::map` is defined](https://en.cppreference.com/w/cpp/container/map) to be:

> std::map is a sorted associative container that contains key-value pairs with unique keys. Keys are sorted by using the comparison function Compare.

spec/facter/facter_spec.rb
```diff
@@ -56,13 +59,31 @@ def stub_no_fact
...
     end
+
+    context 'when there is no user query' do
+      let(:hash_fact_no_user_query) do
+        Facter::ResolvedFact.new('mountpoints', { '/boot' => {}, '/' => {} }, :core, '')
+      end
+
+      it 'returns a stable order hash' do
+        stub_facts([hash_fact_no_user_query])
+        expect(Facter.resolve('')).to eq('mountpoints' => { '/' => {}, '/boot' => {} })
```
joshcooper, line 84: Might be good to verify it sorts the entire tree, like:

```ruby
`mountpoints' => {
  '/' => {  'device' => 'dev1', 'filesystem' => 'ext4' }, 
  '/boot' => { 'device' => 'dev2', 'filesystem' => 'ext4' }
}
```

And then have the stub return `device` and `filesystem` in the reverse order.

joshcooper (approved):

mhashizume (approved):

PR #2564, opened 2023-04-26
(FACT-3136) Split GCE ssh keys on newlines

joshcooper:
GCE returns ssh keys as a string with newlines separating each key. When parsing
the project and VM instance data, convert to an array of keys like Facter 3 did.

Note Facter 3 used a streaming JSON parser and it matched all hash keys named
"sshKeys" (at any level), whereas we are only matching the keys for the project
and VM instance, but not custom fields that happen to be called "sshKeys".

GCE deprecated "sshKeys" and recommends "ssh-keys", so transform both.

[1] https://cloud.google.com/compute/docs/metadata/default-metadata-values

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,8 @@ require:
 
 Layout/LineLength:
   Max: 120
+  Exclude:
+    - spec/facter/resolvers/gce_spec.rb
 
 Lint/RaiseException:
   Enabled: true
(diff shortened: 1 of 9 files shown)


joshcooper:
Once this is merged and passed CI, I'll revert https://github.com/puppetlabs/puppet-agent/commit/938df7d389cc12cdf5b3727e31287477ffbbb9d7

mhashizume (approved):

PR #2565, opened 2023-04-27
(FACT-3192) facter -p --no-external-facts should not raise

AriaXLi:
Facter will raise an error saying `no-external-facts and external-dir options conflict: please specify only one` if the -p/--puppet and --no-external-facts options are used together. The -p option adds another directory to external-dir which causes Facter to think the --external-dir option was also passed in. This commit ensures -p and --no-external-facts can be used together without raising and --external-dir is either set to false or any explicitely passed in external directories and adds a spec test to test this behavior.


joshcooper (approved):
Verified `facter --no-external-facts` and `facter -p --no-external-facts` works as expected

PR #2566, opened 2023-04-28
(FACT-2772) Add env var for facter recursion detection

tvpartytonight:


spec_integration/facter_resolve_spec.rb
```diff
@@ -3,6 +3,27 @@
 require_relative 'spec_helper'
 
 describe Facter do
+  context 'when recursively calling facter in an external/custom fact' do
+    # this will need to be wrapped into a method that will install a posix or windows custom fact
```
tvpartytonight, line 7: @joshcooper I am expecting to have to refactor this part here to create an executable fact on windows; are there any examples or documentation you can point me to?

joshcooper: @tvpartytonight I think facter on Windows solely relies on the file extension (instead of file mode) to determine if it's executable or not. I'd suggest creating a bat file like:

```ruby
ext_fact_path = File.join(ext_dir, 'extfact.bat')

File.write(ext_fact_path, <<~SCRIPTEND)
facter --external-dir #{ext_dir} os
SCRIPTEND
```

One thing to watch out for is (I think) `ext_dir` will need to contain backslashes when used inside a `bat` file/ So you may need to do something like:

```ruby
ext_fact_path.tr!('/', '\\')
```

tvpartytonight: oh, hmm, I am surprised you can just call facter here in the test without specifying which facter to load, but it seems to work just fine on my macos box too.

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -8,6 +8,12 @@ module Execution
       class Base
         STDERR_MESSAGE = 'Command %s completed with the following stderr message: %s'
         VALID_OPTIONS = %i[on_fail expand logger timeout].freeze
+        # Set LC_ALL and LANG to force i18n to C for the duration of this exec;
+        # this ensures that any code that parses the
+        # output of the command can expect it to be in a consistent / predictable format / locale
+        # Set INSIDE_FACTER here to ensure no recursive facter calls are initiated
+        CLI_ENV_VARS = { 'LC_ALL' => 'C', 'LANG' => 'C', 'INSIDE_FACTER' => 'true' }.freeze
```
joshcooper, line 15: I was kind of assuming the entry point to resolving all facts was in the `FactLoader#load` method so that we wouldn't have to special case this. Or does facter sometimes call execute outside of the `load` method?

lib/facter/framework/core/fact_loaders/fact_loader.rb
```diff
@@ -19,6 +19,10 @@ def initialize
     end
 
     def load(user_query, options)
+      if ENV['INSIDE_FACTER']
+        @log.debug('INSIDE_FACTER env var detected, forcing custom/external facts off to prevent recursion')
+        OptionStore.make_options_recursion_safe
+      end
       @internal_facts = load_internal_facts(user_query, options)
       @custom_facts = load_custom_facts(options)
       @external_facts = load_external_facts(options)
```
joshcooper, line 28: Would it make sense to just not call `load_external_facts` if `INSIDE_FACTER` is set, like Facter 3 used to do: https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/exe/facter.cc#L282-L292? Might be good to include that warning message too so people know how to fix the problem in their external facts.

That way you don't need to mutate the options.

lib/facter/custom_facts/core/execution/base.rb
joshcooper: Ah so environment variables set in the current facter process, e.g. `ENV['INSIDE_FACTER'] = "true", will be inherited by any/all child processes it invokes. That's true for `Open3`, `Kernel.exec`, `system`. One exception is with [`Open3.spawn`](https://docs.ruby-lang.org/en/3.0/Process.html#method-c-spawn), as it's possible to override an environment variable:

```ruby
spawn({"FOO"=>"BAR", "BAZ"=>nil}, command)
```
Or clear environment variables that aren't explicitly specified:
```ruby
spawn({"FOO"=>"BAR"}, command, :unsetenv_others=>true)
```

I'm thinking it'd be clearer to check for `INSIDE_FACTER`, and if it's unset, set it:

```ruby
if ENV['INSIDE_FACTER']
  # we're recursing, skip external facts
else
  ENV['INSIDE_FACTER'] = 'true'
  # call external facts
end
```

tvpartytonight: Wait, so is this conversation more appropriate for the comment [here](https://github.com/puppetlabs/facter/pull/2566#discussion_r1185628488) then?

joshcooper: yeah, I'm thinking we shouldn't need to modify any of the execution code, just the part where we load external facts below

tvpartytonight:
@joshcooper I think this is ready for another round of review when you get a chance.

lib/facter/framework/core/fact_loaders/fact_loader.rb
```diff
@@ -20,8 +20,12 @@ def initialize
 
     def load(user_query, options)
       @internal_facts = load_internal_facts(user_query, options)
-      @custom_facts = load_custom_facts(options)
-      @external_facts = load_external_facts(options)
+      if ENV['INSIDE_FACTER']
+        @log.debug('INSIDE_FACTER env var detected, not loading custom/external facts off to prevent recursion')
```
joshcooper, line 24: Could you include a warning message that tells users to specify the `--no-external-facts` option like Facter 3 did https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/exe/facter.cc#L286?

lib/facter/framework/core/fact_loaders/fact_loader.rb
```diff
@@ -20,8 +20,12 @@ def initialize
...
+      else
```
joshcooper, line 25: I'm thinking it would be better to set `ENV['INSIDE_FACTER'] = true` here (prior to loading custom/external facts) instead of in the `execute` method, in case a custom fact did something like:

```ruby
Facter.add(:recurse) do
  setcode { `facter os` }
done
```

The environment variable `INSIDE_FACTER` will automatically be passed to any/all child process no matter which ruby method is used to create the process (spawn, backticks, etc).

lib/facter/framework/core/fact_loaders/fact_loader.rb
joshcooper: oh I think this part was missed in the last commit. I think it'd be good to continue warning folks (like Facter 3) that they need to fix their external facts.

spec_integration/facter_resolve_spec.rb
```diff
@@ -3,6 +3,53 @@
...
+        # This iterates over the supported entry points that would attempt to resolve
+        # external facts; if any one of them descends into recursion, it will exceed
+        # the DEFAULT_EXECUTION_TIMEOUT; otherwise, each of these calls should be only
+        # be a few seconds.
+        facter_entrypoints.each do |entrypoint|
+          entrypoint.call
+          Facter.reset
+          Facter.clear
+          Facter::OptionStore.reset
+          LegacyFacter.clear
```
joshcooper, line 46: We recently found some more global state when running `spec` unit tests. I'm not sure if the same is needed here https://github.com/puppetlabs/facter/commit/b78365653f7248327e28e2dc9aba3b7d5f6efb07#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94? Or in the `spec_integration/spec_helper.rb`?

lib/facter/framework/core/fact_loaders/fact_loader.rb
```diff
@@ -21,7 +21,12 @@ def initialize
     def load(user_query, options)
       @internal_facts = load_internal_facts(user_query, options)
       @custom_facts = load_custom_facts(options)
-      @external_facts = load_external_facts(options)
+      if ENV['INSIDE_FACTER']
```
tvpartytonight, line 24: @joshcooper  I have reworked this but seem to be running into errors with `ENV['INSIDE_FACTER']` leaking between spec tests; it always leak in the same way, which is good. I think it happens when we mock external facts, like this(but still digging into the root cause):
```
      allow(external_fact_loader_double).to receive(:external_facts).and_return([loaded_env_custom_fact])
```

tvpartytonight: OK, I just deleted the ENV var after we set it during external fact collection, and it seems to solve my leaky ENV vars.

spec_integration/facter_resolve_spec.rb
tvpartytonight: I used some debug output in the test to verify that the external fact was indeed being called repeatedly throughout this loop of entrypoints; without these, subsequent calls to facter within this single test would just use cached values and not exercise the external fact collection.

joshcooper (approved):
Oh nice, I didn’t know about to_stderr_from_any_process. Good to know!

PR #2568, opened 2023-05-04
(FACT-3201) Report ProductVersionExtra on macOS 13 and later

ccaviness:
Starting with macOS 13, Apple can use a "Rapid Security Response" process to push small system updates out. When one of these is applied, a new `ProductVersionExtra` field is output from the `sw_vers` command:

```
$ sw_vers
ProductName:        macOS
ProductVersion:        13.3.1
ProductVersionExtra:    (a)
BuildVersion:        22E772610a
```

If no Rapid Security Response update is installed, the `ProductVersionExtra` field does not appear:

```
$ sw_vers
ProductName:        macOS
ProductVersion:        13.4
BuildVersion:        22F5049e
```

The `os.macosx.version` fact should report this. Currently it only reports `full`, `major`, `minor`, and `patch` fields.

This change adds an additional `extra` field for macOS 13 and later, if `ProductVersionExtra` is reported.

With this patch, on a 13.3.1 machine with the recent RSR installed, `facter os.macosx.version` returns:

```
{
  extra => "(a)",
  full => "13.3.1",
  major => "13",
  minor => "3",
  patch => "1"
}
```

### DIFF
--- a/lib/facter/facts/macosx/os/macosx/version.rb
+++ b/lib/facter/facts/macosx/os/macosx/version.rb
@@ -10,23 +10,27 @@ module Facts
                        macosx_productversion_patch].freeze
 
           def call_the_resolver
-            fact_value = Facter::Resolvers::SwVers.resolve(:productversion)
-            ver = version_hash(fact_value)
+            version_value = Facter::Resolvers::SwVers.resolve(:productversion)
+            extra_value = Facter::Resolvers::SwVers.resolve(:productversionextra)
+            ver = version_hash(version_value, extra_value)
 
             [Facter::ResolvedFact.new(FACT_NAME, ver),
-             Facter::ResolvedFact.new(ALIASES[0], fact_value, :legacy),
+             Facter::ResolvedFact.new(ALIASES[0], version_value, :legacy),
              Facter::ResolvedFact.new(ALIASES[1], ver['major'], :legacy),
              Facter::ResolvedFact.new(ALIASES[2], ver['minor'], :legacy),
              Facter::ResolvedFact.new(ALIASES[3], ver['patch'], :legacy)]
           end
 
-          def version_hash(fact_value)
-            versions = fact_value.split('.')
+          def version_hash(version_value, extra_value)
+            versions = version_value.split('.')
             if versions[0] == '10'
-              { 'full' => fact_value, 'major' => "#{versions[0]}.#{versions[1]}", 'minor' => versions[-1] }
-            else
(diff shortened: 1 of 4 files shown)


ccaviness:
Two possible questions:

1. Should we call this `extra`? That's what it's referred to in `sw_vers`, but of course this doesn't alphabetize neatly in `facter os.macosx.productversion` output. Perhaps `rsr`? Or just stick with `extra` and don't feel bothered by the alphabetization.
2. Should we only report `extra` when `ProductVersionExtra` has a value? That is, the `extra` field will only be added to the `os.macosx.productversion` fact hash if it has a non-`nil` value.

ccaviness:
Any chance of getting this merged? Also, I can no longer access FACT-3201 since it was migrated to atlassian (https://puppet.atlassian.net/browse/FACT-3201). Just "You don't have access to this issue"

mhashizume:
Hi @ccaviness , thanks for your contribution. We will be taking a look at this PR during our team's sprint. I'm going to close then re-open this PR to rekick tests.

mhashizume:
To answer your questions @ccaviness:

- `extra` makes sense to me, I would prioritize accurately reflecting the output and naming from Apple over alphabetization.
- I think omitting this fact when it isn't present is the right call, rather than setting it to an empty string. Facter usually just drops a fact altogether when it's not present on the system (I'm having a hard time thinking of an example where we set it to an empty string or something instead).
- Our Atlassian admins are working on restoring public access to our Jira instance and that should hopefully be back available soon.

Would you be able to update your PR to omit the `extra` fact when it's not present? 

Thanks again!

joshcooper:
Also please rebase on `main` due to an update to the jruby 9.3.9.0 check

ccaviness:
I changed it set to `nil` if not found; do you want the entire `extra` key removed instead?

ccaviness:
@joshcooper Git frightens and confuses me. Did I rebase it correctly?

mhashizume:
> I changed it set to `nil` if not found; do you want the entire `extra` key removed instead?

Removing the key makes sense to me, omitting a key entirely when it's not populated is consistent with other Facter behavior, and `sw_vers` doesn't display `extra` when it's not there either. You could probably just update the second conditional to something like `elsif /11|12/.match?(versions[0]) || extra_value.nil?`

ccaviness:
@mhashizume Done.

mhashizume:
Thanks so much @ccaviness ! This should go out in the next version of Facter, which we plan on releasing sometime next month

mhashizume (approved):

PR #2571, opened 2023-05-05
(FACT-3196) Ensure file descriptors are closed

joshcooper:
If ioctl returns non-zero then we were leaking file descriptors.

Always return a mutable string from `Facter::Core::Execution::Base.execute_command`


joshcooper:
jenkins please tests this on solaris11-64a

cthorn42 (approved):
The rest of this looks good, the spec tests are passing for me on my end when I'm testing this.

PR #2573, opened 2023-05-11
link to puppet 8 release notes for latest facter changes

evgeni:
4.4+ is released with Puppet 8


joshcooper:
Hi @evgeni , facter 4 is released in both Puppet 7 and 8 streams. So maybe add both links?

evgeni:
As of now, https://www.puppet.com/docs/puppet/7/release_notes_facter.html doesn't mention 4.4 (probably because there is no P7 release with it yet? idk), so I've just changed it.

joshcooper:
> As of now, https://www.puppet.com/docs/puppet/7/release_notes_facter.html doesn't mention 4.4

I'm not sure I follow. Facter 4 shipped in Puppet 7.0 and is mentioned in the link you mentioned https://www.puppet.com/docs/puppet/7/release_notes_facter.html

evgeni:
The PR is 4 months old. When opening it, the link did not mention Facter 4.

Feel free to close, as it now indeed mentions it.

joshcooper:
ah, ok that makes sense, thanks for taking the time to submit this!

PR #2574, opened 2023-05-11
(FACT-3202) Add is_virtual and virtual support for crio

lollipopman:
Prior to this change facter returned:

    $ facter is_virtual
    false
    $ facter virtual
    physical

After this change facter returns:

    $ facter is_virtual
    true
    $ facter virtual
    crio

This change separates out reading pid 1's environment from proc and reading the cgroup information. It also adds a container_other type when the container runtime is not explicitly supported.


lollipopman:
friendly bump for a review please

lollipopman:
@joshcooper any chance you would have time to take a gander?

lollipopman:
friendly bump for a review!

lollipopman:
Would love a review, any takers?

lollipopman:
would love a review!

lollipopman:
friendly bump!

AriaXLi:
Hi @lollipopman thanks for the PR, could you update your PR so there's no conflicts? Thanks!

lollipopman:
Happy to, pushed, thanks for looking @AriaXLi

lollipopman:
@AriaXLi tests are passing, do you think you could take a look? thanks!

lollipopman:
@AriaXLi any chance you could look at this again, before it bitrots? <3

lib/facter/resolvers/containers.rb
```diff
@@ -14,45 +14,59 @@ class << self
...
+          info, vm = extract_vm_and_info(output_docker, output_lxc)
           @fact_list[:vm] = vm
           @fact_list[:hypervisor] = { vm.to_sym => info } if vm
           @fact_list[fact_name]
         end
 
-        def extract_vm_and_info(output_docker, output_lxc, lxc_from_environ)
+        def read_environ(fact_name)
+          container = Facter::Util::Linux::Proc.getenv_for_pid(1, 'container')
+          return unless container && container != ''
```
joshcooper, line 37: I'm ok with simple guard clauses like `return unless container`, but mixing `&&` and `!=` in there gets confusing. I think this is easier to grok:

```ruby
return if container.nil? || container.empty?
```

lib/facter/util/file_helper.rb
```diff
@@ -15,12 +15,14 @@ def safe_read(path, default_return = '')
           default_return
         end
 
-        def safe_readlines(path, default_return = [])
-          return File.readlines(path, encoding: Encoding::UTF_8) if File.readable?(path)
+        # rubocop:disable Style/SpecialGlobalVars
+        def safe_readlines(path, default_return = [], sep = $/, chomp: false)
+          return File.readlines(path, sep, chomp: chomp, encoding: Encoding::UTF_8) if File.readable?(path)
```
joshcooper, line 20: Can you add the `safe_readlines` changes to single commit (along with changes to call sites and tests)? Also we generally try to avoid mixing positional and keyword args.  Perhaps make `sep` a keyword arg? We can't change `path` or `default_return` as there are probably custom facts that call it.

lib/facter/util/linux/proc.rb
```diff
@@ -0,0 +1,22 @@
...
+    module Linux
+      class Proc
+        class << self
+          def getenv_for_pid(pid, field)
+            path = "/proc/#{pid}/environ"
+            lines = Facter::Util::FileHelper.safe_readlines(path, [], "\0", chomp: true)
+            lines.each do |line|
+              if line.slice(0, field.length) == field && line.slice(field.length) == '='
+                return line.slice(field.length + 1, line.length - (field.length + 1))
+              end
```
joshcooper, line 14: `line.slice(0, field.length) == field` seems to be checking if the `line` starts with `field`. Suggest simplifying to:

```
key, value = line.split('=', 2)
return value if key == field
```

which I think gives the desired result in both the "match" and "non-match" cases:

```
irb(main):013:0> key, value = "foo=bar".split('=', 2)
=> ["foo", "bar"]
irb(main):014:0> puts key; puts value
foo
bar
=> nil
irb(main):015:0> key, value = "foobar".split('=', 2)
=> ["foobar"]
irb(main):016:0> puts key; puts value.inspect
foobar
nil
=> nil
```

lib/facter/resolvers/containers.rb
```diff
@@ -14,45 +14,59 @@ class << self
...
+
+          info = {}
+          case container
+          when 'lxcroot'
+            vm = 'lxc'
+          when 'podman'
+            vm = 'podman'
```
joshcooper, line 44: podman seems to be new, could you mention that in your commit message?

lib/facter/resolvers/containers.rb
```diff
@@ -14,45 +14,59 @@ class << self
...
+          when 'crio'
+            vm = 'crio'
+          when 'systemd-nspawn'
+            vm = 'systemd_nspawn'
+            info = { 'id' => Facter::Util::FileHelper.safe_read('/etc/machine-id', nil).strip }
+          else
+            vm = 'container_other'
```
joshcooper, line 51: Could you add a test for this case? Should we log a warning that it's an unknown container?

lib/facter/resolvers/containers.rb
lollipopman: I agree, fixed

lib/facter/util/file_helper.rb
lollipopman: I broke this out into a single commit. I kept sep as a positional argument, since that mirrors the function readlines, which it is wrapping. However, if you would still like it to be a keyword arg, I am happy to make that change.

lib/facter/util/linux/proc.rb
lollipopman: fixed

lib/facter/resolvers/containers.rb
lollipopman: yup, added.

lib/facter/resolvers/containers.rb
```diff
@@ -14,45 +14,59 @@ class << self
...
-          info, vm = extract_for_nspawn(output_environ) unless vm
+          info, vm = extract_vm_and_info(output_docker, output_lxc)
           @fact_list[:vm] = vm
           @fact_list[:hypervisor] = { vm.to_sym => info } if vm
           @fact_list[fact_name]
         end
 
-        def extract_vm_and_info(output_docker, output_lxc, lxc_from_environ)
+        def read_environ(fact_name)
+          container = Facter::Util::Linux::Proc.getenv_for_pid(1, 'container')
```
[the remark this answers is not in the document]
lollipopman: good idea, fixed.

lib/facter/resolvers/containers.rb
lollipopman: podman worked before, via cgroups, but this adds environ support, added a note to the commit message.

lollipopman:
@joshcooper patch updated, ready for another review

lib/facter/util/file_helper.rb
joshcooper: ah no worries, thanks!

joshcooper:
Thanks @lollipopman looks good, just running tests

lollipopman:
Tests look good, anything else to do @joshcooper?

joshcooper (approved):

joshcooper:
Thank you @lollipopman!

lib/facter/resolvers/containers.rb
```diff
@@ -14,45 +14,65 @@ class << self
         private
 
         def post_resolve(fact_name, _options)
-          @fact_list.fetch(fact_name) { read_cgroup(fact_name) }
+          @fact_list.fetch(fact_name) do
+            read_environ(fact_name) || read_cgroup(fact_name)
```
joshcooper, line 18: If `/proc/1/environ` contains `container=<value>` where `<value>` is non-empty and is not one of the values in the `case`, then we set `vm=container_other` and never fall back to `read_cgroup` .

lollipopman: true, perhaps `read_environ()` should only log a warning and not set the value to `container_other`

raybellis:
Please fix this patch so that it doesn't warn when invoked on non-Linux systems (see issue 2742)

lib/facter/resolvers/containers.rb
joshcooper: @lollipopman prior to your change, we checked `/proc/1/cgroups` for `docker` 

https://github.com/puppetlabs/facter/blob/d47f8790baf41156c534dfcf5210374cd76a593d/lib/facter/resolvers/containers.rb#L39

And then 'lxc' or fell back to checking the `container` environment variable in `/proc/1/environ`

https://github.com/puppetlabs/facter/blob/d47f8790baf41156c534dfcf5210374cd76a593d/lib/facter/resolvers/containers.rb#L42

Your commit reversed the order, so it looks in `/proc/1/environ` first. https://github.com/puppetlabs/facter/blob/7bc38ccb58e6abf66d6141a3b2e9f60eb0077d71/lib/facter/resolvers/containers.rb#L18

Was that intentional? Would it be a problem if I reversed it?

One thing I didn't realize when I merged this is the `VirtualDetector` calls 
https://github.com/puppetlabs/facter/blob/d47f8790baf41156c534dfcf5210374cd76a593d/lib/facter/util/facts/posix/virtual_detector.rb#L19

So the `Containers` resolver can't hard code `container_other` https://github.com/puppetlabs/facter/blob/0656d9a34ce4790129f1fd6eba5bb4d49a9b9ad1/lib/facter/resolvers/containers.rb#L60C19-L60C34

lollipopman: I don't think I had a strong reason for changing the order, and my fault for not adding a comment if I did. I assume my rationale was that the environment variable was more authoritative, but I don't think that argument is very strong at present. As for the hard coding of `container_other` I agree that should be dropped and just return nil for the fact.

joshcooper: No worries @lollipopman, thanks for confirming!

PR #2575, opened 2023-05-19
(FACT-3186) Handle partitions fact when none exist

mhashizume:
Prior to this commit, when a diskless machine used Facter the partitions fact errored out with an undefined method error. This was because the machine did not have any partitions and Facter was unable to dig into the partitions hash because the hash didn't exist.

This commit updates Linux's partition fact to return nil early in the partitions method if the partitions resolver returns nothing.

### DIFF
--- a/lib/facter/facts/linux/partitions.rb
+++ b/lib/facter/facts/linux/partitions.rb
@@ -11,6 +11,10 @@ module Facts
 
       def partitions
         parts = Facter::Resolvers::Partitions.resolve(:partitions)
+        # We should exit early if we are running on a node with no partitions
+        # (e.g. a diskless, netbooted VM)
+        return nil unless parts
+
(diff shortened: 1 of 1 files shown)


mhashizume:
Opening as a draft PR until I can do more testing

joshcooper:
Wondering if these other facts have the same issue:

```
❯ git grep ' mountpoints\.each ' lib
lib/facter/facts/freebsd/mountpoints.rb:        mountpoints.each do |mnt|
lib/facter/facts/linux/mountpoints.rb:        mountpoints.each do |mnt|
lib/facter/facts/solaris/mountpoints.rb:        mountpoints.each do |mnt|
```

mhashizume:
@joshcooper I don't think we need to worry about those facts because this should only be an issue with partitions, right?

mhashizume:
Confirmed with the user in the ticket that this fixes the issue.

joshcooper (approved):

PR #2576, opened 2023-05-23
(FACT-3178) Ensure Facter escapes metacharacters when searching for fact

AriaXLi:
Previously, when searching for a fact that matches the user query, if the fact name contains any metacharacters like dots (.) then the regex comparing the fact_name and processed_equery_fact would incorrectly match and result in the wrong fact being resolved. 

For example, when running `facter ldom.domainrole.impl`, the legacy alias fact, ldom_domainrole_impl is returned instead of the ldom structured core fact. This leads to issues later when extracting the fact value because you cannot dig into ldom_domainrole_impl and nil would be returned as its value. 

This commit ensures any special meta characters are escaped in the regex so the correct fact is found and resolved.

### DIFF
--- a/spec/facter/query_parser_spec.rb
+++ b/spec/facter/query_parser_spec.rb
@@ -79,6 +79,25 @@ describe Facter::QueryParser do
         contain_exactly(an_instance_of(Facter::SearchedFact).and(having_attributes(fact_class: path_class)))
     end
 
+    context 'when user query contains metacharacters (like dots)' do
+      it 'finds the structured fact and not the alias' do
+        query_list = ['ldom.domainrole.impl']
+        ldom_class = 'Facter::Solaris::Ldom'
+
+        loaded_fact_ldom = double(Facter::LoadedFact, name: 'ldom', klass: ldom_class, type: :core, file: nil)
+        ldom_fact_ldom_alias = double(Facter::LoadedFact, name: 'ldom_domainrole_impl', klass: ldom_class,
+                                                          type: :legacy, file: nil)
+        loaded_facts = [loaded_fact_ldom, ldom_fact_ldom_alias]
+
+        matched_facts = Facter::QueryParser.parse(query_list, loaded_facts)
+
+        expect(matched_facts).to be_an_instance_of(Array).and \
+          contain_exactly(an_instance_of(Facter::SearchedFact)
+          .and(having_attributes(name: 'ldom', user_query: 'ldom.domainrole.impl', fact_class: ldom_class,
+                                 type: :core)))
+      end
+    end
+
     context 'when fact does not exist' do
       let(:query_list) { ['non_existing_fact'] }
       let(:loaded_facts) { [] }
--- a/lib/facter/framework/parsers/query_parser.rb
+++ b/lib/facter/framework/parsers/query_parser.rb
@@ -85,7 +85,8 @@ module Facter
 
         return false if fact_with_wildcard && !query_fact.match("^#{fact_name}$")
 
-        return false unless fact_with_wildcard || fact_name.match("^#{processed_equery_fact}($|\\.)")
+        # Must escape metacharacters (like dots) to ensure the correct fact is found
+        return false unless fact_with_wildcard || fact_name.match("^#{Regexp.escape(processed_equery_fact)}($|\\.)")
 
         true
       end


AriaXLi:
I used Josh's Facter branch ([here](https://github.com/joshcooper/facter/tree/ldom_debugging)) that copies the outout of virtinfo and creates fake ldom resovler/facter for linux to test my changes:

```
$ bundle exec facter ldom.domainrole.impl
LDoms

$ bundle exec facter solaris_ldom_domainrole_impl
LDoms
```

lib/facter/facts/solaris/ldom.rb
```diff
@@ -5,15 +5,15 @@ module Solaris
...
-        ldom_domainuuid
+        solaris_ldom_domainchassis
+        solaris_ldom_domaincontrol
+        solaris_ldom_domainname
+        solaris_dom_domainrole_control
+        solaris_ldom_domainrole_impl
+        solaris_dom_domainrole_ios
+        solaris_ldom_domainrole_root
+        solaris_ldom_domainrole_service
+        solaris_ldom_domainuuid
```
joshcooper, line 16: Are these aliases public facing, as in could someone call `Facter.resolve("solaris_ldom_domainname")` and expect to get a value back? Or are they just internal constructs?

lib/facter/facts/solaris/ldom.rb
```diff
@@ -5,15 +5,15 @@ module Solaris
...
-        ldom_domainrole_io
-        ldom_domainrole_root
-        ldom_domainrole_service
-        ldom_domainuuid
+        solaris_ldom_domainchassis
+        solaris_ldom_domaincontrol
+        solaris_ldom_domainname
+        solaris_dom_domainrole_control
+        solaris_ldom_domainrole_impl
+        solaris_dom_domainrole_ios
```
joshcooper, line 13:
```suggestion
        solaris_ldom_domainrole_ios
```

lib/facter/facts/solaris/ldom.rb
```diff
@@ -5,15 +5,15 @@ module Solaris
...
-        ldom_domainrole_control
-        ldom_domainrole_impl
-        ldom_domainrole_io
-        ldom_domainrole_root
-        ldom_domainrole_service
-        ldom_domainuuid
+        solaris_ldom_domainchassis
+        solaris_ldom_domaincontrol
+        solaris_ldom_domainname
+        solaris_dom_domainrole_control
```
joshcooper, line 11:
```suggestion
        solaris_ldom_domainrole_control
```

lib/facter/facts/solaris/ldom.rb
AriaXLi: I'm not too sure, I can run:

```
$ bundle exec facter solaris_ldom_domainrole_impl
LDoms
```

Would more work be needed to make the aliases public facing? I've been having a hard time wrapping my head around ALIASES

joshcooper: I was thinking more of the case where someone is referencing a legacy fact by its alias on the command line (`bundle exec`) `facter ldom_domainname` or in a puppet manifest:

```puppet
notice($facts['ldom_domainname'])
```

Note facter#main ships in puppet-agent#7.x where legacy facts aren't disabled by default.

I'm thinking this fix needs to be made in the code that matches the user query (`ldom.domainrole.control`) against the various `Facter::ResolvedFact` names. My guess is facter is not escaping the `.` in the user query when converting it to a regex?

spec/facter/query_parser_spec.rb
```diff
@@ -79,6 +79,25 @@
         contain_exactly(an_instance_of(Facter::SearchedFact).and(having_attributes(fact_class: path_class)))
     end
 
+    context 'when user query contains metacharacters (like dots)' do
+      it 'finds the structured fact and not the alias' do
+        query_list = ['ldom.domainname.impl']
+        ldom_class = 'Facter::Macosx::Ldom'
```
joshcooper, line 85: I think this class only exists in my PR? That said, I'm not 100% sure how to stub a loaded fact that only exists on a single platform (unlike `os.name`)

AriaXLi: Oh shoot, good catch

AriaXLi: I think I can just set it to Facter::Solaris::Ldom instead

spec/facter/query_parser_spec.rb
```diff
@@ -79,6 +79,25 @@
         contain_exactly(an_instance_of(Facter::SearchedFact).and(having_attributes(fact_class: path_class)))
     end
 
+    context 'when user query contains metacharacters (like dots)' do
+      it 'finds the structured fact and not the alias' do
+        query_list = ['ldom.domainname.impl']
+        ldom_class = 'Facter::Solaris::Ldom'
+
+        loaded_fact_ldom = double(Facter::LoadedFact, name: 'ldom', klass: ldom_class, type: :core, file: nil)
+        ldom_fact_ldom_alias = double(Facter::LoadedFact, name: 'ldom_domainrole_impl', klass: ldom_class,
```
joshcooper, line 88: I noticing the test still passed when I reverted the `query_parser.rb` fix, I think because we're querying for `domainname` on line 84, but stubbing `domainrole` here?

```suggestion
        ldom_fact_ldom_alias = double(Facter::LoadedFact, name: 'ldom_domainname_impl', klass: ldom_class,
```

AriaXLi: Shoot, keep on making small mistakes. I fixed it so it's `domainrole` and the test no longer passes when the changes in `query_parser.rb` are reverted but will pass with the changes

joshcooper (approved):
awesome, nice work!

PR #2578, opened 2023-05-25
(FACT-3089) Retain full uname output for ISA

mhashizume:
Previous to this commit, the processor instruction set architecture (ISA) fact on Linux would split the output for `uname -p` on periods and return only the last value. Depending on the Linux distribution and hardware, this could result in unexpected output.

This commit changes the ISA fact on Linux to return the full output from `uname -p`, which is in line with how this fact is handled on other *nix operating systems (i.e. macOS, AIX, Solaris).


mhashizume:
I haven't found a machine type to be able to get a full processor name from `uname -p`. Setting this PR as a draft til I can find one and reproduce the user's original issue.

joshcooper (approved):

PR #2583, opened 2023-06-01
(FACT-3167) Handle when an AIX network interface is down

cthorn42:
Previously if an interface was down we'd attempt to set the MTU of an interface that didn't exist. We now check for that.


lib/facter/resolvers/aix/networking.rb
```diff
@@ -48,7 +48,7 @@ def populate_with_mtu_and_mac!(interfaces)
               info = line.split("\s")
               interface_name = info[0]
               mac = info[3][/^([0-9a-f]{1,2}[\.:-]){5}([0-9a-f]{1,2})$/]
-              interfaces[interface_name][:mtu] = info[1].to_i
+              interfaces[interface_name][:mtu] = info[1].to_i unless interfaces[interface_name].nil?
               interfaces[interface_name][:mac] = format_mac_address(mac) if mac
```
joshcooper, line 52: Wondering if we'll have a similar issue if the interface is down, but there is a `mac`? Maybe guard both of these lines with:

```ruby
intf = interfaces[interface_name]
if intf
  intf[:mtu] = ...
  intf[:mac] = ... if mac
end
```

spec/facter/resolvers/aix/networking_spec.rb
```diff
@@ -79,4 +79,21 @@
...
+      {
+        'en0' => { bindings: [{ address: '10.32.77.1', netmask: '255.255.255.0', network: '10.32.77.0' }],
+                   ip: '10.32.77.1', mac: '0a:c6:24:39:41:03', netmask: '255.255.255.0',
+                   network: '10.32.77.0' }
+      }
+    end
+    let(:primary) { 'en0' }
+
+    it 'returns a nil MTU' do
+      expect(networking_resolver.resolve(:mtu)).to be_nil
```
joshcooper, line 96: Unfortunately this test still passes if I revert the change to `lib/facter/resolvers/aix/networking.rb` I think it's because the list of interfaces (as stubbed in https://github.com/puppetlabs/facter/pull/2583/files#diff-25535a36374203ce766151e8678356763aff589490306fa3b99a41d53f14206cR25) still contains the `en0` interface. In order to repro, I think you'll need to exclude `en0` from the output of `netstat -in`, which can be done by specifying:

```ruby
let(:netstat_in) { load_fixture('netstat_in').read }
let(:netstat_rn) { load_fixture('netstat_rn').read }
let(:ffi_interfaces) do
  {
    'lo0' => { bindings: [{ address: '127.0.0.1', netmask: '255.0.0.0', network: '127.0.0.0' }],
               bindings6: [{ address: '::1', netmask: '::', network: '::', scope6: 'host' }], ip: '127.0.0.1',
               ip6: '::1', mtu: 16_896, netmask: '255.0.0.0', netmask6: '::', network: '127.0.0.0',
               network6: '::', scope6: 'host' }
  }
end
```

Doing that does trigger the bug, but the all resolvers inherit from BaseResolver and it rescues `NoMethodError` https://github.com/puppetlabs/facter/blob/b64cd028869fa613a341d5d87fe83b789a7dc12c/lib/facter/resolvers/base_resolver.rb#L30-L31 If I add some debug then I see the exception being rescued and logged to the "log_spy"

```
#<NoMethodError: undefined method `[]=' for nil:NilClass

              interfaces[interface_name][:mtu] = info[1].to_i
                                        ^^^^^^^^>
```

That is unfortunate, I really don't like blindly catching NoMethodError like that... I think maybe for this test you can assert that the log spy never receive "undefined method", something like:

```ruby
expect(log_spy).to receive(:debug).with(/undefined method/).never
```

joshcooper: I filed https://tickets.puppetlabs.com/browse/FACT-3207

joshcooper (approved):

PR #2584, opened 2023-06-07
(FACT-3204) Revert "(maint) Skip vlans test for Ubuntu2204"

tvpartytonight:
This reverts commit fb4d13de157c7961e27d874f16affe390fc2707f.


tvpartytonight:
The test passes now in our jenkins CI without any alteration when run in isolation from my laptop; I think it's worth re-enabling in ci to see if it still fails.

mhashizume (approved):

PR #2586, opened 2023-06-12
(FACT-1902) Ensure Facter validates that facts output proper UTF-8

AriaXLi:
This commit adds spec tests to ensure that all fact types output proper UTF-8 and if unable to, an exception will be raised with a message that specifies which fact has invalid data.

### DIFF
--- a/spec/custom_facts/util/normalization_spec.rb
+++ b/spec/custom_facts/util/normalization_spec.rb
@@ -56,6 +56,23 @@ describe LegacyFacter::Util::Normalization do
         expect do
           normalization.normalize(str)
         end.to raise_error(LegacyFacter::Util::Normalization::NormalizationError,
+                           /String \"\\xADay!\" doesn't match the reported encoding UTF-8/)
+      end
+
+      it 'rejects strings that only have the start of a valid UTF-8 sequence' do
+        str = "\xc3\x28"
+        expect do
+          normalization.normalize(str)
+        end.to raise_error(LegacyFacter::Util::Normalization::NormalizationError,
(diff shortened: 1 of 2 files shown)


spec/custom_facts/util/normalization_spec.rb
```diff
@@ -126,4 +134,16 @@ def utf8(str)
...
+    it 'correctly matches encoding for UTF-16LE byte sequence' do
+      str = 'Hello World!'.encode(Encoding::UTF_16LE)
+      expect(str.encoding).to eq(Encoding::UTF_16LE)
+    end
+
+    it 'correctly matches encoding for string that contains binary data' do
+      str = 'Hello World!'.encode(Encoding::BINARY) + 'Hello again!'
+      expect(str.encoding).to eq(Encoding::BINARY)
+    end
+  end
```
joshcooper, line 148: Wasn't sure about this part, were you wanting to pass a non-UTF8 string to `normalize` and verify you get back UTF8, something like?

```ruby
str = 'Hello World!'.encode(Encoding::UTF_16LE)
expect(normallization.normalize(str.encoding)).to eq(Encoding::UTF_8)
```

AriaXLi: Ah I was trying to make tests for these examples in the ticket:

- String contains a valid UTF-16LE byte sequence, but the string's "encoding" method returns UTF-8 (Windows Registry)
- String contains binary data, but "encoding" returns UTF-8 (EC2 userdata)

joshcooper (approved):

PR #2589, opened 2023-06-23
Add is_virtual and virtual support for Nutanix AHV

eman-koch:
Prior to this change facter returned:

    $ facter is_virtual
    false
    $ facter virtual
    physical

After this change facter returns:

    $ facter is_virtual
    true
    $ facter virtual
    ahv

Systems running on Nutanix AHV are correctly recognized now.

### DIFF
--- a/lib/facter/resolvers/windows/virtualization.rb
+++ b/lib/facter/resolvers/windows/virtualization.rb
@@ -13,7 +13,7 @@ module Facter
           # Is_Virtual
 
           MODEL_HASH = { 'VirtualBox' => 'virtualbox', 'VMware' => 'vmware', 'KVM' => 'kvm',
-                         'Bochs' => 'bochs', 'Google' => 'gce', 'OpenStack' => 'openstack' }.freeze
+                         'Bochs' => 'bochs', 'Google' => 'gce', 'OpenStack' => 'openstack', 'AHV' => 'ahv' }.freeze
 
           private
 


eman-koch:
@h0tw1r3 could you give it a look? It would be very helpful to have it in the code and I think puppetlabs/phoenix is no longer active.

h0tw1r3:
@eman-koch I'm not on the team, and I'm definitely not familiar with Nutanix AHV. 

General feedback:
Seems to be missing some things. The only file modified is a windows resolver.  The virtual fact uses virt-what under the hood.
I would also expect a hypervisor fact, see some examples here: https://github.com/puppetlabs/facter/tree/main/lib/facter/facts/linux/hypervisors
Don't forget to add spec tests!

kenyon:
I also see this bug on Windows VMs running on Nutanix AHV with Puppet 8.3.1, facter 4.5.1, PE 2023.5.0.

@h0tw1r3 this is a fix for facter running on Windows only, not Linux, so virt-what isn't applicable. These facts are already correct for Linux VMs running under Nutanix AHV, but they are incorrect for Windows VMs on Nutanix AHV. The `hypervisors` fact is already correctly reported as `kvm` on Windows (Nutanix AHV is really just a rebranded KVM).

Looks like AHV should be added to some tests though.

kenyon:
Seems like this might have gotten lost somehow in the conversion from C++ to Ruby: #1817.

kenyon:
I rebased this on current `main` and added a test in #2692.

joshcooper:
This commit was merged in the linked PR so closing this one.

PR #2590, opened 2023-06-23
(FACT-3198) Do not allow null bytes in names and values

tvpartytonight:
When storing reports in puppetdb, postgres does not allow for the null utf-8 byte. This change normalizes all strings for both fact keys and values to disallow that null byte.

### DIFF
--- a/spec_integration/facter_spec.rb
+++ b/spec_integration/facter_spec.rb
@@ -55,6 +55,38 @@ describe 'Facter' do
     end
 
     context 'with custom facts' do
+      context 'with null bytes' do
+        let(:logger) { instance_spy(Facter::Log) }
+
+        normalization_error = LegacyFacter::Util::Normalization::NormalizationError
+
+        it 'logs an error when a fact keyname contains a null byte' do
+          facter_code = proc do
+            Facter.add("key\0bad_name") do
+              setcode { 'valid_value' }
+            end
+          end
+          facter_code.call do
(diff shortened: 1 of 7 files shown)


spec_integration/facter_spec.rb
```diff
@@ -99,6 +123,27 @@ def tmp_filename(tmp_filename)
...
+      context 'with facts that have null bytes' do
+        before do
+          Facter.search_external([ext_facts_dir])
+          data = { "bad_key\u0000" => 'valid value',
+                   'valid_key' => "\x00" }
+          write_to_file(tmp_filename('bad_facts.yaml'), YAML.dump(data))
+        end
+
+        it 'shows errors in the logs when trying to read in the facts with bad keys' do
+          expect { Facter.value("bad_key\u0000") }.to raise_error(ArgumentError)
```
tvpartytonight, line 135: When trying to read in fact keys, facter ends up calling `load` in the fact loader, which attempts to read in a file name with the key with the null byte [here](https://github.com/puppetlabs/facter/blob/85797f4/lib/facter/custom_facts/util/loader.rb#L26). The error comes from `File.join`, which raises an Argument error `string contains null byte`. Since the error seemed reasonable, I left it alone for now. I realize now it doesn't contain any indication of *which* string contains a null byte, which could make it harder for debugging later.

cthorn42 (approved):
LGTM 👍

PR #2597, opened 2023-07-25
(FACT-3188) Guard against nil for solaris networking fact 

AriaXLi:
This commit updates the Solaris networking resolver to guard against
nil when interacting with expanding primary interfaces.

### DIFF
--- a/lib/facter/util/resolvers/networking/networking.rb
+++ b/lib/facter/util/resolvers/networking/networking.rb
@@ -36,7 +36,9 @@ module Facter
             interfaces = networking_facts[:interfaces]
 
             expand_interfaces(interfaces) unless interfaces.nil?
-            expand_primary_interface(networking_facts, primary) unless primary.nil? || interfaces.nil?
+            return if primary.nil? || interfaces.nil? || networking_facts.nil?
+
+            expand_primary_interface(networking_facts, primary)
           end
 
           def get_scope(ip)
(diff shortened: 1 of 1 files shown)


lib/facter/util/resolvers/networking/networking.rb
```diff
@@ -36,7 +36,8 @@ def expand_main_bindings(networking_facts)
             interfaces = networking_facts[:interfaces]
 
             expand_interfaces(interfaces) unless interfaces.nil?
-            expand_primary_interface(networking_facts, primary) unless primary.nil? || interfaces.nil?
+            expand_primary_interface(networking_facts, primary) unless
+              primary.nil? || interfaces.nil? || networking_facts.nil?
```
mhashizume, line 40: To me, ending a statement with `unless` makes sense when it's brief; once it's long enough to be broken up with a line break, it seems more readable to use the full syntax:
```
unless primary.nil? || interfaces.nil? || networking_facts.nil?
  expand_primary_interface(networking_facts, primary)
end
```

AriaXLi: Oh I didn't know that was the full expression!

AriaXLi: Hmm...Rubocop isn't happy if there isn't also an else after

AriaXLi: Ah Rubocop suggested to use a guard clause instead so my change will be a lil different but still same idea!

mhashizume (approved):

PR #2600, opened 2023-08-02
(FACT-3162) Add 'normalize_os_name' to os release resolver

cthorn42:
When facter resolves the os.name for 'SLES SAP' OS it would include the entire 'SLES_SAP' as the name.
This would break things further down stream like the zypper package provider. This PR fixes the os.name fact so SLES_SAP will correctly resolve as 'SLES'.

### DIFF
--- a/lib/facter/resolvers/os_release.rb
+++ b/lib/facter/resolvers/os_release.rb
@@ -75,9 +75,19 @@ module Facter
         def process_name
           return unless @fact_list[:name]
 
+          normalize_os_name
           join_os_name
           capitalize_os_name
           append_linux_to_os_name
+        end
+
+        def normalize_os_name
+          os_name = @fact_list[:name]
+          @fact_list[:name] = if os_name.downcase.start_with?('sles')
+                                'SLES'
+                              else
+                                os_name
+                              end
         end
 
         def join_os_name
--- a/spec/facter/resolvers/os_release_spec.rb
+++ b/spec/facter/resolvers/os_release_spec.rb
@@ -138,6 +138,26 @@ describe Facter::Resolvers::OsRelease do
     end
   end
 
+  context 'when on SLES' do
+    let(:os_release_content) { load_fixture('os_release_sles').readlines }
+
+    it 'returns os NAME' do
+      result = Facter::Resolvers::OsRelease.resolve(:name)
+
+      expect(result).to eq('SLES')
+    end
+  end
+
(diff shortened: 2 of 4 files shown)


lib/facter/resolvers/os_release.rb
```diff
@@ -75,11 +75,22 @@ def process_id
...
+          normalize_os_name
           join_os_name
           capitalize_os_name
           append_linux_to_os_name
         end
 
+        def normalize_os_name
+          os_name = @fact_list[:name]
+          @fact_list[:name] = if os_name.downcase.start_with?('sles')
+                                os_name = 'SLES'
```
joshcooper, line 87: Not sure if this matters, but should this be `Sles`? based on:

```
irb(main):001:0> "sles".capitalize
=> "Sles"
```

Mostly care about matching whatever Facter 3 did.

cthorn42: I'm just piggy backing off of the other helper methods that look like they just downcase to make the string comparison a bit easier.
I'm not great at reading facter 3, but it looks like at https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/inc/internal/facts/linux/os_osrelease.hpp#L46-L47, facter 3 looked for the `ID`, not the `NAME`.

From one of our sles-15-x86_64 hosts, here is what the os-release file looks like:
```
cat /etc/os-release
NAME="SLES"
VERSION="15"
VERSION_ID="15"
PRETTY_NAME="SUSE Linux Enterprise Server 15"
ID="sles"
ID_LIKE="suse"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:15"
```

In the case of the `ID`, that looks to be in lower case.
For the `NAME`, the value of that is set to upper case.

I think in this case matching what we did in Facter 3 isn't very clear cut. Unless I'm misreading what is going, which I could very well be.

joshcooper: Ah yeah if `ID == "sles"` then we return `os::suse_enterprise_server` which is defined in https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/inc/facter/facts/os.hpp#L103C1-L104C1 So this PR behaves the same.

joshcooper (approved):

PR #2602, opened 2023-08-23
(FACT-3424) squish me

tvpartytonight:


tvpartytonight:
It's difficult to test running facter without development dependencies in either the spec tests or acceptance tests. Spec tests by their very nature have rspec defined as a development dependency, so we cannot test for this in rspec. The acceptance tests install the agent which will install sys-filesystem as a vanagon defined component, so testing at the acceptance level also uncover this error.

Instead, I just added a step in the `presuite.rb` to `bundle install --without development` and then run facter once to ensure that it can run without error.  You can see this test failing [here](https://github.com/puppetlabs/facter/actions/runs/5957255123/job/16159611888?pr=2602) before I moved the sys-filesystem to a runtime-dependency.

mhashizume (approved):

asottile:
fwiw this brings in a dependence on the `ffi` gem to `puppet` -- meaning puppet can no longer be installed without a function C compiler, make toolchain, ffi headers, etc.

this greatly inhibits the ability to use `puppet` to bootstrap a machine because it now requires a whole bunch of manual setup beforehand

joshcooper:
@asottile yep thanks for letting us know. We're working on a fix in https://github.com/puppetlabs/facter/pull/2604

PR #2604, opened 2023-08-29
(FACT-3423) sys-filesystem updates

mhashizume:
This PR:

- Moves sys-filesystem back from a runtime dependency to a development dependency.
- Rescues LoadError in resolvers that depend on sys-filesystem.
- Adds ffi as a development dependency, as some resolvers depend on ffi but it's only implicitly required through sys-filesystem.
- Updates resolvers that depend on ffi to log a debug message when ffi is not present.
- Fixes a minor Rubocop issue

### DIFF
--- a/facter.gemspec
+++ b/facter.gemspec
@@ -29,16 +29,20 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
+  # While we require sys-filesystem in filesystem_helpers we leave it as a
+  # development, not runtime, dependency. We do this so we do not also pull in
+  # sys-filesystem's dependency ffi, which contains native extensions which in
+  # turn would mean Facter would require a compiler to install as a gem.
   spec.add_development_dependency 'rake', '~> 13.0', '>= 13.0.6'
   spec.add_development_dependency 'rspec', '~> 3.0'
   spec.add_development_dependency 'rubocop', '~> 0.81.0'
   spec.add_development_dependency 'rubocop-performance', '~> 1.5.2'
   spec.add_development_dependency 'rubocop-rspec', '~> 1.38'
   spec.add_development_dependency 'simplecov', '~> 0.17.1'
+  spec.add_development_dependency 'sys-filesystem', '~> 1.4'
   spec.add_development_dependency 'webmock', '~> 3.12'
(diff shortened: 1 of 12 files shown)


lib/facter/resolvers/aix/ffi/ffi_helper.rb
```diff
@@ -1,8 +1,16 @@
...
 require_relative 'ffi'
+
+# ffi is only a development dependency and may not be present
+begin
+  require 'ffi'
+rescue LoadError => e
+  log = Facter::Log.new(self)
+  log.debug("The ffi gem has not been installed: #{e}")
+end
+
```
joshcooper, line 13: In theory, `ffi` related code is never required in the main entry point https://github.com/puppetlabs/facter/blob/4c69cb9aa08ecec4e3ab56cb47656c4c826d38dc/lib/facter/framework/core/file_loader.rb#L28-L97

Instead, we lazily load `ffi_helper`, etc such as above when trying to resolve the fact at runtime, like the AIX networking facts: https://github.com/puppetlabs/facter/blob/4c69cb9aa08ecec4e3ab56cb47656c4c826d38dc/lib/facter/resolvers/aix/networking.rb#L63 

I think I'd expect `require 'facter'` to never cause `ffi` to be loaded unless a specific fact depends on it. And if that happens, I think an `error` message should be displayed that some facts will be omitted because ffi is not installed.

There's a possible catch-22 though because facter does execute some code to figure out which OS it's on and therefore, which OS specific resolvers to load. So maybe that is triggering ffi to be loaded?

mhashizume: It looks like Windows always loads the identity resolver:

https://github.com/puppetlabs/facter/blob/4c69cb9aa08ecec4e3ab56cb47656c4c826d38dc/lib/facter/framework/core/file_loader.rb#L902

Which in turn uses FFI to find usernames: 

https://github.com/puppetlabs/facter/blob/main/lib/facter/resolvers/windows/identity.rb#L19

So I can try rescuing LoadError in the `find_username` method.

tvpartytonight (approved):
I'm not sure if there is a better pattern than the implementing a LoadError rescue for each of these locations, seems like a whack-a-mole issue that could end up being problematic later on, but I think it works for now to remove the unwanted transitive dependency.

joshcooper (approved):

PR #2606, opened 2023-09-12
(FACT-3165) Report correct location of where a fact was resolved from

tvpartytonight:


lib/facter/custom_facts/core/aggregate.rb
```diff
@@ -101,7 +108,22 @@ def options(options)
...
         instance_eval(&block)
+
+        # Ruby 1.9+ provides the source location of procs which can provide useful
+        # debugging information if a resolution is being evaluated twice. Since 1.8
+        # doesn't support this we opportunistically provide this information.
+        @last_evaluated = if block.respond_to? :source_location
+                            block.source_location.join(':')
+                          else
+                            true
+                          end
```
mhashizume, line 126: I don't think we need to account for versions of Ruby older than 1.9 since we only claim to support >= 2.5 in the Facter gemspec

tvpartytonight: good point!

lib/facter/custom_facts/util/fact.rb
```diff
@@ -223,8 +223,7 @@ def find_first_real_value(resolutions)
       end
 
       def log_fact_path(resolve)
-        fact = resolve.fact
-        log.debug("#{resolve.fact_type} fact #{fact.name} got resolved from: #{fact.location}")
+        log.debug("#{resolve.fact_type} fact #{resolve.fact.name} was resolved from: #{resolve.last_evaluated}")
```
joshcooper, line 226: Does this test both aggregate and non-aggregate custom facts?

tvpartytonight: I believe the tests run here are sufficient? https://github.com/puppetlabs/facter/blob/9a72dd5/acceptance/tests/custom_facts/custom_facts_aggregate_resolutions.rb

joshcooper (approved):
Using the repro from the ticket, the `one` fact is registered first and the fact value resolves to `one`, but facter reports the location as coming from `two`:

```
[2023-09-15 18:08:57.510196 ] INFO Facter - executed with command line: foo --debug --custom-dir custom_facts 
registering one.rb
registering two.rb
[2023-09-15 18:08:57.519856 ] DEBUG Facter::QueryParser - List of resolvable facts: [#<Facter::SearchedFact:0x000055e484b74490 @name="foo", @fact_class=nil, @user_query="foo", @type=:custom, @file=nil>] 
[2023-09-15 18:08:57.519972 ] DEBUG Facter::Util::Fact - custom fact foo got resolved from: ["custom_facts/two.rb", 1] 
[2023-09-15 18:08:57.521208 ] DEBUG Facter::FactManager - fact "foo" has resolved to: one 
```

The new behavior is facter reports the fact as having been resolved from `one` with proper file name and line number:

```
[2023-09-15 18:08:50.847471 ] INFO Facter - executed with command line: foo --debug --custom-dir custom_facts 
registering one.rb
registering two.rb
[2023-09-15 18:08:50.854295 ] DEBUG Facter::QueryParser - List of resolvable facts: [#<Facter::SearchedFact:0x000055f7493b20e0 @name="foo", @fact_class=nil, @user_query="foo", @type=:custom, @file=nil>] 
[2023-09-15 18:08:50.854388 ] DEBUG Facter::Util::Fact - custom fact foo was resolved from: custom_facts/one.rb:1 
[2023-09-15 18:08:50.855584 ] DEBUG Facter::FactManager - fact "foo" has resolved to: one 
```

PR #2608, opened 2023-09-14
(FACT-3425) Remove #dir_children method

mhashizume:
09cfa94 added a new method #dir_children to Facter to provide the functionality of Ruby's Dir.children (added in Ruby 2.5) while still supporting older (< 2.5) versions of Ruby.

Because Facter no longer supports versions of Ruby older than 2.5, this commit removes the #dir_children method in favor of Ruby's built-in Dir.children.


lib/facter/util/file_helper.rb
```diff
@@ -22,16 +22,6 @@ def safe_readlines(path, default_return = [])
...
-        def dir_children(path)
-          children = if RUBY_VERSION.to_f < 2.5
-                       Dir.entries(path).reject { |dir| ['.', '..'].include?(dir) }
-                     else
-                       Dir.children(path)
-                     end
-
-          children
-        end
-
```
joshcooper, line 34: Since this is a public method and callable by custom facts, I think we need to keep the method and just have it call `Dir.children(path)`

mhashizume:
I've updated this PR to keep the dir_children method, but simplify it.

joshcooper (approved):

PR #2609, opened 2023-09-16
(FACT-1592) A key/value pair for Interface Speed/Duplex

joshcooper:
Collect the speed and duplex settings for physical network interfaces on Linux using `/sys/class/net`.

This supersedes https://github.com/puppetlabs/facter/pull/2480


mhashizume:
I see that this PR adds tests for the state of duplex on an interface, could we add some tests for interface speed as well?

mhashizume (approved):

mhashizume:
LGTM, thanks @jcpunk !

PR #2610, opened 2023-09-20
(maint) Bump JRuby version in unit tests

mhashizume:
This commit updates JRuby in our GitHub Actions unit tests to match the version of the most recent version of puppetserver 7.


mhashizume:
After this is approved, we'll need to update branch protection rules to change the required check from the old JRuby to the new one.

joshcooper (approved):

PR #2611, opened 2023-09-22
(FACT-3427) Fix logger arguments

mhashizume:
Prior to this commit, the define_resolution method was incorrectly passing two arguments to Facter::Log#log_exception when raising. This commit updates the call to log_exception to no longer pass a string and instead only pass an exception.

Additionally, this enables a test that was pending.

### DIFF
--- a/lib/facter/custom_facts/util/fact.rb
+++ b/lib/facter/custom_facts/util/fact.rb
@@ -85,7 +85,9 @@ module Facter
 
         resolve
       rescue StandardError => e
-        log.log_exception("Unable to add resolve #{resolution_name.inspect} for fact #{@name}: #{e.message}")
+        msg = "Unable to add resolve #{resolution_name.inspect} for fact '#{@name}': #{e.message}"
+        msg += "\n" + e.backtrace.join("\n") if Options[:trace]
+        log.error(msg, true)
         nil
       end
 
--- a/spec/custom_facts/util/fact_spec.rb
+++ b/spec/custom_facts/util/fact_spec.rb
@@ -107,12 +107,15 @@ describe Facter::Util::Fact do
       expect(fact.resolution('named')).to be_a_kind_of Facter::Core::Aggregate
     end
 
-    # it "raises an error if there is an existing resolution with a different type" do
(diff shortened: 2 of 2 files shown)


lib/facter/custom_facts/util/fact.rb
```diff
@@ -85,7 +85,7 @@ def define_resolution(resolution_name, options = {}, &block)
 
         resolve
       rescue StandardError => e
-        log.log_exception("Unable to add resolve #{resolution_name.inspect} for fact #{@name}: #{e.message}")
+        log.log_exception(e)
```
joshcooper, line 88: I'm wondering if the exception will always contain enough information to troubleshoot which fact name and ideally file location needs to be fixed? If not, may be we keep the message and just pass the exception as the first argument like: 
```ruby
log_exception(e, "Unable to add resolve #{resolution_name.inspect} for fact #{@name}: #{e.message}")
```

mhashizume: I tried that at first, but this method only takes an exception as an argument and gave me an ArgumentError when I tried to pass the message as well: https://github.com/puppetlabs/facter/blob/163e1bb6b7440d62c8326cd7e916f2394aad3ab2/lib/facter/framework/logging/logger.rb#L163

mhashizume:
Rebased on main and force pushed to get latest checks

lib/facter/custom_facts/util/fact.rb
```diff
@@ -85,8 +85,9 @@ def define_resolution(resolution_name, options = {}, &block)
 
         resolve
       rescue StandardError => e
-        log.log_exception("Unable to add resolve #{resolution_name.inspect} for fact #{@name}: #{e.message}")
-        nil
+        msg = "Unable to add resolve #{resolution_name.inspect} for fact '#{@name}': #{e.message}"
+        msg += "\n" + e.backtrace.join("\n") if Options[:trace]
+        log.error(msg, true)
```
joshcooper, line 90: Because of ruby's "return whatever was last evaluted" I think we want to do to preserve compatibility:

```suggestion
        log.error(msg, true)
        nil
```

joshcooper (approved):

PR #2612, opened 2023-09-25
(maint) Update beaker-hostgenerator version

mhashizume:
beaker-hostgenerator released 2.0 earlier this year, and since that release all new platforms, features, and fixes have been going into  > 2.0.

This commit updates the acceptance Gemfile to depend on beaker-hostgenerator ~> 2 instead of ~> 1.1.


joshcooper (approved):
could you check/update `puppet` & `puppet-agent` (7.x and main) too?

PR #2620, opened 2023-10-04
(FACT-3441) improve reliability of uname resolver

jonathannewman:
In the case that the uname command fails to return expected content, this adds validation to the output looking for both an empty result, or missing results.

Logging is added to help identify the condition.

### DIFF
--- a/lib/facter/resolvers/uname.rb
+++ b/lib/facter/resolvers/uname.rb
@@ -28,12 +28,16 @@ module Facter
         def build_fact_list(output)
           uname_results = output.split("\n")
 
-          @fact_list[:machine] = uname_results[0].strip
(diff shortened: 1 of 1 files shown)


jonathannewman:
Note that in testing, we are seeing occasional failures with puppetserver failing to start because of nil dereferences related to this code.

jonathannewman:
@joshcooper done!

joshcooper:
@jonathannewman could you rebase to pick up this change https://github.com/puppetlabs/facter/pull/2625?

jonathannewman:
Sorry @joshcooper didn't see your ping.  rebased.

mhashizume (approved):

mhashizume:
Thanks for this @jonathannewman !

PR #2621, opened 2023-10-09
(FACT-3436) Standardize Debian architecture fact

mhashizume:
Prior to this commit the os.architecture fact on Debian and Debian-based distrubitions resolved inconsistently. The fact used Debian's architecture naming scheme for some platforms (i386 and amd64) but not others.

This commit updates the os.architecture fact on Debian and Debian-based distrubitions to consistently use Debian's naming convention by relying on the output of `dpkg --print-architecture`.

### DIFF
--- a/lib/facter/facts/debian/architecture.rb
+++ b/lib/facter/facts/debian/architecture.rb
@@ -8,9 +8,7 @@ module Facts
         ALIASES = 'architecture'
 
         def call_the_resolver
-          fact_value = Facter::Resolvers::Uname.resolve(:machine)
-          fact_value = 'amd64' if fact_value == 'x86_64'
-          fact_value = 'i386' if /i[3456]86|pentium/.match?(fact_value)
+          fact_value = Facter::Core::Execution.execute('dpkg --print-architecture')
 
           [Facter::ResolvedFact.new(FACT_NAME, fact_value), Facter::ResolvedFact.new(ALIASES, fact_value, :legacy)]
         end
--- a/spec/facter/facts/debian/os/architecture_spec.rb
+++ b/spec/facter/facts/debian/os/architecture_spec.rb
@@ -4,61 +4,16 @@ describe Facts::Debian::Os::Architecture do
   describe '#call_the_resolver' do
     subject(:fact) { Facts::Debian::Os::Architecture.new }
 
-    context 'when resolver does not return x86_64' do
-      let(:value) { 'i86pc' }
+    %w[amd64 arm64 i386].each do |arch|
+      it 'calls Facter::Core::Execution and returns architecture fact' do
+        allow(Facter::Core::Execution).to receive(:execute).and_return(arch)
 
-      before do
-        allow(Facter::Resolvers::Uname).to receive(:resolve).with(:machine).and_return(value)
-      end
-
-      it 'calls Facter::Resolvers::Uname' do
         fact.call_the_resolver
-        expect(Facter::Resolvers::Uname).to have_received(:resolve).with(:machine)
-      end
+        expect(Facter::Core::Execution).to have_received(:execute).with('dpkg --print-architecture')
 
-      it 'returns architecture fact' do
         expect(fact.call_the_resolver).to be_an_instance_of(Array).and \
-          contain_exactly(an_object_having_attributes(name: 'os.architecture', value: value),
-                          an_object_having_attributes(name: 'architecture', value: value, type: :legacy))
-      end
-    end
-
-    context 'when os is 32-bit' do
-      let(:value) { 'i586' }
-      let(:fact_value) { 'i386' }
-
-      before do
-        allow(Facter::Resolvers::Uname).to receive(:resolve).with(:machine).and_return(value)
-      end
-
-      it 'calls Facter::Resolvers::Uname' do
-        fact.call_the_resolver
-        expect(Facter::Resolvers::Uname).to have_received(:resolve).with(:machine)
-      end
-
-      it 'returns architecture fact' do
-        expect(fact.call_the_resolver).to be_an_instance_of(Array).and \
-          contain_exactly(an_object_having_attributes(name: 'os.architecture', value: fact_value),
-                          an_object_having_attributes(name: 'architecture', value: fact_value, type: :legacy))
-      end
-    end
-
-    context 'when resolver returns x86_64' do
-      let(:value) { 'x86_64' }
-
-      before do
-        allow(Facter::Resolvers::Uname).to receive(:resolve).with(:machine).and_return(value)
-      end
-
-      it 'calls Facter::Resolvers::Uname' do
-        fact.call_the_resolver
-        expect(Facter::Resolvers::Uname).to have_received(:resolve).with(:machine)
-      end
-
-      it 'returns architecture fact' do
-        expect(fact.call_the_resolver).to be_an_instance_of(Array).and \
-          contain_exactly(an_object_having_attributes(name: 'os.architecture', value: 'amd64'),
-                          an_object_having_attributes(name: 'architecture', value: 'amd64', type: :legacy))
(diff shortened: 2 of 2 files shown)


mhashizume:
Hi @jcharaoui , we're in the process of adding official Debian aarch64/arm64 support in Puppet and since I know that you're a Debian maintainer, we were hoping you might be able to confirm that this is the desired behavior.

It seemed like from the [Supported Architectures doc](https://wiki.debian.org/SupportedArchitectures) that Debian generally prefers "arm64" over "aarch64," but I know architecture names can vary depending on the context.

Thank you!

jcharaoui:
I think the appropriate thing would be for facter to report the architecture as `arm64` but I'll consult with my Debian colleagues and let you know!

jcharaoui:
For what it's worth, running `facter architecture` now, using facter from Debian's own archive on an `arm64` installation will report `aarch64`.

jcharaoui:
One suggestion that's been made, if the goal is to report the Debian architecture, would be for facter to simply supply the output of `dpkg --print-architecture`.

mhashizume:
Thanks for the feedback @jcharaoui ! We're going to discuss this a bit internally. Do you feel the benefits of having this fact be consistent with Debian's convention and the `amd64`/`i386` fact would outweigh the potentially breaking change of going from `aarch64` to `arm64`?

bastelfreak:
Hey, module author here. In puppet modules we often need to check the architecture. Our live is much much easier when the usage of aarch64 vs arm64 is consistent for every operating system. I don't care which value is used, but it's easier to develop modules when it's the same. Otherwise we always need to check the architecture and the operating system.

jcharaoui:
> Thanks for the feedback @jcharaoui ! We're going to discuss this a bit internally. Do you feel the benefits of having this fact be consistent with Debian's convention and the `amd64`/`i386` fact would outweigh the potentially breaking change of going from `aarch64` to `arm64`?

I think a breaking change here is probably fine as long as it's properly documented. Since `aarch64` is only used by Debian (and derivatives) we can make sure the change enters the next release and comes with a user-facing announcement (`NEWS` file, in Debian packaging terms).

> Hey, module author here. In puppet modules we often need to check the architecture. Our live is much much easier when the usage of aarch64 vs arm64 is consistent for every operating system. I don't care which value is used, but it's easier to develop modules when it's the same. Otherwise we always need to check the architecture and the operating system.

Right, there's consistency within Debian, but also consistency with other operating systems. I'm not sure how other facter reports the architecture for operating systems . The [schema](https://github.com/puppetlabs/facter/blob/main/lib/schema/facter.yaml#L39) seems to suggest there is a conversion to `amd64` already happening on several operating systems when `uname` reports `x86_64`.

ekohl:
Another perspective on a breaking change: normalize the architecture on what `uname` outputs, meaning getting rid of `amd64` in favor of `x64_64`. Today in Foreman we already normalize that back:
https://github.com/theforeman/foreman/blob/d8200eeae33ff730cf85fcae9c47e92466c64a6a/app/services/puppet_fact_parser.rb#L58-L60

Because if you have a homogeneous fleet of servers you don't want those differences.

ekohl:
Perhaps you want `os.architecture` for the OS specific architecture code while `architecture` is what `uname` returns.

mhashizume:
We also already massage `uname` output for the generic Linux fact: https://github.com/puppetlabs/facter/blob/a866afe3c4eccc74d1ae519aeecd8dc51b686820/lib/facter/facts/linux/os/architecture.rb#L13

While `os.hardware` seems to be the output of `uname -m` without any modification, at least for all Linux platforms, macOS, and FreeBSD.

It seems like using the latter could be the more uniform option across many nodes/platforms (as @bastelfreak described in his comment)

ekohl:
That looks like a useful fact.

mhashizume:
I've updated this PR with @jcharaoui 's suggestion of using `dpkg` to determine architecture.

This should address inconsistencies between architectures on Debian and will prevent us from having to periodically update mappings between `uname -m` and what Debian uses.

`os.hardware` should be able to address the issue of this type of fact varying between distros.

lib/facter/facts/debian/architecture.rb
```diff
@@ -8,9 +8,7 @@ class Architecture
         ALIASES = 'architecture'
 
         def call_the_resolver
-          fact_value = Facter::Resolvers::Uname.resolve(:machine)
-          fact_value = 'amd64' if fact_value == 'x86_64'
-          fact_value = 'i386' if /i[3456]86|pentium/.match?(fact_value)
+          fact_value = Facter::Core::Execution.execute('dpkg --print-architecture')
```
joshcooper, line 11: Does this command return `i386` for all 32-bit Intel processors? I assume yes, but wanted to verify so we don't regress.

Also is there any user configuration that could cause `dpkg --print-architecture` to return 0 or more than 1 result? It believe 0 can never happen because:

> --remove-architecture architecture
> The architecture dpkg is built for (i.e. the  output  of --print-architecture) can never be removed from that list

But I wasn't sure if adding an entry via `--add-architecture` could cause the `--print-architecture` option to print multiple values? Does enabling `multi-arch` change that?

jcharaoui: > Does this command return i386 for all 32-bit Intel processors? I assume yes, but wanted to verify so we don't regress.

Very likely yes, because the command returns the operating system architecture, it's not dependent on any specific hardware being present.

> But I wasn't sure if adding an entry via --add-architecture could cause the --print-architecture option to print multiple values? Does enabling multi-arch change that?

Architectures added in this way are considered "foreign", and can be shown with `dpkg --print-foreign-architectures`.

In addition, all architectures present on the system are available in `/var/lib/dpkg/arch` with the first line being the native architecture.

mhashizume: I confirmed that adding additional architectures via `dpkg --add-architecture` does not affect the output of `dpkg--print-architecture`

```
root@ip-10-227-4-242:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 12 (bookworm)
Release:	12
Codename:	bookworm
root@ip-10-227-4-242:~# dpkg --print-architecture
arm64
root@ip-10-227-4-242:~# dpkg --print-foreign-architectures
root@ip-10-227-4-242:~# dpkg --add-architecture i386
root@ip-10-227-4-242:~# dpkg --add-architecture ppc64el
root@ip-10-227-4-242:~# dpkg --print-foreign-architectures
i386
ppc64el
root@ip-10-227-4-242:~# dpkg --print-architecture
arm64
```

spec/facter/facts/debian/os/architecture_spec.rb
```diff
@@ -4,61 +4,16 @@
   describe '#call_the_resolver' do
     subject(:fact) { Facts::Debian::Os::Architecture.new }
 
-    context 'when resolver does not return x86_64' do
-      let(:value) { 'i86pc' }
+    it 'calls Facter::Core::Execution and returns architecture fact' do
+      %w[amd64 arm64 i386].each do |arch|
```
ekohl, line 8: Isn't it better to swap the order? So do the each and then `it`, making them present as separate examples.

joshcooper:
@mhashizume could you rebase to pick up the Thor fix?

joshcooper (approved):

mhashizume:
This had further reaching implications than we'd realized and we will be reverting this change before it's part of a Facter release. While official Puppet support for Debian on ARM is forthcoming, Ubuntu on ARM has been supported since 2021 and this change has the potential to break a lot of integration that's already been built for those platforms.

I would still encourage users looking for an architecture scheme that's consistent between OSes to use `os.hardware` since `os.architecture` has historically been different for Debian-based OSes and will report `i386` and `amd64` (as opposed to `x86` and `x86_64`) even after this change is reverted.

PR #2626, opened 2023-10-19
(maint) Update user interaction diagram

mhashizume:
Previously, a conceptual diagram for Facter was stored as an image file and embedded in the README.

This commit updates that diagram to a Mermaid-based Markdown format written directly in the README and deletes the docs directory and its only contents, the image file of the diagram.


joshcooper:
@mhashizume could you rebase to pick up the Thor change?

joshcooper (approved):

PR #2632, opened 2023-10-26
(FACT-3446) Call close(2) with Solaris FFI

mhashizume:
Previously, when Facter ran on Solaris it would use shutdown(2) to attempt to close open sockets. However, this would leak file descriptors and occasionally cause an issue where the system would hit its maximum open files.

This commit updates Facter to instead use close(2) on Solaris to ensure that file descriptors are not leaked.


joshcooper (approved):

joshcooper:
For posterity, facter 3 used a `scoped_descriptor` to manage the socket's lifetime and it called `close` when the `scoped_descriptor` went out of scope: https://github.com/puppetlabs/facter/blob/6d7ffc6efdfbc3b1fc79311cdeb4581ac2098d9c/lib/src/util/posix/scoped_descriptor.cc#L16

PR #2635, opened 2023-11-08
Fix ifconfig parse logic failing on some tunnels

Woellchen:
Tunnel interfaces that have a line similar to
"tunnel inet 192.0.2.1 --> 192.0.2.2" would fail the network resolver
because IPs were extracted through two separate IP + mask patterns.
However, in the example above, there is no such thing as a netmask.

This fix merges the patterns and extracts the IPs and netmasks
together so that no mismatches can occur. It might also fix some
cases where only the netmask would be nil and therefore produce an
invalid binding that could in turn resolve to the wrong (outer) IP of
the tunnel instead of the inner IP.

### DIFF
--- a/spec/facter/resolvers/networking_spec.rb
+++ b/spec/facter/resolvers/networking_spec.rb
@@ -5,6 +5,10 @@ describe Facter::Resolvers::Networking do
 
   let(:log_spy) { instance_spy(Facter::Log) }
 
+  before do
+    pending 'contains failing tests to reproduce a bug'
+  end
+
   describe '#resolve' do
     before do
       networking.instance_variable_set(:@log, log_spy)
@@ -156,6 +160,7 @@ describe Facter::Resolvers::Networking do
 
       it 'returns dhcp server ip as nil' do
         expect(networking.resolve(:dhcp)).to be(nil)
+        raise
       end
     end
 
@@ -164,6 +169,7 @@ describe Facter::Resolvers::Networking do
 
       it 'returns interfaces as nil' do
         expect(networking.resolve(:interfaces)).to be(nil)
+        raise
       end
     end
   end
--- a/spec/facter/resolvers/networking_spec.rb
+++ b/spec/facter/resolvers/networking_spec.rb
@@ -5,10 +5,6 @@ describe Facter::Resolvers::Networking do
 
   let(:log_spy) { instance_spy(Facter::Log) }
 
-  before do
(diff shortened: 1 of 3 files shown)


joshcooper (commented):
Thanks @Woellchen! In the first commit that demonstrates the failure, can you mark the failing test(s) as `pending`? And then remove the `pending` in the second commit. This is important so that each commit passes if we later need to `git bisect`.

Or you can keep the commits as-is and I can squash them when merging. Let me know which sounds better.

Woellchen:
Hey @joshcooper, thanks for getting back to me!
I see your point and I have marked the tests as `pending` in the first commit, and unmarked them in the following one.
Please check one more time if this looks good to you now :)

spec/facter/resolvers/networking_spec.rb
```diff
@@ -5,6 +5,10 @@
 
   let(:log_spy) { instance_spy(Facter::Log) }
 
+  before do
+    pending 'contains failing tests to reproduce a bug'
```
joshcooper: This will cause all passing tests to fail, because they're "no longer" pending. See https://rspec.info/features/3-12/rspec-core/pending-and-skipped-examples/pending-examples/ 

You'll want to move the call to `pending` to the specific tests that aren't working yet (instead of callling `raise` in the test) like:

```ruby
 it 'returns dhcp server ip as nil' do
  pending('not working yet')
  expect(networking.resolve(:dhcp)).to be(nil)
end
```

joshcooper:
Closing and reopening to trigger the right set of checks

spec/facter/resolvers/networking_spec.rb
Woellchen: Hey @joshcooper! I have put the `pending` call on purpose in the `before` clause because all tests are actually failing with the updated `ifconfig` fixture to demonstrate that it really breaks Facter.
To not put the `pending` call in each and every test case I have moved it to `before`, but for that reason I also had to temporarily add the `raise`s in these two tests to make `rake spec` succeed.
In the following commit I revert the `pending` and `raise`s. Do you want me to move the `pending` call to every test case instead?

Woellchen:
Hey @joshcooper! To avoid further confusion and delays I have just merged the two commits into one and skipped the failing tests commit that was meant to demonstrate the failure. I have also rebased the branch once again. I hope that this way we can conclude this PR for good :)

PR #2642, opened 2023-11-18

buzzdeee:
    
    Most inspiration taken from FreeBSD, rest from other OS,
    and some own creations. Not all core facts supported, but
    makes puppet facts works well, as well as puppet agent --test
    
    also disable RuboCop Metrics/PerceivedComplexity
    and Metrics/CyclomaticComplexity
    as suggested by @joshcooper
    
    Additionally:
    Case insensitive OS detection
    
    ran into trouble when facter installed via puppetserver gem install,
    there the OS returned seems to be OpenBSD instead of all lowercase.


buzzdeee:
@joshcooper 

this is the follow-up of #2531 

I believe what was missing, and what confused the random order was that the OpenBSD related tests were missing in ./spec/facter/util/facts/posix/virtual_detector_spec.rb because in lib/facter/util/facts/posix/virtual_detector.rb some OpenBSD related detections were added. Also seems, the "before do" has to be kind of in-order with how the fact checks are called in virtual_detector.rb

Now everything aligned, I can't reproduce the issue anymore. I did let the tests run for some time in random order.

joshcooper:
Thanks @buzzdeee can you squash your commits?

buzzdeee:
@joshcooper sure, squashed and rebased, hope all test going to pass again (:

buzzdeee:
OpenBSD went into ABI lock for 7.5 release, if this would be merged, and make it into the next release hopefully getting out before OpenBSD ports tree freeze, would be awesome.

smortex (approved):
This look good and there seems to be decent test coverage for the new code so LGTM.

As somebody who did similar work for FreeBSD, I feel like this can be merged as-is. Maybe some corner cases are not correctly handled by this code, but it is a move in the right direction and will allow small and easy to review  pull-request to fix them if needed in the future.  OpenBSD not being officially supported by Puppet (just like FreeBSD), a "not-perfect-from-the-first-release" seems more acceptable than for officially supported operating systems.

Well done @buzzdeee ! :clap: :tada:

AriaXLi (approved):

PR #2643, opened 2023-11-21
(FACT-3448) Add Amazon to base fact utility

mhashizume:
This commit adds Amazon as a condition to the os_processors_and_kernel_expected_facts method.


mhashizume:
I haven't not been able to test this yet as there isn't a passing puppet-agent build that includes Amazon Linux.

Marking this as a draft until there is a passing build I can test against.

cthorn42 (approved):

Issue #2648, opened 2023-11-27
Cloud fact does not resolve on Amazon Linux 2023


mhashizume:
## Describe the Bug
`$ facter cloud` on Amazon Linux 2023 does not return anything:

```
$ facter --debug --trace cloud
[2023-11-27 18:59:57.339847 ] INFO Facter - executed with command line: --debug --trace cloud
[2023-11-27 18:59:57.339934 ] DEBUG Facter - Facter version: 4.5.1
[2023-11-27 18:59:57.345168 ] DEBUG Facter::FactManager - Resolving facts sequentially
[2023-11-27 18:59:57.345351 ] DEBUG Facter::FactLoader - Loading all internal facts
[2023-11-27 18:59:57.345468 ] DEBUG Facter::FactLoader - Loading custom facts
[2023-11-27 18:59:57.346790 ] DEBUG Facter::FactLoader - Loading external facts
[2023-11-27 18:59:57.348652 ] DEBUG Facter::QueryParser - List of resolvable facts: [#<Facter::SearchedFact:0x0000ffffab44e700 @name="cloud.provider", @fact_class=Facts::Linux::Cloud::Provider, @user_query="cloud", @type=:core, @file=nil>]
[2023-11-27 18:59:57.349621 ] DEBUG Facter::Core::Execution::Posix - Executing command: virt-what
[2023-11-27 18:59:57.352435 ] DEBUG Facter::Core::Execution::Posix - Failed while executing 'virt-what': No such file or directory - virt-what
[2023-11-27 18:59:57.352876 ] DEBUG Facter::Core::Execution::Posix - Executing command: vmware -v
[2023-11-27 18:59:57.355203 ] DEBUG Facter::Core::Execution::Posix - Failed while executing 'vmware -v': No such file or directory - vmware
[2023-11-27 18:59:57.355405 ] DEBUG Facter::Core::Execution::Posix - Executing command: virt-what
[2023-11-27 18:59:57.357737 ] DEBUG Facter::Core::Execution::Posix - Failed while executing 'virt-what': No such file or directory - virt-what
[2023-11-27 18:59:57.358147 ] DEBUG Facter::Util::FileHelper - File at: /proc/xen/capabilities is not accessible.
[2023-11-27 18:59:57.358288 ] DEBUG Facter::Core::Execution::Posix - Executing command: ["/usr/sbin/xl", "/usr/sbin/xm"] list
[...]

PR #2650, opened 2023-12-04
(FACT-3452) Make Xen resolver more strict

mhashizume:
Prior to this commit, Facter could misidentify non-Xen systems as being Xen-based.

In some circumstances, a non-Xen system will have /dev/xvd* files present as symlinks to other devices. This was observed on an Amazon Linux 2023 machine, which was running on a Nitro (KVM-based) hypervisor but symlinked /dev/xvda1 to an nvme device.

This commit updates the Xen resolver to check if /dev/xvda1 is a symlink to avoid misidentifying if a system is Xen-based.

### DIFF
--- a/lib/facter/resolvers/xen.rb
+++ b/lib/facter/resolvers/xen.rb
@@ -26,7 +26,9 @@ module Facter
 
         def detect_xen_type
           xen_type = 'xen0' if File.exist?('/dev/xen/evtchn')
-          xen_type = 'xenu' if !xen_type && (File.exist?('/proc/xen') || File.exist?('/dev/xvda1'))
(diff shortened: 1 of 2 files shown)


mhashizume:
On an Amazon Linux 2023 machine running on a t4g.medium instance on Nitro, this is the current behavior:

```
$ git describe
4.2.5-320-g421f9731c
$ bundle exec facter cloud

$ bundle exec facter virtual
xenu
```

Note that Facter doesn't resolve the cloud fact and misidentifies the machine as Xen.

With this PR:

```
$ git describe
4.2.5-375-ga11f4f8a4
$ bundle exec facter cloud
{
  provider => "aws"
}
$ bundle exec facter virtual
kvm
```

This should resolve https://github.com/puppetlabs/facter/issues/2648

joshcooper (approved):

PR #2652, opened 2023-12-05
(FACT-3110) support for ecdsa 384 and 521 bit keys

h0tw1r3:


joshcooper:
Thanks for your contribution @h0tw1r3 I've added your commit to a new PR and added some tests, so I'm going to close this out.

PR #2653, opened 2023-12-07
(FACT-3451) extend amazon linux os release fact

### LINKED ISSUE #2647: Provide more os.distro.release information for Amazon Linux
## Use Case
[Starting with Amazon Linux 2023](https://docs.aws.amazon.com/linux/al2023/ug/naming-and-versioning.html), Amazon provides a minor release every three months and provides a timestamp within specific minor versions.

Currently, Facter only provides the major version of the OS for Amazon Linux:

```
$ facter os.distro
{
  codename => "Amazon Linux",
  description => "Amazon Linux release 2023 (Amazon Linux)",
  id => "Amazon",
  release => {
    full => "2023",
    major => "2023"
  }
}
```

## Describe the Solution You Would Like
Facter should return as much information about the OS version as it can, including the minor release version and, within the `os.distro.release.full` fact, the timestamp (e.g. `2023.2.20230901`). Something like this seems ideal:

```
$ facter os.distro
{
  codename => "Amazon Linux",
  description => "Amazon Linux release 2023 (Amazon Linux)",
  id => "Amazon",
  release => {
    full => "2023.1.20230912",
    major => "2023"
    minor => "1"
  }
}
```

It seems like Amazon does not provide this OS version information in the typical places that most vendors use, like `/etc/os-release`. It does look like querying the `system-release` package provides the full OS version information we'd need:

```
$ rpm -qa system-release
system-release-2023.1.20230912-0.amzn2023.noarch
```

h0tw1r3:
Referencing a package version seems to be the only way to do this on Amazon Linux 2023.
Fixes #2647

Semi-related, there's no codename in AL 2023, should 'n/a' really be a default value?

### DIFF
--- a/lib/facter/framework/core/file_loader.rb
+++ b/lib/facter/framework/core/file_loader.rb
@@ -244,6 +244,8 @@ os_hierarchy.each do |os|
     require_relative '../../facts/amzn/os/distro/release.rb'
     require_relative '../../facts/amzn/os/release.rb'
 
+    require_relative '../../resolvers/amzn/os_release_rpm'
+
   when 'bsd'
     require_relative '../../facts/bsd/kernelmajversion.rb'
     require_relative '../../facts/bsd/kernelversion.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -119,6 +119,11 @@ Style/ClassVars:
   Exclude:
     - !ruby/regexp /(?:(?!.+_resolver.rb).)*$/
 
+Style/FormatStringToken:
+  Exclude:
+    - 'lib/facter/resolvers/amzn/os_release_rpm.rb'
+    - 'spec/facter/resolvers/amzn/os_release_rpm_spec.rb'
+
 Style/FrozenStringLiteralComment:
   Exclude:
     - 'spec/custom_facts/util/normalization_spec.rb'
--- a/lib/facter/facts/amzn/os/distro/release.rb
+++ b/lib/facter/facts/amzn/os/distro/release.rb
@@ -20,10 +20,10 @@ module Facts
           end
 
           def determine_release_version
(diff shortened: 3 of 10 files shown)


lib/facter/framework/core/file_loader.rb
```diff
@@ -74,6 +74,7 @@
 require_relative '../../resolvers/networking'
 require_relative '../../resolvers/open_vz'
 require_relative '../../resolvers/os_release'
+require_relative '../../resolvers/os_release_rpm'
```
mhashizume, line 77: I'm assuming that this new resolver will only be used for Amazon Linux for now, so we can probably just include it in the Amazon-specific stanza on line 238.

h0tw1r3: fixed

lib/facter/resolvers/os_release_rpm.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Facter
+  module Resolvers
+    class OsReleaseRpm < BaseResolver
+      init_resolver
+
+      QUERY_FORMAT = %()
```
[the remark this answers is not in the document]
h0tw1r3: nothing :) removed

mhashizume:
Closing and reopening to rekick checks

mhashizume (approved):
Thanks for this @h0tw1r3 !

joshcooper:
This changed the `os.release.{full,major}` facts on Amazon 2 (2017), which breaks the `systemd` service provider (as it's no longer the default):

```ruby
 defaultfor 'os.name' => :amazon, 'os.release.major' => ["2", "2023"]
```

I'll put up a PR to restore the old fact behavior for Amazon 1 and 2

PR #2656, opened 2023-12-13
(FACT-3110) Support for ecdsa 384 and 521 bit keys

joshcooper:
* Adds support for ecdsa 384 and 521 bit keys
* Remove overstubbing in the ssh resolvers
* Convert Ssh and FingerPrint to value objects using Struct

Supersedes #2652


cthorn42:
@joshcooper do you know if we'd get any coverage or would see any benefit of build puppet-agent with this and putting that through adhoc? Would that at least validate previous keys or whatever we default still work?

cthorn42:
Talked with Josh about testing this offline, he did this before hand based upon Jeff's work. He was able to validate that Facter functioned with the new RSA and ECDSA keys.

cthorn42 (approved):

PR #2661, opened 2024-01-02
(FACT-3451) Restore os facts for AmazonLinux 2

joshcooper:
Commit abba8950c changed the `os.release` and `os.distro.release` facts on Amazon Linux to query the os-release rpm, effectively running this command:

    # rpm -q --qf '%{VERSION}\n' -f /etc/os-release
    2023.1.20230912

This was appropriate for AL2023, but it introduced a regression on AL2 because the `os.release.major` and `os.distro.release.major` facts changed from "2" to a year-based version like:

    # rpm -q --qf '%{VERSION}\n' -f /etc/os-release
    2017.12

Since the `os.release.major` fact changed from 2 to 2017, the systemd service provider was no longer the default on AL2[1]

AL1 was seemingly unaffected because we were already using a year-based version string and both `VERSION` and `VERSION_ID` were the same:

    # rpm -q --qf '%{VERSION}\n' -f /etc/os-release
    2017.03
    # grep VERSION_ID /etc/os-release
    VERSION_ID="2017.03"

This commit restores the behavior for AL1 and AL2 to use `/etc/system-release`. On AL2, the `major` version will be reported as `2` instead of `20xx`.

This commit preserves the AL2023 behavior so it uses the os-release rpm to report the `major` version `2023`.

[1] https://github.com/puppetlabs/puppet/blob/8.3.1/lib/puppet/provider/service/systemd.rb#L22

### DIFF
--- a/lib/facter/facts/amzn/os/distro/release.rb
+++ b/lib/facter/facts/amzn/os/distro/release.rb
@@ -20,7 +20,12 @@ module Facts
           end
 
           def determine_release_version
-            version = Facter::Resolvers::Amzn::OsReleaseRpm.resolve(:version)
+            # For backwards compatibility, use system-release for AL1/AL2
+            version = Facter::Resolvers::ReleaseFromFirstLine.resolve(:release, release_file: '/etc/system-release')
+            if !version.nil? && version != '2'
+              # Use os-release for AL2023 and up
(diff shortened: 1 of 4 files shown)


amitkarsale (approved):
LGTM

PR #2674, opened 2024-01-23
Drop Ruby 2.5 support

mhashizume:
This PR:

- Raises Facter's minimum Ruby version from 2.5 to 2.6 in the gemspec.
- Updates unit tests to use Ruby 2.6 as their minimum version.
- Update Rubocop and related gems to the last version to support Ruby 2.6 and sets Ruby 2.6 as the TargetRubyVersion in rubocop.yml.
- Addresses new Rubocop offenses from the above updates.

PR #2686, opened 2024-03-06
Re-add Ruby 2.5 support

mhashizume:
We dropped support for Ruby 2.5 in Facter in https://github.com/puppetlabs/facter/pull/2674.

However, this broke compatibility with Puppetserver 7 in rare circumstances in which Facter 4.6.0/4.6.1 was being used in conjunction with Puppetserver < 7.9.0. Up until 7.9.0, Puppetserver used JRuby 9.2.y.z, which is compatible with MRI Ruby 2.5.

This PR reverts the changes made in https://github.com/puppetlabs/facter/pull/2674 and limits Thor to < 1.3.0, as newer versions drop support for Ruby 2.5.

### DIFF
--- a/.github/workflows/unit_tests.yaml
+++ b/.github/workflows/unit_tests.yaml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '2.6'
+          - '2.5'
           - '2.7'
           - '3.0'
           - '3.2'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,7 +2,7 @@
 inherit_from: .rubocop_todo.yml
 
 AllCops:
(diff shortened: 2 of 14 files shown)


joshcooper (approved):
Since puppet#main requires ruby >= 3.1 https://github.com/puppetlabs/puppet/blob/62b1cd63e096c01883cc1a9e1a05b03d4497598d/.gemspec#L24, once puppet#7.x goes EOL, I'm thinking a future version of facter should require >= 3.1 too. Maybe Facter 5? Could you file an issue so we don't forget?

ekohl:
Is this really correct? I'd expect bundler to figure out that it can't use thor 1.3 if they have the correct metadata. Or is Ruby 2.5's bundler broken?

Issue #2693, opened 2024-03-23
Wrong networking.fqdn in the presence of `search .` in `/etc/resolv.conf`


Geod24:
## Describe the Bug

```
$ /opt/puppetlabs/bin/facter networking
{
  dhcp => "xxx.xxx.xxx.xxx",
  domain => ".",
  fqdn => "xxx-xxx-prd-vm03..",
  hostname => "xxx-xxx-prd-vm03",
  [...]
```

This happens because:
```
$ ls -l /etc/resolv.conf
lrwxrwxrwx 1 root root 37 Mar 23 10:47 /etc/resolv.conf -> /run/systemd/resolve/stub-resolv.conf
$ cat /run/systemd/resolve/stub-resolv.conf
# This is /run/systemd/resolve/stub-resolv.conf managed by man:systemd-resolved(8).
# Do not edit.
#
# This file might be symlinked as /etc/resolv.conf. If you're looking at
# /etc/resolv.conf and seeing this text, you have followed the symlink.
#
# This is a dynamic resolv.conf file for connecting local clients to the
# internal DNS stub resolver of systemd-resolved. This file lists all
# configured search domains.
#
# Run "resolvectl status" to see details about the uplink DNS servers
# currently in use.
#
# Third party programs should typically not access this file directly, but only
# through the symlink at /etc/resolv.conf. To manage man:resolv.conf(5) in a
# different way, replace this symlink by a static file or a different symlink.
#
# See man:systemd-resolved.service(8) for details about the supported modes of
# operation for /etc/resolv.conf.

nameserver 127.0.0.53
options edns0 trust-ad
search .
```

Commenting out the `search` entry fixes the issue. This happens because this machine is configured without a domain, but with a search path, leading to this being executed:
https://github.com/puppetlabs/facter/blame/7bfd75b1c76bb5d531c35131b717fb0697a54426/lib/facter/resolvers/linux/hostname.rb#L101-L103

And then a wrong concatenation:
https://github.com/puppetlabs/facter/blame/7bfd75b1c76bb5d531c35131b717fb0697a54426/lib/facter/resolvers/linux/hostname.rb#L110

## Expected Behavior

facter should not read `search` in the first place. `search` is a client lookup option, and it is not intended to set the machine's hostname.

## Steps to Reproduce

See `resolv.conf` configuration above. Make sure the machine has no domain configured.

## Environment

```
# /opt/puppetlabs/bin/facter --version
4.6.1
 # uname -a
[...]

PR #2694, opened 2024-03-24
networking: Workaround 2693 by ignoring resolv.conf entries starting with dot

Geod24:
Currently, systems without a FQDN can be mishandled by facter for certain `/etc/resolv.conf` content. This was initially noticed when systemd-resolved was installed on a host without domain.

systemd-resolved stub resolver sets 'search .' as a search domain, which results in the following hostname/domain/fqdn triplet: `foo`, `.`, `foo..`. See: https://github.com/systemd/systemd/blob/v255/src/resolve/resolv.conf#L19

This is wrong on multiple levels: first, facter does not seem to handle '.' (the root level) well, and there have been PRs to remove the trailing dot in FQDN as far back as 2012 (PR 200), leaving user with a convenient, but sometimes ambiguous string that is not fully qualified.

More fundamentally, facter sould not be using 'search' or 'domain' in resolv.conf to begin with, as those are client resolution options, and usually set for convenience, but are not part of the hostname. However, removing this lookup is much more involved, as facter has been doing this for years.

Hence, this commit implements a a middle ground solution to support top-level/domainless servers without making a breaking change.

### DIFF
--- a/lib/facter/resolvers/linux/hostname.rb
+++ b/lib/facter/resolvers/linux/hostname.rb
@@ -96,9 +96,9 @@ module Facter
 
           def read_domain
             file_content = Facter::Util::FileHelper.safe_read('/etc/resolv.conf')
-            if file_content =~ /^domain\s+(\S+)/
+            if file_content =~ /^domain\s+([^.]\S+)/
               Regexp.last_match(1)
-            elsif file_content =~ /^search\s+(\S+)/
+            elsif file_content =~ /^search\s+([^.]\S+)/
               Regexp.last_match(1)
             end
           end
--- a/lib/facter/resolvers/hostname.rb
+++ b/lib/facter/resolvers/hostname.rb
@@ -68,9 +68,9 @@ module Facter
 
         def read_domain
           file_content = Facter::Util::FileHelper.safe_read('/etc/resolv.conf')
-          if file_content =~ /^domain\s+(\S+)/
+          if file_content =~ /^domain\s+([^.]\S+)/
             Regexp.last_match(1)
-          elsif file_content =~ /^search\s+(\S+)/
+          elsif file_content =~ /^search\s+([^.]\S+)/
             Regexp.last_match(1)
           end
         end
--- a/spec/facter/resolvers/linux/hostname_spec.rb
+++ b/spec/facter/resolvers/linux/hostname_spec.rb
@@ -79,6 +79,30 @@ describe Facter::Resolvers::Linux::Hostname do
 
             it_behaves_like 'detects values'
           end
+
+          context 'when /etc/resolv.conf has "search ."' do
+            let(:resolv_conf) { "search .\n" }
+            let(:domain) { nil }
+            let(:fqdn) { hostname }
+
+            it_behaves_like 'detects values'
+          end
+
+          context 'when /etc/resolv.conf has "search ." with multiple entires' do
+            let(:resolv_conf) { 'search . foo.bar' }
+            let(:domain) { nil }
+            let(:fqdn) { hostname }
+
+            it_behaves_like 'detects values'
+          end
+
+          context 'when /etc/resolv.conf has "search" with multiple entires' do
+            let(:resolv_conf) { 'search foo.bar example.com' }
+            let(:domain) { 'foo.bar' }
+            let(:fqdn) { "#{hostname}.#{domain}" }
+
+            it_behaves_like 'detects values'
+          end
         end
 
         context 'when FFI is not installed' do


Geod24:
@cthorn42 any chance you could take a look at this ?

ekohl:
> More fundamentally, facter sould not be using 'search' or 'domain' in resolv.conf to begin with, as those are client resolution options, and usually set for convenience, but are not part of the hostname. However, removing this lookup is much more involved, as facter has been doing this for years.

I don't think this is true.

Quoting `man 1 hostname`:
> The  FQDN  (Fully  Qualified Domain Name) of the system is the name that the resolver(3) returns for the host name, such as, ursula.example.com.  It is usually the hostname followed by the DNS domain name (the part after the first dot).  You can check the FQDN using hostname --fqdn or the domain name using dnsdomainname.

Quoting `man 3 resolver`:
> The res_ninit() and res_init() functions read the configuration files (see resolv.conf(5)) to get the default domain name and name server address(es).  If no server is given, the local host is tried.  If no domain is given, that associated with the local host is used.  It can be overridden with the environment variable LOCALDOMAIN.  res_ninit() or res_init() is normally executed by the first call to one of the other functions.  Every call to res_ninit() requires a corresponding call to res_nclose() to free memory allocated by res_ninit() and subsequent calls to res_nquery().

So on Linux libc can certainly use it.

> However, removing this lookup is much more involved, as facter has been doing this for years.

You could replace it with a native implementation. Quoting `man 1 hostname` again:
> Technically:  The  FQDN  is the name getaddrinfo(3) returns for the host name returned by gethostname(2).  The
DNS domain name is the part after the first dot.

I think that means you can use:
```ruby
Socket.getaddrinfo(Socket.gethostname, nil, nil, nil, nil, Socket::AI_CANONNAME)
```

h0tw1r3:
I think most sys admins would expect puppet's fqdn / hostname to return exactly what the `hostname` command does.

This is a quick port of the standard linux [hostname](https://sources.debian.org/src/hostname/3.23%2Bnmu2/hostname.c/) C code with the addition of falling back to parsing /etc/resolve.conf.

```ruby
#!/opt/puppetlabs/puppet/bin/ruby

require 'socket'

def fqdn
	# Addrinfo.getaddrinfo
	result = Addrinfo.getaddrinfo(Socket.gethostname, nil, nil, Socket::SOCK_DGRAM, nil, Socket::AI_CANONNAME)
		.map { |a| a.canonname }
		.find { |n| !n.nil? and n.include? '.' }

	return result if !result.nil?

	# Socket.getnameinfo
	result = Socket.getifaddrs().filter_map do |ifap|
		next unless ifap.addr
		next unless (ifap.flags & Socket::IFF_LOOPBACK) == 0
		next if (ifap.flags & Socket::IFF_UP) == 0

		family = ifap.addr.afamily
		next unless (family == Socket::AF_INET || family == Socket::AF_INET6)

		next if family == Socket::AF_INET6 && ifap.addr.ipv6_linklocal?

		n = Socket.getnameinfo(ifap.addr, Socket::NI_NAMEREQD)
		n[0] if n and !n[0].nil? and n[0].include? '.'
	end.first

	return result if !result.nil?

	# Fallback to parsing /etc/resolve.conf
	content = File.read('/etc/resolv.conf', chomp: true)
	domain = if content =~ /^domain\s+(\S+)/
			 Regexp.last_match[1]
		 elsif content =~ /^search\s+(\S+)/
			 Regexp.last_match[1]
		 end
	# strip leading and trailing .
	domain.sub!(/^\.?(.*?)\.?$/, '\1') if domain

	result = [Socket.gethostname]
	result.append(domain) if domain and domain.length > 1

	result.join('.')
end
```

b4ldr:
> > More fundamentally, facter sould not be using 'search' or 'domain' in resolv.conf to begin with, as those are client resolution options, and usually set for convenience, but are not part of the hostname. However, removing this lookup is much more involved, as facter has been doing this for years.
> 
> I don't think this is true.
I think this is definitely true for the search path but i agree using the domain part seems reasonable if there is no better info

lib/facter/resolvers/linux/hostname.rb
```diff
@@ -96,9 +96,9 @@ def hostname_and_no_domain?(hostname, domain)
 
           def read_domain
             file_content = Facter::Util::FileHelper.safe_read('/etc/resolv.conf')
-            if file_content =~ /^domain\s+(\S+)/
+            if file_content =~ /^domain\s+([^.]\S+)/
               Regexp.last_match(1)
-            elsif file_content =~ /^search\s+(\S+)/
+            elsif file_content =~ /^search\s+([^.]\S+)/
```
b4ldr, line 101: AFAICT The following would have the affect of stripping all trailing dot's
```suggestion
            elsif file_content =~ /^search\s+(\S+[^.])/
```

joshcooper: @b4ldr I didn't follow this comment. Are you suggesting we strip trailing dots?

joshcooper:
Thanks for your PR @Geod24! 

I think the "facter has been doing this wrong all this time" issue is valid, but  a different issue that would need to be addressed in Facter 5. Thanks @ekohl and @h0tw1r3 for the sample code.

Your PR makes sense to me. To get this merged could you:

1. Remove the "More fundamentally, ..." comment from your commit message based on the discussion in this PR?

2. Facter appears to be missing test coverage for the case where `search` contains multiple entries. From `man 5 resolver`:

```
       search Search list for host-name lookup.
              By  default,  the search list contains one entry, the local domain name. 
```

Note this isn't true for `domain`

```
              The domain directive is an obsolete name for the search directive that handles one search list entry only.
```

Could you add a test case for that?

3. Facter also assumes there's only one search directive (and reads the first one that matches). However, there may be multiple directives:

```
              If there are multiple search directives, only the search list from the last instance is used.
```

I'd be fine filing a separate issue for this.

Geod24:
> Facter appears to be missing test coverage for the case where search contains multiple entries. From man 5 resolver:

What is the desired outcome here ? 
If `resolv.conf` has `search foo.bar example.com`, do we assume the first entry or the last entry is the domain, or do we error out ?
Does this selection mechanism gets affected by what the entry is ? For example, if we pick the first entry, would we not do so if the first entry is `.` ?

In the meantime, I have rebased the commit and amended the message to remove the aforementioned section.

lib/facter/resolvers/linux/hostname.rb
b4ldr: @joshcooper 

> Are you suggesting we strip trailing dots as was done in #200
yes

> in addition to stripping the leading dot?
Im not sure if we need to do that, i think that would be a change in behaviour?  

> Or that stripping the leading and trailing dot are equivalent when we're talking about root domains
also yes :)

Geod24: Do we want to strip all trailing dots ? Since empty labels are an error, I would be wary of making changes that make their usage valid in Facter's context.

AriaXLi:
Hi @Geod24, thank you for your patience!

Regarding your questions:
```
If resolv.conf has search foo.bar example.com, do we assume the first entry or the last entry is the domain, or do we error out ?
```
We should assume the resolver will pick the first entry and should not error out

```
Does this selection mechanism gets affected by what the entry is ? For example, if we pick the first entry, would we not do so if the first entry is . ?
```
If the first entry is `.` followed by something else, the resolver should pick `.`

Let me know if anything is unclear or if you have any other questions, thank you!

AriaXLi:
This PR addresses issue https://github.com/puppetlabs/facter/issues/2693.

lib/facter/resolvers/linux/hostname.rb
b4ldr: Sorry for the delay in responding.

AriaXLi: Hi @b4ldr this makes sense to me and I was also able to verify a domain with a leading dot is valid. I'll incorporate your suggestions now. Thanks for the response and suggestion!

AriaXLi: Hi @b4ldr thanks for your suggestion, I was also able to verify a domain with a leading dot is valid, like `.example.org` but with your suggested changes, it causes `Regexp.last_match(1)` to be something unexpected.

For example, if resolv.conf had `search foo.bar example.com` with `foo.bar` as the domain, then `Regexp.last_match(1)` should be `"foo.bar"` but with your suggestions, it returns `"foo.bar "`

b4ldr: ahh yes of course this is getting more and more complicated, how about the following?
```ruby
elsif file_content =~ /^search\s+\.?(\S+[^.\s])/
```

b4ldr: the other options is to just use `Regexp.last_match(1).strip` that might be a bit more readable

AriaXLi (approved):

PR #2700, opened 2024-04-03
Ignore proxy for cloud metadata retrieval

mhashizume:
Prior to this commit when a user set the http_proxy environment variable, the Net::HTTP#new in the HTTP resolver would use it when establishing new connections. While this is the desired behavior in most cases, this also interfered with fetching cloud metadata for services like AWS and GCP.

Facter had previously dealt with this behavior in earlier versions (2.x) starting with 23bf7e5, but that was lost during Facter's transition from C++ back to Ruby.

This commit sets Net::HTTP#new's p_addr (proxy address) argument to nil when Facter detects that the http_proxy environment variable is present when attempting to resolve cloud service metadata.

### DIFF
--- a/lib/facter/util/resolvers/http.rb
+++ b/lib/facter/util/resolvers/http.rb
@@ -9,36 +9,45 @@ module Facter
         class << self
           CONNECTION_TIMEOUT = 0.6
           SESSION_TIMEOUT = 5
+          DEFAULT_HOST = Object.new.freeze
 
           # Makes a GET HTTP request and returns its response.
           #
+          # @note This method uses an empty DEFAULT_HOST constant to provide a
+          #   way to override the http_proxy environment variable without
+          #   interfering with other behaviors (such as connecting to a
+          #   loopback address or using the no_proxy environment variable).
           # @param url [String] the address to which the request will be made.
           # @param headers [Hash] the headers you need to add to your request.
           #   Defaults to an empty hash.
           # @param timesouts [Hash] Values for the session and connection
           #   timeouts.
+          # @param http_port [Integer] The port number used to make the
+          #   request. Defaults to 80.
+          # @param proxy_addr [String] The address of a proxy host.
           # @returns [String] the response body if the response code is 200.
           #   If the response code is not 200, an empty string is returned.
           # @example
           #   get_request('https://example.com', { "Accept": 'application/json' }, { session: 2.4, connection: 5 })
-          def get_request(url, headers = {}, timeouts = {})
-            make_request(url, headers, timeouts, 'GET')
+          def get_request(url, headers = {}, timeouts = {}, http_port: 80, proxy_addr: DEFAULT_HOST)
+            make_request(url, headers, timeouts, 'GET', http_port, proxy_addr)
           end
 
           # Makes a PUT HTTP request and returns its response
           # @param (see #get_request)
           # @return (see #get_request)
-          def put_request(url, headers = {}, timeouts = {})
-            make_request(url, headers, timeouts, 'PUT')
+          def put_request(url, headers = {}, timeouts = {}, http_port: 80, proxy_addr: DEFAULT_HOST)
+            make_request(url, headers, timeouts, 'PUT', http_port, proxy_addr)
           end
 
           private
 
-          def make_request(url, headers, timeouts, request_type)
+          # rubocop:disable Metrics/ParameterLists
+          def make_request(url, headers, timeouts, request_type, http_port, proxy_addr)
             require 'net/http'
 
             uri = URI.parse(url)
-            http = http_obj(uri, timeouts)
+            http = http_obj(uri, timeouts, http_port, proxy_addr)
             request = request_obj(headers, uri, request_type)
 
             # The Windows implementation of sockets does not respect net/http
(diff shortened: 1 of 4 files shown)


lib/facter/util/resolvers/http.rb
```diff
@@ -57,7 +57,19 @@ def make_request(url, headers, timeouts, request_type)
...
           def http_obj(parsed_url, timeouts)
-            http = Net::HTTP.new(parsed_url.host)
+            # If Facter is querying a cloud metadata service and a proxy has
+            # been configured by the http_proxy environment variable, set
+            # Net::HTTP#new's p_addr argument to nil to allow the metadata
+            # service to properly resolve.
+            # Net::HTTP#new takes positional arguments, so we have to
+            # explicitly set all arguments up to and including p_addr.
+            http =
+              if ENV['http_proxy'] && parsed_url.to_s.match?(Facter::Resolvers::Ec2::EC2_METADATA_ROOT_URL || Facter::Resolvers::Ec2::EC2_USERDATA_ROOT_URL || Facter::Resolvers::Gce::METADATA_URL)
```
joshcooper, line 67: I think it'd be cleaner to expose `http_host` and `http_port` as keyword arguments to the `get_request` and `put_request` methods and have the various cloud resolvers pass in `nil` values for those (always).

lib/facter/util/resolvers/http.rb
```diff
@@ -10,34 +10,39 @@ class << self
...
+          # @param timesouts [Hash] Values for the session and connection
+          #   timeouts.
+          # @param http_port [Integer] The port number used to make the
+          #   request. Defaults to 80.
+          # @param proxy_addr [String] The address of a proxy host.
+          # @returns [String] the response body if the response code is 200.
+          #   If the response code is not 200, an empty string is returned.
+          # @example
+          #   get_request(https://example.com, { "Accept": 'application/json' }, { session: 2.4. connection: 5 })
+          def get_request(url, headers = {}, timeouts = {}, http_port: 80, proxy_addr: ENV['http_proxy'])
```
joshcooper, line 27: Previously, if the `http_proxy'` environment variable was set, then the call to `Net::HTTP.new(parsed_url.host)` might or might not use the proxy, depending on whether we're connecting to the loopback:

https://github.com/ruby/ruby/blob/1f4d4558484b370999954f3ede7e3aa3a3a01ef3/lib/uri/generic.rb#L1533

or the target host matches the `no_proxy` environment variable:

https://github.com/ruby/ruby/blob/1f4d4558484b370999954f3ede7e3aa3a3a01ef3/lib/uri/generic.rb#L1541

If we always pass `ENV['http_proxy']` through to the `Net::HTTP.new` method, then I think(?) it will always use that explicit proxy.

Maybe define a `DEFAULT_HOST` object and default to the proxy_addr option to that:

```
DEFAULT_HOST = Object.new.freeze
def get_request(url, headers = {}, timeouts = {}, http_port: 80, proxy_addr: DEFAULT_HOST)
```

Then check if the option has been overridden (to possibly `nil`) like:

```ruby
if proxy_addr == DEFAULT_HOST
  Net::HTTP.new(host, port)
else
  Net::HTTP.new(host, port, proxy_addr, proxy_port)
```

Then have the cloud resolvers explicitly pass `proxy_addr: nil`

mhashizume:
I think this is introducing some unwanted behavior related to positional and keyword arguments.

Some of the methods being called here use a hash as what's intended to be a positional argument but, with the addition of keyword arguments in this PR, older Rubies seem to now misinterpret the keys in that hash as individual keyword arguments.

For example in the ec2 resolver, the headers and timeouts arguments now seem to get wrongly interpreted as keyword arguments:
https://github.com/puppetlabs/facter/blob/dbfb450c0ad03d77ee351049ba0a569128afe248/lib/facter/resolvers/ec2.rb#L52-L56

I think we can resolve this by using strings instead of symbols in these positional hash arguments and being explicit about the hash syntax(`{ 'foo' => 'bar' }` instead of `{ foo: 'bar'}`), but that may have other implications. I'll do more testing and research on this.

mhashizume:
When attempting to add new keyword arguments to the get_request and put_request methods, I'm running into issues with different versions of Ruby and how they deal with the separation of positional and keyword arguments.

The last argument of these methods (timeouts) is an optional hash that uses symbols as keys. In Ruby < 3.0, Ruby interprets the keys in the timeouts hash as keyword arguments if we add any keyword arguments to the get_request/put_request methods.

For example, running this rspec test on Ruby 2.7.8:

```
  1) Facter::Util::Resolvers::Http when windows #get_request behaves like a http request on windows when timeout is configured uses the desired value
     Failure/Error:
       def get_request(url, headers = {}, timeouts = {}, http_port: 80, proxy_addr: DEFAULT_HOST)
         make_request(url, headers, timeouts, 'GET', http_port, proxy_addr)

     ArgumentError:
       unknown keyword: :connection
     Shared Example Group: "a http request on windows" called from ./spec/facter/util/resolvers/http_spec.rb:163
     # ./lib/facter/util/resolvers/http.rb:32:in `get_request'
     # ./spec/facter/util/resolvers/http_spec.rb:131:in `block (4 levels) in <top (required)>'
```

Ruby >= 3.0 properly separates positional and keyword arguments:

```
$ bundle exec rspec ./spec/facter/util/resolvers/http_spec.rb:131
Run options: include {:locations=>{"./spec/facter/util/resolvers/http_spec.rb"=>[131]}}
..

Finished in 0.19973 seconds (files took 8.66 seconds to load)
2 examples, 0 failures
```

But because we still support Facter on Ruby >= 2.5, we need to account for Ruby < 3.0.

We _could_ solve this issue by making sure that the keys in the timeouts hash are strings intead of symbols. However, other parts of the code access values in the hash by looking for symbol keys specifically (like in [here](https://github.com/puppetlabs/facter/blob/dbfb450c0ad03d77ee351049ba0a569128afe248/lib/facter/util/resolvers/http.rb#L61-L62)) and get_request/put_request are public methods. The fact that those methods are public mean that while I think additive changes like new keyword arguments are probably okay, changes or removals risk breaking things for any users who may rely on those methods.

I'm going to instead try to implement this with a boolean positional argument instead which, while isn't as clean, seems easier to implement across Ruby versions and retains compatibility with these public methods.

See also: https://www.fastruby.io/blog/fix-sneaky-argument-error-when-upgrading-ruby.html

mhashizume:
Closing in favor of https://github.com/puppetlabs/facter/pull/2704

PR #2704, opened 2024-04-11
Allow cloud metadata to ignore proxy settings 

mhashizume:
Prior to this commit, if the http_proxy environment variable was set for Net::HTTP, cloud resolvers (ec2, GCE, Azure) would fail to resolve metadata.

This commit adds a new positional argument to Facter::Util::Resolvers::Http.get_request and #put_request to allow those resolvers to disable proxy settings.

### DIFF
--- a/lib/facter/util/resolvers/http.rb
+++ b/lib/facter/util/resolvers/http.rb
@@ -10,23 +10,24 @@ module Facter
           CONNECTION_TIMEOUT = 0.6
           SESSION_TIMEOUT = 5
 
-          # Makes a GET http request and returns its response.
+          # Makes a GET HTTP request and returns its response.
           #
-          # Params:
-          # url: String which contains the address to which the request will be made
-          # headers: Hash which contains the headers you need to add to your request.
-          #          Default headers is an empty hash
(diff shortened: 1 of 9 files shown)


lib/facter/util/resolvers/http.rb
```diff
@@ -56,8 +59,16 @@ def make_request(url, headers, timeouts, request_type)
...
           end
 
-          def http_obj(parsed_url, timeouts)
-            http = Net::HTTP.new(parsed_url.host)
+          def http_obj(parsed_url, timeouts, proxy)
+            # If get_request or put_request are called and set proxy to false,
+            # manually set Net::HTTP.new's p_addr (proxy address) positional
+            # argument to nil to override anywhere else a proxy may be set
+            # (e.g. the http_proxy environment variable).
+            http = if proxy == false
```
joshcooper, line 67: nit checking for `== false` could be simplified to:

```ruby
if proxy
  Net::HTTP.new(parsed_url.host)
else
  Net::HTTP.new(parsed_url.host, 80, nil)
end
```

joshcooper (approved):

PR #2706, opened 2024-04-22
(FACT-3460) Add puppetlabs specific gem build task

tvpartytonight:


Rakefile
```diff
@@ -13,6 +14,18 @@ task :changelog, [:version] do |_t, args|
   sh "./scripts/generate_changelog.rb #{args[:version]}"
 end
 
+namespace :pl do
+  desc 'build the gem and place it at the directory root'
+  task :gem_build do
```
joshcooper, line 19: Maybe this was intentional(?) but the packaging repo already defines the `pl` namespace. Should we use a different namespace so as to not accidentally conflict with `pl::gem`?

tvpartytonight: I was just trying to not interfere with what might possibly be in `bundler/gem_tasks`; that defined a top level `:build` task, so I just defined it in PL to keep away from that. I guess we could just use an arbitrary new namespace to keep away from what the packaging gem defines too, but not sure what that should be called.

Rakefile
```diff
@@ -13,6 +14,18 @@ task :changelog, [:version] do |_t, args|
...
+    stdout, stderr, status = Open3.capture3('gem build facter.gemspec')
+    if !status.exitstatus.zero?
+      puts "Error building facter.gemspec \n#{stdout} \n#{stderr}"
```
tvpartytonight, line 22: Should we just `exit(1)` or should we depend on CI to handle this errors in gem building?

joshcooper: exit(1) or raise seems fine, I don't know if there's a preference for rake tasks?

AriaXLi (approved):

PR #2710, opened 2024-05-06
(PA-6392) Let 'unknown' be an acceptable processor.isa fact value for Fedora

shubhamshinde360:
- In os_processors_and_kernel test case, for fedora (intel) we expect processor.isa fact to be x86_64.
 - This value is derived from 'uname -p' command on the system.
 - But Fedora 40 returns 'unknown' for 'uname -p', so we make the test case accept 'unknown' as well.


mhashizume (approved):
LGTM. We already do this with other platforms, and IIRC `uname -p` is non-POSIX so it's not surprising that it's not always available.

PR #2711, opened 2024-05-06
(PA-6393) Let '' (empty string) be an acceptable networking dhcp fact for all fedora versions

shubhamshinde360:
- Fedora is susceptible to a bug in NetworkManager that causes it to fail networking_facts acceptance test.
 - This commit makes '' the acceptable value for networking dhcp fact in Fedora, so CI passes the test.
 - Since all Fedora versions seem to be having the issue, its been made generic to all versions in this commit.
 - See https://github.com/puppetlabs/facter/pull/2515

### DIFF
--- a/acceptance/tests/facts/networking_facts.rb
+++ b/acceptance/tests/facts/networking_facts.rb
@@ -11,7 +11,7 @@ test_name 'C59029: networking facts should be fully populated' do
 
   agents.each do |agent|
     expected_networking = {
-      %w[networking dhcp] => agent['platform'] =~ /fedora-32|fedora-34|fedora-36|el-8-|el-9-/ ? '' : @ip_regex, # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426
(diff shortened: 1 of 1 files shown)


acceptance/tests/facts/networking_facts.rb
```diff
@@ -11,7 +11,7 @@
 
   agents.each do |agent|
     expected_networking = {
-      %w[networking dhcp] => agent['platform'] =~ /fedora-32|fedora-34|fedora-36|el-8-|el-9-/ ? '' : @ip_regex, # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426
+      %w[networking dhcp] => agent['platform'] =~ /fedora-32|fedora-34|fedora-36|fedora-40|el-8-|el-9-/ ? '' : @ip_regex, # https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/issues/426
```
mhashizume, line 14: I think it would be simpler to replace all of the Fedora versions here with just `fedora-` since this is seemingly an issue with all supported versions of Fedora.

shubhamshinde360: Thanks @mhashizume! I have incorporated your suggestion. Please review again.

mhashizume (approved):

PR #2720, opened 2024-05-20
Add Puppet Server 8.x JRuby to unit tests

mhashizume:
This commit adds JRuby 9.4.7.0 to unit tests, which is the same version of JRuby that Puppet Server 8.x uses. (See:
puppetlabs/jruby-utils@1180711.)


joshcooper (approved):

joshcooper (commented):
unfortunately looks like valid test failures (maybe a regression in JRuby 9.4.7?

PR #2519, opened 2022-08-31
(FACT-3142) Add support for Mariner to Facter 4.x

mminkus:
Add support for CBL-Mariner, Microsoft's Linux distribution for
cloud infrastructure and edge products and services.

PR #2723, opened 2024-05-23
Add support for Microsoft Azure Linux to Facter 4.x

### LINKED ISSUE #2727: Add support for Microsoft Azure Linux to Facter 4.x
Add support for Microsoft Azure Linux, previously known as CBL-Mariner, Microsoft's Linux distribution for cloud infrastructure and edge products and services.

mminkus:
Add support for Microsoft Azure Linux, previously known as CBL-Mariner, Microsoft's Linux distribution for cloud infrastructure and edge products and services.

Before:
```
# facter os
{
  architecture => "x86_64",
  family => "Linux",
  hardware => "x86_64",
  name => "Linux",
  release => {
    full => "6.6.22.1-2.azl3",
    major => "6",
    minor => "6"
  },
  selinux => {
    enabled => false
  }
}
```

After:
```
# facter os
{
  architecture => "x86_64",
  family => "Redhat",
  hardware => "x86_64",
  name => "AzureLinux",
  release => {
    full => "3.0.20240401",
    major => "3",
    minor => "0"
  },
  selinux => {
    enabled => false
  }
}
```

Fixes #2727

### DIFF
--- a/lib/facter/config.rb
+++ b/lib/facter/config.rb
@@ -40,7 +40,8 @@ module Facter
             'Slackware',
             'Mageia',
             'Openwrt',
-            'Mariner'
+            'Mariner',
+            'Azurelinux'
           ]
         },
         {
--- a/lib/facter/facts/azurelinux/os/release.rb
+++ b/lib/facter/facts/azurelinux/os/release.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Facts
+  module Azurelinux
+    module Os
+      class Release
+        FACT_NAME = 'os.release'
+        ALIASES = %w[operatingsystemmajrelease operatingsystemrelease].freeze
+
+        def call_the_resolver
+          version = from_specific_file || from_os_release
+
+          return Facter::ResolvedFact.new(FACT_NAME, nil) unless version
+
+          [Facter::ResolvedFact.new(FACT_NAME, version),
+           Facter::ResolvedFact.new(ALIASES.first, version['major'], :legacy),
+           Facter::ResolvedFact.new(ALIASES.last, version['full'], :legacy)]
+        end
+
+        def from_specific_file
+          version = Facter::Resolvers::SpecificReleaseFile.resolve(:release,
+                                                                   { release_file: '/etc/azurelinux-release',
+                                                                     regex: /AZURELINUX_BUILD_NUMBER=([0-9.]+)/ })
+          Facter::Util::Facts.release_hash_from_matchdata(version)
+        end
+
+        def from_os_release
+          version = Facter::Resolvers::OsRelease.resolve(:version_id)
+
+          Facter::Util::Facts.release_hash_from_string(version)
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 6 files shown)


lib/facter/facts/azurelinux/os/release.rb
```diff
@@ -0,0 +1,35 @@
...
+        ALIASES = %w[operatingsystemmajrelease operatingsystemrelease].freeze
+
+        def call_the_resolver
+          version = from_specific_file || from_os_release
+
+          return Facter::ResolvedFact.new(FACT_NAME, nil) unless version
+
+          [Facter::ResolvedFact.new(FACT_NAME, version),
+           Facter::ResolvedFact.new(ALIASES.first, version['major'], :legacy),
+           Facter::ResolvedFact.new(ALIASES.last, version['full'], :legacy)]
```
AZaugg, line 17: Can you add support for minor?

lib/facter/facts/azurelinux/os/release.rb
```diff
@@ -0,0 +1,35 @@
...
+        end
+
+        def from_specific_file
+          version = Facter::Resolvers::SpecificReleaseFile.resolve(:release,
+                                                                   { release_file: '/etc/azurelinux-release',
```
AZaugg, line 22: Should we be using the /etc/os-release which is the standard for systemd based system?

mminkus: We could, but the base linux resolver does not parse the full version correctly so I parse `/etc/azurelinux-release` similarly to what I did with mariner parsing `/etc/mariner-release` in PR https://github.com/puppetlabs/facter/pull/2519

joshcooper: > the base linux resolver does not parse the full version correctly

Could you expand on that? What does it parse currently and what would you expect it to?

mminkus: Sorry for the late reply, was on vacation.
With just the base resolver, we get:
```
  release => {
    full => "3.0",
    major => "3.0"
  },
```
vs. the more desirable:
```
  release => {
    full => "3.0.20240401",
    major => "3",
    minor => "0"
  },
```

lib/facter/facts/azurelinux/os/release.rb
joshcooper: I think this is still an open question?

mminkus: We're actually getting minor in the facter output already, but I'm not sure how that's working.

lib/facter/config.rb
```diff
@@ -40,7 +40,8 @@ module Config
             'Slackware',
             'Mageia',
             'Openwrt',
-            'Mariner'
+            'Mariner',
+            'Azurelinux'
```
mhashizume, line 44: Nit, but shouldn't both Mariner and AzureLinux be nested under Rhel (line 20)?

mhashizume:
This looks good to me other than the hierarchy in `lib/facter/config.rb` and the minor conflict in `lib/facter/framework/core/file_loader.rb`. Today is feature complete for our next release of Puppet (during which we will also be including a new release of Facter), so I've created a new PR with your work and those minor changes: https://github.com/puppetlabs/facter/pull/2738

PR #2733, opened 2024-06-14
(FACT-3205) Ignore ENOENT when cleaning up temporary test files

joshcooper:
Facter tests sometimes fail on JRuby with the following, even when using
the same seed:

    $ bundle exec rspec --order random:22464 spec_integration
    ...
    Errno::ENOENT:
      No such file or directory - /tmp/external_facts20240617-2780089-h18bte/XFZQDAMC_data.yaml
    # ./spec_integration/facter_spec.rb:13:in `write_to_file'
    # ./spec_integration/facter_spec.rb:752:in `block in <main>'

Since ENOENT means the entry doesn't exist, ignore it like puppet does[1].
However, we don't want to use `rm_rf` because that will ignore all
StandardErrors, not just ENOENT[2].

[1] https://github.com/puppetlabs/puppet/blob/8.6.0/spec/lib/puppet_spec/files.rb#L11-L15
[2] https://bugs.ruby-lang.org/issues/18784

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,9 +53,6 @@ Naming/VariableNumber:
     - 'lib/facter/resolvers/windows/ffi/networking_ffi.rb'
     - 'lib/facter/util/facts/windows_release_finder.rb'
 
-RSpec/MultipleMemoizedHelpers:
-  Enabled: false
-
 RSpec/StubbedMock:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


.rubocop.yml
```diff
@@ -53,27 +53,6 @@ Naming/VariableNumber:
     - 'lib/facter/resolvers/windows/ffi/networking_ffi.rb'
     - 'lib/facter/util/facts/windows_release_finder.rb'
 
-RSpec/MultipleMemoizedHelpers:
```
cthorn42, line 56: Was this, `RSpec/MultipleMemoizedHelpers` meant to be removed? I'm not tracking if the removal is directly related to the main issue of this PR.

cthorn42:
For the commit, 69030d149a3d94f53153925f2bd77ca602c6f4db, could we just include the removal of `RSpec/MultipleMemoizedHelpers`?

cthorn42 (approved):

PR #2734, opened 2024-06-18
(FACT-3463) Add server to mountpoint device for AIX

tvpartytonight:


tvpartytonight:
I have manually checked this change in `AIX-7.2` and the mountpoints successfully returns the server name.

lib/facter/resolvers/aix/mountpoints.rb
```diff
@@ -31,9 +31,14 @@ def read_mount(fact_name)
           def add_mount_points_fact(line)
             elem = line.split("\s")
 
-            elem.shift unless line[0] == ' '
+            if line[0] != ' '
+              server = elem.shift
+              device = "#{server}:#{elem[0]}"
+            else
+              device = elem[0]
+            end
```
joshcooper, line 39: Ah interesting, there was a previous fix 6c6ef2df386c416e65914048e0ad4749da78b8a1 to just ignore the `server`. Could you mention that in your commit message?

joshcooper (approved):

PR #2740, opened 2024-07-26
JRuby test bumps

mhashizume:
The JRuby versions in both streams of puppetserver were recently updated. This PR bumps the versions of JRuby used in tests accordingly.


mhashizume:
We should update branch protection rules after approval to reflect this change.

Issue #2742, opened 2024-07-30
Resolving the `virtual` fact on FreeBSD produce warning


smortex:
## Describe the Bug

```sh-session
root@puppet:~ # facter --version
4.8.0
root@puppet:~ # facter virtual
[2024-07-30 09:30:11.629353 ] WARN Facter::Resolvers::Containers - Unable to getenv for pid 1, 'uninitialized constant Facter::Util::Linux' 
jail
```

## Expected Behavior

```sh-session
root@puppet:~ # facter virtual
jail
```

No warning about something that looks suspiciously Linux-related :-)

## Steps to Reproduce

Run facter 4.8.0 on a FreeBSD system

## Environment
 - Version 4.8.0
 - Platform FreeBSD

## Additional Context

The problem is incidentally introduced by #2574 where Linux specific code is added to `lib/facter/resolvers/containers.rb`, a file which based on its name is supposed to be platform-agnostic.

After further investigation, this file is really platform-specific and not in the right directory.  A few other files have the same problem.  I will open two PR to:

1. Fix files located in the wrong directory to make it explicit if they are specific to a platform or not;
2. Change the way the virtual fact is resolved on FreeBSD to stop the warning from being produced.

bibi21000:
same problem here in a lxc container with Ubuntu 22.04 and Ubuntu 20.04

> Warning: Facter: Container runtime, 'lxc', is unsupported, setting to, 'container_other'

bibi21000:
@joshcooper ty for information

buzzdeee:
same here on OpenBSD with facter 4.8.0:

 facter33 virtual
[2024-08-19 08:04:38.279003 ] WARN Facter::Resolvers::Containers - Unable to getenv for pid 1, 'uninitialized constant Facter::Util::Linux' 
physical

dhs-rec:
Still seeing this with facter 4.10.0 on FreeBSD 14.

PR #2743, opened 2024-07-30
Ensure platform-specific code is correctly namespaced

smortex:
Facter facts and providers are sometimes tied to a specific platform (e.g. Linux, FreeBSD, etc).  When it is the case, they are placed in a directory named after this platform.

Some files are incorrectly stored in the base directory making it less obvious that they are platform-specific.

This PR fix this issue and is part 1/2 of a fix for #2742


joshcooper:
Thanks @smortex! Moving the `containers` resolver makes sense based on my reading of https://wiki.freebsd.org/Containers, but what about the others like `dmi`? What about them makes them linux specific? Is it the `/sys/class` filesystem?

When adding resolvers in the future, what's the best way to determine if something is linux-specific or not? Wondering if maybe there's a good reference/porting guide we could refer to?

smortex:
Hey @joshcooper, I am catching up after long holiday away from home

> Thanks @smortex! Moving the `containers` resolver makes sense based on my reading of [wiki.freebsd.org/Containers](https://wiki.freebsd.org/Containers), but what about the others like `dmi`? What about them makes them linux specific? Is it the `/sys/class` filesystem?

Indeed, `/sys` is something that is not available on FreeBSD and probably some other namespacing could be done if dmi facts are collected there.  My goal in this PR is to eventually fix a warning that is produced with the latest code.  As far as I can see, the dmi facts do not produce warnings when attempting to access `/sys/...`.

> When adding resolvers in the future, what's the best way to determine if something is linux-specific or not? Wondering if maybe there's a good reference/porting guide we could refer to?

Ughhhhh :frowning: No real idea. Some acceptance tests that would run on various operating systems and ensure that no warning / error is produced may help… but that's non-trivial.  Maybe creating a team with people interested in "exotic" operating systems and pinging that team when new provider are added may be simpler :shrug: ?

> Also could you resolve conflicts?

Done!

PR #2744, opened 2024-07-30
Fix resolution of the `virtual` fact on FreeBSD

smortex:
We should not call Linux-specific code from a non-Linux system.  Before #2743, the Linux-specific code was unexpectedly loaded, and the Linux resolver was non-functional on FreeBSD, causing a warning to be output.  With #2743, the Linux specific code is not loaded anymore, breaking further the fact resolution.

In order to fix this, the virtual_detector code needs to either be split into platform-specific pieces, or take care to only call platform-specific code on these platforms.  Given this is a utility class, I chose the second route, and since the Linux-specific code is now only loaded on Linux, rely on the namespace existing or not to determine if we should run that code.

This is part 2/2 of a fix of #2742


smortex:
Interesting failures with jRuby… I don't understand what is happening wrongly :monocle_face:.  Suggestions are welcome!

ltning:
Hola! What is keeping this from getting merged, I wonder?

PR #2745, opened 2024-08-05
Add type GUIDs to partitions resolver

mhashizume:
This PR adds partition type GUIDs to the partitions resolver.

In order to add this information, this PR also changes the partitions resolver to prefer `lsblk` over `blkid`. Previously, Facter would use `blkid` and fall back to `lsblk` only if `blkid` did not return anything, for whatever reason. However, there are several reasons to prefer `lsblk`:

- `lsblk` can return more information (like partition type GUIDs) than `blkid`.
- `lsblk` does not have to necessarily run as root, `blkid` does.
- Newer versions of `blkid`'s man page specifically recommend using `lsblk`: 
> **It is recommended to use lsblk(8) command to get information about block devices, or lsblk --fs to get an overview of filesystems, or findmnt(8) to search in already mounted filesystems.**
>
> lsblk(8) provides more information, better control on output formatting, easy to use in scripts and it does not require root permissions to get actual information. blkid reads information directly from devices and for non-root users it returns cached unverified information. blkid is mostly designed for system services and to test libblkid(3) functionality.


mhashizume:
I still need to add tests and make sure that `lsblk` returns information in the same order as `blkid` for a stable API.

lib/facter/resolvers/partitions.rb
```diff
@@ -121,7 +122,18 @@ def execute_and_extract_blkid_info
...
 
-          blkid_and_lsblk[:lsblk] ||= Facter::Core::Execution.execute('lsblk -fp', logger: log)
+          lsblk_version_raw = Facter::Core::Execution.execute('lsblk --version 2>&1', logger: log)
+          lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f
+
+          # The -p or --paths option was added in lsblk 2.23, anything before that it makes more sense to return and fall back to blkid
+          return {} if lsblk_version < 2.23
+
+          # lsblk 2.25 added support for GPT partition type GUIDs
+          blkid_and_lsblk[:lsblk] ||= if lsblk_version >= 2.25
```
jcpunk, line 132: Might it be easier to run `lsblk` with the longer args and `rescue` if it errors out?

mhashizume:
Closing and reopening to grab latest tests.

mhashizume:
I've updated the branch protection rule to no longer require integration tests against JRuby 9.3.90 or 9.4.7.0 (these tests were updated in https://github.com/puppetlabs/facter/pull/2740)

cthorn42 (approved):

PR #2746, opened 2024-08-07
(FACT-3474) Update facter schema, CONTRIBUTING.md, and add acceptance test to verify facts conform to schema

AriaXLi:
This PR:
- Updates facter.yaml to include any facts added since the schema was last
  updated
- Updates CONTRIBUTING.md so contributors know to add new facts to facter.yaml
  schema
- Adds an acceptance test that tests the facter output conforms to the schema
  on each platform

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,9 @@
+# How to contribute #
+
+Third-party patches are essential for keeping Puppet great. We simply can't access the huge number of platforms and myriad configurations for running Puppet. We want to keep it as easy as possible to contribute changes that get things working in your environment. There are a few guidelines that we need contributors to follow so that we can have a chance of keeping on top of things.
+
+## Adding New Facts ##
+
+When adding new facts, they need to be added to the [schema](lib/schema/facter.yaml). The fact name, description, and type must be specified in the [schema](lib/schema/facter.yaml).
+
+Learn more about how to contribute in our [Contribution Guidelines](https://github.com/puppetlabs/.github/blob/main/CONTRIBUTING.md).
--- a/acceptance/tests/facts/schema.rb
+++ b/acceptance/tests/facts/schema.rb
@@ -1,0 +1,115 @@
+test_name "Validate facter output conforms to schema" do
+  tag 'risk:high'
+  confine :except, :platform => 'windows' # See FACT-3479, once resolved this line can be removed
+
+  require 'yaml'
+  require 'ipaddr'
+
+  # Validates passed in output facts correctly conform to the facter schema, facter.yaml.
+  # @param schema_fact The schema fact that matches/corresponds with output_fact
+  # @param schema_fact_value The fact value for the schema fact
+  # @param output_fact The fact that is being validated
+  # @param output_fact The fact value of the output_fact
+  def validate_fact(schema_fact, schema_fact_value, output_fact, output_fact_value)
+    schema_fact_type = schema_fact ? schema_fact_value["type"] : nil
+    fail_test("Fact: #{output_fact} does not exist in schema") unless schema_fact_type
+
+    # For each fact, it is validated by verifying that output_fact_value can
+    # successfully parse to fact_type and the output fact has a matching schema
+    # fact where both their types and name or regex match.
+    case output_fact_value
+    when Hash
+      fact_type = "map"
+    when Array
+      fact_type = "array"
+    when TrueClass, FalseClass
+      fact_type = "boolean"
+    when Float
+      fact_type = "double"
+    when Integer
+      fact_type = "integer"
+    when String
+      if schema_fact_type == "ip"
+        begin
+          IPAddr.new(output_fact_value).ipv4?
+        rescue IPAddr::Error
+          fail_test("Invalid ipv4 value given for #{output_fact} with value #{output_fact_value}")
+        else
+          fact_type = "ip"
+        end
+      elsif schema_fact_type == "ip6"
+        begin
+           IPAddr.new(output_fact_value).ipv6?
+        rescue IPAddr::Error
+          fail_test("Invalid ipv6 value given for #{output_fact} with value #{output_fact_value}")
+        else
+          fact_type = "ip6"
+        end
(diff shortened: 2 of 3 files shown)


acceptance/tests/facts/schema.rb
```diff
@@ -0,0 +1,88 @@
...
+          puts "no schema elements: #{schema_elements} \n\n"
+          next
+        end
+        schema_fact, schema_fact_value = get_fact(schema_elements, fact)
+        validate_fact(schema_fact, schema_fact_value, fact, value)
+      end
+    end
+    if fact_type != schema_fact_type
+      puts "#{output_fact} has value: #{output_fact_value} and type: #{fact_type} does not conform to schema fact value type: #{schema_fact_type} \n\n"
+    end
```
joshcooper, line 48: Thinking we should also validate that it's possible to parse `output_fact_value` based on the `fact_type`. For example, `identity.uid` is supposed to be an `integer`, so I would expect a fact value of `uid => "bob"` to fail validation. Doing `Integer(output_fact_value)` will either successfully parse the value as an integer or raise an exception.

For Hash/Map, I think you want to verify that `output_fact_value.is_a?(Hash)` before calling `output_fact_value.each do |fact, value|`.

For IP addresses, you may want to use ruby's built in classes like:

```ruby
if schema_fact_type == "Ip"
  fail_test("...") unless IPAddr.new(output_fact_value).ipv4?
elsif schema_fact_type == "Ip6"
  fail_test("...") unless IPAddr.new(output_fact_value).ipv6?
...
```

AriaXLi:
ubuntu-20.04 failed because the disks.<devicename>.wwn fact doesn't currently exist in the schema...

AriaXLi:
On Windows, the `networking.interfaces.dhcp` fact is returning "" instead of a valid IPv4 value. This causes the schema.rb acceptance test to fail since the `dhcp` fact is not conforming to the schema. Due to this, the test will temporarily not be run on Windows until [FACT-3479](https://perforce.atlassian.net/browse/FACT-3479) is resolved

acceptance/tests/facts/schema.rb
```diff
@@ -0,0 +1,117 @@
...
+  # @param output_fact The fact value of the output_fact
+  def validate_fact(schema_fact, schema_fact_value, output_fact, output_fact_value)
+    fact_type = output_fact_value.class.to_s.downcase
+    schema_fact_type = schema_fact ? schema_fact_value["type"] : nil
+    fail_test("Fact: #{output_fact} does not exist in schema") unless schema_fact_type
+
+    # For each fact, it is validated by verifying that output_fact_value can
+    # successfully parse to fact_type and the output fact has a matching schema
+    # fact where both their types and name or regex match.
+    if fact_type == "hash"
```
joshcooper, line 21: A common pattern in ruby is to use `case/when`. It uses triple equals (`===`) to detect if `output_fact_value` is an instance of each class, so that you don't have to call `output_fact_value.class.to_s.downcase` and then list out each stringified class name:

```suggestion
    case output_fact_value
    when Hash
      fact_type = 'map'
    when Array
      fact_type = 'array'
    ...
    else
       fail_test("Invalid fact type given: #{fact_type}")
    end
```

acceptance/tests/facts/schema.rb
```diff
@@ -0,0 +1,117 @@
...
+      fact_type = "map"
+    elsif fact_type == "array"
+      Array(output_fact_value)
+    elsif fact_type == "trueclass" || fact_type == "falseclass"
+      fact_type = "boolean"
+    elsif fact_type == "float"
+      Float(output_fact_value)
+      fact_type = "double"
+    elsif fact_type == "integer"
+      Integer(output_fact_value)
```
joshcooper, line 32: Since the `output_fact_value` is an instance of `Hash`, `Array`, `Boolean`, `Float`, `Integer` or `String` I don't think we need to validate the `output_fact_value` by creating a copy, e.g. `Integer(...)`.

I do think we want to continue validating IP and mac addresses, since not all `Strings` are valid IP/mac addresses.

acceptance/tests/facts/schema.rb
joshcooper: Sorry, I think I got a little confused here. If facter were to emit `uid => "bob"` then the String "bob" would be passed as the `output_fact_value`. We'd set the `fact_type = 'string'`, but that won't match the type as specified in the schema. So it would be caught by the `assert_match(fact_type, schema_fact_type, ...`

CONTRIBUTING.md
```diff
@@ -0,0 +1,9 @@
+# How to contribute #
+
+Third-party patches are essential for keeping Puppet great. We simply can't access the huge number of platforms and myriad configurations for running Puppet. We want to keep it as easy as possible to contribute changes that get things working in your environment. There are a few guidelines that we need contributors to follow so that we can have a chance of keeping on top of things.
+
+## Adding New Facts ##
+
+When adding new facts, they need to be added to the [schema](facter.yaml). The fact name, description, and type must be specified in the [schema](facter.yaml).
```
bastelfreak, line 7: should it be `lib/schema/facter.yaml` instead of `facter.yaml`?

bastelfreak:
@AriaXLi thanks for the great PR! Do you know if there's an easy way for us to validate a factset, in json, against this schema? I'm asking because in [FacterDB](https://github.com/voxpupuli/facterdb) we collect a lot of factsets from running machines, but some are also generated with sed statements or manipulated.

AriaXLi:
> @AriaXLi thanks for the great PR! Do you know if there's an easy way for us to validate a factset, in json, against this schema? I'm asking because in [FacterDB](https://github.com/voxpupuli/facterdb) we collect a lot of factsets from running machines, but some are also generated with sed statements or manipulated.

Hi @bastelfreak thanks, glad I can help! Hmm I chose yaml since `lib/schema/facter.yaml` is yaml but I imagine you could probably use a factset in json by loading `lib/schema/facter.yaml` as a json file instead of yaml 🤔 not 100% sure though as I haven't tried it myself

joshcooper (approved):

PR #2749, opened 2024-08-28
Add error handling for older lsblk

mhashizume:
The partitions resolver uses `lsblk --version` to determine the version of lsblk and what features it has available to it. The --version flag was not added until lsblk 2.22 in 2012, so versions prior to that will respond with `lsblk: unrecognized option '--version'`.

Puppet still supports older operating systems like SUSE Enterprise Linux 11 that may ship versions of lsblk that do not support --version.

This commit updates the partitions resolver to handle errors when determining the version of lsblk and to fall back to blkid when lsblk is too old to support --version.

### DIFF
--- a/lib/facter/resolvers/partitions.rb
+++ b/lib/facter/resolvers/partitions.rb
@@ -123,7 +123,8 @@ module Facter
           return {} unless available?('lsblk', blkid_and_lsblk)
 
           lsblk_version_raw = Facter::Core::Execution.execute('lsblk --version 2>&1', logger: log)
-          lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f
+          # Return if the version of lsblk is too old (< 2.22) to support the --version flag
+          lsblk_version_raw.match?(/ \d\.\d+/) ? lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f : (return {})
 
           # The -p/--paths option was added in lsblk 2.23, return early and fall back to blkid with earlier versions
           return {} if lsblk_version < 2.23


mhashizume:
SLES 11 behavior before this PR:

```
cagey-dilution:~ # cat /etc/os-release
NAME="SLES"
VERSION="11.4"
VERSION_ID="11.4"
PRETTY_NAME="SUSE Linux Enterprise Server 11 SP4"
ID="sles"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:suse:sles:11:4"
cagey-dilution:~ # facter partitions
[2024-08-28 14:50:25.327269 ] ERROR Facter::Resolvers::Partitions - Resolving fact partitions, but got undefined method `[]' for nil:NilClass at /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/facter/resolvers/partitions.rb:126:in `populate_from_lsblk'

```

After this PR:

```
cagey-dilution:~ # facter partitions
{
  /dev/sda1 => {
    filesystem => "swap",
    size => "1.06 GiB",
    size_bytes => 1141899264,
    uuid => "aaf307c8-7759-4908-a7c1-f7d91a42a97e"
  },
  /dev/sda2 => {
    filesystem => "ext3",
    mount => "/",
    size => "18.92 GiB",
    size_bytes => 20317208576,
    uuid => "b9df5794-4316-4e3e-bcbb-6e3b85976c06"
  }
}
cagey-dilution:~ #
```

lib/facter/resolvers/partitions.rb
```diff
@@ -122,8 +122,15 @@ def execute_and_extract_blkid_info
...
 
-          lsblk_version_raw = Facter::Core::Execution.execute('lsblk --version 2>&1', logger: log)
-          lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f
+          # Catch exceptions and return for old versions of lsblk (< 2.22) that do not support the --version flag
+          # We only really need this for one of the oldest OSes that Puppet supports, SLES 11. We can probably remove this error handling
+          # when we drop SLES 11 support in Puppet.
+          begin
+            lsblk_version_raw = Facter::Core::Execution.execute('lsblk --version 2>&1', logger: log)
+            lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f
+          rescue NoMethodError
```
joshcooper, line 131: Does the `execution` method return an exit code so you can know if it succeeded or not? I'm thinking the NoMethodError is raised when calling `[0]` on nil. Maybe check for a match explicitly?

mhashizume: While `lsblk --version` returns 1 when run in Bash, it doesn't look like `Facter::Core::Execution.execute` returns any exit code:

https://github.com/puppetlabs/facter/blob/b6cf607a746e02e040ce5ca2dc85cc14c0982803/lib/facter/custom_facts/core/execution/base.rb#L110

joshcooper: Would it make sense to handle the nil case explicitly?

```ruby
lsblk_version_raw = ...
lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)&.first&.to_f || 0
return {} if lsblk_version < 2.23
```

lib/facter/resolvers/partitions.rb
```diff
@@ -123,10 +123,10 @@ def populate_from_lsblk(partition_name, blkid_and_lsblk)
           return {} unless available?('lsblk', blkid_and_lsblk)
 
           lsblk_version_raw = Facter::Core::Execution.execute('lsblk --version 2>&1', logger: log)
-          lsblk_version = lsblk_version_raw.match(/ \d\.\d+/)[0].to_f
+          lsblk_version = lsblk_version_raw.match(/\d\.\d+/)&.to_f
```
joshcooper, line 126: I think maybe?

```suggestion
          lsblk_version = lsblk_version_raw.match(/\d\.\d+/)&.first&.to_f
```

mhashizume: I gave this a shot, but it didn't work. It's weird that MatchData acts like an Array in some ways (uses `[]` syntax) but does not support methods like `first`.

I tried another approach, let me know what you think

joshcooper (approved):

Issue #2752, opened 2024-09-05
Tests fail on Arch Linux


bastelfreak:
## Describe the Bug

Hi,
when running `rspec`  on Arch Linux, a few tests fail:

```
rspec ./spec/framework/detector/os_detector_spec.rb:127 # OsDetector initialize when host_os is linux detects linux distro
rspec ./spec/framework/detector/os_detector_spec.rb:131 # OsDetector initialize when host_os is linux calls Facter::OsHierarchy with construct_hierarchy
rspec ./spec/framework/detector/os_detector_spec.rb:137 # OsDetector initialize when host_os is linux calls Facter::Resolvers::OsRelease with id
rspec ./spec/framework/detector/os_detector_spec.rb:143 # OsDetector initialize when host_os is linux calls Facter::Resolvers::RedHatRelease with id
rspec ./spec/framework/detector/os_detector_spec.rb:159 # OsDetector initialize when host_os is linux when distribution is not known falls back to linux
rspec ./spec/framework/detector/os_detector_spec.rb:164 # OsDetector initialize when host_os is linux when distribution is not known when no hierarchy for os identifier logs debug message
rspec ./spec/framework/detector/os_detector_spec.rb:198 # OsDetector initialize when host_os is linux when distribution is not known when family is known logs debug message
```

errors:

```
Failures:

  1) OsDetector initialize when host_os is linux detects linux distro
     Failure/Error: hierarchy = @os_hierarchy.construct_hierarchy(identifier)

       #<InstanceDouble(Facter::OsHierarchy) (anonymous)> received :construct_hierarchy with unexpected arguments
         expected: (:redhat)
              got: (:archlinux)
       Diff:
       @@ -1 +1 @@
       -[:redhat]
       +[:archlinux]

        Please stub a default value first if message might be received with other args as well.
     # ./lib/facter/framework/detector/os_detector.rb:46:in `detect_hierarchy'
     # ./lib/facter/framework/detector/os_detector.rb:41:in `detect'
     # ./lib/facter/framework/detector/os_detector.rb:13:in `initialize'
     # ./spec/framework/detector/os_detector_spec.rb:128:in `block (4 levels) in <top (required)>'

  2) OsDetector initialize when host_os is linux calls Facter::OsHierarchy with construct_hierarchy
     Failure/Error: hierarchy = @os_hierarchy.construct_hierarchy(identifier)

[...]

PR #2753, opened 2024-09-06
(#2752) Stub File.readable? for Devuan/Arch Linux

bastelfreak:
This is required to run the test suite on Devuan/Arch Linux boxes.


joshcooper:
rubocop failure will be fixed in https://github.com/puppetlabs/facter/pull/2754

joshcooper (approved):

PR #2754, opened 2024-09-06
rubocop: autofix

bastelfreak:
latest rubocop was complaining about this. The safe autofix corrected it.

### DIFF
--- a/lib/facter/custom_facts/core/execution/base.rb
+++ b/lib/facter/custom_facts/core/execution/base.rb
@@ -25,8 +25,8 @@ module Facter
             # set the new (temporary) value for the environment variable
             ENV[var] = value
           end
-          # execute the caller's block, capture the return value
-          rv = yield
+          # execute the caller's block, returning its value
+          yield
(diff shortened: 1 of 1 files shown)


lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -39,7 +39,6 @@ def with_env(values)
             end
           end
           # return the captured return value
```
joshcooper, line 41: could you remove the stale comment?

lib/facter/custom_facts/core/execution/base.rb
```diff
@@ -26,7 +26,7 @@ def with_env(values)
             ENV[var] = value
           end
           # execute the caller's block, capture the return value
```
joshcooper, line 28:
```suggestion
          # execute the caller's block, returning its value
```

lib/facter/custom_facts/core/execution/base.rb
bastelfreak: done

joshcooper (approved):

PR #2755, opened 2024-09-06
dependabot: check for github actions and bundler

bastelfreak:


joshcooper:
rubocop failure will be fixed in https://github.com/puppetlabs/facter/pull/2754

joshcooper (approved):

PR #2758, opened 2024-09-19
Support 'xenhvm' for AWS instances

treydock:


joshcooper:
> $ sudo facter -p cloud.provider
> $

I would have expected this to return `aws`

Can you update your commit message to describe how facter behaved before? And how `xen-on-nitro` is different than `xen`?

Does `/opt/puppetlabs/puppet/bin/virt-what` return the expected output?

treydock:
Updated commit message, will paste here for completeness:

```
$ sudo facter -p cloud

$ sudo facter -p virtual
xenhvm

$ sudo /opt/puppetlabs/puppet/bin/virt-what
xen
xen-hvm
aws
```

We have pushed this custom fact to our systems that mimics the logic Facter uses with this PR's addition:

```ruby
Facter.add('cloud') do
  setcode do
    value = nil
    provider = Facter::Util::Facts::Posix::VirtualDetector.platform
    if ['xenhvm', 'xen', 'kvm'].include?(provider)
      metadata = Facter::Resolvers::Ec2.resolve(:metadata)
      value = 'aws' unless metadata.nil? || metadata.empty?
    elsif ['gce'].include?(provider)
      metadata = Facter::Resolvers::Gce.resolve(:metadata)
      value = 'gce' unless metadata.nil? || metadata.empty?
    end
    if value.nil?
      value
    else
      { 'provider' => value }
    end
  end
end
```

Using that custom fact that accounts for `xenhvm` we get this:

```
$ sudo facter -p cloud
{
  provider => "aws"
}
$ sudo puppet facts show cloud
{
  "cloud": {
    "provider": "aws"
  }
}
```

treydock:
@joshcooper Let me know if these changes and the updated commit messages are looking like they could be merged.

joshcooper (approved):

PR #2762, opened 2024-09-30
FACT-3459: Workaround 2693 by ignoring resolv.conf entries starting with dot

AriaXLi:
Currently, systems without a FQDN can be mishandled by facter for certain `/etc/resolv.conf` content. This was initially noticed when systemd-resolved was installed on a host without domain.

systemd-resolved stub resolver sets 'search .' as a search domain, which results in the following hostname/domain/fqdn triplet: foo, ., foo... See: https://github.com/systemd/systemd/blob/v255/src/resolve/resolv.conf#L19

This is wrong on multiple levels: first, facter does not seem to handle '.' (the root level) well, and there have been PRs to remove the trailing dot in FQDN as far back as 2012 (PR 200), leaving user with a convenient, but sometimes ambiguous string that is not fully qualified.

This commit implements a a middle ground solution to support top-level/domainless servers without making a breaking change. Additionally, it adds more coverage for various search cases.

(cherry picked from commit 3bffeff1e0061fa27954cf8db5df7a19fb3f8da4)

### DIFF
--- a/lib/facter/resolvers/hostname.rb
+++ b/lib/facter/resolvers/hostname.rb
@@ -68,9 +68,9 @@ module Facter
 
         def read_domain
           file_content = Facter::Util::FileHelper.safe_read('/etc/resolv.conf')
-          if file_content =~ /^domain\s+(\S+)/
+          if file_content =~ /^domain\s+([^.]\S+)/
             Regexp.last_match(1)
-          elsif file_content =~ /^search\s+(\S+)/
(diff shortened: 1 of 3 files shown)


AriaXLi:
Work was originally done in #2762, this PR just adds 2 more tests to cover more cases with `search`. This addresses issue https://github.com/puppetlabs/facter/issues/2693.

AriaXLi:
I was able to push directly to Geod24's PR instead of opening a new one so I'm closing this.

PR #2765, opened 2024-10-02
(FACT-3483) Temporarily pin sys-filesystem to <= 1.5.0 to resolve jruby integration check failure

AriaXLi:


AriaXLi:
This should be reverted once FACT-3483 is resolved.

mhashizume:
Could it make sense to keep 1.4 as the minimum version for now?

AriaXLi:
> Could it make sense to keep 1.4 as the minimum version for now?

Yeah that makes sense to me!

mhashizume (approved):

PR #2767, opened 2024-10-03
Add Ruby 3.3 to CI

bastelfreak:
Ruby 3.3 got released months ago and more and more distributions are picking it up and also building facter against it. So I think it makes sense to add it to the CI.


bastelfreak:
While checking I also noticed that https://github.com/puppetlabs/facter/blob/main/.github/workflows/acceptance_tests.yml#L17 uses quite ancient operating systems. But I don't know if that's on purpose or not.

PR #2771, opened 2024-10-17
Fix: IMDSv2 token cannot be retrieved through a proxy

scrill:
AWS instance metadata V2 requires a token, but if a proxy is used, the token cannot be obtained.


scrill:
Any updates?

PR #2778, opened 2024-11-07
make sure facts for windows server 2025 is discovered as 2025

brajjan:
Windows Server 2025 was released 2024-11-01 and the current facts discovers 2025-servers as `11` instead of `2025`

This PR makes sure that 2025 server operating systems resolve to 2025 instead of 11

Manually tested on a Windows server 2025 server and Windows 11 client

### DIFF
--- a/lib/facter/util/facts/windows_release_finder.rb
+++ b/lib/facter/util/facts/windows_release_finder.rb
@@ -25,10 +25,12 @@ module Facter
           def check_version_10_11(consumerrel, kernel_version)
             build_number = kernel_version[/([^.]*)$/].to_i
 
-            return '11' if build_number >= 22_000
-            return '10' if consumerrel
+            return '11' if consumerrel && build_number >= 22_000
+            return '10' if consumerrel && build_number >= 20_348
 
-            if build_number >= 20_348
+            if build_number >= 26_100
+              '2025'
+            elsif build_number >= 20_348
               '2022'
             elsif build_number >= 17_623
               '2019'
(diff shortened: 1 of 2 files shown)


joshcooper:
Thanks @brajjan! Do you know of documentation from Microsoft where they specify the relationship between build numbers and 2025?

brajjan:
Sure. This is the official documentation of all LTSC releases. https://learn.microsoft.com/en-us/windows-server/get-started/windows-server-release-info

joshcooper:
Looks like a valid spec failure for Windows 10 detection? https://github.com/puppetlabs/facter/actions/runs/11720779220/job/32817170282?pr=2778#step:4:14

brajjan:
It sure does.. was no check for version number for Windows 10 before my PR - only checking if it was a client less than version 11

Seems like some of the version checks in the [code](https://github.com/puppetlabs/facter/blob/0656d9a34ce4790129f1fd6eba5bb4d49a9b9ad1/lib/facter/util/facts/windows_release_finder.rb#L31-L34) are against the LTSC/GA-releases and some checks are for Insiders/beta-releases. 
[2019-Build 17623 - Insiders](https://techcommunity.microsoft.com/discussions/windowsserverinsiders/announcing-windows-server-2019-preview-build-17623/173715) (LTSC = 17763)
2022-Build 20348 - LTSC

What do you think is the proper way to do this? Check the first beta release of each version and have that as the check, or is it acceptable to only check for GA/LTSC-releases? If acceptable I will update all version checks to check for the earliest GA/LTSC-release for that OS

I can probably get a hold of all beta build numbers if I do a bit of research.

dhs-rec:
While testing Puppet on Windows Server 2025 in AWS, I made the same change to get the correct release information, only to find this PR afterwards. Would be great to have it in the next Puppet 7/8 releases.

PR #2783, opened 2024-12-06
facter: change sys-filesystem to a runtime dep

lollipopman:
Prior to this change sys-filesystem was listed as a development dependency, to avoid requiring a C compiler when performing a gem install. However, this produces a version of facter whose mountpoints fact and partitions facts are broken. Re-visit the decision and prefer to produce a fully functional facter version.

This issue was discovered when using facter in Debian, where the lack of a dependency on sys-filesystem produces a package with broken mountpoints and partitions facts.


lollipopman:
Debian bug for reference, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1089174

Issue #2781, opened 2024-11-27
Incorrect `hypervisors` and `virtual` Facts with Unsupported Container Runtime Warning in Docker


molhamalnasr:
## Describe the Bug

When running Puppet Facter inside a Docker container, the `hypervisors` key and `virtual` fact are incorrectly set. Additionally, a warning is displayed:  
`Warning: Facter: Container runtime, 'docker', is unsupported, setting to, 'container_other'`.

### Example Output:
```bash
# facter -p hypervisors
[2024-11-27 13:49:35.972235 ] WARN Facter::Resolvers::Containers - Container runtime, 'docker', is unsupported, setting to, 'container_other'

# facter -p virtual
[2024-11-27 13:49:42.488042 ] WARN Facter::Resolvers::Containers - Container runtime, 'docker', is unsupported, setting to, 'container_other'
container_other
```

## Expected Behavior

In previous versions (e.g., Tag 4.7.0), the `hypervisors` and `virtual` facts returned the correct values:

```bash
# facter -p hypervisors
{
  docker => {
    id => "<STRING_ID>"
  }
}

# facter -p virtual
docker
```

## Steps to Reproduce

1. Run Puppet Agent/Facter inside a Docker container.
2. Execute the command `facter -p hypervisors` or `facter -p virtual`.

## Environment

- **Version:** Puppet Agent 7.32.1  
- **Platform:** Ubuntu 24.04  
- **Facter Version:** 4.8.0  

## Additional Context

After debugging the repository, I identified the root cause in the following line within `lib/facter/resolvers/containers.rb`:

```ruby
read_environ(fact_name) || read_cgroup(fact_name)
```

The call to `read_environ(fact_name)` always returns a value, which prevents `read_cgroup(fact_name)` from being executed. This causes the resolver to misidentify the container runtime and set incorrect values for the `hypervisors` and `virtual` facts.

mhashizume:
Thank you for reporting this @molhamalnasr , we have added this issue to our backlog.

joshcooper:
@molhamalnasr I can't reproduce this ever working. In particular, `read_environ` returns nil, because there isn't a `container` environment variable defined for pid 1:

```
❯ docker run -it ubuntu:noble cat /proc/1/environ 
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/binHOSTNAME=00ddb9b41876TERM=xtermHOME=/root
```

[...]

Issue #2784, opened 2024-12-19
Unsupported container runtime 'openvzve' with facter 4.8.0+


Viktor0014:
## Describe the Bug
Since facter 4.8.0 we getting unsupported container runtime 'openzve'

Problem origin here [facter source](https://github.com/puppetlabs/facter/blame/8dfaabf794dab16267120e906002c8147a27ef8c/lib/facter/resolvers/containers.rb#L46)

## Expected Behavior
Support 'openvzve' as runtime again

## Steps to Reproduce
1. Setup virtuozzo 7 server (any version) 
2. Install puppet-agent 7.32 or newer
3. Simple run facter resulting in following output (stripped down)
[root@test.priv resolvers]# facter virtual
[2024-12-19 11:11:27.557706 ] WARN Facter::Resolvers::Containers - Container runtime, '0ae2eb17-876d-46b1-a093-1ee716f7a75a', is unsupported, setting to, 'container_other'
container_other

## Environment
 - puppet-agent-7.32.0-1.el8.x86_64 or newer 
 - Virtuozzo Linux release 7.9 
 - Facter 4.8.0

## Additional Context
see also fact for openvz in source code https://github.com/puppetlabs/facter/blob/8dfaabf794dab16267120e906002c8147a27ef8c/lib/facter/resolvers/open_vz.rb

joshcooper:
This is caused by 7bc38ccb58e6abf66d6141a3b2e9f60eb0077d71 because if the container type is not recognized by the `Container` resolver, then it returns `container_other`

https://github.com/puppetlabs/facter/blob/7bc38ccb58e6abf66d6141a3b2e9f60eb0077d71/lib/facter/resolvers/containers.rb#L56

However, the `VirtualDetector` calls the `Container` resolver

https://github.com/puppetlabs/facter/blob/0656d9a34ce4790129f1fd6eba5bb4d49a9b9ad1/lib/facter/util/facts/posix/virtual_detector.rb#L19

So `container_other` prevents it from falling back to other methods like `check_open_vz`

https://github.com/puppetlabs/facter/blob/0656d9a34ce4790129f1fd6eba5bb4d49a9b9ad1/lib/facter/util/facts/posix/virtual_detector.rb#L11-L12

We had the same issue with Illumox, see 31d32da513aae094bdbcc011155c2e247be11284

The `Container` resolver should ignore unknown container types so that `VirtualDetector` can try other methods

BuhtigithuB:
[...]

PR #2786, opened 2025-01-14
fix(Containers resolver): properly detect Docker runtime in containers (#2781)

molhamalnasr:
Previously, the `Containers` resolver would always return a value from `read_environ`, preventing `read_cgroup` from being called. As a result, the `hypervisors` and `virtual` facts were often set incorrectly to `container_other` and displayed a misleading warning.

Changes in this commit:
- `read_cgroup` now returns early unless Docker or LXC matches are found.
- `read_environ` returns `nil` when encountering an unsupported container.
- If both methods return `nil`, `vm` defaults to `container_other` with a warning.

These updates ensure that Facter correctly detects Docker and properly reports the `virtual` and `hypervisors` facts when running inside a Docker container.

issue: (#2781)

### DIFF
--- a/lib/facter/resolvers/containers.rb
+++ b/lib/facter/resolvers/containers.rb
@@ -15,7 +15,17 @@ module Facter
 
         def post_resolve(fact_name, _options)
           @fact_list.fetch(fact_name) do
-            read_environ(fact_name) || read_cgroup(fact_name)
+            environ_result = read_environ(fact_name)
+            cgroup_result = read_cgroup(fact_name)
+
+            if environ_result.nil? && cgroup_result.nil?
+              @fact_list[:vm] = 'container_other'
+              @fact_list[:hypervisor] = ''
+              log.warn("Container runtime is unsupported, setting to 'container_other'")
+              @fact_list[fact_name]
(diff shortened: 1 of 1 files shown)


lollipopman:
looks good to me, thanks

lib/facter/resolvers/containers.rb
```diff
@@ -15,7 +15,17 @@ class << self
 
         def post_resolve(fact_name, _options)
           @fact_list.fetch(fact_name) do
-            read_environ(fact_name) || read_cgroup(fact_name)
+            environ_result = read_environ(fact_name)
+            cgroup_result = read_cgroup(fact_name)
+
+            if environ_result.nil? && cgroup_result.nil?
+              @fact_list[:vm] = 'container_other'
```
joshcooper, line 22: `VirtualDetector` assumes this resolver will return nil if it doesn't recognize the container type (or if there is none), so that the detector can try other methods. See my comments in https://github.com/puppetlabs/facter/issues/2784#issuecomment-2620234191

danielchristianschroeter:
+1

PR #2799, opened 2026-04-05
fix: include cloud.provider in EC2 and GCE fact groups to respect blocklist

### LINKED ISSUE #2690: The cloud.provider fact still calls EC2 despite EC2 being in the block list
## Describe the Bug
Even if EC2 is in the blocklist for facter, the `cloud.provider` fact will still make a call to EC2. The spot where it calls EC2 is [here](https://github.com/puppetlabs/facter/blob/7bfd75b/lib/facter/facts/linux/cloud/provider.rb#L14).

## Expected Behavior
When EC2 is in the blocklist for facts, no calls should be made to EC2.

## Workaround
Add the `cloud.provider` fact in the blocklist as well as EC2 to prevent the call.

Is related to #2672

raman1236:
## Summary

Fixes #2690

When `EC2` or `GCE` is added to the `blocklist` in `facter.conf`, the `cloud.provider` fact continues to run and makes HTTP requests to cloud metadata services (IMDS/GCE metadata).

## Root Cause

The `FACT_GROUPS` in `lib/facter/config.rb` defines the `EC2` group as only `[ec2_metadata, ec2_userdata]` and the `GCE` group as only `[gce]`. The `cloud.provider` fact is not listed in either group, so blocking them has no effect on `cloud.provider` — which directly calls `Facter::Resolvers::Ec2.resolve(:metadata)` and `Facter::Resolvers::Gce.resolve(:metadata)`.

## Fix

Add `cloud.provider` to both the `EC2` and `GCE` `FACT_GROUPS` entries. This way, blocking either group also prevents `cloud.provider` from running and making network requests to metadata services.

## Testing

Added two rspec tests in `spec/framework/config/fact_groups_spec.rb`:
- `with blocked EC2 group` → verifies `cloud.provider` appears in blocked facts
- `with blocked GCE group` → verifies `cloud.provider` appears in blocked facts

Verified with direct Ruby evaluation that the groups now include `cloud.provider`.


raman1236:
Friendly ping — this PR ensures `cloud.provider` is included in EC2 and GCE fact groups so that blocklist/allowlist filtering works correctly. Would appreciate a review when you have a moment. Thanks!

